PR #34, opened 2016-11-02
Iss15

JeremyMcCormick:
Minor cleanup to event model classes.


JeremyMcCormick:
Hi, Omar.

I will make the changes you suggested to detector.gdml and put "magic numbers" in their own variable definitions.

I also need to change the size of the HCal envelope box as it is too big right now, so that subdetector is not in the right place.

I'm still not seeing your comments here when I pull up this PR page...very strange.

--Jeremy

JeremyMcCormick:
Okay, finally your comments have shown up....

PR #42, opened 2016-11-07
Hcal digi processor & Hcal geometry tweaks

awhitbeck:
- Adding code to EventProc to simulate HCal digis from LDMX_Event tree.

### DIFF
--- a/EventProc/include/EventProc/HcalDigiProcessor.h
+++ b/EventProc/include/EventProc/HcalDigiProcessor.h
@@ -4,9 +4,13 @@
 #include "TTree.h"
 
 #include "Event/SimEvent.h"
+#include "DetDescr/DetectorID.h"
+#include "DetDescr/DefaultDetectorID.h"
 
 using event::SimEvent;
-using event::SimCalorimeterHit;;
+using event::SimCalorimeterHit;
+using detdescr::DetectorID;
+using detdescr::DefaultDetectorID;
 
 #include "EventProc/EventProcessor.h"
 #include "Event/SimCalorimeterHit.h"
@@ -20,23 +24,21 @@ class HcalDigiProcessor : public EventProcessor {
   
 public:
   
-    HcalDigiProcessor(TString outputFileName_,bool verbose_=false):outputFileName(outputFileName_),verbose(verbose_){};
+    HcalDigiProcessor(TTree* outputTree_,bool verbose_=false):outputTree(outputTree_),verbose(verbose_){};
   
-    int getLayer( SimCalorimeterHit* hcalHit ); 
     void initialize();
     void execute();      
     void finish();
    
  private:
   
-    TFile* outputFile;
     TTree* outputTree;
-    TString outputFileName;
+
     std::vector<int> *hcalDetId_,*hcalLayerNum_,*hcalLayerPEs_;
     std::vector<float> *hcalLayerEdep_,*hcalLayerTime_,*hcalLayerZpos_;
     std::map<layer,zboundaries> hcalLayers;
     bool verbose;  
-
+    DetectorID* detID;
     static const float firstLayerZpos;
     static const float layerZwidth;
     static const int numHcalLayers;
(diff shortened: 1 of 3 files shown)


JeremyMcCormick:
Hi, Owen.

There are some minor code style issues you should fix before I will merge this:

- The `EventProc/test/hcalDigiProducer.cxx` file is missing the corresponding header.  You shouldn't define the class in the `.cxx` file.

- The file name `hcalDigiProducer.cxx` should be capitalized to match the `HcalDigiProducer` class.

- I'm not sure what indent you are using throughout but it should always be 4 spaces (I thought there was some inconsistency but I can't if it is just the editor).

- The `main` defined in `hcalDigiProducer.cxx` should have its own `.cxx` file similar to how this is done in the _SimApplication_ module e.g. add a `EXECUTABLES src/ldmx_hcal_digi_producer.cxx` to the _EventProc_ `CMakeLists.txt` file.

- We don't indent for the namespace, but we are indenting for `public` and `private` (see existing code for this style).

I can spot at least one optimization.  These variables within `execute()` can be defined statically or as `static const` class members:

``` C++
float MeVperMIP = 1.40;
float PEperMIP = 13.5*6./4.;
float meanNoise = 2.;
float depEnergy;
float meanPE;
```

The layer numbers are relatively easily read using the `DetectorID` so eventually you could change this code that is using the Z position; this was probably was clear right now due to lack of doc.

You can make an ID usable for your job like:

``` C++
IDField::IDFieldList fieldList;
fieldList.push_back(new IDField("subdet", 0, 0, 3));
fieldList.push_back(new IDField("layer", 1, 4, 11));
DetectorID* detID = new DetectorID(&fieldList);
```

Then set the value on the ID and unpack it:

``` C++
detID->setRawValue(simCalHit->getID());
detID->unpack();
```

And finally you can get the layer back from the field values:

``` C++
int layerID = detID->getFieldValue("layer");
```

For now you would need to define your own ID but I plan to add a default ID type that you could directly instantiate and use and so you would not need to define in this way.

If possible, you may want to separate your geometry and code changes into two branches.  I am tracking the updates to the calorimeter geometries in a specific github issue with its own branch so it would help me.

Thanks.

--Jeremy

omar-moreno:
Hi @JeremyMcCormick 

A suggestion. If you want users to follow your coding style (as terrible as it is ;)), you should write a style guide.  Otherwise, how are we suppose to know what you think is acceptable or not.

JeremyMcCormick:
Hi, @omar-moreno.

I have a preliminary style guide I'll move from my personal space to the LDMX Confluence.

JeremyMcCormick:
Once #43 is merged then you can easily get the layer number from the encoded ID ...

``` C++
void initialize() {
    DetectorID* detID = new DefaultDetectorID();
}

void execute() {
    for( int iHit = 0 ; iHit < numHCalSimHits ; iHit++ ){
        SimCalorimeterHit* hcalHit = (SimCalorimeterHit*) getEvent()->getCollection(event::HCAL_SIM_HITS)->At(iHit);    
        detID->setRawValue(hcalHit->getID());
        detID->unpack();
        int layer = detID->getFieldValue("layer");
    }
}
```

I think then that you can get rid of the code you wrote which tries to figure out the layer by Z position of the hit.

JeremyMcCormick:
I spotted one more minor code optimization you can make.

You can get the TClonesArray pointer before your event loop and then use it directly like ...

``` C++
TClonesArray* hcalHits = getEvent()->getCollection();
for( int iHit = 0 ; iHit < numHCalSimHits ; iHit++ ) {
    SimCalorimeterHit* hcalHit = hcalHits->At(iHit);
}
````

This should provide a small performance increase over accessing it through the string method every time.

In fact, that pointer should be the same address for the whole job as it is only deleted at the end, so I think you could even cache this value in the _initialize_ method.

omar-moreno:
@JeremyMcCormick since you optimizing (prematurely as that should be a diff issue) then ...
 
Using getEvent()->getCollection(); is a problem because you use a std::map to map a TClonesArray collection to a string ID.  You should use std::unordered_map instead: 

http://www.cplusplus.com/reference/unordered_map/unordered_map/

I bet if you use this, then the performance boost that you would get from using your above suggestion would be minimal.

awhitbeck:
Notes on updated PR:

- made an honest attempt to conform to style standards; I suspect I missed things
- changed the file structure around
- implemented optimization suggestions
- have not yet attempted to rework the code to rely solely on the detID for determining the layer.  I have to review the code before I can change anything.  -- can we merge this and open an issue on this?
- reverted the hcal geometry changes; these are not in another PR

PR #60, opened 2016-11-11
pdg in SimParticle should be pdgID. 

### LINKED ISSUE #59: pdg in SimParticle should be pdgID.
This also applies to setter and getter methods.

omar-moreno:
Resolves #59 .


JeremyMcCormick:
I'm planning to merge this in tomorrow after the meeting.

JeremyMcCormick:
I think we should increment the class version of SimParticle from 1 to 2 which I didn't see in the code changes.

PR #85, opened 2016-12-07
Implement ECal hit readout.

ocolegro:


JeremyMcCormick:
Hi, Owen.  

Thanks for the updates.

I would like to see the following changed before merging this:

- The ReadoutCalorimeterHit class appears to exactly duplicate SimCalorimeterHit, and I believe it is not needed.  Several of your methods can be added to SimCalorimeterHit instead such as the method which adds an edep to an existing hit.  I do not think we want to add a new class to the event model, especially if it isn't adding any new functionality.  You also have not added the necessary changes to the Event module for your class to work with the ROOT persistency, so let's just go with the existing class and make changes to it as needed.

- The ProcessHits() method of CalorimeterSD should not be a pure virtual (but it does need to be changed to virtual as you have done for inheritance).  This class is used for writing hits from the target and trigger pad, so it should be left as is.  Your EcalSD and HcalSD classes can either call this method or overload it, as needed.

- Related to previous point, it looks like for now the HcalSD could just call the base CalorimeterSD ProcessHits method instead of implementing its own version.

- The hex readout component that you have coded within EcalSD should be put into a separate class with a clearly defined interface for getting a position from a cell ID and vice versa.  This functionality is going to be needed in recon, so it needs to be separated from the SD class which will not be available there.

- Instead of adding to the DefaultDetectorID you should make a new class like EcalDetectorID, because not every subdetector will have a "cellid".  You can also just call the field "cell" because all values from the DetectorID are for some kind of ID.  You should basically copy the code from the DefaultDetectorID constructor in EcalDetectorID and add your additional field there.

-Code style is not quite right in a few places (mostly indent levels).  Please clean this up on the branch if you can (there's a code style guide on the ldmx-sw wiki here on github).  This is not a priority compared to the above structural changes, and I can do some cleanup myself on the master after merging if necessary.

Let me know if you have questions/comments.

PR #103, opened 2016-12-19
Cleanup calorimeter geometry in v1 detector

awhitbeck:
- Cleaned up hcal geometry
- Decoupled wrap-around hcal from backing hcal
- Put ecal and hcal in separate parent geometries
- Put ECal in separate files
- Adding lots of configuration parameters to hcal geometry for optimization studies

### DIFF
--- a/Detectors/data/ldmx-det-full-v1-fieldmap/hcal.gdml
+++ b/Detectors/data/ldmx-det-full-v1-fieldmap/hcal.gdml
@@ -113,52 +113,30 @@
     <element name="Mn">
       <fraction n="1" ref="Mn55"/>
     </element>
-    <isotope N="50" Z="24" name="Cr50">
-      <atom unit="g/mole" value="49.946"/>
+    <isotope N="54" Z="26" name="Fe54">
+      <atom unit="g/mole" value="53.9396"/>
     </isotope>
-    <isotope N="52" Z="24" name="Cr52">
-      <atom unit="g/mole" value="51.9405"/>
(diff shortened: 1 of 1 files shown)


JeremyMcCormick:
Can you put back the field map in detectors.gdml?  It is currently commented out.

JeremyMcCormick:
Also, you can take this out I believe as it does not appear to be used ...

```
<box lunit="mm" name="expHall_box0x2731e20" x="20000" y="20000" z="40000"/>
```

JeremyMcCormick:
There is a conflict showing on this branch.  Can you merge in the master to your branch and resolve this before I approve the PR?  I think there were some minor changes to detector.gdml that were merged into master that you should incorporate.

awhitbeck:
For the conflict, there was some white space that it complained about and then git, for some reason, complained about replacing the calorimeter physvol tags with the separate ecal and hcal physvol tags.  The only conflicts were in detector.gdml and you can see my resolution commit here:

https://github.com/awhitbeck/ldmx-sw/commit/8480bcf5312f5beaaf98d1d4852b2f1e50c27e35

Since there were no changes made to my branch, I don't think I can resolve this?

There are two other issues that we want to fix before we merge this though:

 - replace stainless steel with plain steel.  This was found to have a sizable difference in the performance and steel is what we were using for the "LOI".
 - add copy numbers to the wrap around hcal so that the detids are unique.

JeremyMcCormick:
I'm going to make another PR from a branch off LDMXAnalysis/ldmx-sw.

PR #105, opened 2016-12-20
Replaces PR 103 for updating calorimeter geometries

JeremyMcCormick:


awhitbeck:
@JeremyMcCormick I think this addresses all of you suggestions.   I ran the overlap test last night with the hole detector and everything looked good.  I ran some muon, electron, and neutron events and everything seems to be working as expected.

JeremyMcCormick:
@awhitbeck I noticed a minor discrepancy.  The Z thickness of the ECal is 325 in the ecal.gdml file.

```
<box lunit="mm" name="Wbox0x2730a60" x="520." y="520." z="325."/>
```

But you use 326 in the detector.gdml file.

```
<constant name="em_calorimeter_thickness" value="326.0*mm"/>
```

Shouldn't these be exactly the same?

Issue #118, opened 2017-01-14
Configurable EventProcessor System


jmmans:
Based on the development in iss113:

Update the event processor core by adding a Framework package which allows the set of Processors to be configured dynamically.  

Adjust the class heirarchy so that EventImpl is a subclass of Event.

Introduce Analyzer and Producer event Processor types.

JeremyMcCormick:
I'll close this as this issue will be resolved by PR #122.  Further comments/changes will be managed from the PR page.

PR #122, opened 2017-01-18
Add configurable event processing system

jmmans:
This PR is for issue #118 and is using the branch [iss114](https://github.com/LDMXAnalysis/ldmx-sw/tree/iss114).

The top level CMakeFiles change is clearly unnecessary, but I'm not smart enough with git to drop it.

To test digitization: 
 (1) set the environment variable PYTHONPATH to include the ldmxcfg.py directory and the current working directory.  
 (2) Copy Configuration/python/sample_digi.py to the working directory
 (3) Edit the file to set the input and output file names as desired
 (4) Run ldmxsw sample_digi (without the .py, for now)

The last bit (without the py) is something which I should fix in the next round of updates -- if there is a py, the path contaiing the file should be added automatically to the pypath and the stub should be used as the name of the module to run.

### DIFF
--- a/Framework/src/EventImpl.cxx
+++ b/Framework/src/EventImpl.cxx
@@ -44,10 +44,8 @@ void EventImpl::add(const std::string& collectionName, TClonesArray* tca) {
 
 void EventImpl::addToCollection(const std::string& name, const TObject& obj) {
     std::string branchName;
-    if (name == "EventHeader")
-        branchName = name;
-    else
-        branchName = makeBranchName(name);
+    if (name == "EventHeader") return; // no adding to the event header... 
+    branchName = makeBranchName(name);
 
     auto location = objectsOwned_.find(branchName);
     if (location == objectsOwned_.end()) {
@@ -204,7 +202,6 @@ TTree* EventImpl::createTree() {
     outputTree_ = new TTree("LDMX_Events", "LDMX Events");
 
     eventHeader_ = new event::EventHeader();
-    add("EventHeader", eventHeader_);
 
     return outputTree_;
 }


JeremyMcCormick:
"The top level CMakeFiles change is clearly unnecessary,"

You added the Framework module to the build list there, so that change is necessary.

JeremyMcCormick:
I can handle your points 1 and 2 about setting up the PYTHONPATH in the .sh script that is generated for the build.  I will go ahead and add that on the branch.

For now, why don't we just have people edit or copy the .cfg files which could live in Configuration/python as a standard location?  That way you can just add this dir to the PYTHONPATH and you would not need the current working dir to also be added.

JeremyMcCormick:
Here's my preliminary feedback on what I'd like to see changed:

- The "edepSim" field should be changed to "energy" in CalorimeterHit.  There is already an MC cal hit class which has the sim edep, so this name is confusing.

- Doxygen doc needs to be added for all new classes including EcalHit, HcalHit, Producer, Analyzer, ConfigurePython, rewritten EventProcessor class, etc.

- I don't understand this macro at the end of Framework/CMakeLists.txt

```
install(FILES DESTINATION ${CMAKE_INSTALL_PREFIX}/lib)
```

There are no files in the list so I guess it isn't doing anything?

- I don't think we want an "ldmxsw" namespace.  Following the current way it should be "framework".  However, I plan to normalize everything by changing the namespace to "ldmx" across all files (covered in another github issue), so I guess we can leave it for now.  I can change around the namespaces on another branch once we have merged this.

- Code style is not following what is already in ldmx-sw in many small ways, including indentation, spacing between characters, etc.  If you do not want to fix this yourself, I can fix this up by using my Eclipse formatting on the files which I can do directly on the branch.  It would touch a lot of files, so it would be something we'd need to coordinate.

- The include guards are not following the style we have used, e.g. you have

```
#ifndef ldmxsw_Framework_EventProcessorFactory_h_included
#define ldmxsw_Framework_EventProcessorFactory_h_included
```

which should be changed to

#ifndef FRAMEWORK_EVENTPROCESSORFACTORY_H_
#define FRAMEWORK_EVENTPROCESSORFACTORY_H_

- I don't think the actual app people run should be called 'ldmxsw' because that refers to the entire project.  I don't like overloading terms like this.  Can the actual program people run be changed to something more descriptive like 'ldmx-app' ?

JeremyMcCormick:
When I run this without arguments I get a cryptic error message:

```
[1006 $] ldmxsw
terminate called after throwing an instance of 'std::logic_error'
  what():  basic_string::_S_construct null not valid
Aborted (core dumped)
```

There should be a print out showing program usage instead.

JeremyMcCormick:
So that users do not need to set the PYTHONPATH themselves, I propose that we copy all the required py files to the python installation directory (under the user install prefix dir).

That way the setup script can add python/Framework and python/Configuration to the PYTHONPATH so that the user need not set this manually in their environment.

If they copy a config to some other dir and want to run it, then they'd have to add the current working directory to PYTHONPATH themselves.

JeremyMcCormick:
When I run the sample_digi module I get the following error:

```
[1083 $] ./ldmxsw sample_digi
Process with pass name 'digi'
Processor sequence:
  Producer(ecaldigi of class EcalDigiProducer)
  Producer(hcaldigi of class HcalDigiProducer)
Files:
   'ldmx_sim_events.root' -> 'ldmx_digi_events.root'
Shared libraries to load:
   EventProc/libEventProc.so
Framework Error [LibraryLoadFailure] : Error loading library 'EventProc/libEventProc.so':EventProc/libEventProc.so: cannot open shared object file: No such file or directory
  at /u/ey/jeremym/ldmx-dev/ldmx-sw/Framework/src/EventProcessorFactory.cxx:50 in loadLibrary
```

Does this relate to the issue you had mentioned on Slack where this would need to be added to the extra link libraries of the ldmxsw executable?

jmmans:
Regarding the crash -- that is related to shared library sequencing (map isn't ready before the object gets inserted into it), but I thought it was resolved by some work I did today.  I will keep working on that -- in the worst case, there is a somewhat less magical way to do it.

JeremyMcCormick:
Based on your suggestion, I tried the following for the library in the config file...

```
p.libraries.append("libEventProc.so")
```

This seems to load EventProc from the load library path, so it works!

Can you change the existing/sample configs to use this?  That way they should be portable and don't need to be run from the build dir.

jmmans:
At the head of this commit chain, one now _needs_ the .py explicitly on the script which is going to be run.

jmmans:
These latest patches adjust how the pythonic part is handled and introduce per-event-processor templates.  Seems to work well.

jmmans:
Ok, the very initial draft version of the ECAL trigger is merged.  The threshold needs to be tuned and we need to add a requirement on the trigger pad -- future issue/PR.

I don't plan any further functionality changes.  Everything from here on out is style cleanup.  Josh and Jeremy Mc are very welcome to join in .. :-)

jmmans:
Ok, I've added the documentation to the critical classes.  I'm going to turn to some other priorities now, but I'll come back to address issue #126 and #135.

JeremyMcCormick:
I committed a bunch of code cleanup fixes for style, removal of tabs, and header guard format.  I also added doxygen class and file tags where they were missing.  Most functions have full doc now as well.

I also changed the CMake generation of the python module init files so they are written first to the build dir and then the python directory in the installation area.  This keeps the source tree clean of build files.

There is a bug now where EventHeader is not being filled correctly after the update which fixed the duplicate EventHeade objects being added (I assume it is a consequence of this or related to it).  This needs to be fixed and then I think we are ready to merge this branch into master.

jmmans:
EventHeader issue fixed, tested.  I think we're done.

JeremyMcCormick:
With current HEAD of iss114, the first event header is not filled correctly...

```
[ Process ] :  Processing 1 Run -1 Event -1  (2017-01-23 12:53:49.425409000-0800) [ EcalDigiProducer ] : Got 84 ECal hits in event -1
```

Subsequent ones look fine e.g. 

```
[ Process ] :  Processing 2 Run 0 Event 1  (2017-01-23 12:53:49.427703000-0800) [ EcalDigiProducer ] : Got 93 ECal hits in event 1
```

We should fix this before merging, if possible.

JeremyMcCormick:
The timestamp from event header also does not look correct.  It seems to be set from the current time with resolution of nanoseconds during event processing e.g.

[ Process ] :  Processing 1 Run -1 Event -1  (2017-01-23 15:55:53.068369000-0800)                          
[ EcalDigiProducer ] : Got 84 ECal hits in event -1                                                        
[ Process ] :  Processing 2 Run 0 Event 1  (2017-01-23 15:55:53.071661000-0800)                            
[ EcalDigiProducer ] : Got 93 ECal hits in event 1                                                         
[ Process ] :  Processing 3 Run 0 Event 2  (2017-01-23 15:55:53.072115000-0800)                            
[ EcalDigiProducer ] : Got 96 ECal hits in event 2                                                         
[ Process ] :  Processing 4 Run 0 Event 3  (2017-01-23 15:55:53.072536000-0800)                            
[ EcalDigiProducer ] : Got 91 ECal hits in event 3                                                         
[ Process ] :  Processing 5 Run 0 Event 4  (2017-01-23 15:55:53.072973000-0800)                            
[ EcalDigiProducer ] : Got 92 ECal hits in event 4                                                         
[ Process ] :  Processing 6 Run 0 Event 5  (2017-01-23 15:55:53.073367000-0800)                            
[ EcalDigiProducer ] : Got 107 ECal hits in event 5

jmmans:
The timestamp shown in the log is NOT the time from the event header.  It is the time while the processing is ongoing, so that one can tell how long the event processing is taking.

JeremyMcCormick:
From what I can tell in rootbrowse, it looks like the timestamp from the simulation is not being stored correctly.  The run should span over many seconds but I see only 1 bin filled.  (I can look into this further in the ROOT terminal if needed to confirm.)

JeremyMcCormick:
I wonder if the timestamp print out here *should* be from the event header and we should instead have a "chrono" Analyzer class which the user can add to their job instead for event processing times.  I think it is confusing, because this time is printing right next to the event header info (e.g. event and run numbers) so it looks like it goes along with it.

jmmans:
The timestamp should be ok in the file except for the first-event problem, where the time seconds are very small.

JeremyMcCormick:
The first event issue looks fixed now.

Confirmed timestamps look fine with a debug print in event processing.

Going to merge now!

Issue #140, opened 2017-01-21
Add a mainpage to the Doxygen doc.


JeremyMcCormick:
This can just point to the ldmx-sw github page where the primary documentation is located.

JeremyMcCormick:
This will be resolved by PR #142.

PR #147, opened 2017-01-26
Support hosting of Doxygen doc on Github Pages

JeremyMcCormick:
This PR makes changes to how the Doxygen doc is generated so it can be hosted easily on Github Pages.

- The documentation is generated into the docs directory within the source tree.

- The generated doc files were checked into the repository.

- There is a docs/index.html page that redirects to html/index.html which is the documentation index.

- Moved the Doxygen config from config/doxygen to the docs dir.

This should allow us to easily host our Doxygen doc on Github Pages by pointing it to the docs directory in the repository.


JeremyMcCormick:
I will resolve issue #140 on this branch too (it is a minor change).

JeremyMcCormick:
I turned the generation of the doxygen doc off by default.  I think this should be ready for merging now.

PR #155, opened 2017-02-04
Iss138

awhitbeck:
This PR adds a specialized class for hcal detector IDs and adds noise hits to all hcal cells.  

The new hcal detector ID class specifies the bit assignments for hcal subdetectors (back, wrap_top, wrap_bot, wrap_left, wrap_right) and implements functionality to generate a list of all detector cells in the from or map<int,int> and map<int,float> -- the key of the maps is the raw detID.  

To test this, one can generate some sim data and run this configuration file with ldmx-app:

https://github.com/LDMXAnalysis/ldmx-sw/blob/iss138/EventProc/python/hcalDigis.py

### DIFF
--- a/DetDescr/include/DetDescr/EcalDetectorID.h
+++ b/DetDescr/include/DetDescr/EcalDetectorID.h
@@ -9,6 +9,7 @@
 
 // LDMX
 #include "DetDescr/DefaultDetectorID.h"
+#include "DetDescr/EcalHexReadout.h"
 
 namespace ldmx {
 
@@ -18,12 +19,16 @@ namespace ldmx {
  */
 class EcalDetectorID : public DefaultDetectorID {
 
+    private:
+        EcalHexReadout* hexReadout_{nullptr};
+
     public:
 
         /**
          * Adds a cell field and re-initializes the ID.
          */
         EcalDetectorID() {
+            hexReadout_ = new EcalHexReadout();
             this->getFieldList()->push_back(new IDField("cell", 2, 12, 31));
             init();
         }
@@ -35,6 +40,8 @@ class EcalDetectorID : public DefaultDetectorID {
         int getCellID() {
             return this->getFieldValue(2);
         }
+
+        std::map<int,float> getMap();
 };
 
 }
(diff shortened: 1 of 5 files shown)


jmmans:
Similar support is still needed for ECAL... :-)

JeremyMcCormick:
Thanks Andrew.  I'll look this over on Monday.

awhitbeck:
I tried to also address this for the EcalDigiProducer.  I have noted some things below that I either changed or think we should change below:

- switching definition of amplitude and energy -- I think amplitude should always include noise, but I am not so sure we have a consistent way of defining energy
- adding EcalHexReadout to EcalDetectorID as data member
- adding getMap function to EcalDetectorID
- timing of EcalHit is set to time of last simHit? -- This doesn't sound correct
- dropping print statement

JeremyMcCormick:
Minor style correction:

In DetDescr/include/DetDescr/HcalDetectorID.h can you name your private member variables using camelcase convention and not underscores to match the rest of the code?

Also, static variables should be all caps separated by underscores.

JeremyMcCormick:
Also, you should delete `hexReadout_` in the destructor of `EcalDetectorID` to avoid a memory leak.

Alternately you could define the `EcalHexReadout` as a non-pointer.

JeremyMcCormick:
The method `HcalDetectorID::getMap()` is actually creating a new map, so you might want to name it `createMap` as `get` implies it is just returning an existing value.  It also isn't entirely clear what the map is for just from the method name so maybe `createReadoutMap` would be more clear.

JeremyMcCormick:
You do not need the methods `int HcalDetectorID::getSubDet()` or `int HcalDetectorID::getLayer()` as these are already methods on the `DefaultDetectorID` class you are extending.

JeremyMcCormick:
It may not make sense to have `void setNumLayers(int num_back_hcal_layers_, int num_wrap_hcal_layers_);` on `HcalDetectorID` because this doesn't seem like something that should be changed once this object is created.  These values would be based on the geometry and should be fixed for the analysis job.  (If it is more convenient to have this you can leave it but likely it is not something that a user should have access to in a public interface.)

JeremyMcCormick:
I'm closing this for now.  Please re-open this later or make a new PR for this issue.

Issue #156, opened 2017-02-05
Fieldmap offset is incorrect.


omar-moreno:
Currently, the field is offset by -10 cm.  Instead the field should be offset by -40 cm.

omar-moreno:
Attached is a comparison between the old fieldmap with the -100 mm offset and the new one with the -400 mm offset. 

![20170204_fieldmap_comparisons](https://cloud.githubusercontent.com/assets/3708508/22624425/c7e5f22e-eb30-11e6-80cd-75502f2e8632.png)

JeremyMcCormick:
This should be resolved now with PR #157 which was merged.

PR #157, opened 2017-02-05
Apply correct offset to the fieldmap.

omar-moreno:


omar-moreno:
See #156 for a plot showing the old and new fields.

PR #198, opened 2017-03-01
Iss183

ocolegro:


JeremyMcCormick:
This branch is not in a mergeable state.  We'll instead merge these changes with fixes from the iss190 branch.

PR #209, opened 2017-03-07
Enhancements for saving of tracks to SimParticle collection.

JeremyMcCormick:
- Added a new SimCore module which for now only has a UserTrackInformation class with a save flag.

- In UserTrackingAction, created trajectories in post-tracking for flagged tracks so they are automatically persisted.

- Corrected the momentum and energy calculations in the Trajectory class and in the constructor added a trajectory point for tracks that have already been stepped so their endpoint is available.

- Added SimCore module dependency to SimPlugins for access to the UserTrackInformation (new module is also a dep of SimApplication now, too).

- Linear search of trajectory container is replaced by a map created during event processing for optimization.

- Add an example user sim plugin called TrackSaver which flags all the tracks in the event so they are saved.

- Confirmed that simulation output collections look reasonable when running ldmx-sim in "normal" mode and with TrackSaver enabled (checked persistency manager output with verbose level 3).

- Added the TrackProcessSaver sim plugin which can be used to save tracks created by specific physics processes.

- Added process type to Trajectory class, which is copied from value of `G4VProcess::GetProcessSubType()`.

- Added process type to SimParticle class which is copied from the Trajectory.

- Incremented SimParticle ClassDef version to 3.

This new feature can be used to find PN reaction products by checking for process type 121 on the SimParticle.

Saving tracks by process can be activated using a macro like the following:

```
/ldmx/plugins/load TrackProcessSaver
/ldmx/plugins/TrackProcessSaver/verbose 2
/ldmx/plugins/TrackProcessSaver/addProcess photonNuclear
```

This PR replaces #204 and #208.

**The change to the SimParticle class makes this backward incompatible with existing data because of changes to the event model.**

### DIFF
--- a/Event/src/SimParticle.cxx
+++ b/Event/src/SimParticle.cxx
@@ -51,7 +51,8 @@ void SimParticle::Print(Option_t *option) const {
             "momentum: ( " << px_ << ", " << py_ << ", " << pz_ << " ), " <<
             "mass: " << mass_ << ", " <<
             "nDaughters: " << daughters_->GetEntries() << ", "
-            "nParents: " << parents_->GetEntries() <<
+            "nParents: " << parents_->GetEntries() << ", "
+            "processType: " << processType_ <<
             " }" << std::endl;
 }
 


JeremyMcCormick:
TODO: In Trajectory constructor need to get vertex kinetic energy + mass.

JeremyMcCormick:
TODO: Make string comparison a "contains" rather than match so don't need two processes for biased.

JeremyMcCormick:
TODO: Fix SimParticle persistency so old data can be read.

JeremyMcCormick:
TODO: Add particle endpoint momentum to resolve #240.

PR #226, opened 2017-03-14
Add HcalVetoProcessor.

omar-moreno:


awhitbeck:
I know you're probably trying to just get this in, but probably we should make the cut value configurable.

JeremyMcCormick:
I ran ldmx-app using `hcalVeto = ldmxcfg.Producer("hcalVeto", "ldmx::HcalVetoProcessor")` and it ran fine.  The veto doesn't pass most of the time (which I assume is expected).

We should make the number of PE a configurable parameter.  Then I think this can be merged.

omar-moreno:
What kind of events did you process?  It's strange that the veto doesn't
pass ``most of the time''.

JeremyMcCormick:
Most events do not have any activity in the HCal so you would expect all those to fail the veto.

omar-moreno:
pass the veto.

On Wed, Mar 15, 2017 at 11:46 AM, Jeremy McCormick <notifications@github.com

JeremyMcCormick:
Oh right, I had that flipped around....could just have been the events I was running (I think they were biased PN events).

omar-moreno:
@awhitbeck Hcal veto is now configurable.  @JeremyMcCormick This is ready to be merged.

PR #233, opened 2017-03-21
Configuration updates and improvements.

JeremyMcCormick:
Configuration updates from #232 issue.

Also resolves #152.

Merge this before #234.

Issue #232, opened 2017-03-20
Improvements to configuration.


JeremyMcCormick:
- Change LSF setup script so it automatically sources the env setup.

- Copy Configuration/data/* to data/config installation dir.

- Install detectors to data/detectors dir.

JeremyMcCormick:
Resolved by PR #233.

PR #234, opened 2017-03-21
Add field map as sub-module.

JeremyMcCormick:
Adds MagFieldMap sub-module that is cloned during installation if INSTALL_FIELDMAP CMake option is set to ON.

B-field data is stored as a .tar.gz file in the repo to get around git's 100 MB file size limit.

All dat files are unzipped and copied to data/fieldmap during installation (currently just one file).

Merge this after #233 as the #232 branch was merged into this one to avoid conflicts in several files.

### DIFF
--- a/scripts/run_ldmx_sim.py
+++ b/scripts/run_ldmx_sim.py
@@ -7,9 +7,6 @@ import os
 import getpass
 import subprocess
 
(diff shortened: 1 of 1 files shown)


JeremyMcCormick:
Replaced by PR #238.

PR #236, opened 2017-03-21
Check for dependencies in top level CMake file.

JeremyMcCormick:
Also adds XercesC dep check requiring setting of XercesC_DIR or XercesC_LIBRARIES and XercesC_INCLUDE_DIR in CMake arguments.

Merge this after PR #233 and #234.


JeremyMcCormick:
Replaced by PR #238.

PR #265, opened 2017-04-06
Implement flexible plugin framework for track filtering and saving

JeremyMcCormick:
Implements a framework for making filters for saving tracks.

The following criteria can be applied as filters for track saving:

- KE at track vertex

- PDG code or name of particle

- Creator physics process

- Creator physics process of daughter particles

- Name of detector region

- Name of geometric logical volume

- Saving of all tracks

Aside from the KE threshold and "save all" filters, the filters can be composed of one or more item,s e.g. one or more PDG code, multiple physics processes, etc.  Only one of these needs to pass for the filter to return true.

Filters are combined together into filter chains.  All filters in a chain must pass for the processed track to be flagged for saving by the plugin.

Multiple filters may be active per event and they are run separately, so if any filter chain passes a track then it will be flagged for saving.

Filter chains may be activated in the pre or post tracking hooks (post is the default).

There is an extensive set of Geant4 macro commands for creating and configuring the filters and making the filter chains.  (Guidance has been implemented so usage should be fairly self-explanatory.)

```
Idle> ls                                                                      
Command directory path : /ldmx/plugins/TrackFilterPlugin/                     


Guidance :
Commands for the sim plugin TrackFilterPlugin

 Sub-directories : 
 Commands :        
   verbose * Set the verbosity level of the sim plugin (1-4).
   threshold * Set kinetic energy threshold at track vertex  
   pdgid * Filter on PDG ID code                             
   particle * Filter on the name of a particle               
   process * Filter on creator physics process name          
   parent * Filter on creator physics process of daughter tracks
   region * Filter on name of detector region                   
   volume * Filter on name of logical volume at track vertex    
   pass * Filter that always passes                             
   create * Create a filter chain from the current set of filters
   print * Print out filter chain information                    
   action * Set the tracking action ('pre' or 'post')          
```

Here are a few examples...

Save all tracks in the event:

```
/ldmx/plugins/TrackFilterPlugin/pass
/ldmx/plugins/TrackFilterPlugin/create SaveAllTracks
```

Save neutrons above 5 MeV KE coming from PN reactions:

```
/ldmx/plugins/TrackFilterPlugin/threshold 5 MeV
/ldmx/plugins/TrackFilterPlugin/pdgid 2112
/ldmx/plugins/TrackFilterPlugin/process photonNuclear true
/ldmx/plugins/TrackFilterPlugin/action pre
/ldmx/plugins/TrackFilterPlugin/create NeutronFilter
```

Save all parents of PN secondaries:

```
/ldmx/plugins/TrackFilterPlugin/parent photonNuclear true
/ldmx/plugins/TrackFilterPlugin/action post
/ldmx/plugins/TrackFilterPlugin/create PNParentFilter
```

Save all tracks originating in the calorimeters:

```
/ldmx/plugins/TrackFilterPlugin/region CalorimeterRegion
/ldmx/plugins/TrackFilterPlugin/action pre
/ldmx/plugins/TrackFilterPlugin/create CalFilter
```

This should probably be tested rather extensively as it is a fairly complex sub-framework.  (I have done some basic testing and will continue to do so.)

Please let me know if you have any questions/comments about the implementation.


JeremyMcCormick:
TODO: Add detailed doxygen doc with instructions, examples, etc. to TrackFilterPlugin class

JeremyMcCormick:
TODO: Add filter for saving on track ID. (e.g. when reseeding and you see a track that you want to investigate)

JeremyMcCormick:
> Does this create a single filter that is configurable via a macro, or is it creating multiple filters? 

You can create as many filter chains as you want within your Geant4 macro.  They are all run independently.  Essentially all the individual filters in a chain must pass but the filter chains themselves are OR'd together.  Hope that answers your question.

> Do all tracks get passed through the various filters? 

Yes, all tracks are passed through every filter.  I don't see how it would work otherwise.

> How much overhead is involved when doing this? 

Not much.  I don't see significant slow-down from running these filters, and most of the actual filter checks are one-liners or only a few LoC.

> Also, 'I don't see much of a difference' is not an acceptable answer :).

I apologize, but the context and relevance of this particular statement you've quoted escapes me right now.  :)

omar-moreno:
'I don't see much of a difference' ---> "I don't see significant slow-down from running these filters".  

Can you run 10,000 events using your example macro and time them?  Also run 10,000 events without the filtering so we have something to compare against.  I just want to make sure we aren't significantly slowing things down.

JeremyMcCormick:
> Can you run 10,000 events using your example macro and time them? Also run 10,000 events without the filtering so we have something to compare against.

Run time of 1000 events with filtering:

```
real    20m22.330s
user    19m28.434s
sys     0m3.547s
```

And without filtering:

```
real    18m59.336s
user    18m20.189s
sys     0m2.315s
```

So it is about a 10% slow down.

I ran with these two filter chains:

```
# save neutrons above 5 MeV KE from PN reactions
/ldmx/plugins/TrackFilterPlugin/threshold 5 MeV
/ldmx/plugins/TrackFilterPlugin/pdgid 2112
/ldmx/plugins/TrackFilterPlugin/process photonNuclear true
/ldmx/plugins/TrackFilterPlugin/action pre
/ldmx/plugins/TrackFilterPlugin/create NeutronFilter

# save all parents of PN secondaries
/ldmx/plugins/TrackFilterPlugin/parent photonNuclear true
/ldmx/plugins/TrackFilterPlugin/action post
/ldmx/plugins/TrackFilterPlugin/create PNParentFilter
```

I don't consider that too significant, and there is not built-in overhead to the simulation because this is all done with a plugin which is only optionally activated.

PR #270, opened 2017-04-14
Add end point momentum to SimParticle

JeremyMcCormick:


JeremyMcCormick:
I can see in the output that there are some zeroes flagged as negative:

`endPointMomentum: ( -0, 0, -0 )`

I don't know if this is an issue or not for analysis code.  

If it needs to be fixed, then let me know.

omar-moreno:
Is this the expected Geant4 behavior? I want to make sure it's not
indicative of an underlying issue.

JeremyMcCormick:
> Is this the expected Geant4 behavior? I want to make sure it's not indicative of an underlying issue.

I was getting negative zeroes directly from the values in the G4Track, so that is Geant4's doing.  

I pushed an update so that the setter method of Trajectory flips the sign off for them, which seems to work fine.  I double checked this in the output using a test processor.

I doubt it is anything to worry about in general but I did clean it up for our own output/analysis.

omar-moreno:
"I doubt it is anything to worry about in general." :)

PR #275, opened 2017-05-09
beamspot smearing, some first tools for doing pileup

nhanvtran:


nhanvtran:
@JeremyMcCormick 
I removed the static inputs for the PrimaryGeneratorAction but actually I didn't find a clean way to do the same for the MultiParticleGunGenerator so I'm left with these lines.  
https://github.com/LDMXAnalysis/ldmx-sw/blob/iss268/SimApplication/src/PrimaryGeneratorMessenger.cxx#L52-L53

Let me know if you have a good idea.  I can continue to customize the MultiParticleGunGenerator if you want, right now it's a skeleton.  Also, in the meantime, I'll work on documenting our various primary generators in the Twiki section

JeremyMcCormick:
I would just have set methods for all the gen parameters on UserPrimaryGeneratorAction which are called by the messenger to update it.  I think it is fine if some of them are unused by any particular generator.  Then they can be easily accessed rather than needing to go through some static interface on a messenger class.

nhanvtran:
Ok, I'll try to fix this in the next PR where the pileup is more usable

nhanvtran:
Actually @JeremyMcCormick I'm not sure how you are proposing to do this.  There is no UserPrimaryGeneratorAction is there?  Do you mean PrimaryGeneratorAction?  Then how does the G4PrimaryGenerator access that information?

PR #279, opened 2017-05-23
Update 4 GeV electron macro. 

omar-moreno:


omar-moreno:
I just realized that this has the changes from iss170b merged into it. I wont merge this in until I can confirm that the event headers can be loaded.

PR #305, opened 2017-07-13
Iss134: Implement basic geometry API

JeremyMcCormick:
Implements the following additions and changes:

- Adds basic `DetectorElement` class with `DetectorElementImpl` implementation

- Provides a `DetectorDataService` which loads exported full GDML files into ROOT, loads them into the `TGeoManager` and automatically sets up the detector element hierarchy

- Provides detector element implementations for all major components and subsystems

- Adds a test `DefaultDetectorDataService_test` which loads a detector and prints out information from the detector elements

- Adds simple script for exporting GDML to single output file for ROOT, stripping the 'magnet.gdml' module automatically from the output

- Adds `GDMLMessenger` for exporting geometries without pointers in names

- Adds ability to exclude specific GDML modules when reading in detector files

- Resolves GDML module files by reading them from the detector data dir if they are not found from the current dir

- Resolves field map files from the fieldmap data dir if they are not found from the current dir

- Adds a test processor that matches ECal sim hits with their sensors which can be run using the config at `Configuration/python/det_id_check.py`


JeremyMcCormick:
I already know one thing I need to revert.  I had changed the SD classes to use the volume copynumber values instead of the SubdetID from the auxiliary data in userinfo for setting the subdetector IDs, but this is not backward compatible with existing detector models.  So I'll go ahead and revert it.

omar-moreno:
What is the point of the test detector?

JeremyMcCormick:
"What is the point of the test detector?"

It was for testing my code while v3 was still being worked on but I deleted it now.

JeremyMcCormick:
TODO:

- Change LDMX GDML macro command dir to `/ldmx/gdml` for simplicity and update macros

- Add doc and doxygen tags

- Cleanup `GDMLReadStructure` by moving method implementations to .cxx file

- Verify recon and analysis output on the `v3-fieldmap` and `v3-fieldmap-magnet` detectors

EventProc/src/DetIDAnalyzer.cxx
```diff
@@ -0,0 +1,178 @@
...
+
+                    hcalID_->setRawValue(chit->getID());
+                    auto hitID = hcalID_->unpack();
+                    std::cout << "Hcal hit: " << hitID << std::endl;
+
+                    // Remove strip ID so the lookup will work.
+                    hcalID_->clear();
+                    hcalID_->setFieldValue(0, hitID[0]);
+                    hcalID_->setFieldValue(1, hitID[1]);
+                    ecalID_->setFieldValue(2, hitID[2]);
```
nhanvtran: trivial typo?

nhanvtran:
Why are you adding the `detector_full.gdml`? what are those for?

nhanvtran:
I checked basic functionality, it looks good.  I have not done an extensive review of how it maps to the expected geometry behavior...but to what extent should this be checked?  if it's working now, should we merge it in so that we don't have to fix the conflicts constantly? then do more checks once in the master? (it's a standalone functionality)

omar-moreno:
​I would say yet, however, I don't want this to become a distraction in the
case that it breaks something.  There is a history of this happening ...​

JeremyMcCormick:
The recon and sim are both confirmed to work fine with this branch.  What more do you want?

omar-moreno:
Check that the output matches the master ... "running fine" is never going
to be enough of a check.

JeremyMcCormick:
"Check that the output matches the master"

@nhanvtran and I both confirmed the recon is the same.

omar-moreno:
Plots?

omar-moreno:
If this is up to date, I'm going to go ahead and merge it in.  I'm going to start another round of debug in preparation for target photo-nuclear so I should be able to catch any issues if bugs are introduced. @JeremyMcCormick Let me know if I should hit the merge button.

JeremyMcCormick:
Just give me to the early afternoon for additional testing (everything
working to my knowledge but would like to double check).

PR #307, opened 2017-07-15
Revise SimParticle process type to use custom enum list instead of G4 process sub-type

JeremyMcCormick:
I verified that the values are being set correctly in Trajectory for processes in the enum list, but the ROOT output needs to be checked.

### DIFF
--- a/Event/include/Event/SimParticle.h
+++ b/Event/include/Event/SimParticle.h
@@ -13,6 +13,7 @@
 
 // STL
 #include <vector>
+#include <map>
 
 namespace ldmx {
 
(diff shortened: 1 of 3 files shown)


ASPatterson:
Looks good. Before and After attached for unbiased + SaveAllTracks, using identical tracks. Only oddity is initial electron ...

Using [some processtype table](https://twiki.cern.ch/twiki/bin/view/CMSPublic/SWGuideMCTruth#Geant4_process_numbering_scheme), these are confirmed to work:
Ionization (was 2, now 6)
brem (was 3, now 5)
photoelectric (was 12, now 8)
Compton (was 13, now 2)
gamma->2e (was 14, now 3)
positron annihil (was 5, now 1)
ill-defined-nuclear (was 121, now 9)

I didn't check our new enum values 4, 7, or 10 (electronNuclear, multiple scattering, gamma->2mu).

And it quickly uncovered a case of where "121" failed! (see tails of Before and After).



[after.txt](https://github.com/LDMXAnalysis/ldmx-sw/files/1150028/after.txt)
[before.txt](https://github.com/LDMXAnalysis/ldmx-sw/files/1150029/before.txt)

Issue #254, opened 2017-03-31
Ability to select only a subset of events for recording to the output file


jmmans:
Add the ability for EventProcessor modules to indicate if a given event should be written to the file or not.

In the CMS framework, this is done by a new subclass of EventProcessor, the "Filter" which is also used in the software trigger.  For this to work, we'd need to reconsider the event processing from being a single linear sequence to being a set of multiple parallel sequences -- basically a tree of event processors.  I think this might be overkill/inappropriate for this purpose.

Instead, I propose to add to the EventProcessor a call which can be made to mark that the given event should or should not be kept.  Then, the framework python can contain the information as to which event processors should be consulted (and whether to apply an "OR" or "AND" logic between them).

omar-moreno:
This was resolved by PR 7dbb94d0cb8b148d7bf3e9841d6172568f505030.

PR #310, opened 2017-07-19
Skimming support for ldmx-app

jmmans:
This PR resolves issue #254 , allowing any event processor to indicate that a given event should be kept or dropped.  The user can specify in the python file which event processors are to listened to, and the framework will keep only the specified set of events in the output file.

### DIFF
--- a/Framework/src/EventFile.cxx
+++ b/Framework/src/EventFile.cxx
@@ -117,7 +117,7 @@ namespace ldmx {
                 if (storeCurrentEvent) tree_->Fill(); // fill the clones...
             }
             if (event_) {
-            event_->Clear();
+                event_->Clear();
                 event_->onEndOfEvent();
             }
         }
--- a/Framework/src/EventProcessor.cxx
+++ b/Framework/src/EventProcessor.cxx
@@ -18,7 +18,7 @@ namespace ldmx {
     }
 
     void EventProcessor::setStorageHint(ldmx::StorageControlHint hint, const std::string& purposeString) {
-    process_.getStorageController().addHint(name_,hint,purposeString);
+        process_.getStorageController().addHint(name_,hint,purposeString);
     }
   
     TDirectory* EventProcessor::getHistoDirectory() {


nhanvtran:
Going through this now @jmmans 
It would be nice to add a few examples to this PR.  
I'm not sure if they should go in `EventProc/python` or `Configuration/python` -- @JeremyMcCormick   @omar-moreno what's the agreed convention here?  

Based on a first reading, I guess it's some examples of `p.skimRules = []`.  I'm testing it now and can put some examples in myself, but @jmmans guidance would be welcome.

omar-moreno:
I would say put them in EventProc/python.

JeremyMcCormick:
Full example configs should go in Configuration/python .

The EventProc/python scripts contain classes to be instantiated, not actual
runnable examples.

jmmans:
Added a sample python file, ready for merge!

nhanvtran:
Did a functionality test, looks good.  I'd say this is ready to merge.  However, @jmmans, it would also be nice to add some documentation on a bunch of different skim rules -- not obvious to the user all the possible options...

JeremyMcCormick:
@jmmans The indentation level in the code you've changed is mostly just wrong, and there are tabs in it as well.  I will fix it this time, but please in the future use spaces only and match the existing indentation.  (There's a bunch of methods you've added here whose indentation doesn't even line up with the existing ones.)

jmmans:
@JeremyMcCormick : I would certainly appreciate a working solution to meet the tabbing requirements.  I have tried a few things in emacs to acheive this, but they have all been a failure -- that how the tabs are appearing.

Jeremy

JeremyMcCormick:
You can put this into your `.emacs` file:

```
(setq-default indent-tabs-mode nil)
```

And set the tab width using the variable:

```
tab-width
```  

If worse comes to worse, then hit the spacebar instead of the tab key.  :-)

JeremyMcCormick:
The indentation should be fixed now and I exterminated all the tabs.

@nhanvtran  and I were wondering if there could be more complex examples added as the sample one seems to only test the most basic functionality (dropping/keeping via modulus rule).

For instance, I assume you tested the mechanism of the drop/keep consensus amongst processors through their "votes" but we don't have a good way to test this or other more complex rules to know that it actually works.  Probably that should all be tested before this is merged.

jmmans:
I've tested the basic support of allowing "any" to keep or reject.  The other code is there to build on for the future.  We need this for operations purposes!

PR #320, opened 2017-07-24
Iss313

arickman:
Enable Electronuclear biasing


omar-moreno:
Let's leave this code on a branch until it is tested further.

PR #333, opened 2017-08-02
Add public getters for recoil momentum and position.

omar-moreno:


GavinNiendorf:
I have verified these changes locally, everything works.

PR #343, opened 2017-08-06
Noise hit generator.

omar-moreno:


JeremyMcCormick:
Correct me if I'm wrong, but I don't think this class belongs in this package.  SimCore is for small utilities used by SimApplication.  This appears to be a utility class that would be used in post-processing of the simulation data to add noise hits, so it should go into EventProc.

PR #347, opened 2017-08-09
Generate an event from the scoring plane trajectories

nhanvtran:
.mac for testing in SimApplication/data/g4macros/generators/mu_gun*

There are 2 steps: 
step1, muon gun shoots through v3 geom with scoring planes
step2, find the muon exiting the back of the ecal scoring plane and regenerate that

This also fixes the bug (in `UserActionPlugin.cxx`) so that it always saves particles with gen status == 1.

### DIFF
--- a/SimApplication/data/g4macros/generators/e_gun_rootseed_step2.mac
+++ b/SimApplication/data/g4macros/generators/e_gun_rootseed_step2.mac
@@ -2,6 +2,7 @@
 /run/initialize
 
 /ldmx/generators/root/open test-input.root
+/ldmx/generators/root/mode/regenerate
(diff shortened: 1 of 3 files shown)


omar-moreno:
@nhanvtran Did you check that you get the same Hcal information back after you regenerate?

nhanvtran:
ah darn, will try to do tonight

omar-moreno:
Great, thanks!

nhanvtran:
hi all finally did this check...no problems.  events are the same after regenerating, we're good to go here i think

omar-moreno:
Alright, I'll merge it in.

PR #367, opened 2017-08-15
Iss366: fix non-unique tracker IDs

JeremyMcCormick:
- Add `TrackerID` class that includes module number

- Introduce new ID scheme in SDs for trackers where `layer = copyNum / 10` and `module = copyNum % 10`

- Update copy numbers in GDML files to follow new ID scheme


omar-moreno:
I reverted the changes to the size of the recoil tracker envelope in order to avoid overlaps with the magnet bore. Overlap check shows no overlaps.

PR #369, opened 2017-08-16
Iss368

omar-moreno:
The `HcalDigiProducer` should be double checked on this branch because of changes to the ID.  It appears that mostly the raw ID values are being used and the output should not change but it should be verified.

### DIFF
--- a/Event/include/Event/HcalHit.h
+++ b/Event/include/Event/HcalHit.h
@@ -54,6 +54,16 @@ namespace ldmx {
                 return pe_;
             }
 
+            /** Decode the section associated with the hit from the ID. */
+            int getSection() const {
+                return (getID() & 0x7000) >> 12;
+            }
+
+            /** Decode the strip associated with the hit from the ID. */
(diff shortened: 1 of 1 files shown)


nhanvtran:
I guess it would be good to access the subdetector (back, top, bottom, right, left) from the hit collection -- right now there's only a `getLayer()` method.  you can get the raw ID but you cannot unpack it in standalone analysis becasue the `HcalID` class is not wrapped

nhanvtran:
by the way, just putting down the conversation with @JeremyMcCormick for posterity...if we can do the wrapping of the ldmx classes in parallel to the production that also works.  Or else can do the bitshifting (as @JeremyMcCormick outsie of ldmx-sw)

omar-moreno:
Adding the methods would change the event model so I say we put them in
now.  It's not that difficult to do, so might as well.

@nhanvtran would you want a method for all information packed into the ID?
i.e. getSection(), getStrip() ...

PR #379, opened 2017-08-25
Iss378

bechenard:
Ready to merge

### DIFF
--- a/EventProc/src/HcalDigiProducer.cxx
+++ b/EventProc/src/HcalDigiProducer.cxx
@@ -135,7 +135,7 @@ namespace ldmx {
         } 
         
         
-        event.add("HcalDigis", hits_);
+        event.add("hcalDigis", hits_);
     }
 
 }
--- a/Configuration/python/sample_recon_Hcal.py
+++ b/Configuration/python/sample_recon_Hcal.py
@@ -1,0 +1,41 @@
+#!/usr/bin/python
+
+import sys
+
+# we need the ldmx configuration package to construct the object
+from LDMX.Framework import ldmxcfg
+
+# first, we define the process, which must have a name which identifies this
+# processing pass ("pass name").
+p=ldmxcfg.Process("digi")
+
+# Currently, we need to explicitly identify plugin libraries which should be
+# loaded.  In future, we do not expect this be necessary
+p.libraries.append("libEventProc.so")
+
+# load the template hcalDigis configuration from its python file
+from LDMX.EventProc.hcalDigis import hcalDigis
+
+# switch to Strip aggregation (for testing)
+hcalDigis.parameters["doStrip"] = 1
+
+# load the template ecalDigis configuration from its python file
+from LDMX.EventProc.ecalDigis import ecalDigis
+
+# load the template ecalDigis configuration from its python file
+from LDMX.EventProc.simpleTrigger import simpleTrigger
+
+
+# Define the sequence of event processors to be run
+p.sequence=[ecalDigis, hcalDigis, simpleTrigger ]
+
+# Provide the list of input files to run on
+p.inputFiles=["output.root"]
+
+# Provide the list of output files to produce, either one to contain 
+# the results of all input files or one output file name per input file name
+p.outputFiles=["ldmx_digi_events.root"]
+
+# Utility function to interpret and print out the configuration to the screen
+p.printMe()
+
--- a/EventProc/python/hcalDigis.py
+++ b/EventProc/python/hcalDigis.py
@@ -5,8 +5,8 @@ from LDMX.Framework import ldmxcfg
 hcalDigis = ldmxcfg.Producer("hcalDigis", "ldmx::HcalDigiProducer")
 
 # set the mean noise in PE units
-hcalDigis.parameters["meanNoise"] = 2.0
+hcalDigis.parameters["meanNoise"] = 1.5
 hcalDigis.parameters["mev_per_mip"] = 1.4
 hcalDigis.parameters["pe_per_mip"] = 13.5
-hcalDigis.parameters["doStrip"] = 1
+hcalDigis.parameters["doStrip"] = 0
 


JeremyMcCormick:
I don't think there should be a separate HcalStripHit class.  The existing HcalHit class should likely be modified with your changes.  It should work fine as long as the class version is incremented.  Was there some good reason not to do this?

bechenard:
HI Jeremy,

Sorry for the delayed reply, I don't check that mailbox often. Since Nhan
wanted to keep the old scheme to cross-validate the results, we decided to
go with separate products and modules for the time being. The trick is that
it is annoying to recover HcalHits from HcalStripHits as we randomly sample
the background in the strips or full plane.

We will likely replace the HcalHit with the HcalStripHit in the future
(when Nhan is happy).

Is that ok for you?
Cheers,
Bertrand

nhanvtran:
Confirming...

EventProc/src/HcalStripDigiProducer.cxx
```diff
@@ -0,0 +1,141 @@
...
+#include "DetDescr/HcalID.h"
+#include "Event/HcalStripHit.h"
+
+namespace ldmx {
+
+    const float HcalStripDigiProducer::FIRST_LAYER_ZPOS = 569.5;
+    const float HcalStripDigiProducer::LAYER_ZWIDTH = 60.;
+    const int HcalStripDigiProducer::NUM_HCAL_LAYERS = 15;
+    const float HcalStripDigiProducer::MEV_PER_MIP = 1.40;
+    const float HcalStripDigiProducer::PE_PER_MIP = 13.5 * 6. / 4.;
```
nhanvtran: So looking back at old talks, this should be 13.5 not 13.5 * 1.5...can we fix this and also in HcalDigiProducer.cxx?

SimApplication/src/HcalSD.cxx
```diff
@@ -43,23 +44,43 @@ namespace ldmx {
...
         int copyNum = aStep->GetPreStepPoint()->GetTouchableHandle()->GetHistory()->GetVolume(layerDepth_)->GetCopyNo();
-        int layer = copyNum % 1000;
         int section = copyNum / 1000;
+        int layer = copyNum % 1000;
+
+        //stripID: back Hcal, segmented along y direction for now every 10 cm -- alternate x-y in the future?
+        //         left/right side hcal: segmented along x direction every 10 cm
+        //         top/bottom side hcal: segmented along y direction every 10 cm
+        int stripID = -1;        
+        if (section==HcalSection::BACK) stripID = int( (localPosition.y()+scint->GetYHalfLength())/100.0);
```
nhanvtran: Isn't this 1cm strips, not 10cm strips?

nhanvtran:
General comment: this will break the old layer-wise HCAL veto. 
The general scheme is sum over all simhits and aggregate them in the same detIDraw.  Then for a given detIDraw, make a hit.  Once we add the strips into the detIDraw, then one cannot do the same thing for layers.  

Should we update the layer-wise HCAL veto at the same time?

awhitbeck:
@nhanvtran I think we should make the changes to the old digi producer in this PR.  I think the simplest fix is to change `detIDraw` to `layer` for the map keys.   But I think this just effects the digi producers, not the veto processor.

@bechenard, I have a few minor comments for the `HcalStripDigiProducer`, some also apply equally to the `HcalDigiProducer`, which I can take care of, if you want:

 - This is wrong and probably should be removed:
https://github.com/LDMXAnalysis/ldmx-sw/blob/iss378/EventProc/src/HcalStripDigiProducer.cxx#L19-L21

 - Change layer to strip in comment: 
https://github.com/LDMXAnalysis/ldmx-sw/blob/iss378/EventProc/src/HcalStripDigiProducer.cxx#L83

 - Change var name to something more vague, or change layer to strip:
https://github.com/LDMXAnalysis/ldmx-sw/blob/iss378/EventProc/src/HcalStripDigiProducer.cxx#L83

I am still running some test, so there might be more to come, but after a quick look, it seems that everything for `HCalStripDigiProducer` works as expected.

nhanvtran:
@awhitbeck yes I agree, we should put the fix in this PR
@awhitbeck for your 3rd point, I think the link is wrong

awhitbeck:
I ran a test by generating muons which were randomly distributed in the x-y plane of the hcal.  I am seeing half the number of hits I would expect with strip id=30.  A plot is included which shows the hit multiplicity for all cells in the back section. 

image

bechenard:
Weird I'll look into it. What are the bounds you use?

awhitbeck:
If I understand the code, its -1500,1500 mm in both x and y directions.

bechenard:
Hi Nhan,

Now, this is actually the correct definition. The strip 0 goes from -1550
to -1450, and if I have a point at -1500, then
localPosition.y()+scint->GetYHalfLength() = 50,
so dividing by 100 is actually the correct thing to do (dividing by 10
gives 1 cm strips).

bechenard:
Hi Andrew,

For some (weird) reason, this has been coded as -1550 -> 1550. This might
explain the deficit, but I'll run more checks.

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -12,108 +12,130 @@
...
+            hit->setYpos(hcalYpos[detIDraw]);
+            hit->setZpos(hcalZpos[detIDraw]);
             ihit++;
-        } // end loop over detIDs (layers)
-        // put it into the event
-        event.add("hcalDigis", hits_);
+        } 
+        
+        
+        event.add("HcalDigis", hits_);
```
nhanvtran: bug here...this should be `hcalDigis`, not `HcalDigis`

nhanvtran:
Besides the one bug above, it looks great.  I would add some example configuration for usres in
`Configuration/python/sim_recon_prod.py.tmpl`

The default values should be:
```
hcalDigis.parameters["meanNoise"] = 1.5
hcalDigis.parameters["mev_per_mip"] = 1.4
hcalDigis.parameters["pe_per_mip"] = 13.5
hcalDigis.parameters["doStrip"] = 0
```

bechenard:
Hi Nhan,

Should doStrip=0 be the default? Maybe for now.
I've fixed the bug and default values, and added a script to show how to use it (Configuration/python/sample_recon_Hcal.py) for free!

nhanvtran:
Sure, @bechenard thanks, everything looks good on my side.
@awhitbeck did you want to run another muon spray validation?  or should we merge this guy?
@omar-moreno @JeremyMcCormick is that ok?

bechenard:
Let Andrew spray first😀andrew could you post your script to do so

Thanks
Bertrand

awhitbeck:
Yep, I'll rerun the test and post a recipe.

omar-moreno:
Go for it.

awhitbeck:
Looks ready to go to me.

PR #381, opened 2017-08-26
Added HCAL display in Event Display

bechenard:
Need to merge iss378 first to include dependencies on the new Hcal hit data product.


bechenard:
Just updated and tested the code. The HCAL is OFF by default, and all the geometry is currently hard-coded in the file (like the other systems). Iss378 has already been merged.

PR #385, opened 2017-09-05
revert change that causes empty event to be read

JeremyMcCormick:
Reverts change that introduced bug where an empty/default event buffer was being read in EventFile.


jmmans:
I swear I've tried to get this in multiple times, but I just don't seem to commit it to the right branch at the right time... :-)
:+1:

PR #394, opened 2017-09-12
Energy calibration and trigger skim

omar-moreno:


omar-moreno:
cal_energy

Looks good.

PR #397, opened 2017-10-01
Iss391: Update the BDT

GNiendorf:
This PR will update the BDT with a new pkl file, updated recon files, and an updated bdtMaker.py file (all with calibrated energy). You can find train and test root files for signal (with individual mass files) and PN bkg at /nfs/slac/g/ldmx/data/mc/cal_bdt_data/. https://github.com/LDMXAnalysis/ldmx-sw/issues/391


omar-moreno:
@GavinNiendorf Can you update the recon config files found here so they work with the new BDT: 

https://github.com/LDMXAnalysis/ldmx-sw/blob/master/Configuration/python/

omar-moreno:
Also, once you update the configurations, I'm going to move the branch on your fork into ldmx-sw so I can do a couple of test runs.

GavinNiendorf:
I looked through the files and it looks like only the sim_recon_prod file uses the BDT, but let me know if I missed one.

PR #400, opened 2017-10-06
Sorting SimCalorimeterHit collections, Iss399

awhitbeck:
I am adding producer that can sort either ecal or hcal sim hits based on edep.  

There is also a configuration file that can be run, I think out of the box, to test it.


nhanvtran:
Tested, works as desired!

PR #404, opened 2017-10-11
Save generated particles if not inluded in trajectories

bechenard:
My own understanding is that we would  add SimParticles for each generated particle if the SimParticles build from the trajectories do not contain any entries from a generated particle. This can be changed to produce entries for all generated particles every time. There are also some SimParticle fields that don't make sense for generated particles, default values could be changed.

### DIFF
--- a/SimApplication/include/SimApplication/SimParticleBuilder.h
+++ b/SimApplication/include/SimApplication/SimParticleBuilder.h
@@ -81,6 +81,12 @@ namespace ldmx {
              */
             void buildParticleMap(TrajectoryContainer* trajectories, TClonesArray* simParticleColl);
 
+            /**
+             * Add a SimParticle for each generated particle if no generated particle has been saved 
+             * @param simParticleColl The output SimParticle collection.
+             */
+            void saveGeneratedParticle(TClonesArray* simParticleColl);
+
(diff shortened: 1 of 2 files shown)


JeremyMcCormick:
I don't think is the best way to accomplish this.  All you really need to do is turn on trajectory storage if the track is associated to a primary within the UserTrackingAction class.

There is currently a check like this:

```
if (curGenStatus == 1) {
    storeTrajectory(aTrack);
}        
```

This saves gen particles that are not doc particles (e.g. from an LHE file).

Now that I'm thinking about it, you could probably just as well use this check:

```
if (aTrack->GetDynamicParticle()->GetPrimaryParticle() != NULL) {
    // turn on trajectory save for primary here
}
```

We would need to confirm that the doc particles are already being filtered out by the event generation (I think they are but it should be confirmed).

bechenard:
Hi Jeremy,

Not sure I answered to this. I'll admit I;m not an expert, so if there is a
better solution, I'll take it. Did you implement it, or do we need to do
this?
Thanks,
Bertrand

JeremyMcCormick:
No I didn’t implement it but it is only a few lines of code....

PR #410, opened 2017-10-17
Added HCAL in event display - off by default

bechenard:
This is for issue 380... Sorry for the mix up


omar-moreno:
@bechenard Is this something that still needs to be merged?

bechenard:
Hi Omar,
Sorry, I haven't seen your message before. If this still works, then we can commit it. Otherwise, I won;t have time to go back to it right now.

PR #413, opened 2017-10-20
Iss412

awhitbeck:
incorporating sorted sim hit collections into default work flow.  

I tested the code by checking the digi amplitudes for both the ecal and hcal collection using both the sorted and unsorted sim hit collections.  They are highly correlated, but not exactly the same.  It looks like there are a few rechits that are slightly different when using the sorted sim hits, but I am not sure why.    See plots below. 

image

image


omar-moreno:
@awhitbeck So this needs #424 to be fixed before merging right?

awhitbeck:
correct.  I'll try to get a PR in for that soon.

omar-moreno:
Sounds good.

Issue #419, opened 2017-12-19
Seg fault when setting endpoint position in Trajectory


omar-moreno:
Problem occurs when traj->getEndPoint gets called in SimParticleBuilder. It started occurring when compiling with gcc6.3.1.  It's fine when an older version of GCC isused.

omar-moreno:
Changing Tracjectory::getEndPoint to return a G4ThreeVector instead of a const G4ThreeVector& fixes the crash.

omar-moreno:
This issue was resolved by PR 59711a3a13349e7dfa6b384b23eab4082293c56e.

PR #420, opened 2017-12-19
iss419: Seg fault when setting endpoint position in Trajectory

omar-moreno:
This resolves issue #419.


omar-moreno:
Tested the changes using a different compiler and they are fine.  Ready to merge.

PR #425, opened 2018-01-28
Iss424

awhitbeck:
This fixes an issue with dropping collections.   It also added a 'keep' rule function for ignoring collections in the input, if needed.

### DIFF
--- a/Framework/include/Framework/EventFile.h
+++ b/Framework/include/Framework/EventFile.h
@@ -71,6 +71,13 @@ namespace ldmx {
             void addDrop(const std::string& rule);
 
             /**
+             * Add a rule for ignoring collections from the input.
+             * @param rule The rule for ignoring collections.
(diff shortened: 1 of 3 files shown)


omar-moreno:
Can you post a config file that  I can use to test this?

awhitbeck:
[test_sort.txt](https://github.com/LDMXAnalysis/ldmx-sw/files/1683515/test_sort.txt)

Here is a configuration that runs the sorting, then you can test the different drop, keep, and ignore commands (currently commented out) at the bottom of the configuration.

omar-moreno:
I ran the test macro provided above and noticed that events from a branch were being dropped but the branch itself wasn't.  @awhitbeck said he would look into it.

omar-moreno:
I'm closing this pull request for now until we have more time to test it.

PR #426, opened 2018-01-29
Iss138

awhitbeck:
This address issue 138 and adds pure noise hits to `HcalDigiProducer` following the same procedure as the `EcalDigiProducer`.

### DIFF
--- a/EventProc/src/HcalDigiProducer.cxx
+++ b/EventProc/src/HcalDigiProducer.cxx
@@ -194,7 +194,7 @@ namespace ldmx {
         }
 
         // simulate noise hits in side, top/bottom hcal
-        noiseHits_PE = noiseGenerator_->generateNoiseHits((STRIPS_SIDE_TB_PER_LAYER_*NUM_SIDE_TB_HCAL_LAYERS_)*4-numSigHits_side_tb);
+        noiseHits_PE = noiseGenerator_->generateNoiseHits((STRIPS_SIDE_TB_PER_LAYER_*NUM_SIDE_TB_HCAL_LAYERS_)*2-numSigHits_side_tb);
         for( auto noise : noiseHits_PE ){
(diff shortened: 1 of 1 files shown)


omar-moreno:
@awhitbeck @nhanvtran @bechenard Has this been reviewed?  Does it matter if this isn't included in the next recon round?

awhitbeck:
Your description is not quite right, at least not what is intended, I only simulate the noise for the emtpy cells, so there should be no edge cases like you mention in your second to last bullet.  Note, the noise for the cell with sim energy is handled differently.  

I did test this, but was hoping either you or Nhan could also  try to run it.

omar-moreno:
@awhitbeck @nhanvtran @bechenard Is this PR still on hold until further testing?

awhitbeck:
I am reviving this... I am hoping someone else can try to run it.

awhitbeck:
So, to get the noise fully correct for the case in which we readout each strip twice, we need to extend the HcalID to include a photodetector field.  However, this will require some thought on how we spread the signal over each of these detector element.  So, I proposed we merge this without the multi photodetector functionality and leave this for future development.

I just need to add one more feature so that the noise simulation can handle to different geometry for the top/bottom and left/right side hcal.  I'll commit this shortly.  Then we should merge this and test.

awhitbeck:
Okay, I have fixed one last issue.  I think the noise simulation is now ready.  Below are some plots of the PE distribution, [1], and the section [2], layer  [3], and strip ID [4]. 


[1]
This is using parameters in Configuration/python/sim_recon_prod.py.tmpl.  One expects, 3e5 hits with 1 PE, 6e2 with 2 PEs, and .8 hits with 3 PEs.  Rates are statistically compatible.  

image
[2]
image

[3]
back: 
image
side, bottom: 
image
side, left: 
image
side, right:
image
side, top:
image

[4]
back:
image
side, bottom:
image
side, left:
image
side, right:
image
side, top:
image

PR #438, opened 2018-02-27
Iss134: Implementation of basic geometry system

JeremyMcCormick:
As of today, merged master into this branch and resolved (minor) conflicts.

The `ldmx-app` binary loads okay (getting an error when it runs but it looks to be unrelated and having to do with an out-of-date recon config I am using).

Independent testing of this branch before merging would be good.  It is unlikely to break anything fundamental in the event processors because the new geometry functionality is not actually used in the current recon anyplace..

There were also some minor changes (addition of addition tags) to the detector GDML files.  We should verify that sim and recon output looks the same as master using the same detector model (v3).

### DIFF
--- a/Framework/include/Framework/Process.h
+++ b/Framework/include/Framework/Process.h
@@ -129,6 +129,13 @@ namespace ldmx {
              * Access the storage control unit for this process
              */
             StorageControl& getStorageController() { return m_storageController; }
+
+        private:
+
+            /**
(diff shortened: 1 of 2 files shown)


omar-moreno:
This branch has gotten difficult to review so I'm going to break it up into smaller pull request.  It will also allow me to understand what is being done.

PR #442, opened 2018-03-21
Adding a seed for the beam smearing

nhanvtran:
This takes the Geant seed so that we can have reproducibility


omar-moreno:
Is this just setting a random number as a seed or making use of the Geant4 run seed?

nhanvtran:
This grabs the geant seed defined in the .mac file

PR #443, opened 2018-03-22
Ecal Clustering and Event Display Additions

JHiltbrand:
I realize now that I snuck in a very simple change to Event.h to this pull request. The getCollection method that takes a collection name and a pass name did not have a "const" between the function arguments and the function body causing an error on my end.


omar-moreno:
This looks fine, I just need to run a few test.  @JHiltbrand I'm thinking we move everything into an Ecal module instead of having the code live in a Cluster module?  Since this is Ecal specific code, it make sense  that it lives in an Ecal module.

JHiltbrand:
@omar-moreno ,

That sounds reasonable. What specifically did you have in mind?

PR #449, opened 2018-04-04
Iss444

bechenard:
New Hcal geom for v5 detector. I kept the side HCAL length to 290 mm instead of 300mm to avoid messing with the virtual detectors.

### DIFF
--- a/Detectors/data/ldmx-det-full-v5-fieldmap-magnet/hcal.gdml
+++ b/Detectors/data/ldmx-det-full-v5-fieldmap-magnet/hcal.gdml
@@ -1,288 +1,366 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no" ?>
 <gdml xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://service-spi.web.cern.ch/service-spi/app/releases/GDML/schema/gdml.xsd">
 
+<!-- The posiiton of the ECAL is wrong on detector.gdml. If should be at 215 mm, and 
+   name="ecal_front_z" value="215*mm
+   name="em_calorimeter_thickness" value="290.0*mm
+   name="em_calorimeter_pos_z" value="ecal_front_z + em_calorimeter_thickness/2.
+   name="hadron_calorimeter_thickness" value="3300.0*mm"
+   name="hadron_calorimeter_pos_z" value="ecal_front_z + hadron_calorimeter_thickness/2."
+
+  then the hcal size ("hadron_calorimeter_thickness") is simply side_dz + bach_hcal_thickness
+-->
+
   <define>
     
+    <!-- loop variables - must be dedfined here -->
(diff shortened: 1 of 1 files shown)


awhitbeck:
I looked at the new geometry, it looks good to me.  I ran an overlap check and everything passed.  It does seem like the back scoring plane is a bit far from the last layer. See plot below.  If we are going to change the scintillator size, we might want to adjust this.  I have no strong feelings about the scintillator thickness, by the way.

image

Blue is sim hit z position, red is scoring plane hit z position

bechenard:
HI Andrew,


It is true I never looked at the virtual detector. I don't think the gap 
is a big issue if the scoring plane is large enough, though.

Cheers,

Bertrand

omar-moreno:
Also ran the overlap check and can confirm things look fine.  I think as long as the new hcal geometry falls inside of the scoring planes, it should be fine.  I can do a quick check of this.

omar-moreno:
Looks like the Hcal falls inside of the Hcal scoring plane volumes (blue below)

image

image

omar-moreno:
Zooming in a bit on the front face, it looks like the Ecal extends outside of the scoring plane volume now.  Actually, it looks like the whole Ecal/Hcal system got shifted forward a bit.  I need to check to make sure if there isn't overlap with the magnet.

bechenard:
Hi Omar,


Yes, I also noticed that, hence my previous comment. I kept the geometry 
the same size as I wasn't sure if this would mess up with the 
re-simming. I believe the proper way to fix that is to get rid of the 
different offsets introduced (+1 for the ECAL bounding box, and a -2.5 
for the HCAL z position).

Let me know if this messes up with the re-simming - doing some work on 
that right now.

Cheers,

Bertrand

omar-moreno:
@bechenard As long as the non-parallel world geometry has no overlaps (which at the moment it doesn't), resiming shouldn't be an issue. The Hcal can be whatever size it needs to be. 

I'm fixing the overlaps with the scoring planes now.  Those are easy fixes.

omar-moreno:
Well, giving that we have no concrete answer as to what the Hcal size should be, sticking with a giant Hcal is the way to go.

awhitbeck:
By the way, I thought the plan was that the strips allow the side hcal would go along the z-direction so that the fibers would come out the side instead of the front?

bechenard:
The strips need to go along the x or y direction to come out of the 
sides. Along the z direction, you either end up in the magnet of the 
back HCAL. Or an I missing something?

awhitbeck:
Sorry I meant segmented along z (which has a width of 29cm).  I think currently they are segmented along x and y.

bechenard:
I need to check this. It could very well be that I didn't update the 
code when we decided to change the segmentation.

omar-moreno:
This branch was merged into v4_hcal_dev and will be merged into master when that branch gets merged.  This PR is no longer necessary.

PR #451, opened 2018-04-04
Iss450

bechenard:
Added Birks correction form HCAL. This was given to me from H. Wenzel, based on the paper NIM 80 (1970) 239-244. 
The current structure of the code has forced me to include the Birks correction directly in Geant4, this can;t be done at the reco level without changing the hit making model.

### DIFF
--- a/SimApplication/include/SimApplication/HcalSD.h
+++ b/SimApplication/include/SimApplication/HcalSD.h
@@ -31,6 +31,10 @@ namespace ldmx {
             virtual ~HcalSD();
 
             G4bool ProcessHits(G4Step* aStep, G4TouchableHistory* ROhist);
+            
+        private:
+          double birksc1_;
+          double birksc2_;    
     };
 
 }
--- a/SimApplication/src/HcalSD.cxx
+++ b/SimApplication/src/HcalSD.cxx
@@ -10,11 +10,16 @@
 #include "G4SDManager.hh"
 #include "G4Geantino.hh"
 #include "G4ChargedGeantino.hh"
+#include "G4ParticleDefinition.hh"
+#include "G4ParticleTypes.hh"
 
 namespace ldmx {
 
     HcalSD::HcalSD(G4String name, G4String theCollectionName, int subdetID, DetectorID* detID) :
-            CalorimeterSD(name, theCollectionName, subdetID, detID) {
+            CalorimeterSD(name, theCollectionName, subdetID, detID),
+            birksc1_(1.29e-2),
+            birksc2_(9.59e-6)
+    {
     }
 
     HcalSD::~HcalSD() {}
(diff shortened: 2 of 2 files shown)


awhitbeck:
Are we splitting this commit?  

Looking at the implementation of Birks's correction, it looks like this is not the same implementation that is used in CMS [1], where the 2nd order correction coefficient seems to depend on both birks1 and birks2.  There is also a different parameterization for alpha particles; should we consider adding this as well?

Is the block comment really necessary?  It deosn't even seem correct.  We ignore the efficiency term and it doesn't include the higher order quenching terms.  In general, I don't like having a giant comment in the middle of the code.  

[1] https://github.com/cms-sw/cmssw/blob/master/SimG4CMS/Calo/src/CaloSD.cc#L439-L459

bechenard:
Hi Andrew,

I'm ok to split. I got the Birks implementation from Hans Wenzel, based 
on the paper quoted in the comment. I don't know what is used for CMS, 
and how this has been determined. Is that described somewhere? Since I'm 
using this implementation of Birks for my current studies, I say we keep 
it and we can revisit after the white paper. I imagine both 
paramterizations should lead to fairly similar results. (I don't have 
time to study the differences, but feel free to look at it).

I included the block comment for you and Nhan, but feel free to remove 
it (just keep the reference to the paper).

Cheers,

Bertrand

awhitbeck:
I put a link to the CMS implementation, which cites the same paper.  I haven't been able to find this paper, since NIM doesn't have online access to papers from the 70's.

bechenard:
Ok, will look at it later.

Bertrand

bechenard:
Hi Andrew,

Except for the alpha correction, the form is functionally equivalent for 
both implementation. My birk2 constant is rewritten as 
birk2'*birk1*birk1 in the CMS code. I don't know which values are used, 
but itmight very well be the same result if both are based on the same 
paper. We could ad the alpha correction if you wish, it is simply 
reweighing both birk1 and birk2.

Cheers,

Bertrand

awhitbeck:
For CMSSW, the parameters are set here, [1], it seems.  While the second order terms seem consistent, the leading order correction is a factor of 2 smaller.

Have you looked at this NIM paper to check what these constants should be?  If so, can you send it around?

Andrew

[1]
https://github.com/cms-sw/cmssw/blob/master/SimG4Core/Application/python/g4SimHits_cfi.py#L291-L293

omar-moreno:
@bechenard Did you attach the paper?

omar-moreno:
I'm going to go ahead and do a test re-recon on the BDT skim.  I've gotten the number of files down to 522 so re-siming should be quick.

bechenard:
Yes, here it is once again

Bertrand

awhitbeck:
I still don't see a paper.

omar-moreno:
Still can't see this paper.  I'm going to go ahead and seperate this out so I can merge it into the dev branches.  I'll reopen the PR once it's ready.

PR #458, opened 2018-04-11
Abort the run (instead of the event) if the end of a file is reached.

omar-moreno:


omar-moreno:
This PR resolves #458 .

PR #465, opened 2018-04-27
Fiducial Tagging Code, New BDT's, Multiple BDT Support

GNiendorf:
edit: I'm going to update the default cut values for the BDT's tomorrow (Monday) morning after discussing it with Giulia and Joe, so please wait until then before pushing this in.

1 fiducial BDT and 4 non-fiducial BDT's were generated by Giulia on v5 data. I updated EcalVetoProcessor to allow for multiple BDT's and also to include code for the fiducial definition. See https://github.com/LDMXAnalysis/ldmx-sw/issues/464 for the corresponding issue. I also added a "cellxy.txt" file which has the x,y coordinates of all cells, which must be included when using EcalVetoProcessor for the fiducial definition, and added functionality to drop fiducial or non-fiducial events.

I can explain the changes easier through looking at the updated recon file. You now specify a fiducial bdt and 4 non-fiducial BDT's (in order of increasing mass). You must also have the cellxy.txt file in your working area (with the 5 BDT's) for the recon to run properly. You now also specify 5 bdt cuts, one disc cut for the fiducial bdt and one for each non-fiducial BDT (with the first element corresponding to the fiducial disc cut and the next four to the non-fiducial cut in increasing mass). To drop fiducial events you would set "ecalVeto.parameters["drop_fid_nf"] = [1,0]" and to drop non-fiducial events "ecalVeto.parameters["drop_fid_nf"] = [0,1]", both of which are assuming that you are listening to ecalVeto and drop as default. If you set one of these options with the BDT turned off (do_bdt = 0) it will still drop non-fiducial or fiducial events, which may be useful for skimming samples of either class. If you have do_bdt on, then for the non-fiducial events (assuming default drop and listening to ecalVeto) it will save the event if any of the four BDT's is above threshold. For the fiducial case it will save the event if the event is above threshold for the single BDT, like before.

Some internal changes, disc_value is now a vector with -1 as the default value for each element, in which one element is replaced by the disc value of the fiducial BDT or four elements are replaced by the disc values of the four non-fiducial BDT's, depending if the event is classified as fiducial or non-fiducial. A new "Inside" variable has also been added to the tree (which specifies if the recoil electron hit inside (value 1) the ECAL or hit outside (value 0) the ECAL), as well as the projected position of the recoil electron on the ECAL face as "FaceX" and "FaceY" which are used in the fiducial code.

This code should not be a burden to use with old recon-scripts. Just make sure that all BDT's and cellxy.txt are in your working area and specify if you want to drop fiducial or non-fiducial events (or neither).

ecalVeto = ldmxcfg.Producer("ecalVeto", "ldmx::EcalVetoProcessor")
ecalVeto.parameters["num_ecal_layers"] = 34
ecalVeto.parameters["do_bdt"] = 1
ecalVeto.parameters["fid_bdt_file"] = "fid_bdt.pkl"
ecalVeto.parameters["nf_bdt_files"] = ["p001_nf_bdt.pkl", "p01_nf_bdt.pkl", "p1_nf_bdt.pkl", "p0_nf_bdt.pkl"]
ecalVeto.parameters["cellxy_file"] = "cellxy.txt"
ecalVeto.parameters["disc_cut"] = [0.96, 0.95, 0.95, 0.95, 0.95]
#Set to 1 to drop fid or non-fid events respectively (Must listen and drop by default).
ecalVeto.parameters["drop_fid_nf"] = [0, 0]

### DIFF
--- a/EventProc/include/EventProc/NonFidEcalVetoProcessor.h
+++ b/EventProc/include/EventProc/NonFidEcalVetoProcessor.h
@@ -1,11 +1,11 @@
 /**
- * @file EcalVetoProcessor.h
- * @brief Class that determines if event is vetoable using ECAL hit information
+ * @file NonFidEcalVetoProcessor.h
+ * @brief Class that determines if events outside fiducial region are vetoable
  * @author Owen Colegrove, UCSB
  */
 
-#ifndef EVENTPROC_ECALVETOPROCESSOR_H_
-#define EVENTPROC_ECALVETOPROCESSOR_H_
+#ifndef EVENTPROC_NONFIDECALVETOPROCESSOR_H_
+#define EVENTPROC_NONFIDECALVETOPROCESSOR_H_
 
 // ROOT
 #include "TString.h"
@@ -15,7 +15,7 @@
 // LDMX
 #include "DetDescr/EcalHexReadout.h"
 #include "DetDescr/EcalDetectorID.h"
-#include "Event/EcalVetoResult.h"
+#include "Event/NonFidEcalVetoResult.h"
 #include "Event/SimTrackerHit.h"
 #include "Framework/EventProcessor.h"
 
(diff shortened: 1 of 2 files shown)


omar-moreno:
@GNiendorf Is this backwards compatible with the current ldmx-sw BDT? I'm wondering if pushing this is going to require us to re-recon everything. You are changing the event model so that might make the current recon files useless.

GNiendorf:
I'm going to close this branch and make a new PR so there are no useless commits from my older code.

PR #479, opened 2018-07-30
nvt/hcal reco v4

nhanvtran:
- change to 5 mm wide sensitive detectors
- alternating XY strips
- add possibility of super strips
- double-sided strip readout with new "minPE()" method and double-sided noise simulation

### DIFF
--- a/EventProc/python/hcalDigis.py
+++ b/EventProc/python/hcalDigis.py
@@ -6,11 +6,11 @@ hcalDigis = ldmxcfg.Producer("hcalDigis", "ldmx::HcalDigiProducer")
 
 hcalDigis.parameters["meanNoise"] = 0.02
 hcalDigis.parameters["readoutThreshold"]= 1
-hcalDigis.parameters["strips_side_lr_per_layer"] = 3
-hcalDigis.parameters["num_side_lr_hcal_layers"] = 17
-hcalDigis.parameters["strips_side_tb_per_layer"] = 3
-hcalDigis.parameters["num_side_tb_hcal_layers"] = 17
-hcalDigis.parameters["strips_back_per_layer"] = 62 # n strips correspond to 5 cm wide bars
+hcalDigis.parameters["strips_side_lr_per_layer"] = 6
+hcalDigis.parameters["num_side_lr_hcal_layers"] = 31
+hcalDigis.parameters["strips_side_tb_per_layer"] = 6
+hcalDigis.parameters["num_side_tb_hcal_layers"] = 31
+hcalDigis.parameters["strips_back_per_layer"] = 60 # n strips correspond to 5 cm wide bars
 hcalDigis.parameters["num_back_hcal_layers"] = 40
 hcalDigis.parameters["super_strip_size"] = 1 # 1 = 5 cm readout, 2 = 10 cm readout, ...
 hcalDigis.parameters["mev_per_mip"] = 1.4*20./6.


nhanvtran:
It was not a visual illusion, it was not the magnetic field, it was just a bug :(

I shifted the hits by 150 mm, not smeared.   Now it's fixed.  

image

@awhitbeck let me know if you need some raw numbers for computing poisson probabilities -- we can talk about it on Slack

omar-moreno:
@nhanvtran @awhitbeck What's the ETA on this?

bechenard:
I've just started the review on this one...

Cheers,

Bertrand

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -24,21 +24,24 @@ namespace ldmx {
...
         mev_per_mip_ = ps.getDouble("mev_per_mip");
         pe_per_mip_  = ps.getDouble("pe_per_mip");
         doStrip_     = ps.getInteger("doStrip");
         noiseGenerator_ = new NoiseGenerator(meanNoise_,false);
-        noiseGenerator_->setNoiseThreshold(readoutThreshold_);
+        //noiseGenerator_->setNoiseThreshold(readoutThreshold_);
+        noiseGenerator_->setNoiseThreshold(1); // hard-code this number, create noise hits for non-zero PEs! 
     }
 
     unsigned int HcalDigiProducer::generateRandomID(HcalSection sec){
```
bechenard: What do we need that for?

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -65,8 +69,14 @@ namespace ldmx {
...
             SimCalorimeterHit* simHit = (SimCalorimeterHit*) hcalHits->At(iHit);
             int detIDraw = simHit->getID();
+            detID_->setRawValue(detIDraw);
+            detID_->unpack();
+            int layer = detID_->getFieldValue("layer");
+            int subsection = detID_->getFieldValue("section");
+            int strip = detID_->getFieldValue("strip");                 
             std::vector<float> position = simHit->getPosition();
 	    
             //if we aggregate by layer, set all strip to zero and rcalculate the detIDraw
```
bechenard: Do we want to remove this? We won't do full layer anymore, and somebody might accidentally use that option if in there.

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -78,14 +88,22 @@ namespace ldmx {
...
-            }           
+                int newstrip = strip/SUPER_STRIP_SIZE_;
+                detID_->setFieldValue("strip",newstrip);
+                detID_->pack();
+                // get the new raw value
+                detIDraw = detID_->getRawValue();
+            }
             
             // for now, we take am energy weighted average of the hit in each stip to simulate the hit position. 
             // will use strip TOF and light yield between strips to estimate position.
```
bechenard: I think this is the right place to add the hit position filtering if we want it, since we still have the correct location of the deposited energy. The HCAL veto already has a averaged / smeared position, so this would be less reliable to put it in there.

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -128,76 +145,151 @@ namespace ldmx {
...
+            int cur_subsection = curDetId->getFieldValue("section");            
+            int cur_layer      = curDetId->getFieldValue("layer");
+            int cur_strip      = curDetId->getFieldValue("strip");
+            float cur_xpos, cur_ypos; 
+
+            if (cur_subsection != 0){ // for sidecal don't worry about attenuation because it's single readout
+                hcalLayerPEs[detIDraw] = random_->Poisson(meanPE+meanNoise_);
+                hcalLayerMinPEs[detIDraw] = hcalLayerPEs[detIDraw];
+            }
+            if (cur_subsection == 0){// get PEs with attentuation
```
bechenard: Should we make the attenuation a parameter?

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -128,76 +145,151 @@ namespace ldmx {
...
+                    cur_xpos = hcalXpos[detIDraw] + random_->Gaus(0.,150.); 
+                }
+                if (cur_xpos > total_width/2.) cur_xpos = total_width/2.;
+                if (cur_xpos < -1.*total_width/2.) cur_xpos = -1.*total_width/2.;
+                if (cur_ypos > total_width/2.) cur_ypos = total_width/2.;
+                if (cur_ypos < -1.*total_width/2.) cur_ypos = -1.*total_width/2.;
+                // std::cout << "super_strip_width = " << super_strip_width << "\t total_super_strips = " << total_super_strips << "\t total_width = " << total_width << std::endl;
+            }
+
+            if( hcalLayerPEs[detIDraw] >= readoutThreshold_ ){ // > or >= ?
```
bechenard: >= is fine for now I think, this will depend on the readout scheme anyway.

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -24,21 +24,24 @@ namespace ldmx {
...
         readoutThreshold_ = ps.getInteger("readoutThreshold");
         meanNoise_   = ps.getDouble("meanNoise");
         mev_per_mip_ = ps.getDouble("mev_per_mip");
         pe_per_mip_  = ps.getDouble("pe_per_mip");
         doStrip_     = ps.getInteger("doStrip");
         noiseGenerator_ = new NoiseGenerator(meanNoise_,false);
-        noiseGenerator_->setNoiseThreshold(readoutThreshold_);
+        //noiseGenerator_->setNoiseThreshold(readoutThreshold_);
+        noiseGenerator_->setNoiseThreshold(1); // hard-code this number, create noise hits for non-zero PEs! 
     }
```
bechenard: Should we enforce that the number of NUM_BACK_HCAL_LAYERS_  is a multiple of SUPER_STRIP_SIZE_ , i.e. NUM_BACK_HCAL_LAYERS_  %SUPER_STRIP_SIZE_ = 0? It seems strange to have a fraction of a super strip, and that causes problems when we calculate the strip id.

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -128,76 +145,151 @@ namespace ldmx {
...
+                std::cout << "Layer: " << layer << std::endl;
+                std::cout << "Subsection: " << subsection << std::endl;
+                std::cout << "Strip: " << strip << std::endl;
+                std::cout << "Edep: " << hcaldetIDEdep[detIDraw] << std::endl;
+                std::cout << "numPEs: " << hcalLayerPEs[detIDraw] << std::endl;
+                std::cout << "time: " << hcaldetIDTime[detIDraw] << std::endl;
+                std::cout << "z: " << hcalZpos[detIDraw] << std::endl;
+                std::cout << "Layer: " << layer << "\t Strip: " << strip << "\t X: " << hcalXpos[detIDraw] <<  "\t Y: " << hcalYpos[detIDraw] <<  "\t Z: " << hcalZpos[detIDraw] << std::endl;
+            }        // end verbose            
         }
```
bechenard: If I understand well, the noise is simulated per strip, i.e. we have either noise is the two readouts of a given strip (one value might be zero), or no noise in both readouts. Is that the correct model, or should we distribute noise hits in readouts instead of strips?

omar-moreno:
If the changes to the SD are fine, I recommend putting them in a different PR so I can merge them into master and start the production back up. Then you can take your time evaluating the recon changes.

Event/include/Event/HcalHit.h
```diff
@@ -138,6 +155,9 @@ namespace ldmx {
             /** The number of PE estimated for this hit. */
             float pe_{0};
 
+            /** The minimum number of PE estimated for this hit, different from pe_ when you have two ended readout */
+            float minpe_{-99};
```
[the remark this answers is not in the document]
nhanvtran: It's a debug technique because technically minPE can be 0 and the hit will still be read out (where as PE cannot be 0 and read out).  If we want to change anything here, I would initialized `pe_` to -99 to be safe

EventProc/src/HcalDigiProducer.cxx
nhanvtran: I don't think there is an issue with keeping it as it retains backward compatibility

EventProc/src/HcalDigiProducer.cxx
nhanvtran: Do you mean if we allow a hit or not if we want to restrict the size of the hcal for different vetoes?  If yes, it's an interesting issue of where to do this.  I would probably make a check a little later down in the code around L111 then you would have multiple digiProducers for multiple vetoes.  Is that what you were thinking?

EventProc/src/HcalDigiProducer.cxx
nhanvtran: Yes that's a good idea.  Also another parameter for the time smear resolution?

EventProc/src/HcalDigiProducer.cxx
```diff
@@ -128,76 +145,151 @@ namespace ldmx {
...
+                float PE_close     = random_->Poisson(meanPE_close+meanNoise_);
+                float PE_far       = random_->Poisson(meanPE_far+meanNoise_);
+                hcalLayerPEs[detIDraw] = PE_close + PE_far;
+                hcalLayerMinPEs[detIDraw] = std::min(PE_close,PE_far);
+            }
+            // std::cout << "depEnergy = " << depEnergy << "\t hcalLayerPEs[detIDraw] = " << hcalLayerPEs[detIDraw] << "\t hcalLayerMinPEs[detIDraw] = " << hcalLayerMinPEs[detIDraw] << std::endl;
+
+            if (cur_subsection == 0){
+                float super_strip_width = SUPER_STRIP_SIZE_*50.0;
+                float total_super_strips = (STRIPS_BACK_PER_LAYER_ + SUPER_STRIP_SIZE_ - 1)/SUPER_STRIP_SIZE_; // to help the integer round up
```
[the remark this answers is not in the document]
nhanvtran: Well i think that's just the issue -- in my formulation I am able to handle `STRIPS_BACK_PER_LAYER  %  SUPER_STRIP_SIZE == 0` with some offset.  So it works but the last super strip position is offset a little bit.  Unfortunately the number that we chose in the geometry is a little awkward (62 5cm strips) so enforcing a nice round number does not leave you with many options.  So I went with this approach.

EventProc/src/HcalDigiProducer.cxx
nhanvtran: We had a detailed discussion about this in the software meeting maybe a week and a half ago.  The meaning of `readoutThreshold_` now strictly means, at what threshold do we save hits and that's it.  This is because before this parameter was dually used to also help generate noise.  This can get confusing for the 2 SiPM bar hit, so instead the noise generation for the HCal is fixed to generate noise above 1 PE.

EventProc/src/HcalDigiProducer.cxx
nhanvtran: So to do what you are proposing is a little bit tricky -- which is why I went with this more flexible route.  

The core issue is that the information is spread out in different places and cannot necessarily be accessed at a given step.  The gdml file (which has an awkward size layer of 3.1 m now -> 62 strips) can only be accessed during the SIM step.  Hence we define a bar size in the HCal SD.
At RECO level, we don't know anything about the geometry anymore, so I was using the strip ID defined in the SIM step as a proxy for aggregating into superstrips.  If you want to do what you are suggesting, a parameter to the RECO job would have to be the transverse size of the HCal layers for a given geometry, which is a little backwards IMO.

The logic now is:
- gdml defines the geometry
- sim hit SD strip ID is defined at the SIM level using gdml geometry information
- reco hits and (super)strip ID aggregate sim hits using their strip ID

What you are proposing is:
- define the reco strip size and ID based on the geometry, but the issue is that at the RECO step, one does not have access to the geometry unless you hardcode it as a parameter

I'm not sure which is better.  In the former case, you can handle any geometry but can have partial superstrips while in the latter you have to enforce nice round numbers but is geometry dependent.

EventProc/src/HcalDigiProducer.cxx
nhanvtran: Noise is simulated per readout (or SiPM).  And then the total PE and the minPE are saved (such that the maxPE could also be calculated)

nhanvtran:
@bechenard thanks for the comments.  I wrote in some replies but we can also chat for efficiency.

@omar-moreno we could consider splitting off the SD part into a separate PR, but then we would have a little awkward situation where the master would have incompatible hcal SIM and RECO steps.  

Let's have @bechenard and me discuss and see how close we are and re-evaluate?

nhanvtran:
from @bechenard and @nhanvtran chat:
- put in protection for non-regular super strip size
   - more native N super strip definition
- make attentuation length and time smearing configurable
- remove `doStrip` option

@bechenard update v9 geom, 3.1x3.1m to 3.0x3.0m and shrink side hcal accordingly

For next/future PR, add geometrical sim hit filtering for partial detector vetoes

nhanvtran:
@bechenard @omar-moreno the changes that are discussed above are now implemented

nhanvtran:
Ok, updated config now there

omar-moreno:
Should we merge this in?

bechenard:
Hi Omar,

I believe yes, but you want Nhan to give you the final say. You also 
need to make sure you get the latest v9-dev.

Cheers,

Bertrand

nhanvtran:
I'm okay with merging :).   the v9-dev is already merged PR #488

PR #488, opened 2018-08-17
Reduced transvereses HCAL size to 3 m

bechenard:


nhanvtran:
ah probably you should also update the digi configuration to reflect the right number of strips in the back hcal too (though this will be updated again by the PR #479)

bechenard:
Right, but I thought your PR would deal with it, so I didn;t want to 
mess the code in between. Would you mind doing it?
Thanks,
Bertrand

nhanvtran:
I can put it in my PR.  Still a question if we want 31 or 32 layers in the sidecal.

btw, what do we do with the sidecal dz = 290?  we say we have 6 strips wide, but technically one of the strips is not as wide as the other...

bechenard:
Hi Nhan,

Assuming the ECAL takes a total length of 525 mm, and a 
absorber/scintillator thickness of 20mm/15mm, a layer is 39mm (there are 
two 2mm air gaps). With those numbers, you get a total length of 2943 mm 
for 31 layers, and 3021mm for 32 layers. I picked the 31 configuration, 
but if you want to keep 32, that's fine for me.

Cheers,

Bertrand

PR #489, opened 2018-08-17
Addresses iss487

awhitbeck:


omar-moreno:
@awhitbeck  Looking through the code now.  Is there a reason why you use the HCal scoring plane hits to find the recoil electrons? It's not really needed.

PR #490, opened 2018-09-11
V10 dev

bechenard:
Simplified the gdml files to get rid of multiple offsets and simplifying calculation of HCAL length. This is exactly the same geometry as v9 - I checked by running a few events and printing the volume coordinates. I had to include a 0.5mm offset on the HCAL placement to do so. We might want to get rid of that in the future, so that the ECAL and HCAL mother volumes are aligned.


omar-moreno:
@nhanvtran @awhitbeck Any luck with reviewing this PR?

nhanvtran:
@awhitbeck knows gdml much better than I but in case he's tied up , I can try

awhitbeck:
I am totally swamped.  I have a lot of travel and extra things on my plate this month.  If this drags on for a couple of weeks, then I can tackle this then.

bechenard:
Hi guys,

This isn't really urgent, and the rewrite is more for convenience that 
anything, so feel free to do as most convenient.

Cheers,

Bertrand

PR #491, opened 2018-10-04
Adds additional energy calibration weight to account for an extra ECal layer.

omar-moreno:


omar-moreno:
After the fix, hits are showing up in layer 34. 
max_layer_hit
ecal_layers_hit

Issue #499, opened 2019-01-26
Add setting of run description using macro command.


omar-moreno:
At the moment, it's currently hard-coded into the code.

omar-moreno:
This issue was resolved by PR #501

PR #501, opened 2019-01-30
Iss499 - Allow the run description to be settable from a macro.

omar-moreno:


omar-moreno:
This resolves Issue #499 .

PR #552, opened 2019-05-31
Iss551 - Add min PE requirement to back HCal hits.

omar-moreno:


EventProc/python/hcalVeto.py
```diff
@@ -5,3 +5,4 @@
 hcalVeto.parameters['pe_threshold'] = 5.0
 hcalVeto.parameters['max_time'] = 50.0
 hcalVeto.parameters['max_depth'] = 4000.0
+hcalVeto.parameters['min_pe'] = 1.
```
nhanvtran: maybe call this something like "back_min_pe"?  just so it's clear that this is only for back HCal?
Otherwise looks good for me

omar-moreno:
Will update the config.

EventProc/python/hcalVeto.py
omar-moreno: Made the change.

bechenard:
Hi Omar,

The depth and time cut looks fine for now. Good re-simming.

Bertrand

Issue #555, opened 2019-06-08
Enable dropping collections from reco output file only, keeping them during reco 


bryngemark:
Adding ability to disable a branch (collection) from the output file while keeping it active in the input file, such that the input info can be used to generate collections in the output file but then discarded and only the result kept. This was discussed in a specific case in #424 but solved differently there.

The solution here is to ensure the output tree exists when the keep/drop/ignore rules are processed, and in the drop case, disable the branch in the output tree while explicitly keeping it in the input tree.

The "ignore" case is kept for disabling the branch already in the input tree.

bryngemark:
This is an alternative solution to #534. Suggest making a separate issue for the use case of regenerating a collection (e.g. ECalVeto collection).

PR #556, opened 2019-06-08
Ensuring output tree exists before attempting to drop collection

bryngemark:
This fix avoids the seg fault that happens otherwise when a collection is flagged to be dropped 
from the output tree only (case: drop).
The collection can still be dropped already from the input tree (case: ignore).

The use case of regenerating an existing reco collection should have its own
flag "regenerate" where the input file collection is disabled and output
collection enabled, with the same name. Suggest new ticket for that.

As a sanity check, I have verified that the PE distribution etc is unaffected by dropping the HcalSimHits_sim collection from the reco file. So the input collection is kept and used as it should.

This pull request is in response to issue #555 (forgot to set the branch name accordingly).

### DIFF
--- a/Framework/src/EventFile.cxx
+++ b/Framework/src/EventFile.cxx
@@ -113,7 +113,13 @@ namespace ldmx {
             parent_->tree_->SetBranchStatus(srule.c_str(),1);
             tree_->SetBranchStatus(srule.c_str(),1);
         } else if( isDrop ){
-            parent_->tree_->SetBranchStatus(srule.c_str(),0);
+	  if (! tree_) { 
+	    //this prevents creating a "ghost" branch in the output tree from the enabled branch in the input tree 
+	    parent_->tree_->SetBranchStatus(srule.c_str(),0);
+	    tree_ = parent_->tree_->CloneTree(0);
+	  }
+	  tree_->SetBranchStatus(srule.c_str(),0);
+	  parent_->tree_->SetBranchStatus(srule.c_str(),1);
         } else if( isIgnore ){
             parent_->tree_->SetBranchAddress(srule.c_str(),0);
         } else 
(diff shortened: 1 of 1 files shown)


Framework/src/EventFile.cxx
```diff
@@ -113,7 +113,13 @@ namespace ldmx {
             parent_->tree_->SetBranchStatus(srule.c_str(),1);
             tree_->SetBranchStatus(srule.c_str(),1);
         } else if( isDrop ){
-            parent_->tree_->SetBranchStatus(srule.c_str(),0);
+	  if (! tree_) { 
+	    //this prevents creating a "ghost" branch in the output tree from the enabled branch in the input tree 
+	    parent_->tree_->SetBranchStatus(srule.c_str(),0);
+	    tree_ = parent_->tree_->CloneTree(0);
+	  }
+	  tree_->SetBranchStatus(srule.c_str(),0);
```
tomeichlersmith: This line throws the warning:
```
Error in <TTree::SetBranchAddress>: unknown branch -> DROP_BRANCH_NAME
```
This is because `tree_` was cloned when that branch was turned off. This error can be removed by putting this line in the `else` branch of the above `if` tree so this line is only called when `tree_` is already created.

But I don't see a situation where `tree_` is already created. The drop/keep rules are added to the output Event file before any events are read, this means this code is *always* called before the other location where `tree_` could be cloned (at the beginning of `nextEvent`). Maybe put it in an `else` branch with a comment/warning about this being a weird case?

Framework/src/EventFile.cxx
```diff
@@ -126,7 +132,9 @@ namespace ldmx {
             if (!parent_->tree_) {
                 EXCEPTION_RAISE("EventFile", "No event tree in the file");
             }
-            tree_ = parent_->tree_->CloneTree(0);
+	    if ( isOutputFile_ && !tree_) {
```
tomeichlersmith: This part will conflict with PR #566 , but keeping the other PR's version of these lines will allow both to work.

PR #476, opened 2018-07-17
Iss475

tomeichlersmith:
More detailed comments can be viewed in issue tracker with corresponding issue number.

I have written a python script to create template processors and edited the main CMakeLists.txt file to automatically generate list of modules in correct order consider dependencies.

PR #561, opened 2019-06-19
Iss475 - New Processor Template and Automating Cmake Module Listing

tomeichlersmith:
Same purpose as (closed) Pull Request #476 but I spent some time cleaning it up with git.

NewProcessor.py prints a new Analyzer/Producer Template given a name for the processor and a module to put it in.

CMakeLists.txt is edited to construct module list automatically. Look at comments in that file to see method.


tomeichlersmith:
I can separate out the python script and put it in ldmx-sw-scripts, but the other portion of this branch contains the additions to CMakeLists.txt that allow cmake to automatically construct a list of modules and then put them in dependency order.

omar-moreno:
The CMake changes we should keep.

tomeichlersmith:
Moved script to ldmx-sw-scripts/utils.

PR #566, opened 2019-07-03
Iss253 - merge mode: multiple input files to one output file

tomeichlersmith:
This patch allows for the user to have only one output file no matter what the number of input files is.

When switching input files, it copies the addresses of the new input file into the output tree so that the new input data is accessible (done by `EventFile::updateParent`). Then if new branches are being added and they already exist in the output tree, it resets the addresses of those new branches instead of re-creating them (done in `EventImpl::add` functions).

A boolean flag `isSingleOutput_` is used to tell the output `EventFile` that it should operate in merge-mode.

Other small changes were required so that the output tree is not changed when a new input file is reached.

On this branch, I was able to:
 - Still use ldmx-app with one output file for each input file
 - Combine all of the events in several (<10) input files into one output file with and without producers (`hcalDigis`, `ecalDigis`, and `simpleTrigger`)
 - Skim events from several input files and put passing events in one output file with `DummyAnalyzer` providing the keep/drop decision

On this branch, I was NOT able to:
 - Test merging capabilities on a large number (>100) of input files because my machine was running out of RAM. I am assuming that ROOT is holding all of the input files/trees until the end of processing (which could be another issue).
 - Check that the `RunHeader` was copied correctly. I don't know how to produce an event file with a different `RunHeader`, so I wasn't able to run tests on input files that have different `RunHeader`s.


tomeichlersmith:
I was able to find the source of the memory leak causing the machine to run out of RAM. The framework is not forgetting to close input files/trees, the processors are just poorly designed in terms of memory management. Most of them have a small memory leak (forgetting to `delete random_` in the deconstructor), but the worst offender was `HcalDigiProducer` which was forgetting to delete a new `HcalID` instance that was created *on every event*. After patching up `HcalDigiProdocuer`, I am now able to skim over millions of events and create one output file. 

Merging all of the events from several input files into one output file is possible now but the output file grows with the size of the input files (obviously), so it takes longer and uses much more memory.

tomeichlersmith:
I should also say that I wasn't planning on including the changes to `HcalDigiProducer` in this merge because, to me, it is a separate problem that most processors have.

awhitbeck:
do you want to just make issues for all of the memory leaks you find and the relevant people can make PRs?

tomeichlersmith:
I already patched all of the memory leaks I was able to find #569 , but I can separate those into different issues if they might need specialized solutions (I just use smart pointers and a few `delete`s).

omar-moreno:
@tomeichlersmith Can you point me to a python configuration that you use to test this?

tomeichlersmith:
I tested using a (slightly) modified `sample_skim.py` and `sample_digis.py` where I used more than one input file, but left only one output file. The modifications I made were mainly focused on testing other producers (like `simpleTrigger`). Would you like me to send you the exact config file I used?

PR #569, opened 2019-07-08
Iss568 - update memory management of processors

tomeichlersmith:
Unfortunately, we are stuck with using raw pointers when dealing with ROOT functions, but I was able to change other raw points to smart pointers that will deallocate memory automatically when they go out of scope.

I did not touch the sim modules because of the interaction with Geant4 (similar reasons as ROOT).

This also patched the memory leak found in `HcalDigiProducer` that was generating ~60 MB per 1000 events.


bechenard:
Hi Tom. Yes, I tend to declare objects in the constructor whenever possible instead of using smart pointers, that makes the code cleaner. The effect should be the same, though.

omar-moreno:
I prefer the use of smart pointers which, in my opinion, is a cleaner
solution.  Also, it helps prevents memory leaks.

bechenard:
I'm not sure I understand why. The data member will be automatically 
destroyed by the object's destructor when it goes out of scope, so there 
are no memory leak. Am I missing something here?

omar-moreno:
@bechenard Although your approach is correct and has been the standard prior to c++11, smart pointers has the advantage of moving the memory management burden away from the developer.  They also help alleviate a lot of the issues (e.g. memory leaks, pointer indirection, unsafe memory access) that raw pointers have.  For these reasons, the use of raw pointers is now strongly discouraged. Besides, it looks like raw pointers are going to be deprecated in c++20 anyway ;). 

The point is, if we would have used smart pointers from the beginning, we likely wouldn't have this PR ... 

Either @tomeichlersmith or I can update your raw pointers before merging.

bechenard:
Hi Omar,

Yes I completely agree, and I transformed raw pointers into smart 
pointers as well (except the one needed for ROOT). The point is that I 
replaced a pointer by a plain object initialized in the constructor. i.e.

In the hh file:
    private:
      HCalSectionID hcalID_

In the cc file
   HcalDigiProduced::HcalDigiProduced() : hcalID_(),..... {}

That way, everything is taken care for you. Since we don't copy the 
producer, I don't think there is any harm to have a slightly bigger 
object, and you can gain a bit in performance if the data members are 
declared in the right order. I'll let you look at the file and let me know.

Cheers,

Bertrand

tomeichlersmith:
I've undone the changes to `HcalDigiProducer` so that Bertrand's PR won't conflict, and then I merged the master branch in to make sure it was up to date.

omar-moreno:
Thanks Tom. I'll take a look today.

PR #579, opened 2019-07-25
New code and pkl file for BDT with containment variables

vdutta:
Adding containment variables and Gabrielle pkl file.


omar-moreno:
Great! I'll take a look.

omar-moreno:
@vdutta We should also make sure I can still use the old BDT's given the changes to ECalVetoProcessor.  I don't see any reason why we wouldn't be able to, but it's definitely worth a check.

PR #581, opened 2019-07-29
Iss576

bechenard:
Pull request for HCal fixes:
  - memory leaks fixed
  - smeared/quantized position fixed, this also affected attenuation calculation
  - noise to right/bottom side HCal added

Note that the left/right side HCal were inverted in the gdml file, so I changed the HcalSection ids to reflect that. In the next iteration, we should fix the gdml file and revert back.


tomeichlersmith:
I think this looks good! It will conflict with some of the changes I made in PR #569 , but I think these changes should take precedence.

awhitbeck:
this is on my to-do list today...

awhitbeck:
Code seems good.  I have one minor set of comments so far:  

-drop commented lines:
https://github.com/LDMX-Software/ldmx-sw/blob/iss576/EventProc/src/HcalDigiProducer.cxx#L43
https://github.com/LDMX-Software/ldmx-sw/blob/iss576/EventProc/src/HcalDigiProducer.cxx#L223-L245
https://github.com/LDMX-Software/ldmx-sw/blob/iss576/EventProc/src/HcalDigiProducer.cxx#L211-L215

I am going to run this on some events and just check some distributions...

bechenard:
Hi Andrew,
Could we keep the comments around line 211 and 223 (you can delete 43 and lines 223-225). This is what should be done if we were to use the fully reconstructed position from digitized info only. I'm keeping the position from the MC information for now to study the detector size, but we'll need to move towards this scheme in the future and I want to remember what should be done.
Thanks

awhitbeck:
sure that sounds reasonable.  I'll drop the other comment lines before I am done.

omar-moreno:
What's the status of this?

bechenard:
I made a PR, Andrew, Nhan or you should be reviewing it. Andrew said he 
was running some tests  yesterday but I haven't heard back from him.

- Bertrand

awhitbeck:
bah, sorry.  I'll finish this up this weekend.

bechenard:
Hi Andrew,

That would certainly be a good idea, though the ECal geometry changes, 
so the comparison might look weird.

Cheers,

Bertrand

PR #585, opened 2019-08-01
ADD LDMX detector with Ti target.

omar-moreno:


omar-moreno:
This resolves issue #585

Issue #266, opened 2017-04-12
Update SimApplication for Geant4  10.5 compatibility


JeremyMcCormick:
Possible that some of the initialization code may need to be rearranged for compatibility with Geant4 10.3, particularly for physics classes.

omar-moreno:
This has been resolved by PR #590

PR #590, opened 2019-09-13
Iss266 - Allows compiling ldmx-sw against Geant > 10.3

omar-moreno:
Although I checked that this worked locally, it would be great to get someone else to verify what I have already verified :).


omar-moreno:
This fixes issue #266

tomeichlersmith:
Where did you compile it locally? I don't have a Geant4 10.3 installation at UMN, so the quick way for me to check this would be to check on SLAC computers. If you've already checked on SLAC computers, I can install Geant4 10.3 at UMN but that would take some more time.

omar-moreno:
I have checked at SLAC computer so ideally it would be great if you can
check at UMN.  There is no rush on this PR so time isn't an issue.

If you decide to test at UMN, you might as well compile against the latest
version of Geant (10.5.1).

tomeichlersmith:
Is there a specific patch of Geant 10.5.1 that we are using (like how we were using 10.02.p03_v0.3)?

omar-moreno:
The latest is patch 1. You can find the latest version here:
https://geant4.web.cern.ch/support/download

omar-moreno:
Nice. If you have any kind  of macro available, can you just run a quick
test when you get a chance? Before, it wasn't even able to run over an
inclusive sample.

tomeichlersmith:
I was able to run over 100 events with a simple inclusive 4GeV electron macro. Any other filters/plugins that need to be checked? or is that the main hurdle?

bechenard:
I'll test it later today, I have Geant4 10.4 and 10.5 installed locally.

omar-moreno:
The main hurdle was whether the physics list was being loaded properly
which that you just verified to work by generating 100 events.  I guess one
last test you can try is compiling against our current version of Geant4
and seeing that it remains backwards compatible.  Again, this isn't time
crucial so it can be done whenever you have time.

omar-moreno:
Thanks!

bechenard:
Great, thanks Tom, one less thing to do for me!

Bertrand

PR #806, opened 2020-06-16
HistogramPool and NtupleManager (slight) enhancements

### LINKED ISSUE #799: HistogramPool and NtupleManager redesign
One issue that the user runs into when using the HistogramPool or the NtupleManager is that the names of the histograms or variables can be conflicting across different processors. In order to be robust against this, I have a plan to redesign the HistogramPool and NtupleManager.

### HistogramPool
Instead of being a singleton, the HistogramPool will be a class object that each processor has a copy of. Then the HistogramPool can be used to `create` and `fill` histograms in a processor without worry about conflicting with other processors. This class object will be created in the base `EventProcessor` class as a protected member variable so all derived processors have access to it.

### NtupleManager
Similar to HistogramPool, each processor can have it's own manager that has exactly one tree. This tree could be named the processor name. Then the `addVar` and `setVar` functions don't have to worry about conflicting with other processors. There will still be a central NtupleManager that is a singleton that handles when the trees are filled and/or cleared, but this central NtupleManager will not be seen by a normal user writing an analyzer.

---
Overall, I don't see a reason to _not_ do away with the singletons. One possible argument against this strategy is that there would be more memory overhead because there will be more maps of strings to histogram pointers, but the STL maps are light enough and the number of processors small enough that I don't think this is too large of a concern.

tomeichlersmith:
This resolves #799 .

## HistogramPool
- Changed singleton style to more modern structure (reference instead of pointer)
- Added helper class `HistogramHelper` that handles interface between user `EventProcessor` and `HistogramPool`
  - Each `EventProcessor` has its own `HistogramHelper` called `histograms_`.
  - Prepends name of processor to all ROOT histogram references so that there aren't any name conflicts.
  - `fill` methods for more-readable code (e.g. `histograms_.fill( 'hist' , 0.2)`).
  - Still has a `get` method for more specific/advanced configuration of the histograms
  - Handles a weights used when filling all histograms. This is in `HistogramHelper` so it does not affect histograms made by other processors. If left unset, it defaults to one. Helpful for using an event weight in histograms, e.g.
```
histograms_.setWeight( theEventWeight );
... other calculations ...
histograms_.fill( 'my_histogram' , 0.2 ); //filled with theEventWeight as weight
```
- Able to configure both 1D and 2D histograms from python
  - Can now use any combination of uniform and variable binning that you wish
  - Histograms configured in python are created before process starts (but after `configure`), so that user can modify them in `onProcessStart` (if they wish).
- Cleaned up DQM analyzers
- Removed unused `Histogram1DBuilder`
- Add in base `EventProcessor` class in `ldmxcfg.py` so that there is less code-repetition

## NtupleManager
- Changed singleton style to more modern structure (reference instead of pointer)
- Removed need for map (used `toupper` instead)

## Tests
- [x] Successfully ran `ldmx-test`
- [x] Successfully ran all DQM analyzers
  - Histograms are filled and look okay?
- [ ] **Actually make sure histograms are physically reasonable**
  - I did this for the ones I'm familiar with, but not comfortable with all of them

Still To-Do
---

### Not compatible with Python2
I _could_ make these changes compatible with Python2, but I don't want to.

### Need to patch analyzers in `ldmx-analysis`
@vdutta @omar-moreno I think y'all are the ones who wrote the analyzers currently in _ldmx-analysis_. This is just a warning that [I will be changing them](https://github.com/LDMX-Software/ldmx-analysis/tree/iss799) so that they are compatible with these updates. (no major changes, just instead of `ntuple_->` you do `ntuple_.` basically).

### DIFF
--- a/DQM/include/DQM/HCalDQM.h
+++ b/DQM/include/DQM/HCalDQM.h
@@ -31,7 +31,7 @@ namespace ldmx {
             HCalDQM(const std::string &name, Process &process);
 
             /** Destructor */
-            ~HCalDQM();
+            ~HCalDQM() { }
 
             /** 
              * Configure the processor using the given user specified parameters.
@@ -46,9 +46,6 @@ namespace ldmx {
              * @param event The event to analyze.
              */
             void analyze(const Event& event);
-
-            /** Method executed before processing of events begins. */
-            void onProcessStart();
 
         private:
 
--- a/DQM/include/DQM/PhotoNuclearDQM.h
+++ b/DQM/include/DQM/PhotoNuclearDQM.h
@@ -11,8 +11,6 @@ namespace ldmx {
 
     // Forward declarations within the ldmx workspace
     class Event;
-    class HistogramPool; 
-    class Process;
     class SimParticle;
 
     class PhotoNuclearDQM : public Analyzer { 
--- a/DQM/include/DQM/RecoilTrackerDQM.h
+++ b/DQM/include/DQM/RecoilTrackerDQM.h
@@ -30,7 +30,6 @@ namespace ldmx {
 
     class RecoilTrackerDQM : public Analyzer { 
 
-
         public: 
 
             /** Constructor */
--- a/DQM/include/DQM/TrigScintDQM.h
+++ b/DQM/include/DQM/TrigScintDQM.h
@@ -16,8 +16,6 @@
 //----------//
 //   ROOT   //
 //----------//
-#include "TH1F.h"
-#include "TH2F.h"
 #include "TVector3.h"
 
 //----------//
@@ -28,7 +26,6 @@
 #include "DetDescr/TrigScintID.h"
 #include "Tools/AnalysisUtils.h"
 #include "Event/EventDef.h"
-
 
 namespace ldmx { 
 
--- a/DQM/python/PN.py
+++ b/DQM/python/PN.py
@@ -55,3 +55,32 @@ photo_nuclear.build1DHistogram("hardest_n_theta",  "#theta of Hardest Photo-nucl
 photo_nuclear.build1DHistogram("hardest_pi_ke",    "Kinetic Energy Hardest Photo-nuclear #pi (MeV)", 400, 0, 4000)
 photo_nuclear.build1DHistogram("hardest_pi_theta", "#theta of Hardest Photo-nuclear #pi (Degrees)", 360, 0, 180)
 
+photo_nuclear.build2DHistogram("h_ke_h_theta", 
+                    "Kinetic Energy Hardest Photo-nuclear Particle (MeV)",
+                    400, 0, 4000, 
+                    "#theta of Hardest Photo-nuclear Particle (Degrees)",
+                    360, 0, 180)
+
+photo_nuclear.build2DHistogram("1n_ke:2nd_h_ke", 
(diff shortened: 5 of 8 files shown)


Framework/python/ldmxcfg.py
```diff
@@ -36,17 +37,111 @@ def __init__(self, instanceName, className):
...
+
+    This object contains the parameters that are necessary for a ldmx::Producer to be configured.
+
+    See Also
+    --------
+    LDMX.Framwork.ldmxcfg.EventProcessor : base class
+    """
+
+    def __init__(self, instanceName, className):
+        super().__init__(instanceName,className)
```
tomeichlersmith, line 144: This is the line that is not compliant with Python2. I could remove the base class and endure some code copying if we still wanted to support Python2.

scripts/ldmx-env.sh
```diff
@@ -66,5 +66,5 @@ docker pull ${_docker_hub_repo}:${LDMX_DOCKER_TAG}
 #   the container runs the users command from the same place that the user
 #   intended.
 ###############################################################################
-alias ldmx='docker run -i -e LDMX_BASE -v $LDMX_BASE:$LDMX_BASE -u $(id -u ${USER}):$(id -g ${USER}) ${_docker_hub_repo}:$LDMX_DOCKER_TAG $(pwd)'
+alias ldmx='docker run -it -e LDMX_BASE -v $LDMX_BASE:$LDMX_BASE -u $(id -u ${USER}):$(id -g ${USER}) $LDMX_DOCKER_TAG $(pwd)'
```
tomeichlersmith, line 69: Forgot the `t` before. This is not connected to this PR, but it is helpful when using the docker container because then the user can see the colors and use ctrl+C.

DQM/src/TrigScintHitDQM.cxx
```diff
@@ -16,128 +16,86 @@ namespace ldmx {
...
+        histograms_.create("x", "Hit x position", 1000, -100, 100);
+        histograms_.create("y", "Hit y position", 1000, -100, 100);
+        histograms_.create("z", "Hit z position", 1000, -900, 100);
+        
+        histograms_.create("pe", "Pe deposition in a TrigScint bar", 1500, 0, 1500);
+        histograms_.create("hit_time", "TrigScint hit time (ns)", 1600, -100, 1500);
+        
+        histograms_.create("id_noise", "Channel ID of noise hit", 100, 0, 100);
+        histograms_.create("pe_noise", "Pe deposition in a TrigScint bar noise hit", 1500, 0, 1500);
+        histograms_.create("n_hits_noise", "TrigScint noise hit multiplicity in the pad/event", 100, 0, 100);
```
bryngemark, line 36: So, can I take it that you have achieved what I was aiming for, namely, if I in my config specify 
`tag = ldmxcfg.Analyzer("Tag", "ldmx::TrigScintHitDQM")`
and
`up = ldmxcfg.Analyzer("Up", "ldmx::TrigScintHitDQM")`,
the histograms will end up getting names starting with Tag and Up, respectively, keeping them unique?

I guess what is unclear to me is if you meant that it's the name of the instance of the processor, or the name of the processor class, that gets prepended to all histogram names.

tomeichlersmith: Yes, the `HistogramHelper` class prepends the instance name of the processor (so Tag or Up) to the histograms.

PR #807, opened 2020-06-17
Modify Python Configuration Procedure

### LINKED ISSUE #780: Expand python modules with more helpful/robust functions
Our python modules currently are very bare-boned, and I think we can expand them to make the use of ldmx-sw much easier and robust. I am planning to add more interfaces that will make the python configuration less likely to make small mistakes.

Here is a list of planned improvements:
- [x] Add a directory of event files to the process as input files
- [ ] A few pre-written drop/keep rules that can be added
      - e.g. Only Reco, Sim and Reco (no digi), everything
- [ ] Functions to add different chains to the process (e.g. digi pipelines or cutflows)
- [x] Derive python classes with member functions to help set the parameters
      - **Only have the parameter name hardcoded once in python and once in C++**
      - _Biggest Change_: Each processor, primary generator, user action could have its own derived python class with helper functions. Probably won't need to do the classes with a smaller number of parameters or are less complicated.
      - This may be the change the forces us to abandon python2 support (derived class syntax changes from python2 to python3)
      - Put **all** defaults in these derived classes so printing the python configuration is more transparent
- [ ] Try to maintain backwards compatibility for the user's run scripts, but _no promises_.

Isn't this just more code nonsense? _Yes_, but these changes will make ldmx-sw easier to use for newbies and help it be more robust for all of us. Hopefully will drastically reduce the number of small errors like mispelling parameter names.

tomeichlersmith:
This resolves #780 --- I don't complete all the goals there, but it is time to put it in.

 ## Big Idea
Python keeps track of class member variables in the `__dict__` member variable which maps the name of the variable to its value. Each class that needs python configuration will have a mirror class in the python that names the parameters that it needs. Then those parameters are accessed in the C++ using the names from the python.

#### C++
The C++ changes are relatively straightforward. In `ConfigurePython`, I use a `static` function called `getMember` which takes an input python object and extracts all of its member variables into a `std::map<std::string,std::any>` (similar to what Omar wrote before). This is done recursively so that classes can have class members with their own parameters.

The other C++ changes are places where I needed to change the configuration slightly to match this new style. We don't need the structures `ProcessorClass`, `HistogramInfo`, or `Class` anymore because the member variables `instance_name`, `class_name`, etc... are included in the `__dict__` member variable dictionary. I also needed to update some parameter names so that they don't cause trouble in python (the ones with `.` in them).

#### Python
The majority of these changes involves defining a python class for _each_ Producer/Analyzer that needs configuration variables. I also implemented an automatically-generated python module called `include` for each of our modules that attaches the module library to the list of libraries to be installed. This means in the constructor for the processor, you can run, for example
```python
from LDMX.DQM import include
include.library()
```
which means the user doesn't need to interface with the `Process.libraries` member unless they are working with a class that doesn't do this in its constructor.

## Tests
I was able to run all the **most** processors with this new configuration style.
There are some edge-case processors (e.g. PnWeightProcessor) that I'm not sure how to run, so I wasn't able to fully test that it is working. 

## Before Merging
- [x] Document C++ more thoroughly, specifically `ConfigurePython` and `MyProcessor`
- [x] Document Python changes
- [x] Make sure _all_ of the parameters in ldmx-sw have either (1) a corresponing member variable in a Python class or (2) have a default given in the C++.
- [x] Add a page explaining the processor structure to the github-pages site
- [x] Notify everyone that script re-writing is incoming
- [x] Update tutorial scripts and explanations to this big change in interface
- [x] Update NewProcessor.py template-printing script
- [x] Update ldmx-analysis repo with changes to interface

### DIFF
--- a/EventProc/include/EventProc/PnWeightProcessor.h
+++ b/EventProc/include/EventProc/PnWeightProcessor.h
@@ -5,6 +5,8 @@
  * @author Alex Patterson, UCSB
  * @author Omar Moreno, SLAC National Accelerator Laboratory
  * @note
+ * This processor is not used anymore and is only left here for documentation purposes.
+ *
  * PnWeightProcessor calculates an event weight which is persisted as a PnWeight
  * object. This weight is based on the kinematics of a photonuclear reaction and 
  * is intended to correct for the overproduction of events with high-momentum,


tomeichlersmith:
I have changed the behavior of `Parameters` slightly in order to avoid mistakes. Now a parameter either requires a default to be passed in the C++ or requires there to be a parameter in the python.

The script I used to test these changes is [linked here](http://homepages.spa.umn.edu/~eichl008/basic.py)

tomeichlersmith:
@omar-moreno some questions:

1. Do you want me to have a `Biasing` python configuration class? I think this could help enlighten some things, but might hurt others.
2. I can merge in the `dumParameters` method from PR #802 into this PR, so it'd only be one merge. Should I do that?

omar-moreno:
1) Let's leave things as they are now.
2) No, let's keep them separate.  We still haven't tested whether we can
push to  RUCIO which is part of that PR.

omar-moreno:
PNWeightProcessor is deprecated so don't waste your time making a config for it.  It's only in the repo for documentation purposes.

PR #808, opened 2020-06-17
Clearing the track map also deletes the trajectory.

omar-moreno:


tomeichlersmith:
Yay! You found it! (or one of them lmao)

PR #810, opened 2020-06-22
Iss775 - Fixes random seed IO issue

### LINKED ISSUE #775: Remove I/O access for RandomEngine saving and restoration
### Description:
Currently RandomEngine object is serialised to disk, then read from disk, then written to header in string form.

### Idealised solution (better):
Serialise directly the object `*G4Random::getTheEngine()` to the header from `EventUserAction`. Have a function that serialises it back to the `*G4Random::getTheEngine()` object.

### Alternative proposed solution (less good):
Use the current implementation serialise it to string via stringstream then write it to file as a string. This might be potentially bad if serialisation doesn't work in the way that is expected.

The hotfix is as follow

```
std::ostringstream ss;
ss << *G4Random::getTheEngine();
MyEngineStatusString = ss.str(); 
MyEngineStatusString.erase(0, MyGeneratorString.find("\n") + 1); // this line is added to match the output of the G4Function
RootPersistencyManager::getInstance()->SaveEngineStatusToString(MyEngineStatusString);
```

in `EventUserAction` at `BeginOfEventAction` and adding the methods `SaveEngineStatusToString` and `ReturnEngineAsString` to `RootPersistencyManager`, which would be equivalent to passing it to the persistency manager, more precisely at

```
std::string seedString = getEventSeeds();
eventHeader.setStringParameter("eventSeed", seedString);

```
By replacing 
`getEventSeeds();` 
with 
`RootPersistencyManager::getInstance()->ReturnEngineAsString();`
which would then write it as usual instead of making the call to file.

The only thing to pay attention to is if my offset by erasing the first line is correct, but it should be

## Corollary

The lines

```
if ( dynamic_cast<RunManager*>(G4RunManager::GetRunManager())->useRootSeed() ){
        G4Random::restoreEngineStatus("tmpEvent.rndm"); // this line will be needed to read in a set of seeds
}
```

Then could be re-written to never access disk because one can drop the Engine back into place by just adding a method that would read the string from the root file as such

```
std::istringstream iss(RestoreEngineStatusFromString);
iss >> *G4Random::getTheEngine();
```

=> This part is a bit of a bit of a guesswork since it wasn't properly tested

This de facto removes the `SimApplication/RootSimFromEcalSP`disk I/O.

## Sidenote:  
having maybe a good serialisation strategy to drop the object directly into file instead of string might be the best option here, albeit it is not as straight forward as the other one

Note that what I did could be probably also be done [by using the functions available here](http://www.apc.univ-paris7.fr/~franco/g4doxy4.10/html/_random_8cc_source.html#l00165)

Namely probably the `setTheEngine`, or `restoreFullState`, but as far as I understand this is identical with a few more steps.

```
165 HepRandomEngine * HepRandom::getTheEngine()
166 {
167   return theDefaults().theEngine.get();
168 }

170 void HepRandom::setTheEngine (HepRandomEngine* theNewEngine)
171 {
172   theDefaults().theEngine.reset( theNewEngine, do_nothing_deleter() );
173 }

185 std::ostream& HepRandom::saveFullState ( std::ostream & os ) {
186   os << *getTheEngine();
187   return os;
188 }

190 std::istream& HepRandom::restoreFullState ( std::istream & is ) {
191   is >> *getTheEngine();
192   return is;
193}
```

Note that the serialisation method to file is safe by construction but doubling the memory footprint since it takes the double from the Engine and converts it to two longs then input them into file. If that part had to be fixed, then indeed one would need to touch Geant4 directly and this is a bit out of scope at that point.

```
271 std::ostream & HepJamesRandom::put ( std::ostream& os ) const {
272   char beginMarker[] = "JamesRandom-begin";
273   os << beginMarker << "\nUvec\n";
274   std::vector<unsigned long> v = put();
275   for (unsigned int i=0; i<v.size(); ++i) {
276      os <<  v[i] <<  "\n";
277   }
278   return os;  
279 }
280 
281 std::vector<unsigned long> HepJamesRandom::put () const {
282   std::vector<unsigned long> v;
283   v.push_back (engineIDulong<HepJamesRandom>());
284   std::vector<unsigned long> t;
285   for (int i=0; i<97; ++i) {
286     t = DoubConv::dto2longs(u[i]);
287     v.push_back(t[0]); v.push_back(t[1]);
288   }
289   t = DoubConv::dto2longs(c);
290   v.push_back(t[0]); v.push_back(t[1]);
291   t = DoubConv::dto2longs(cd);
292   v.push_back(t[0]); v.push_back(t[1]);
293   t = DoubConv::dto2longs(cm);
294   v.push_back(t[0]); v.push_back(t[1]);
295   v.push_back(static_cast<unsigned long>(j97));
296   return v;
297 }
298 
```

```
106 void HepJamesRandom::saveStatus( const char filename[] ) const
107 {
108   std::ofstream outFile( filename, std::ios::out ) ;
109 
110   if (!outFile.bad()) {
111     outFile << "Uvec\n";
112     std::vector<unsigned long> v = put();
113     for (unsigned int i=0; i<v.size(); ++i) {
114       outFile << v[i] << "\n";
115     }
116   }
117 }
```

But modification of that behaviour would indeed be a modification of Geant4 at this point.

omar-moreno:
Resolves #775

### DIFF
--- a/SimCore/src/RootPersistencyManager.cxx
+++ b/SimCore/src/RootPersistencyManager.cxx
@@ -122,7 +122,6 @@ namespace ldmx {
         // is then extracted and saved to the event header.   
         std::ostringstream stream;
         G4Random::saveFullState(stream);
-        //std::cout << stream.str() << std::endl;
         eventHeader.setStringParameter("eventSeed", stream.str());
     }
 


omar-moreno:
The legacy seed looks like the following: 
`Uvec
2226355706
1070643019
3221225472
1071977356
3758096384
1070733755
2147483648
1072368577
3758096384
1069802555
0
1067090816
0
1071992817
2684354560
1071766538
536870912
1072078996
2147483648
1071919600
1073741824
1064547600
0
1070796147
3221225472
1072399035
3221225472
1072391801
2147483648
1072003677
`

While the new seed is stored as follows: 

`JamesRandom-begin
Uvec
2226355706
1072422976
3221225472
1072318341
2684354560
1071148774
2147483648
1072314965
536870912
1072637123
1610612736
1072429069
0
1072386281
3221225472
1072662700
1610612736
`

Need to understand if the addition of the `JamesRando-begin` in the new seed breaks backwards compatibility.

omar-moreno:
As it stands, this change will not allow us to resim old files because the format of the seed is slightly different i.e. the new seed has a beginTag ("JamesRandom") and the old seed doesn't.  However, if we ever want to resim old files, we can add some code that adds the tag to the beginning of the seed so it can be used with the current seed restoration scheme.  That can be done using a different issue.

PR #811, opened 2020-06-22
Iss643 dev - Boost logging. 

omar-moreno:
This cherry picks the minimal changes pushed into iss643.  That branch is really outdated and still needs some work so it's best to start clean.

### DIFF
--- a/Tools/include/Tools/Logger.h
+++ b/Tools/include/Tools/Logger.h
@@ -1,12 +1,5 @@
-/**
- * @file Logger.h
- * Interface for ldmx-sw to boost logging
- *
- * @author Tom Eichlersmith, University of Minnesota
- */
-
-#ifndef EXCEPTION_LOGGER_H
-#define EXCEPTION_LOGGER_H
+#ifndef TOOLS_LOGGER_H
+#define TOOLS_LOGGER_H
 
 /**
  * Necessary to get linking to work?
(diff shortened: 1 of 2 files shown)


omar-moreno:
@tomeichlersmith Let's finish up putting in the necessary logging classes without changing all processors to make use of the logging.

tomeichlersmith:
Is there a reason you remved the `enableLogging` and `ldmx_log` macros at the end of `Logger.h`?

omar-moreno:
It's likely that they weren't cherry picked.  Let me check.

tomeichlersmith:
The other thing that I noticed is that you copied over no changes to the application source file. (formerly ldmx_app.cxx). This setup is necessary for logging to function anywhere in ldmx-sw, but I could see delaying those changes for a later PR where we actually implement the cleanup of logging in most/all of ldmx-sw.

omar-moreno:
I had a question related to that.  Why not setup the configuration from the
python config?

tomeichlersmith:
I wanted to be able to have all exceptions to be included in the logging. Since we have some exceptions in the `ConfigurePython` class, it needed to be separate.

If we are okay, with the `ConfigurePython` having exceptions _not_ included in the logging, then we could definitely have the logging configured in the python.

omar-moreno:
I think that's fine.

tomeichlersmith:
Ok, I think that is complicated enough, we should put it on a different issue and I can implement that over the next few weeks.

PR #819, opened 2020-06-24
extraction again, this time with workflows

tomeichlersmith:
We just wanted to be careful and make sure the history was preserved as best as possible.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,12 @@
 [submodule "MagFieldMap"]
 	path = MagFieldMap
-	url = git@github.com:LDMX-Software/MagFieldMap.git
+	url = https://github.com/LDMX-Software/MagFieldMap
 [submodule "Framework"]
 	path = Framework
-	url = git@github.com:LDMX-Software/Framework.git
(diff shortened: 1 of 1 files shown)


tomeichlersmith:
@omar-moreno Maybe I'm just getting cold feet, but we could hold off on this extraction until v2.1 release or some other larger milestone. I could work on getting ldmx-analysis pulled in as a submodule and that can be our experimental ground.

omar-moreno:
I think it's fine if we do it now. I'm actively testing this and it seems
to be working fine.

tomeichlersmith:
The biggest issue I'm running into is I can't figure out a way to pull down the submodules inside of the GitHub action. I think it is important to keep the GitHub action so that we can have the building and testing be automatic, but I haven't been able to figure out how to get the submodules to be cloned correctly.

omar-moreno:
We can switch to Travis if that's a big issue.

tomeichlersmith:
Ok, sounds good to me

omar-moreno:
This will be done in a different PR.

PR #820, opened 2020-06-25
Patches for Target EN Biasing and Filtering

tomeichlersmith:
My commit messages explain pretty well what I did, but in short:
- Don't configure the volume and process names for the TargetENProcessFilter from python, they need to be very specific names so they should be hardcoded.
- Patch some of the biasing parameters in `simulator` to be the correct type (Side Note: Why is the biasing factor an integer?)


tomeichlersmith:
Moved these patches to PR #830

PR #829, opened 2020-06-26
Patch simulator biasing 

bryngemark:
The mapping from biasing process to biasing particle was never actually read, only set up.


bryngemark:
hahaha oh no! I just wanted one line in simulator.py to change :D please cherry pick it?

tomeichlersmith:
Should be just these this change:

tomeichlersmith:
The reason this is so messy is because the change you made is in a different git repo now. I have put the required change into the [SimCore module](https://github.com/LDMX-Software/SimCore).

You can access this change by doing the following on ldmx-sw master:
```
cd SimCore;
git pull origin master
```
Or, this might also work (not sure):
```
git submodule update --init --recursive
```

PR #832, opened 2020-06-29
allow for multiple tries for each event during production mode

tomeichlersmith:
I have found this enhancement very helpful when I am working with the biasing/filtering nonsense that is necessary with Geant4 and efficient simulation. When biasing isn't too low (or filtering too harsh), this means when you ask for N events, you get N events, although you might have to wait for (max-tries-1)*N events to be filtered and not saved.

I also moved the `getParameter` calls for Process member variables from `ConfigurePython::makeProcess` to `Process::Process`. This just cleans up the `ConfigurePython` class and follows the idea that the class that uses a parameter is the one that calls `getParmaeter`.

### Other Modules
When using max-tries (or anything that could abort an event), it is helpful to keep track of both the number of events "tried" and the number of events "selected" (i.e. non-aborted). These counts can be added to the `RootPersistencyManager` and `Simulator` classes in the SimCore module.


tomeichlersmith:
I have made the necessary SimCore changes and put them in [its own PR](https://github.com/LDMX-Software/SimCore/pull/1). Again, these _don't need to be merged together_, but it would be nice to merge the SimCore PR first and then add those changes to this PR.

tomeichlersmith:
@omar-moreno Unless you want this feature for your own production workflow, we can put this on the backburner until after the 2.1.0 release.

omar-moreno:
We will add it for the 2.1 release.  I'll get to it today.

PR #837, opened 2020-07-02
Update env script to deal with locally built containers

bryngemark:
Basically, a third argument that can be used to pull from the right repo:
dev, or pro, or 'local' in which case nothing is pulled and a local container can be used (as specified in arg 2).
No checks are implemented to ensure that these are the only names passed. If the user messes this up, pulling the container will fail. I believe the error messages in that case are enough.

In addition, some tweaks to the help command, to cover the most common attempted help strings, and document this additional arg.

### DIFF
--- a/scripts/ldmx-env.sh
+++ b/scripts/ldmx-env.sh
@@ -10,20 +10,19 @@
 if [ -z $3 ]
 then
     _repo_name="dev"
-    #this is the name of the dockerhub repository
-    #   it shouldn't change
-    # except it should when we want a production container
+    # this is the name of the dockerhub repository
+    # it should change when we want a production container 
     # also, if it's set to "local", skip downloading at all and attempt to use a locally built container
 else
     _repo_name="$3"
 fi
-export _docker_hub_repo="ldmx/$_repo_name"
+export _docker_hub_repo="ldmx/${_repo_name}"
+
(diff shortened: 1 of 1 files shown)


scripts/ldmx-env.sh
```diff
@@ -7,9 +7,24 @@
...
+else
+    _repo_name="$3"
+fi
+export _docker_hub_repo="ldmx/$_repo_name"
+#compatibility tweak. we could also suggest that all local builds are named with prefix convention ldmx/local: instead of ldmx/dev:
+if [ ${_repo_name} = "local" ]
+then
+    export _docker_hub_repo="ldmx/dev"
+fi
+echo $_docker_hub_repo
```
tomeichlersmith, line 26: Could you remove this line?

tomeichlersmith: It just looks like a line you used in debugging and would just clutter the output. If you think it is important to show the user the docker hub repo, I think thats cool. Then I would ask for you to fill out this line with a short sentence about what the script has decided.

bryngemark: sorry i forgot to push! now a few things should be prettier, like this intermediate check line being gone

tomeichlersmith: Thank you!

PR #842, opened 2020-07-03
Make trigger output collection name settable

bryngemark:
Until now, only one output trigger collection was possible since the output collection name was hardcoded. Having the output collection name as a settable parameter allows for comparisons across trigger algorithm settings with multiple instances of simpleTrigger in the same reconstruction round. 

Also made input collection name settable for good measure even if I imagine that this will not change very frequently.


bryngemark:
Since testing is broken: I have successfully built and run this and verified that the collection name gets set to the default or whatever I want.

PR #847, opened 2020-07-07
Iss841 - Move Exception class to Framework

omar-moreno:


omar-moreno:
I will fix how the catch header gets included in another PR.  For now, I focused on removing the Exception module.

omar-moreno:
This will require the SimCore PR to be merged first.  Once that is in, then I'll update this PR to include the lastest version of SimCore.

PR #852, opened 2020-07-08
import dictionary file inside Event module

### LINKED ISSUE #851: Add python module to load Event dictionary
A lot of users use python to load event files and do some quick analyses and/or just investigate the event file. Adding a python module in the `Event` module that can do the complex nonsense of loading the Event dictionary into ROOT will be helpful.

tomeichlersmith:
This resolves #851 

Marked as draft because I need to test it and make sure it works.

### DIFF
--- a/Event/python/dictionary.py.in
+++ b/Event/python/dictionary.py.in
@@ -1,0 +1,13 @@
+"""Import EventModel dictionary into PyROOT
+
+This needs us to import ROOT and then we can load
+the Event library which also pulls in the Event dictionary.
+
+Examples
+--------
+Use the following line in your python script to inform PyROOT of our Event Model:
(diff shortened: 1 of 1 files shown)


tomeichlersmith:
Success! Needed to have PyROOT built with python3 inside of the container, but then it works:
```
tom@zuko:~/ldmx$ ldmx python3
Python 3.6.9 (default, Apr 18 2020, 01:56:04) 
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from LDMX.Event import dictionary
>>> import ROOT
>>> from LDMX.Event import dictionary
>>> eh = ROOT.ldmx.EventHeader()
>>> eh.isRealData()
False
>>> quit()
```

PR #853, opened 2020-07-10
Iss850 - Update CMake configs

omar-moreno:


Event/CMakeLists.txt
```diff
@@ -12,40 +21,56 @@ string(REGEX REPLACE "\n$" "" GIT_SHA1 "${GIT_SHA1}")
...
+# Generate the ROOT dictionary.  The following allows the use of the macro 
+# used to generate the dictionary. 
+include("${ROOT_DIR}/RootMacros.cmake")
+
+# Unfortunately, the Event headers need to be included globably for the 
+# dictionary to be generated correctly. 
+include_directories(${PROJECT_SOURCE_DIR}/include)
+
+# Generate the ROOT dictionary
+root_generate_dictionary(EventDic ${PROJECT_SOURCE_DIR}/include/Event/EventDef.h LINKDEF ${PROJECT_SOURCE_DIR}/include/Event/EventLinkDef.h)
```
tomeichlersmith, line 41: The generated root dictionary references the files that you input here, so you should input the installed files. That way, we can delete the source and build files and still use the installed event dictionary.

omar-moreno: Is there a reason why you would delete the src files? Is that what you are doing in the container?

tomeichlersmith: Yes, that allows us to remove the source and build files before compressing the container. So it isn't necessary, just helpful.

PR #859, opened 2020-07-14
More updates to the environment scripts

### LINKED ISSUE #858: Use flagged-parameters instead of positional parameters in ldmx-env.sh
This will be more difficult to code up, but will be very useful from a user point of view.

Currently, there are three parameters:
1. Where ldmx base directory is
2. What docker-hub repo should the image come from: `dev`, `pro`, or `local`
3. What image-tag should be used

The flags could be:
1. `--base`, `-b`
2. `--repo`, `-r`
3. `--tag`, `-t`

This leaves `--help`, `-h` available for a help message.

~~I will implement this using `getopts` as suggested in [this article](https://alvinalexander.com/linux-unix/shell-script-command-line-arguments-linux-unix/), but I will need to request someone with a Mac to test that it works with their shell.~~

tomeichlersmith:
- Patched up a bug with the `ldmx` function
- Switch command parsing to flag-based instead of positional: resolves #858 

@bryngemark You have a laptop with MacOS on it correct? Could you test this to make sure my bash-nonsense isn't Linux-only.

### DIFF
--- a/scripts/ldmx-env.sh
+++ b/scripts/ldmx-env.sh
@@ -171,7 +171,6 @@ function ldmx-container-config() {
 function ldmx() {
     #store current working directory relative to ldmx base
     _pwd=$(pwd -P)
-    _current_working_dir=${_pwd##"${LDMX_BASE}/"} 
     cd ${LDMX_BASE} # go to ldmx base directory outside container
     if hash docker &>/dev/null; then
         docker run \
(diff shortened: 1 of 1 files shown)


tomeichlersmith:
I abandoned `getopts` because it doesn't support long args e.g. `--help`. The building tags using `local` is not enforced, but it is rather easy to add another tag ot any image using:

bryngemark:
sounds like a good addition & solution!

tomeichlersmith:
Still need to test this at SLAC --> putting it on draft because SLAC computer is taking long

PR #862, opened 2020-07-15
Update some ecal geo

### LINKED ISSUE #861: ECal Geometry small bugs
- Gap between module hexagons should be 1.5mm instead of only 1mm
   - Just need to update constants.gdml
- center-to-**corner** radius of module hexagons is set to 85mm in `constants.gdml` while the center-to-**flat** radius is set to 85mm in EcalHexReadout.
   - Need to update EcalHexReadout and all the uses of it.

tomeichlersmith:
This resolves #861 .

Dependent on [SimCore#19](https://github.com/LDMX-Software/SimCore/pull/19)


vdutta:
Hi @tomeichlersmith, could you comment on what the bug actually was? Was the readout area being incorrectly set because the radii were set differently in the gdml vs the EcalHexReadout?

tomeichlersmith:
The readout area is set in the gdml so that wasn't the issue. The radii of the hexagons are used to partition up the hits into the different cells, so since the radius was set wrongly, the cells were not the correct size (they were larger than expected).

PR #879, opened 2020-07-31
Iss876 - Update to DetectorID structure

tomeichlersmith:
@jmmans Can you link the issues that this resolves?

### One important thing to note:
Previously, the ID -> sub-IDs (e.g. layer ID) decoding was done in the Event objects as well as in the DetectorID classes. This led to some issues in processors that mixed which object did the decoding because they didn't match anymore. I decided to remove any ID decoding from the event objects in order to remove this possible confusion.


tomeichlersmith:
We will need to merge in PRs in the submodules and then update the submodule refs before this will work.

bryngemark:
I ran the Trigger Scintillator DQM plots. I will push a fix to the DQM code (where a (broken) check on the existence of a collection that I thought I had deleted apparently made its way back in). 

Apart from that: this all looks normal/unbroken to me. 


TrigScintDigiDQM_id
TrigScintDigiDQM_y
TrigScintDigiDQM_x
TrigScintDigiDQM_n_hits
TrigScintDigiDQM_pe

PR #880, opened 2020-08-04
Conditions subsystem first version ready for integration

jmmans:
The first version of the conditions system is ready to be integrated.  This pull request provides the following:

* The basic core infrastructure for providing conditions information like pedestals, gains, electronics maps, and any other information which is run-dependent not event-dependent.

* An implementation of homogeneous tables keyed by raw detid.  The tables are homogeneous in that they must contain data columns which all either doubles or integers. More complex tables are possible, but not addressed by this pull request.

* An implementation of streaming (store/load) for the tables into CSV files which can be read from either local files or via http.

This pull is developed on top of iss876.


tomeichlersmith:
Since this is dependent on PR #879 , I am going to convert this to a draft until that PR has been merged.

tomeichlersmith:
Still a draft because we need to get the Framework PR merged in.

tomeichlersmith:
Still dependent on https://github.com/LDMX-Software/Framework/pull/14

tomeichlersmith:
@jmmans How can we test this? Is there a simple configuration file that we can run? Would it be possible to write a test that would be automatically run?

Issue #657, opened 2020-03-24
Make the Gap a settable parameter in EcalHexReadout


omar-moreno:
For the simulation, this should be settable via GDML through the use of AuxInfo.  For the reconstruction, it should be passed as a parameter. 

Currently, the default is set to 0.  If a gap other than 0 is specified in the GDML and not propagated through to the hex readout, the simulation will fail.

Issue #873, opened 2020-07-25
Improve Realism of Cellular Layout in EcalHexReadout


tomeichlersmith:
We are not very familiar with how the hex modules will be built, but we know several details that can be implemented in EcalHexReadout to improve the realism of the cellular layout that aren't implemented right now.

Currently, the cellular layout is a uniform hexagonal grid across the module: [Cell_ID_Cell_Position_Map.pdf](https://github.com/LDMX-Software/ldmx-sw/files/4976162/Cell_ID_Cell_Position_Map.pdf)

This positioning has several flaws that need to be rectified:
- [x] Only 397 cells are numbered, there should be 432 <- PR #881 
- [ ] Mousebites: The cells do not penetrate into the corners of the module hexagon to leave room for mounting
- [x] All the cells are shifted by one half of a cell width so that the center of the module is positioned on the border between cells instead of at the center of a cell (198) <- PR #881 
- [x] There needs to be a cell ID --> trigger cell ID mapping. The trigger cells are groups of 9 (3x3) DAQ cells that are added together during the trigger digitization. The outline for them can be seen [in these slides](http://homepages.spa.umn.edu/~eichl008/digi-pipeline/bkgrd/FD_HGCROC3_190702_TowardsV3.pdf). <- PR #916 
- [x] Incorporate a cell ID <-> UV coordinate translation. This will be helpful for defining trigger cell groupings as well as defining neighboring cells. (UV coordinates are coordinate axes parallel to two adjacent sides of the hexagon, so axes that make an angle of 120 degrees) <- PR #903 
- [ ] The cell ID with respect to the x-coordinate will flip between sensitive layers because the modules will be flipped when mounted to make the full layer "sandwich".

@omar-moreno This is _not_ something I can implement in the short-term, so don't expect this change before the next production. I just want to save my notes so that I don't lose them.

@jmmans Can you attach some documents about the HGCAL module geometry? I remember you showing me some diagrams that helped show the cellular layout, but I can't find them.

jmmans:
I've attached the high-density sensor information which is available.

![191122_432ch](https://user-images.githubusercontent.com/4933600/88490029-16865800-cf5e-11ea-9d07-0a3e26d11121.png)

Tom -- you have a script which makes the map drawing.  Does this script work "in general"?  I'm thinking to try to address the 432 cell issue and the shifted cell issue, which are tied together.

tomeichlersmith:
[...]

PR #881, opened 2020-08-04
Iss873 - Improve realism of ecal hex readout and configure through python

tomeichlersmith:
This depends on PR #879 and resolves issue #657 while resolving some of the realism-issues #873 .

I have also written an executable in the DetDescr module that prints the local CellID <-> CellPosition mapping into a pdf image.


tomeichlersmith:
I am marking this as ready for review, but I am unable to test it without the cmake updates in PR #883 that make compiling and running in the container possible.

tomeichlersmith:
We will need to merge in https://github.com/LDMX-Software/SimCore/pull/23 and https://github.com/LDMX-Software/Ecal/pull/10 and then update the submodule refs here.

tomeichlersmith:
I have created a map of the local Cell ID to position relative to module using the executable that is in this PR: 
[Cell_ID_Cell_Position_Map.pdf](https://github.com/LDMX-Software/ldmx-sw/files/5104273/Cell_ID_Cell_Position_Map.pdf)

PR #884, opened 2020-08-14
Trigger scintillator hit clustering and tracking

### LINKED ISSUE #755: Trigger scintillator hit clustering
Hit clustering is the first step towards electron counting.

An algorithm clustering up to three neighbouring channel hits to a cluster and adds it to the Event bus. 

I have something that works pretty well, and has the first-order complexity of a simple scheme for splitting clusters that are larger than three cells.

### LINKED ISSUE #756: Trigger scintillator tracking
A simple straight-track finder for the trigger pads could improve electron counting precision by adding spatial information to cluster counting. 

Further down the line, options like allowing two-hit tracks (out of three possible...) will be explored. 

For now, I have something that takes seeds from the collection of clusters from one pad (which one is configurable) and makes tracks from clusters in the other pads (which ones is, again, configurable, but initially I'm assuming we want two), based on a width tolerance, and using track residual minimization for disambiguation.

### LINKED ISSUE #760: Trigger scintillator noise hits can get duplicated
**Describe the bug**
Every now and then, there are two noise hits in the same channel.

**To Reproduce**
This can be seen by listing trigger scintillator reco hit IDs event by event, for instance I found it as part of my cluster producer validation.

The solution is to make sure the noise hits get pushed to the hit map upon creation. 

Andrew says, this should happen somewhere around (currently) [this line](https://github.com/LDMX-Software/ldmx-sw/blob/master/EventProc/src/TrigScintDigiProducer.cxx#L164).

bryngemark:
A rather simple (given the simple geometry) clustering and (straight-line) tracking algorithm for the trigger scintillator electron counting. Would love to have this in the upcoming release.

It is based on the logic that any straight path beam electron should pass through two bars, adjacent in channel number: one per front/back layers of a pad, and that its y coordinate (channel number) should be close to constant (bending is in x). 

And a fix for the duplicated noise hit IDs. 

Based on the recently merged det ID fix branch (#879 ). Related issues: closes #755 , closes #756 , and closes #760 .


awhitbeck:
does this include Jeremy's detID fixes?  Is there a separate PR for that which is still pending?

bryngemark:
It does. I added this stuff to the branch I used for validation of his changes. The algorithms are independent of that but the duplicate noise ID fix was added to the digi code as it stands after his fixes (the actual fix is one line of code though, and again doesn't explicitly depend on his ID fixes).

bryngemark:
Or wait that's not what you asked – you were asking if _this_ is where Jeremy's changes get merged. I believe they already have been. The commit history is from that branch though since I pushed a fix to the DQM code.

omar-moreno:
Yes, Jeremy's changes are already in master.

tomeichlersmith:
@bryngemark Could you rebase your changes onto the current master? That will help us see what changes you are actually making. _Warning:_ rebasing is difficult so be careful and you may need to re-do some patches/commits that occurred before you merged in Jeremy's ID changes.

bryngemark:
ok, should be done, easier now?

tomeichlersmith:
Yes, much better, thank you!

Event/include/Event/TrigScintCluster.h
```diff
@@ -0,0 +1,191 @@
...
+#include <iostream>
+#include <set>
+
+// ldmx-sw
+#include "Event/EventConstants.h"
+#include "Event/TrigScintHit.h"
+#include "Event/EcalCluster.h"
+// these two can't be found..?
+//#include "Framework/ParameterSet.h" // Needed to import parameters from configuration file
+//#include "Framework/EventProcessor.h"
```
tomeichlersmith, line 26: You can delete these. The Event module _should not_ depend on the Framework module, so if it compiles without these, you can skip them.

Event/include/Event/TrigScintCluster.h
```diff
@@ -0,0 +1,191 @@
...
+ * @author Lene Kristian Bryngemark, Stanford University : adaptation to TrigScint 
+ */
+
+#ifndef EVENT_TRIGSCINTCLUSTER_H_
+#define EVENT_TRIGSCINTCLUSTER_H_
+
+// ROOT
+#include "TObject.h"
+#include "TString.h"
+#include "TRefArray.h"
```
tomeichlersmith, line 14: Please remove TString.h and TRefArray.h includes. I don't think you use them here.

EventProc/python/trigScintClustering.py
```diff
@@ -0,0 +1,50 @@
...
+from LDMX.Framework import ldmxcfg
+
+class TrigScintClusterProducer(ldmxcfg.Producer) :
+    """Configuration for cluster producer for Trigger Scintillators"""
+
+    def __init__(self,name) :
+        super().__init__(name,'ldmx::TrigScintClusterProducer')
+
+        from LDMX.EventProc import include
+        include.library()
```
tomeichlersmith, line 20: In order to update the container with the cmake changes, I removed the automaic generation of the `include.py` module. I replaced it with a change to the constructor:
```
super().__init__(name,'ldmx::TrigScintClusterProducer','EventProc')
```

EventProc/src/TrigScintTrackProducer.cxx
```diff
@@ -0,0 +1,365 @@
...
+	  
+	  return;
+	}
+  
+  
+	void TrigScintTrackProducer::onFileClose() {
+	  
+	  if (verbose_ )
+		{
+		  std::cout << "In TrigScintTrackProducer: Closing file!" << std::endl;
```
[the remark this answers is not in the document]
bryngemark: I've done this all the way up to the Print() method; I don't want to hardwire its verbosity in its definition. Any good suggestions on how to redirect it to the logger?

Event/include/Event/TrigScintCluster.h
```diff
@@ -0,0 +1,175 @@
...
+#include "Event/EcalCluster.h"
+
+
+namespace ldmx {
+
+  /**
+   * @class TrigScintCluster 
+   * @brief Stores cluster information from the trigger scintillator pads. Adds on the ECal cluster functionality
+   */
+  class TrigScintCluster : public EcalCluster {
```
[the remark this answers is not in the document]
bryngemark: Done now, inherits from TObject instead.

bryngemark:
I've done as much as I plan to do here; indentation fixes sound simple but is, I think, a matter of emacs vs vi and not as trivial as it looks at first glance. (I have tried different tab setting solutions and none really works, and honestly, I'm perfectly content with the way emacs normally and beautifully handles this for me and don't want to spend more time on chasing a solution.)

jmmans:
Dear all,

I agree.  I think if LDMX is going to use a very unusual set of tabbing 
rules, the emacs bindings should be provided.  The LDMX rules are 
"close" to the cc-mode in emacs, but LDMX for some reason tabs in things 
like "public:".

Jeremy

omar-moreno:
Before doing this, we need to decide on a style and stick with it.  At the
moment, we don't have a set style other than a few suggestions on the wiki
and 4 space indents.

Given that there isn't a standard style, I suggest we either consider
Bjarne's

https://www.stroustrup.com/bs_faq2.html

Or Google's style

https://google.github.io/styleguide/cppguide.html?cl=282453546

I'm biased towards Google's style since I tend to use it on personal
projects and find it nice. However, both are good and well documented.

Once we decide on a style, I can go through and refactor the code and put
in rules that enforces the style before a PR is merged. We can then provide
whatever bindings people might need to get the style into their editors.

tomeichlersmith (approved):

bryngemark:
So -- where are we on this? I just pushed a tiny tweak, and I see some test failed: Ecal PN simulation, but target PN etc seems to have run just fine. Whose responsibility is it to make my branch pass the tests and also, against what is it tested..? I'm pretty sure none of the automatic (failing) tests relate to my development at all.

tomeichlersmith:
I have updated some of the tests and so it might be a good time to re-merge (or re-rebase) this branch onto master. This should fix up the tests that aren't related to your work. Feel free to slack me if you run into weird git issues.

omar-moreno:
Please merge in master.

Event/include/Event/TrigScintCluster.h
```diff
@@ -0,0 +1,175 @@
...
+     * Take in the hits that make up the cluster.
+     * @param hit The digi hit's entry number in the events digi 
+     * collection.
+     */
+
+    void addHit(uint idx, const TrigScintHit* hit);
+    
+    void setSeed( int idx ) { seed_ = idx ; } 
+	
+	void setEnergy(double energy) {
```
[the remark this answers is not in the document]
omar-moreno: Can you fix the formatting?

Event/include/Event/TrigScintCluster.h
```diff
@@ -0,0 +1,235 @@
...
+//#include "Event/EcalCluster.h"
+
+
+namespace ldmx {
+
+  /**
+   * @class TrigScintCluster 
+   * @brief Stores cluster information from the trigger scintillator pads. Adds on the ECal cluster functionality
+   */
+  class TrigScintCluster : public TObject {
```
omar-moreno, line 29: Inheriting from TObject is not necessary.

Event/src/TrigScintCluster.cxx
```diff
@@ -0,0 +1,41 @@
+#include "Event/TrigScintCluster.h"
+
+ClassImp(ldmx::TrigScintCluster)
+
+namespace ldmx {
+
+  TrigScintCluster::TrigScintCluster() { //: EcalCluster() {
```
omar-moreno, line 7: Since you are using the default cluster, there is no need to have an empty constructor in the src file.  Instead, in the header file just do 
    `TrigScintCluster() = default;`

Event/src/TrigScintCluster.cxx
```diff
@@ -0,0 +1,41 @@
...
+    
+  }
+  
+  TrigScintCluster::~TrigScintCluster() {
+    Clear();
+
```
omar-moreno, line 13: Fix extra space.

Event/src/TrigScintCluster.cxx
```diff
@@ -0,0 +1,41 @@
...
+  void TrigScintCluster::Clear(Option_t*) {
+    
+	hitIDs_.clear();
+	
+	centroidX_ = 0;
+	centroidY_ = 0;
+	centroidZ_ = 0;
+	setEnergy(0);
+	setNHits(0);
+    setCentroid(0);
```
omar-moreno, line 35: Fix formatting.

omar-moreno:
@bryngemark Can you fix the formatting of all the code?

EventProc/src/TrigScintTrackProducer.cxx
```diff
@@ -0,0 +1,365 @@
...
+	  seeding_collection_      = ps.getParameter< std::string >("seeding_collection"); //probably tagger pad, "TriggerPadTagClusters" 
+	  input_collections_       = ps.getParameter< std::vector < std::string > >("further_input_collections"); // {"TriggerPadUpClusters" , "TriggerPadDnClusters" }
+	  output_collection_       = ps.getParameter< std::string >("output_collection"); 
+	  passName_                = ps.getParameter< std::string >("input_pass_name"); 
+	  verbose_                 = ps.getParameter< int >("verbosity");
+
+	  // TO DO: allow any number of input collections
+    
+
+	  if (verbose_)
```
omar-moreno, line 28: Why do you need a check for verbose if you are already setting the log level to info? You can always change the log level to something higher if you want these messages to stop.

EventProc/src/TrigScintTrackProducer.cxx
```diff
@@ -0,0 +1,365 @@
...
+	  // an output collection name
+	  // a verbosity controller
+    
+	  if ( verbose_) {
+		ldmx_log(debug)<< "TrigScintTrackProducer: produce() starts! Event number: " << event.getEventHeader().getEventNumber()  ;
+      
+	  }
+    
+    
+	  const auto seeds{event.getCollection< TrigScintCluster >(seeding_collection_, passName_)};
```
omar-moreno, line 58: Should use the RandomNumberSeedService for this instead.

EventProc/src/TrigScintTrackProducer.cxx
bryngemark: So to my understanding, the logger verbosity level is a global feature (so if I set it to debug, I get debug level messages from all invoked producers). For my use of clustering and tracking, which is still to work out the best parameters, it's very convenient to be able to just make one of the producers verbose. So I'd like to keep this configurable per producer.

omar-moreno: You bring up a good point.  We should be able to set the log level at the processor level. Totally fine to leave as is then.

bryngemark:
Ok.

I've fixed the indentation issues as much as I could. Again it doesn't look the same for me (in emacs) as on github, and I don't want to spend any more time chasing whitespace. I look forward to a common solution like the emacs bindings Jeremy suggested. 

As for the failed tests, which are more recent than my changes = independent of this work, I have attempted rebasing, but the changes (to other modules etc) are substantial, and it fails. I do think this is completely unnecessary for merging in the PR. The tests don't test this code anyway.

The two conflicting files are:
1. `trigScintDigis.py` (governing hit reconstruction) --> this I have replaced by a common python configuration for hit reconstruction, clustering and tracking: `trigScint.py`, so this "conflict" is actually a (to the user, anyway disruptive) file deletion
2. `TrigScintDigiProducer.cxx` --> here all I did was add one line of bugfix, to record the noise hit ID. It _should_ be an easy merge?

Event/include/Event/TrigScintCluster.h
```diff
@@ -0,0 +1,230 @@
...
+#include "Event/TrigScintHit.h"
+
+
+namespace ldmx {
+
+  /**
+   * @class TrigScintCluster 
+   * @brief Stores cluster information from the trigger scintillator pads. Adds on the ECal cluster functionality
+   */
+  class TrigScintCluster : {
```
awhitbeck, line 25: I think this is a typo.  I guess we're not intending this to inherit from anything.  The code also won't compile like this.

omar-moreno:
Thanks for all of the updates, I'll do some final formatting, make sure it compiles, merge in master and merge it in.

omar-moreno (approved):

omar-moreno:
> I have updated some of the tests and so it might be a good time to re-merge (or re-rebase) this branch onto master. This should fix up the tests that aren't related to your work. Feel free to slack me if you run into weird git issues.
> (That being said, @omar-moreno I think this is good and could be merged even though the tests are failing)

@tomeichlersmith where the changes you made pushed to Framework? The issue with the test is still present so was wondering if I needed to update a submodule.

tomeichlersmith:
Yes, there were changes in Framework, so the submodules need to be updated.

tomeichlersmith:
@bryngemark I figured out why the tests are failing! I don't know why the log from the action doesn't show this, but when I ran the test configs on my local machine I got this:
```
  File "/home/tom/ldmx/ldmx-sw/EventProc/test/sim_and_reco.py", line 28, in <module>
    from LDMX.EventProc.trigScintDigis import TrigScintDigiProducer
ModuleNotFoundError: No module named 'LDMX.EventProc.trigScintDigis'
```
It looks like you need to update `EventProc/test/sim_and_reco.py` to match the new module name that you use. (My guess is that the merging of master changed your changes back).

bryngemark:
Thanks for figuring this out, @tomeichlersmith! I have updated the name of the trig scint reco config in the test. I also added in clustering and tracking to the test sequence.

bryngemark:
It does look, though, that the change in how the random seeds are generated was not fully propagated here (tests fail). @tomeichlersmith, @omar-moreno, I'm guessing one of you knows how to make the TS digi code compliant with the new approach?

tomeichlersmith:
Yea, the below code snippet is from EcalDigiProducer and shows how to seed the noise generator while using the RNSS.
https://github.com/LDMX-Software/Ecal/blob/22b54afbe6b4e220508a957a3dae717e9aa48b58/src/EcalDigiProducer.cxx#L55-L59

bryngemark:
Thanks @tomeichlersmith! I added that, and just ran the test locally and it didn't fail -- so, fingers crossed then.

PR #903, opened 2020-09-04
Provides U,V indices for EcalID, some structural fixes to EcalID

jmmans:
This pull request addresses the "U,V" issue from Tom's original issue 873.  It also introduces an intermediate EcalAbstractID to allow for different subtypes of EcalIDs.  This allows the introduction of an EcalTriggerID for the individual cells of the ECAL trigger calculation.

These changes are completely safe for application to the ongoing master as they do not affect the on-disk image, etc.  I have submitted them separately as a pull request to avoid having very complex pull requests (like one I have building regarding EcalHexReadout) tied in with these simple aspects which will help everyone.

### DIFF
--- a/DetDescr/include/DetDescr/DetectorIDInterpreter.h
+++ b/DetDescr/include/DetDescr/DetectorIDInterpreter.h
@@ -100,7 +100,12 @@ namespace ldmx {
             /** 
              * Register a new field interpreter for a given subdetector id
              */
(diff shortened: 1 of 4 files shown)


tomeichlersmith (approved):
This looks good to me. The test failed at the Conditions test that we are patching now.

omar-moreno (approved):

omar-moreno (approved):

PR #914, opened 2020-09-23
RandomNumberSeedService

jmmans:
This pull request introduces the RandomNumberSeedService as a Conditions object.  It also integrates the use of the service into all identified locations where random numbers are currently used in LDMX-sw.

The Service currently supports three master seeding modes -- time seeding, seeding based on run number, and seeding from a seed provided in the python.  Clients request a seed using a string name.  If the seed has not been created already, it is created using the master seed plus a hash of the seed name.


tomeichlersmith:
## Related PRs in Submodules
- https://github.com/LDMX-Software/Conditions/pull/6
- https://github.com/LDMX-Software/Framework/pull/25
- https://github.com/LDMX-Software/SimCore/pull/28

tomeichlersmith (approved):
Looks good. We will just need to be careful when working with other developers to make sure they update their own developments to use the RNSS instead of their own random seed.

omar-moreno (approved):

PR #916, opened 2020-09-24
Provision of EcalTrigPrimDigis

jmmans:
Builds on iss895 and other pull requests to bring in the ability to create trigger primitives for Ecal.

Example python file (hidden as a txt file for github)
[sim_and_reco_example_py.txt](https://github.com/LDMX-Software/ldmx-sw/files/5279207/sim_and_reco_example_py.txt)

### DIFF
--- a/Event/include/Event/HgcrocTrigDigi.h
+++ b/Event/include/Event/HgcrocTrigDigi.h
@@ -9,6 +9,14 @@
 
 namespace ldmx {
 
+// Forward declaration needed by typedef
+class HgcrocTrigDigi; 
+
+  /**
+ * Define the type of collection for trig digis
+ */
+typedef std::vector<HgcrocTrigDigi> HgcrocTrigDigiCollection;
+
 /**
  * @class HgcrocTrigDigi
  * @brief Contains the trigger output for a single trigger hgcroc channel
@@ -131,11 +139,6 @@ private:
   /// ROOT Dictionary class definition macro
   ClassDef(HgcrocTrigDigi, 1);
 };
-
-/**
- * Define the type of collection for trig digis
- */
-typedef std::vector<HgcrocTrigDigi> HgcrocTrigDigiCollection;
 
 } // namespace ldmx
 
--- a/Tools/src/HgcrocTriggerCalculations.cxx
+++ b/Tools/src/HgcrocTriggerCalculations.cxx
@@ -4,11 +4,16 @@
 #include <iostream>
 
 namespace ldmx {
-  
-HgcrocTriggerConditions::HgcrocTriggerConditions(const IntegerTableCondition& ict, bool validate): ict_{ict} {
-  if (!validate) return;
-  if (ict_.getColumnCount()<5) {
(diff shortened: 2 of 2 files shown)


tomeichlersmith:
Depends on PR #914 for updates to the conditions system.

tomeichlersmith:
## Related Submodule PRs
- https://github.com/LDMX-Software/Ecal/pull/17
- https://github.com/LDMX-Software/SimCore/pull/29
- https://github.com/LDMX-Software/Framework/pull/26

tomeichlersmith:
## Still To-Do
- [ ] Integrate the `HgcrocTriggerCalculations` into the `HgcrocEmulator` so that it is available for Hcal and the emulation of the chip is done in one place
- [ ] Allow for passing _separate conditions tables_ for the trigger calculations and the reconstruction. This is important becuase in the future, we can change the recon conditions (e.g. the recon pedestals) but can't change the conditions that the data was gathered under.
- [ ] Cleanup and documentation, specifically focused on announcing this necessary change to any configuration that wants to use the ecal digi pipeline: The user will need to import a set of conditions for the ecal digi pipeline to function properly. Again, this is important because the user can choose different conditions to import from.

jmmans:
Dear Tom,

I don't understand point 2 -- this is already the case, so why does it 
appear here?  The action would be to add the use of condition tables for 
reconstruction, but that shouldn't be part of this pull request.

Jeremy

tomeichlersmith:
Good point, I was just copying down my thoughts to make sure I don't forget anything. I will mark that as "done".

tomeichlersmith (approved):
Looks good to me. Was able to merge the sub-module Ecal PR and update the ref here, so we are good to go.

Event/include/Event/HgcrocTrigDigi.h
```diff
@@ -0,0 +1,152 @@
...
+  HgcrocTrigDigi(uint32_t tid, uint8_t tp=0);
+
+  /**
+   * Destructor
+   *
+   * Needs to be defined for ROOT
+   * dictionary definition, does
+   * nothing right now.
+   */
+  virtual ~HgcrocTrigDigi();
```
omar-moreno, line 45: If this does nothing, then might as well set it equal to default 
`virtual ~HgcrocTrigDigi() = default;`

you can then remove the definition from the src file.

Event/include/Event/HgcrocTrigDigi.h
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Sort the collection to trig digis
+   * by the raw ID.
+   *
+   * @param[in] d another digi to compare against
+   * @returns true if this ID is less than the other ID
+   */
+  bool operator<(const HgcrocTrigDigi &d) {
```
omar-moreno, line 54: I suggest making the variable name a bit clearer i.e. d --> digi

omar-moreno (changes requested):

omar-moreno:
Looks fine overall, but would be useful to have some documentation on how the trigger primitive bits are being set.

omar-moreno (approved):

PR #925, opened 2020-10-12
Adding Hcal as a submodule

cmantill:
Moving old producers to its own Hcal repository. New versions will follow later.


tomeichlersmith:
@cmantill You will need to move the python-configuration classes from `EventProc` into the `Hcal` module as well. After the move, you will also need to change the lines in `__init__` that reference `EventProc` to `Hcal`.

tomeichlersmith (approved):
Looks good. Simple move and the tests are passing.

omar-moreno (approved):

PR #927, opened 2020-10-13
Add beam electron collection producer

### LINKED ISSUE #926: A sim-level filter that selects and keeps only simhits in a collection that are directly associated with incoming beam electrons
Need a way to benchmark beam electron counting against the actual number of beam electrons a system has seen. 

A pure beam electron hit collection provides a clean sample of hits against which to benchmark algorithm performance. It automatically accounts for cases such as completely overlapping trajectories or electrons scattering out of acceptance. 

I have also thought of adding some more labelling to (non-)beam electron hits, but this is not the most pressing need. It would be straightforward to single out other types of hits to other collections. Having a separate collection makes it easy to compare individual events.

bryngemark:
This class sifts through a collection of SimCalorimeterHits and only keeps the ones which can be associated to the beam electrons using generator level information.

This output collection can then be run through reco/digi or any other later processing without any further tweaks.

On a preexisting sim+trigscint reco file, I ran the producer and the trigger scintillator digi on this collection using the following configuration (snippet):

```
passName="truth"
p=ldmxcfg.Process( passName )

from LDMX.EventProc.truthHits import TruthHitProducer
from LDMX.EventProc.trigScintDigis import TrigScintDigiProducer

#-- create truth beam electron collections                                                                                                                       
beTag=TruthHitProducer.tagger()
beUp=TruthHitProducer.up()
beDown=TruthHitProducer.down()

#-- digi of truth beam electron collections                                                                                                                      
digiTag=TrigScintDigiProducer.tagger()
[... similar for up/down, won't state from here] 

digiTag.input_collection="truthBeamElectronsTag"                                                                                                           
digiTag.pass_name=passName

#turn off noise hits in truth collections    -- can't be identically 0 (poisson function error)                                                                                                                    
digiTag.mean_noise=0.00000000001

p.sequence = [ beTag, beUp, beDown, digiTag ]
```

which results in two trig scint reco hit collections: `trigScintDigisTag_sim` from the input file, and `trigScintDigisTag_truth` from running this example. Later processors can then select either by pass name. 

Two example distributions from the tagger pad, comparing the full hit reco in black to the beam electron hit reco in turquoise.

1. the fraction of the energy deposit making up a hit which is associated with beam electrons. The beam electron collection has all its entries at 1, as expected. The full reco has some hits with mixed input, and a spike at 0 (pure secondaries, or pure electronics noise).
beamEfrac

2. the bar ID vs event number (typically we expect 2 adjacent bars to light up when a particle crosses a pad)
barID_evNB

So I would say this works as intended, and closes #926.


EventProc/src/TruthHitProducer.cxx
```diff
@@ -0,0 +1,77 @@
...
+		  bool keep = false;
+			  // check if hits is from beam electron and, if so, add to output collection
+		  for( int i = 0 ; i < simHit.getNumberOfContribs() ; i++){
+			
+			auto contrib = simHit.getContrib(i);
+			if( verbose_ ){
+			  ldmx_log(debug) << "contrib " << i << " trackID: " << contrib.trackID << " pdgID: " << contrib.pdgCode << " edep: " << contrib.edep ;
+			  ldmx_log(debug) << "\t particle id: " << particleMap[contrib.trackID].getPdgID() << " particle status: " << particleMap[contrib.trackID].getGenStatus();
+			}  //beam electron (PDGID = 11, genStatus == 1) 
+			if( particleMap[contrib.trackID].getPdgID() == 11 && particleMap[contrib.trackID].getGenStatus() == 1 ){
```
tomeichlersmith, line 66: The Particle map doesn't store everything generated in the simulation in it, so you should add a simple check that the particle map has that track ID before accessing it. This will avoid a seg fault if (for whatever reason) the sim hit references a particle that doesn't exist in the particle map.
The particle map _always_ has the primary particle(s), so this check shouldn't change the behavior of this producer.

bryngemark: Nice catch, thanks!

omar-moreno:
This should go into the TrigScint repo correct?

bryngemark:
Well I hadn't envisioned it as TrigScint specific but you are right that for other systems to use it, it might need a little bit of expansion. But there is nothing in there that requires any TS stuff.

omar-moreno:
> Well I hadn't envisioned it as TrigScint specific but you are right that for other systems to use it, it might need a little bit of expansion. But there is nothing in there that requires any TS stuff.
> 
> I just exemplify here with TS collections because of a personal bias I guess :D and because this is the imminent use case I have in mind myself.

OK, I mention moving this too because the EventProc module is gone starting v3.0.  If you feel this is generic enough that it should go somewhere else, then we can put it somewhere outside of TrigScint (after testing of course). 

Will look at code later today.

bryngemark:
Hi @omar-moreno, any chance of getting this into the 2.3 release? It would be really useful for TS/multielectron studies.

omar-moreno:
@bryngemark Please move this into the TrigScint module.

bryngemark:
I can't seem to get master merged into my dev branch properly so I can't figure out how to move it (and I can't test run it either because I can't compile). @omar-moreno would you be ok with helping me out here?

omar-moreno:
Yes, I'll take a look at it.

omar-moreno:
OK, I've merged in master, moved the files to TrigScint and pushed an update submodule.  I would of just closed this PR and opened a new one, but I wanted to preserve the commit history.   Once the files get merged into TrigScint, this can be  merged in.

PR #930, opened 2020-10-20
overlay producer 

bryngemark:
So this is the current status. What we can produce right now:

- a SimCalorimeterHit collection that includes both simhits from a simulator or an input file, and the corresponding collection from an overlay file
- that can be shifted in time
- with a fixed or Poisson varied number
- with overlay hits added as contribs to EcalSimHits (which uses them later on. Not true for e.g. TS)
- and overlay contribs have incidentID = trackID -1000, PDG code 0

What I'd need help to finalize:
- verify that there's no double counting of overlay contribs. Right now, if you look at trackID it looks like some retain their old trackID (se figure), and some get trackID = 0..?
- being able to run this over SimTracker hits
- figure out if I need to tweak EventFile to be able to reset the overlay event number counter (currently I'm using nextEvent() as is)


![EcalSimHitsOverlay_sim trackIDContribs_](https://user-images.githubusercontent.com/42861632/96654276-bfd7b280-12ef-11eb-997f-96e6a279b3c6.png)


To run it, first run a bunch of 1e events, for example (these are snippets!! standard imports skipped):
```
p=ldmxcfg.Process("sim")
p.run = 12
mySim =  simulator.simulator("mySim")                                                                          
mySim.randomSeeds = [ 20, 23 ]
mySim.beamSpotSmear = [ 20., 80., 0.]  #mm                                                                                                               
                                                                                                                     
from LDMX.Detectors.makePath import *
mySim.setDetector( 'ldmx-det-v12', True  )
mySim.scoringPlanes = makeScoringPlanesPath('ldmx-det-v12')
                                                                     
mySim.generators = [ generators.single_4gev_e_upstream_tagger() ]
p.sequence = [ mySim ]
                                                                                                               
p.maxEvents=100                                                                           
p.outputFiles=["ldmx_upstream_1e_events.root"]
```
-- and once more with some different run number/seeding to get a sim file to add overlay to, if you don't want to generate sim on the fly. 

Then run overlay:
```
if len(sys.argv) > 1 :
    inputName=sys.argv[1]  #specify an input file name or run sim on the fly                                                                                     
    runSim = False
else :
    runSim = True

mySim =  simulator.simulator("mySim")
mySim.randomSeeds = [ 1, 2 ]
mySim.beamSpotSmear = [ 20., 80., 0.]  #mm                                                                                                               

p.run = 1
p.maxEvents=10

from LDMX.Detectors.makePath import *
mySim.setDetector( 'ldmx-det-v12', True  )
mySim.scoringPlanes = makeScoringPlanesPath('ldmx-det-v12')                                                                                                                                                        
mySim.generators = [ generators.single_4gev_e_upstream_tagger() ]

from LDMX.EventProc.overlay import OverlayProducer                                                                                                              \
overlay = OverlayProducer()
overlay.overlayFileName = "ldmx_upstream_1e_events.root"
overlay.numberOverlaidInteractions = 1.
overlay.timeSpread = 0.     # <-- a fix pileup time offset is easier to see than a Gaussian distribution 
overlay.timeMean   = 1.     # <-- 
overlay.doPoisson  = 0
overlay.overlayHitCollections=[ "TriggerPadUpSimHits", "EcalSimHits"]

if runSim :
    p.sequence = [ mySim, overlay ]
else :
    p.inputFiles= [inputName ]
    p.sequence = [ overlay ]

p.termLogLevel = 0.    # <-- really useful to see all the printouts here 
p.outputFiles=["ldmx_events_withOverlay.root"]
```

And finally, run digi/reco on these collections:
```                                                                          
p=ldmxcfg.Process("sim")

if len(sys.argv) > 1 :
    inputName=sys.argv[1]  #specify the output name if default is not desired                                                                                    
else:
    inputName="ldmx_events_withOverlay.root"                                                                   

import LDMX.Ecal.EcalGeometry                                                                                                                                                
from LDMX.Ecal import digi
from LDMX.Ecal import vetos

p.maxEvents=100
                                                            
tsDigisUp  =TrigScintDigiProducer.up()
tsDigisUp.input_collection = "TriggerPadUpSimHitsOverlay"

ecalDigi   =digi.EcalDigiProducer('EcalDigis')
ecalReco   =digi.EcalRecProducer('ecalRecon')
ecalVeto   =vetos.EcalVetoProcessor('ecalVetoBDT')

ecalDigiOverlay    =digi.EcalDigiProducer('EcalDigisOverlay')
ecalRecoOverlay    =digi.EcalRecProducer('ecalRecoOverlay')
#ecalveto input collection name needs to be settable, otherwise could have done:                                                                                                            
#ecalVetoOverlay    =vetos.EcalVetoProcessor('ecalVetoBDTOverlay')                                                                                               

ecalDigiOverlay.inputCollName  = "EcalSimHitsOverlay"
ecalDigiOverlay.digiCollName   = "EcalDigisOverlay"
ecalRecoOverlay.digiCollName   = "EcalDigisOverlay"
ecalRecoOverlay.recHitCollName = "EcalRecHitsOverlay"
ecalRecoOverlay.simHitCollName = "EcalSimHitsOverlay"
                                                                                       
p.sequence=[ ecalDigi, ecalReco, ecalVeto, ecalDigiOverlay, ecalRecoOverlay, tsDigisUp ]

p.inputFiles= [ inputName ]
outputName=inputName.replace(".root", "_digi.root")
p.outputFiles= [ outputName ]
```


EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,329 @@
...
+	
+	/*
+	  if (! overlayFile_->nextEvent() ) {
+	  std::cerr << "Couldn't read next event!" ;
+	  return;
+	  }
+	*/
+	
+	//get event wherever nextEvent left us 
+	Event * overlayEvent = overlayFile_->getEvent();
```
tomeichlersmith, line 89: Your member variable `overlayEvent_` was attached to the `overlayFile_` using `setupEvent` in `onProcessStart`. This means `overlayEvent_` is the object that is updated when you call `nextEvent`, so you don't need to get a reference to the event here.

bryngemark: Thanks for catching that! I'll push a fix. Indeed this code still has a few remnants of scrambled iterations since May 😄  I hope it still parses well enough for figuring out what goes awry with the contribs though.

EventProc/python/overlay.py
```diff
@@ -0,0 +1,38 @@
...
+
+        Sets all parameters to reasonable defaults.
+
+    Examples
+    --------
+        from LDMX.EventProc.overlay import OverlayProducer
+        p.sequence.append( OverlayProducer() )
+    """
+
+    def __init__(self,name = 'overlay') :
```
tomeichlersmith, line 18: The constructor is the perfect place for all the "required" variables. From my perspective, I would add the overlay file name here so that a user can't construct the producer without specifying a file. e.g.
```
def __init__(self,file_name) : 
        super().__init__(name,'ldmx::OverlayProducer','EventProc')
        self.overlayFileName = file_name
```

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,329 @@
...
+
+		  //we don't need to touch the hard process sim hits, really... but we might need the simhits in the hit map.
+		  for (const SimCalorimeterHit &simHit : simHits ) {
+			
+			simHit.Print();
+
+			//store ids of existing hits. for these, need to add contribs...
+			if (isEcalHitCollection) {
+			  int simHitID = simHit.getID();
+			  hitMap[ simHitID ] = SimCalorimeterHit();
```
tomeichlersmith, line 188: Instead of using the default constructor, you could just copy over the `simHit` object.
```
hitmap[simHitID] = simHit
```
Then you wouldn't have to do the `set` commands after this.

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,329 @@
...
+		  overlayHit.Print();
+		  const float overlayTime = overlayHit.getTime() + timeOffset;
+		  overlayHit.setTime( overlayTime ); 
+
+		  if (! isEcalHitCollection )  //special treatment for ecal
+			outHits.push_back( overlayHit );
+		  else {
+			int overlayHitID = overlayHit.getID();
+			if ( hitMap.find( overlayHitID ) == hitMap.end() ) {
+			  hitMap[ overlayHitID ] = overlayHit;
```
tomeichlersmith, line 215: This line is what is causing the "doubling". If there isn't a regular sim hit in a cell, then we create a new `SimCalorimeterHit`. This line (in its current form) preserves all of the contribs from the overlay hit. Replacing this code block with something like the following should fix it.
```
hitMap[overlayHitID] = SimCalorimeterHit();
hitMap[ overlayHitID ].setID(overlayHitID);
std::vector <float> hitPos = overlayHit.getPosition();
hitMap[ overlayHitID ].setPosition( hitPos[0], hitPos[1], hitPos[2]);
hitMap[ overlayHitID ].addContrib(overlayIncidentID_,overlayTrackID_,overlayPdgCode_, overlayHit.getEdep(), overlayTime);
```

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,329 @@
...
+	   ldmx_log(debug) <<  "[ OverlayProducer ]: onFileOpen() " ;
+	}
+	
+	//      simFile_ = std::make_unique<EventFile*> ( input );
+	//      simFile_=&input ;
+	
+	return;
+  }
+  
+  void OverlayProducer::onFileClose(EventFile&) {
```
tomeichlersmith, line 284: Remove this function please.

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,329 @@
...
+	// this should be added to the sim file, so to "event"
+	for (uint iColl = 0; iColl < outCollNames.size(); iColl++) { 
+	   ldmx_log(debug) <<  "Writing " << outCollNames[iColl] << " to event bus" ;
+	  event.add(outCollNames[iColl], v_outHits[iColl]);
+	}
+	
+	return;
+  }
+  
+  void OverlayProducer::onFileOpen() {//EventFile& input) {
```
tomeichlersmith, line 273: Remove this function please.

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,329 @@
...
+  void OverlayProducer::onProcessStart() {
+	if (verbosity_ > 2) {
+	   ldmx_log(debug) <<  "[ OverlayProducer ]: onProcessStart() " ;
+	}
+
+      //based on the config parameters, set stuff up 
+      // doing this here guarantess that the overlay file(s) gets set up regardless of whether we have an input file or not; onFileOpen gets called for both modes.
+
+      rndm_  = std::make_unique<TRandom2>( seed_ );
+      rndmTime_  = std::make_unique<TRandom2>( seed_ );
```
tomeichlersmith, line 298: I know you know that you need to switch to the RNSS, but I just wanted to be complete in my review.

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,329 @@
...
+		 ldmx_log(debug) <<  "[ OverlayProducer ]: onProcessStart () successful. Used input file: " <<overlayFile_->getFileName() ;
+		 ldmx_log(debug) <<  "[ OverlayProducer ]: onProcessStart () successful. Got event info: " ;
+		overlayFile_->getEvent()->Print( verbosity_ );
+      }
+	  
+	  
+	  return;
+    }
+
+    void OverlayProducer::onProcessEnd() {
```
tomeichlersmith, line 322: Remove this unless you want to print some end-of-run statistics.

EventProc/python/overlay.py
```diff
@@ -0,0 +1,38 @@
+"""Configuration for overlay """
+
+
+
+from LDMX.Framework import ldmxcfg
+
+class OverlayProducer(ldmxcfg.Producer) :
+    """Configuration for pileup overlay
+
+        Sets all parameters to reasonable defaults.
```
tomeichlersmith, line 10: Document your parameters here as well as in the Cpp source. General format:
```
Parameters
----------------
Name : type
     Description

Attributes:
-------------
name : type
    Description
```
Parameters are for the inputs to the constructor. Attributes are for everything else.

EventProc/python/overlay.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self,name = 'overlay') :
+        super().__init__(name,'ldmx::OverlayProducer','EventProc')
+
+
+        self.overlayFileName = "ldmx_upstreamMultiElectron_events.root"
+        self.overlayTreeName = "LDMX_Events"
+        self.numberOverlaidInteractions = 1.
+        self.timeSpread = 0.
+        self.doPoisson = 0
+        self.overlayProcessName = "inclusive"
```
tomeichlersmith, line 27: Since you are just reading from a file. This isn't super necessary to be configurable. I would keep it as a Cpp member variable, but remove it here to clean up the python config.

tomeichlersmith (changes requested):
This looks great! I have a lot of assorted clean up requests (and a possible fix for the "doubling" you are seeing! :smile: )

One general request is just more documentation. This can be a part of your clean-up process and don't be afraid to be verbose and repeat yourself!

EventProc/src/OverlayProducer.cxx
bryngemark: Thanks – actually the duplication of simhits came from copying over the simhits collection in the start of the event, and then again when copying over all hits in the hitmap (after adding in overlay contribs). I've fixed this now. But I did take this suggestion (basically it's the inverse of what I did before) since it makes sure we're not preserving any overlay hit contribs.

EventProc/src/OverlayProducer.cxx
bryngemark: So, I toyed with this a bit, and I'm not sure how to do it when I'm not using a noise generator. It looks like I need to define some Conditions object but I'm not sure how to think about this. I figure the main use case envisioned so far has been electronics noise. Perhaps @tomeichlersmith or whoever ended up implementing the random seed service can comment on how to go about this?

tomeichlersmith: *Good News*: You don't need to define a whole conditions object! Basically all RNSS does is define `seed_` for you (instead of you passing it from python). This makes it centrally generated so that we can redo an entire process with only defining one "master seed".

Since the conditions objects need to be configured themselves, the earliest (safe) place to access conditions object is in `produce`. The method that other producers have taken is leaving your random engines as `nullptr`s until they are seeded, so delete the lines referencing the rndm generators above, and then...
```
//at the beginning of OverlayProducer::produce
if (rndm_.get() == nullptr) {
  //not been seeded yet, get it from RNSS
  const auto& rnss = getCondition<RandomNumberSeedService>(RandomNumberSeedService::CONDITIONS_OBJECT_NAME);
  rndm_ = std::make_unique<TRandom2>(rnss.getSeed("OverlayProducer::rndm_"));
}
if (rndmTime_.get() == nullptr) {
  //not been seeded yet, get it from RNSS
  const auto& rnss = getCondition<RandomNumberSeedService>(RandomNumberSeedService::CONDITIONS_OBJECT_NAME);
  rndmTime_ = std::make_unique<TRandom2>(rnss.getSeed("OverlayProducer::rndmTime_"));
}
```

bryngemark: Thanks! So all I was missing then, was how to refer back to the object using `OverlayProducer::rndm_` – but I would never have realized the `produce` bit myself.

EventProc/python/overlay.py
bryngemark: Good suggestion, taken.

bryngemark:
So, I think this code works as expected now. The ecal contribs (or rather, the sim hits) are no longer double counted:

EcalSimHitsOverlay_sim trackIDContribs_

And distributions look reasonable:
EcalSimHitsOverlay_sim x_
HcalSimHitsOverlay_sim z_
TriggerPadUpSimHitsOverlay_sim y_

I have also implemented shifting the time of the overlay events across several bunches (and within bunches, not seen here. Note that earlier time bunches also occur but the way I plot it in this branch dump, the sim event sets the axes so they're not visible here...)
EcalSimHitsOverlay_sim time_

I also have a tweak in for resetting the event counter once we hit the end of the pileup tree. However this is done in `EventFile::nextEvent()` and that class is a in motion right now as we're restructuring for v3.0. It's not a showstopper, and a one-liner in the overlay code (a new bool added to the EventFile constructor) so I have commented it for now., and can implement it very easily on a branch based off of `3.0.0-dev` later.

Now all that remains is being able to deal with `SimTrackerHits` and, possibly, randomizing the first overlay event number used (can be a later improvement). 

I will continue validating this by running reco and comparing to multiparticle gun samples, as outlined earlier. But please have a look now.

EventProc/include/EventProc/OverlayProducer.h
```diff
@@ -0,0 +1,141 @@
...
+
+	/**
+	 * To use for finding the sim event bus passengers, mostly a disambiguation
+	 */
+	std::string simPassName_;
+	
+	/**
+	 * Let the total number of events be poisson distributed, or fix at the chosen value, poissonMu_
+	 */
+	int doPoisson_{0};
```
tomeichlersmith, line 82: We are now able to pass booleans as parameters. I would suggest using that capability unless this variable is used for something else that requires it to be an int.

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,332 @@
...
+	  //not been seeded yet, get it from RNSS
+	  const auto& rnss = getCondition<RandomNumberSeedService>(RandomNumberSeedService::CONDITIONS_OBJECT_NAME);
+	  rndmTime_ = std::make_unique<TRandom2>(rnss.getSeed("OverlayProducer::rndmTime"));
+	}
+
+	
+	// sample a poisson distribution, or use a deterministic number of overlay events
+	int nEvsOverlay = doPoisson_ ? (int) rndm_->Poisson( poissonMu_ ) : (int)poissonMu_ ;
+	// the poisson samples the total number of events, which is nOverlay + 1 (since it includes the sim event)
+	nEvsOverlay -=1; // now subtract the sim event from the poisson mu
```
tomeichlersmith, line 78: What if there is no sim event before it? Could "inclusion" of the sim event be configurable?
Another option is to test if there is anything in the event bus using `exists`. For example, you could check for `SimParticles` which is a collection that is always produced by the simulation.

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,332 @@
...
+	if (verbosity_ > 2) {
+	   ldmx_log(debug) <<  "starting from overlay event " <<  overlayEvent_.getEventHeader().getEventNumber() ;
+	}
+
+      
+	// using nextEvent to loop, we need to loop over overlay events and in an inner loop, loop over collections, and store them.
+	// after all pileup events have been added, the vector of collections is iterated over and added to the event bus.
+	
+	std::vector< std::vector<SimCalorimeterHit> >  v_outHits;
+	std::map< std::string, int > collectionMap;
```
tomeichlersmith, line 101: Is there a reason that you don't use
`std::map< std::string , std::vector<SimCalorimeterHit> > collectionMap`
?
This would seem to simplify your code somewhat... I will include other comments that offer an alternative using this setup.

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,332 @@
...
+		std::string outCollName = collections_[iColl]+"Overlay";
+		// if we alredy added at least one overlay event, this collection already exists in the output collection map
+		// otherwise, start out by just copying the sim hits, unaltered. 
+		if ( collectionMap.find(outCollName) == collectionMap.end() ) {
+		  
+		  simHits = event.getCollection<SimCalorimeterHit>( collections_[iColl], simPassName_ );
+		  // but don't copy ecal hits immediately: for them, wait until overlay contribs have been added. then add everything through the hitmap
+		  if (! isEcalHitCollection ) { 
+			v_outHits.push_back(simHits);
+			collectionMap[ outCollName ] = v_outHits.size() - 1;
```
tomeichlersmith, line 171: If combining `collectionMap` and `v_outHits`:
```
if (! isEcalHitCollection) collectionMap[outCollName] = simHits;
```

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,332 @@
...
+			}
+			// add the overlay hit (as a) contrib 
+			// incidentID = -1000, trackID = -1000, pdgCode = 0  <-- these are set in the header for now but could be parameters
+			hitMap[ overlayHitID ].addContrib(overlayIncidentID_,overlayTrackID_,overlayPdgCode_, overlayHit.getEdep(), overlayTime);
+		  } // ecal hits
+		  else {
+			std::map<std::string, int>::iterator itr = collectionMap.find( outCollName );
+			if (verbosity_ > 2)
+			  ldmx_log(debug) <<  "Adding non-Ecal overlay hit to outhit vector at " << itr->second ;
+			v_outHits[ itr->second ].push_back( overlayHit );
```
tomeichlersmith, line 232: If combining `collectionMap` and `v_outHits`
```
collectionMap[outCollName].push_back(overlayHit);
if (verbosity_>2) ldmx_log(debug) << "Adding non-Ecal overlay hit to outhit vector '" << outCollName << "'";
```

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,332 @@
...
+		if (verbosity_ > 2)
+		  ldmx_log(debug) <<  "Hits in hitmap after overlay of " << collections_[iColl] << "Overlay :" ;
+		for ( auto &mapHit : hitMap ) {
+		  if (verbosity_ > 2)
+			mapHit.second.Print();
+
+		  if ( collectionMap.find(collections_[iColl]+"Overlay") == collectionMap.end() ) {
+			ldmx_log(debug) <<  "Adding first hit from hit map as first outhit vector to v_outHits";
+			v_outHits.push_back( {mapHit.second} );
+			collectionMap[ collections_[iColl]+"Overlay" ] = v_outHits.size() - 1;
```
tomeichlersmith, line 268: If combining `collectionMap` and `v_outHits`
```
collectionMap[collections_[iColl]+"Overlay"] = { mapHit.second };
```

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,332 @@
...
+		  }
+		  else
+			v_outHits.back().push_back( mapHit.second );
```
tomeichlersmith, line 271: If combining `collectionMap` and `v_outHits`
```
collectionMap[collections_[iColl]+"Overlay"].push_back(mapHit.second);
```

EventProc/src/OverlayProducer.cxx
```diff
@@ -0,0 +1,332 @@
...
+
+	  ldmx_log(debug) <<  "Writing " << itr.first << " to event bus" ;
+	  if (verbosity_ > 2) {
+		ldmx_log(debug) <<  "List of hits added: " ;
+		for ( auto &hit : v_outHits[ itr.second ] )
+		  hit.Print();
+	  }
+
+	  event.add( itr.first, v_outHits[ itr.second ]);
+	}
```
tomeichlersmith, line 291: A nice feature of C++17 which would clean up this code is python-like looping:
```
for (auto const& [ key , val ] : a_map ) {
    bool always_true_by_definition = (val == a_map.at(key));
}
```
If you were to combine `collectionMap` and `v_outHits` then this loop would look really nice:
```
for (auto const& [ name , coll ] : collectionMap ) {
    ldmx_log(debug) << "Writing " << name << " to event bus.";
    if (verbosity_ > 2) {
        ldmx_log(debug) << "List of hits added: ";
        for ( auto const& hit : coll ) hit.Print();
    }
    event.add( name , coll );
}
```

tomeichlersmith (commented):
Looking good! My requests are all related to cleaning up the code, none of it is urgent or necessary.

EventProc/src/OverlayProducer.cxx
bryngemark: So I've though about this, but rather from the angle that in a Poisson( mu < 4 or so), there will often be n_e = 0 events. But in MC, we always generate something that happens. So how to simulate this?

And why?
I think what we're possibly thinking about is an event where out-of-time (OOT) pileup signal triggers the event. An aside: I don't think we need to worry about simulating untriggered events -- we can get the total rate straight by weighting our events according to multiplicity, and accounting for that sometimes (a lot of the time) we will have 0e. And perhaps this is the way forward, just making fixed multiplicity samples and combining them, rather than Poisson fluctuating already at overlay stage (which mostly assures you that you don't waste time on multiplicities you don't care as much about). 

The case where OOT pileup triggers the event though. Wouldn't we have a possibility to study this basically in cases where the sim event itself wouldn't have triggered it, but (intime) pileup does, or when readout thresholds are already crossed by OOT pileup?

Or maybe we could speculate forever but just flying it once is the answer 😄  

I think it's an interesting point and possibly simple enough to implement but I'd like to have more of a direction decision for how to study OOT pileup before I go ahead and implement it. Maybe a separate issue for what you'd like to be able to study is a good idea, to make sure we don't lose track of the idea?

EventProc/src/OverlayProducer.cxx
bryngemark: No reason. I thought about it too but 1. had a vague recollection of more complicated object maps not being allowed (probably an older standard?) and 2. was mostly happy when I got things to work and considered this to be  cosmetics. I'll implement it when I get back to this code (once I have some help on the SimTrackerHits implementation).

EventProc/src/OverlayProducer.cxx
tomeichlersmith: I see your point. I will open another issue because it does get complicated rather quickly. :+1:

EventProc/include/EventProc/OverlayProducer.h
bryngemark: done

bryngemark:
Ok! Now I think this is ready for some more serious validation so I'd like it to go in! Example plot: tagger tracker hits in 500 events of 1+1 overlay.

taggerTracker_px_2e

(Whenever the v3.0 structure has been merged in with master I will also add my small addition to EventFile allowing the event counter to be reset in the case that the file is an overlay input file. This isn't necessary for validating the overlay procedure in itself though. So that can be a separate issue.)

tomeichlersmith (approved):
My only request is to put documentation in the doxygen-style above your function declarations. You've already commented your code source a lot, so this should just be essentially a copy/summary of those comments explaining what is happening.

bryngemark:
Will do -- can you point me to an example though so I understand what it is you want? I've been looking around on our docs pages and some sample code and can't find anything that has longer documentation in the beginning.

bryngemark:
Oh wait you mean adding stuff like @param and a few more words sometimes in the header file. Sure.

bryngemark:
A ping on this one! I think it's ready to go

PR #933, opened 2020-11-03
QIE digis

GNiramay:
I am adding class for storing QIE digis, required for simulation of trigger scintillators.


Event/include/Event/TrigScintQIEDigis.h
```diff
@@ -0,0 +1,111 @@
...
+     * Store cids of all time samples
+     * @param cid_ array of cids
+     */
+    void SetCID(int* cid_);
+
+    // private:
+    /// channel ID
+    int chanID;
+    /// Net input no. of PEs
+    int truePE;
```
awhitbeck, line 97: this should be dropped

Event/include/Event/TrigScintQIEDigis.h
```diff
@@ -0,0 +1,111 @@
...
+    /// Whether or not there is noise
+    bool IsNoisy;
```
awhitbeck, line 99: these lines should be dropped

PR #934, opened 2020-11-03
Iss932 - v2.3.0 Validation

tomeichlersmith:
I needed some ldmx-sw changes to fix up the ECal digitization problems, so I figured we could have a validation-wide PR for any other small bug fixes that other subsystems might stumble across.

## Related PRs
- https://github.com/LDMX-Software/Ecal/pull/23
- https://github.com/LDMX-Software/SimCore/pull/38


omar-moreno:
@vdutta Validation files are at: /nfs/slac/g/ldmx/data/mc/v12/4gev_1e_pn_recon_iss932

tomeichlersmith:
Here is some inclusive validation (without comparison to past reconstruction). 100k 4GeV electrons were shot directly into the center of the ECal (from z=235mm). Some comments:

- The Reconstruction is only on the lowside of the simulation in the ADC region becuase the DIGI measures the maximum signal and not the actual peak. This is a ~1% effect that can be accounted for with tweaking the reconstruction parameters.
- The total reconstructed energy has a mean slightly under the desired 4GeV: 3.973GeV. After dividing out the current second order energy correction, we can see that the current reconstruction gets us within half a percent: (3973/4000)*(4010/4000) = 0.9957. This would mean our new second order energy correction would be ~1.004 which could roughly account for the effect mentioned above. We could include this updated correction factor now or wait until later.

Given the drastic under-the-hood changes that I've made to the DIGI, I am not surprised that we are seeing a shift between the older version and this one. Nevertheless, with the plots below, it looks like we are still faithfully reconstructing a wide range of energies within the ECal. 

## EcalDigiVerify Plots

Recon_v_Sim
TotalReconEnergy

## Ecal Digi Pipeline Test Plots

RecRatio_vs_SimEnergy
TPRatio_vs_SimEnergy
OutputE_vs_InputE_LogScale

PR #938, opened 2020-11-11
Add in electron counter and multielectron capable trigger

### LINKED ISSUE #936: Multielectron-aware missing energy trigger
We should make the trigger decision aware of the number of electrons in the event. (This is needed for multi-electron studies, for trigger skims on pileup overlay files, ... )

I'm planning to create a new trigger processor that evolves "simple trigger" to taking the number of beam electrons into account, using either a formula (likely too naive, given resolution effects at higher multiplicities) or looking up the (configurable) missing energy threshold corresponding to the electron count.
 
I also want to have a separate electron counter. It's not a big deal (the number could be pulled from collections inside the trigger processor) but I think 
1. this is closer to what we will have in reality: the trigger decision will get info on the ecal sums and the electron count from different sources
2. it will make it easier to switch between having the real (simulated) number of beam electrons, or the actually counted beam electrons, as the input to the trigger, which is needed for trigger performance studies 
3. it also modularizes the choice of electron counting algorithm used (which is a TS decision) and how to use that info (a TDAQ/Ecal decision). 

The flow I see is:
Run (Ecal and TS) digi/reco -->  any TS algorithms needed on the digi/reco hits to get the objects needed for an electron count -->  electron counting processor which sets the electron count in the event -->  the trigger algo using the electron count and Ecal energies.

So apart from adding two processors I'd like an electron count variable added to the event. I think this will be a globally useful variable. Actually, that and a "truth" number of beam electrons simulated. The overlay producer could increment that number.

I also considered calculating the electron count (from some configurable input collection) as part of the trigger processor but I think the modular approach will last longer. Advantage: doing it all in one go safeguards you against missing steps but this can be covered by exceptions (and indeed e.g. the Ecal veto already requires that reco was run, etc., so we use dependent flows already).

bryngemark:
Add two general use processors:
1. `ElectronCounter`, which can use a simulated number of electrons or (for now) a TS track collection to count the number of electrons on target.
2. `MultiElectronTriggerProcessor`, which uses the electron count set by the `ElectronCounter` (and/or possibly the overlay producer, later on) to look up which Ecal energy sum threshold to use in the trigger decision. Really, this is a minimal change to `TriggerProcessor`, and this for the 1e case it would run the same and almost transparently, but thresholds have to be configured as a list. Adding as a new class is minimally intrusive, but I'd be in favor of overwriting `TriggerProcessor` with this new implementation if everyone agrees.

Also `Event` got an electron count variable, and setter+getter. I'm thinking that the `ElectronCounter` could have an output collection with true/simulated number of electrons, and counted number of electrons, for performance studies. Right now that isn't implemented. 

The current implementation has a few possible improvements (TODO's in the code, could become feature requests/enhancement issues), but covers what is needed for multielectron and TS performance studies. I have tested that it produces sensible results on a 2e file, using both the simulated actual umber of beam electrons, and the count obtained from running TS tracking on the beamElectron ("truth") collection. I consider it ready for larger-scale validation with a v3.0 release. Closes #936 


Finally, I attach my run script in its entirety, if anyone wants to try it. Run it as:
`ldmx fire runElectronCounter.py nElectrons someInputFile.root `
where `nElectrons` = number of beam electrons in the simulation of `someInputFile.root`
You'd just need a sim level input file. 

```
#!/usr/bin/python                                                                                                                                                  

import sys
import os
                                                                                             
passName="truth"
p=ldmxcfg.Process( passName )
simPassName="truth" #<-- whatever clustering and tracking input  uses. doesn't have to be the pass created here                                                    

#maximum number of events to process                                                                                                                               
nEv = 1000

if len(sys.argv) < 1 :
    print("The number of beam electrons has to be specified. Use (positional) argument 1 for it.")
    exit(1)
else :
    nPart=sys.argv[1]   #number of beam electrons to use                                                                                                           
                                              
if len(sys.argv) > 2 :
    inputName=sys.argv[2]  #specify the input name                                                                                       
else:
     print("An input file name has to be specified. Use (positional) argument 2 for it.")
     exit(1)
print("Using input file name: "+inputName)

if len(sys.argv) > 3 :
    outputName=sys.argv[3]  #specify the output name if default is not desired                                                                                     
else:
    outputName="clustered_truth_"+inputName
print("Using output file name: "+outputName)

#log file 
lFileName = outputName.replace(".root", ".log")


# ------------------- all set; propagate these settings in detail, and run  ---------------                                                                        

#load the default configuration files                                                                                                                              

from LDMX.TrigScint.trigScint import TrigScintClusterProducer
from LDMX.TrigScint.trigScint import trigScintTrack
from LDMX.TrigScint.trigScint import TrigScintTrackProducer
from LDMX.TrigScint.truthHits import TruthHitProducer
from LDMX.EventProc.electronCounter import ElectronCounter
from LDMX.EventProc.multiElectronTrigger import multiElectronTrigger


#-- create truth beam electron collections                                                                                                                         
beTag=TruthHitProducer.tagger()
beUp=TruthHitProducer.up()
beDown=TruthHitProducer.down()

#-- digi of truth beam electron collections                                                                                                                        

digiTag=TrigScintDigiProducer.tagger()
digiUp=TrigScintDigiProducer.up()
digiDown=TrigScintDigiProducer.down()

digiTag.input_collection="truthBeamElectronsTag"                                                                                                                  
digiTag.pass_name=passName
digiUp.input_collection="truthBeamElectronsUp"
digiUp.pass_name=passName
digiDown.input_collection="truthBeamElectronsDn"
digiDown.pass_name=passName

#turn off noise hits in truth collections                                                                                                                          
digiTag.mean_noise=0.00000000001
digiUp.mean_noise=0.00000000001
digiDown.mean_noise=0.000000001

# --- clustering ---                                                                                                                                              \
                                                                                                                                                                   
clTag=TrigScintClusterProducer.tagger()
clUp=TrigScintClusterProducer.up()
clDown=TrigScintClusterProducer.down()

clTag.input_pass_name=simPassName
clUp.input_pass_name=simPassName
clDown.input_pass_name=simPassName


# --- tracking ---                                                                                                                                                \
                                                                                                                                                                   
trigScintTrack.input_pass_name=simPassName


# --- electron counting  ---                                                                                                                                                    

electronCounter = ElectronCounter( int(nPart) )
electronCounter.input_collection = "TriggerPadTracks"
# defaults to false 
#electronCounter.use_simulated_electron_number = True

# --- triggering  ---      

#if you want to change trigger thresholds. the default ones are dummies, chosen to give the same rate as the 1e default (1.5 GeV) and just as a flat number, drawn from valentina's plot, not weighted by expected multiplicity profile  
#multiElectronTrigger.thresholds = [ 1500.0, 6000.0, 8500.0, 1300.0 ]  #toy something up            


# Ecal stuff                                                                                                                                                       
import LDMX.Ecal.EcalGeometry
from LDMX.Ecal import digi

ecalDigi   =digi.EcalDigiProducer('EcalDigis')
ecalReco   =digi.EcalRecProducer('ecalRecon')


p.maxEvents=nEv

p.sequence = [ beTag, beUp, beDown, digiTag, digiUp, digiDown, clTag, clUp, clDown, trigScintTrack, electronCounter, ecalDigi, ecalReco, multiElectronTrigger ]


p.logFileName= lFileName
p.fileLogLevel=0
p.termLogLevel=0

# Provide the list of input files to run on                                                                                                                                                                                                                           
p.inputFiles=[ inputName ]                                                                                                                                                                                                                     
p.outputFiles=[outputName]

# Utility function to interpret and print out the configuration to the screen                                                                                      
print(p)

```


tomeichlersmith (commented):
- I *like* separating the basic trigger from this multi-electron trigger. The whole point of having a Framework that allows dynamic loading of processors is to allow us to have many processors doing a variety of things. Keeping each individual processor as simple as possible is a feature (I believe).
- I can't see the Framework changes on GitHub. This is usally indicative of not commiting changes to that repository and only to ldmx-sw. Since Framework is a separate repo, you need another branch and its own set of commits to track those changes and share them with GitHub.
- I am curious about why you chose to modify Event and not simply add another event bus object. That would be a lot less intrusive and (like you mention), you could include more information about the "electron count" than just one number. I can help you write an event bus object in the new (v3.0) style since Omar and I changed a lot of things in that respect on this branch.

bryngemark:
Thanks Tom! Indeed I had missed to finally push the commits I had done to Framework. This should be all good now.

The reason I'm thinking it would be interesting to change the basic `TriggerProcessor` are
1.  the difference is tiny and the actual logic (calculating energy sums over N layers) is identical. I just moved from using a single threshold to a list, and getting the number of electrons from somewhere. 1e-running as before would require two changes in the config: threshold = 1500. --> threshold = [ 1500.] , and an addition `electronCounter.use_simulated_electron_number = True`

As for modifying event, rather than add a collection and event bus passenger, this was mostly to deal with the case where you don't want to run the electron counter but just access the number of simulated events, that could be modified by e.g. the OverlayProducer directly, and use that count in your trigger directly. But this doesn't require pulling variables from particular collections.

tomeichlersmith:
I think your reasons for merging this processor into the TriggerProcessor are good, from my POV, you can feel free to do that.

On the other note, I am really opposed to the edits to Event. (This makes it sound like I'm angry, but I'm just trying to convey seriousness). I think `Event` needs to stay separate from any physics because it is already complicated enough. I know your specific changes are small and not complicated, but it smudges the line between Framework and physics. I would rather have two processors (e.g. OverlayProducer and ElectronCounter) both produce the same object that is used downstream rather than two processors edit a member variable of Event that could be used downstream.

bryngemark:
Ok, so if nobody objects soon, I'll go ahead and make my additions part of `TriggerProcessor` rather than add a (almost identical) new processor.

As for the event bus object. Would the EventHeader be a good place for that then? Or did you mean that there has to be a new output collection (that either of overlay and electron counting can produce)? I'm concerned though that collections are immutable so if I run overlay first, and routinely set an electron count in that process, and at some later point run electron counting, I'd have to have a different collection on that second go. Or? 

I think I'm also not entirely up to speed with the philosophy of the framework itself, and where clear cut lines should be drawn, so I might have a very different concept of these things than you (and Omar?) do! To me "event" is, well, the representation of a beam on target event, and certain global physics could be reasonable to keep there. Like the simulated number of electrons (which could vary between events if we allow a distribution). But maybe the counted number of electrons should indeed go somewhere else. 

The next logical step for this might be to start thinking in terms of a trigger collection where the different systems share their input. Like, the hcal veto (or PE count or whatever the hcal will communicate to the global trigger?), the electron count, the ecal energy sums.

tomeichlersmith:
You are correct that collections are immutable, so you would have to have a second "electron-count" event bus object added. Downstream processors would need to be able to determine which object to use. You could also have an option for the overlay producer that turns off its "electron-counting" behavior.

From my POV, the `Event` class is the bus that holds all the event information. This means it is just the container to hold the different objects (like the EventHeader, Sim hits, etc..). From this POV, there should be no physics information in that class so that we can isolate the "bus" from the "passengers" on the bus. You could make the electron count a member of EventHeader, but I still would advocate for the electron count to be its own "passenger" on the bus so that you can add more information to it (like e.g. which processor created it).

EventProc/include/EventProc/ElectronCounter.h
```diff
@@ -0,0 +1,90 @@
+
```
omar-moreno: Re-format.  See the command posted in #software-dev

omar-moreno:
> You are correct that collections are immutable, so you would have to have a second "electron-count" event bus object added. Downstream processors would need to be able to determine which object to use. You could also have an option for the overlay producer that turns off its "electron-counting" behavior.
> 
> From my POV, the `Event` class is the bus that holds all the event information. This means it is just the container to hold the different objects (like the EventHeader, Sim hits, etc..). From this POV, there should be no physics information in that class so that we can isolate the "bus" from the "passengers" on the bus. You could make the electron count a member of EventHeader, but I still would advocate for the electron count to be its own "passenger" on the bus so that you can add more information to it (like e.g. which processor created it).

I agree with this perspective.  On top of adding flexibility it will make maintaining much easier.

bryngemark:
Alright, since making an entire output collection for a single number per event didn't really make sense (and that would be a new type of event bus passenger etc), I went for the option to make use of the EventHeader, setting and getting the electron count as an intParameter. 

I've tested it, and it still works. So no need for changes to Event. I'm happy with this and would like it to be in the coming release, v3.0.0 I guess.

omar-moreno:
@bryngemark Adding the count as an int parameter means the changes to Framework are no longer needed correct?

omar-moreno:
So this branch is completely screwed up now. I'm going to merge in trunk, but the changes should be checked carefully in case the merge ends up with issues.

bryngemark:
Goddammit when will I learn that Github mention mails keep ending up in spam! Sorry, I missed this. 
Yes you're right the Framework changes won't be needed with the new approach to storing the electron count.

omar-moreno:
:) I'll patch all this up soon for the beta release.

omar-moreno:
@bryngemark I think this is ready now. Any chance you can run a quick test?

omar-moreno:
Dependent on [Framework#41](https://github.com/LDMX-Software/Framework/pull/41).

omar-moreno:
This is fairly stand alone so I'm just going to merge it in. We can test this with the beta-release.

bryngemark:
Reminder that the changes to the TriggerProcessor for some reason didn't make it in during the merge

bryngemark:
@omar-moreno here's the PR with the missing multielectron trigger changes that got lost. It looks like the relevant commit is [this one](https://github.com/LDMX-Software/ldmx-sw/pull/938/commits/427faa10475eac1dc98c08cfa2cb77ecec978172)

PR #940, opened 2020-11-12
Enabling EaT - Simulation Updates to allow studying the EaT analysis channel

### LINKED ISSUE #800: Partial Energy Sort during Simulation Processing
The default method of stacking during the Geant4 simulation is setup so that the particles are
processed in reverse creation-order. This means that the primary particle is simulated all the way to zero KE and then the last particle it produced is simulated, etc. until the stack is emptied.

A stacking strategy that partial sorts the particles by energy while they are being simulated would be the premier solution. This would allow us to make sure all the high energy particles are processed first, thus, making some filters _a lot_ more effiecient.

A basic first-order particle sort would make sure all particles above a certain energy threshold would be simulated before any particles below that threshold. This would also mean that particles that lose energy and go below the threshold are then suspended and aren't resumed until the other high-energy particles are dealt with.

This sorting method would be a `UserAction` that will need access to several different types: `STACKING`, `STEPPING`, and `TRACKING` I think are what's necessary.

One alternative is to force the particles to simulate in generation-order: primary, all primary's children, all primary's children's children, etc. This doesn't work as well becuase there are a lot of low energy particles that are produced that the simulation is wasting time on in the beginning.

**Additional context**
I got interested in developing this while researching ways to improve the efficiency of the ECal as Target simulation. This is intimately related to the mid-shower PN simulation as well.

### LINKED ISSUE #870: Dark Brem Filter Re-Design
I've been working on improving the UserActions that filter for different types of events used in the ECal as Target (EaT) project.

I am currently in the process of validating these filters for use with dark brems occurring in both the target and the ECal, so I wanted to bring that work into the official sphere.

This development will also include other filters that will be helpful for EaT including
- PrimaryToEcalFilter: Make sure the primary electron reaches the ecal with a certain energy
- EcalBremFilter: Make sure primary electron brems a photon of a minimum energy in the ECal
- EcalENFilter: Make sure primary electron loses at least a certain amount of energy to EN interactions in the ECal

## Still TODO
- [ ] Validate that the filters work for both the target and the ECal cases
- [x] Include and document target and ECal dark brem templates (like PN templates)
- [ ] Get [dependent changes to SimCore](https://github.com/LDMX-Software/SimCore/issues/20) validated and merged in

tomeichlersmith:
This resolves #870 .

## Updates
- Biasing filters to help create signal samples and biased backgrounds
- Dark brem process enhancements to fix up its runtime generation
- A simulation action that has tracks above an input energy threshold processed before all other tracks.

## Still To-Do
- [x] Rebase onto v3.0.0-dev
- [x] Double check xsec calculation against MadGraph/MadEvent
- [ ] Update analysis note on DocDB.

## Submodule PR
- https://github.com/LDMX-Software/SimCore/pull/34


tomeichlersmith:
This got messy and I'm going to work on cleaning it up. WIll open a new PR onto trunk soon.

PR #944, opened 2020-11-20
Add in bugfix to TS track downstream overlap removal

### LINKED ISSUE #943: TS tracking doesn't fully remove duplicates
**Describe the bug**
There are cases where clusters in the two downstream pads are assigned to different tagger seed clusters, creating a duplicate track. This happens because the overlap removal logic assumes that these overlapping tracks will be neighbors in the list of tracks, which is not always the case.

I have a patch for this.

bryngemark:
The TS tracking assumed that tracks sharing downstream clusters
would be at adjacent indices in the track list. This is far from
always the case (especially at higher electron multiplicities).
This patch now loops over all the remaining indices in the overlap
check.

Closes #943


bryngemark:
I just merged this in with the TrigScint main. It's a small fix but a large improvement in accuracy of the electron count out of TS and it would be great to have this in the next release.

PR #952, opened 2021-01-12
Adding trigger submodule

### LINKED ISSUE #950: Adding Trigger submodule
I would like to include a new submodule `Trigger` in ldmx-sw.  The submodule would include sub-directories for two types of algorithm implementations.  The first is pure software, which can be compiled within ldmx-sw, directly using the standard trigger reco objects, supports floating-point calculations, etc.  The second is hls code, which could be run in a standalone setup and (eventually) also directly in ldmx-sw.  This repo would also be included as a submodule in the APx firmware repository.

therwig:
This PR adds a Trigger submodule, addressing issue #950.
The `Algo` area contains the elements to be compiled within cms-sw, while Algo_HLS contains the elements that can be run standalone.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "docs/doxygen.conf/doxygen-awesome-css"]
 	path = docs/doxygen.conf/doxygen-awesome-css
 	url = https://github.com/jothepro/doxygen-awesome-css
+[submodule "Trigger"]
+	path = Trigger
+	url = git@github.com:LDMX-Software/Trigger.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -30,4 +30,4 @@
 	url = https://github.com/jothepro/doxygen-awesome-css
 [submodule "Trigger"]
 	path = Trigger
-	url = git@github.com:LDMX-Software/Trigger.git
+	url = https://github.com/LDMX-Software/Trigger.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,9 @@ if(BUILD_ALL OR BUILD_RECON_ONLY)
     # This module is focused on track reconstruction
     add_subdirectory(Tracking)
 
+    # This module contains trigger algorithms.
+    add_subdirectory(Trigger/Algo)
+
 endif()
 
 if(BUILD_ALL OR BUILD_SIM_ONLY)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,7 @@ add_subdirectory(Ecal ${CMAKE_BINARY_DIR}/EcalEvent)
 add_subdirectory(Hcal ${CMAKE_BINARY_DIR}/HcalEvent)
 add_subdirectory(TrigScint ${CMAKE_BINARY_DIR}/TrigScintEvent)
 add_subdirectory(Tracking ${CMAKE_BINARY_DIR}/TrackingEvent)
+add_subdirectory(Trigger/Algo ${CMAKE_BINARY_DIR}/TriggerEvent)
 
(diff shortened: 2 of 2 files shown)


omar-moreno:
I recommend splitting out Algo and Algo_HLS into their own modules.  I strongly prefer just pulling in code that will be used directly with ldmx-sw.

therwig:
Sorry to split the discussion across a few places, but the idea of the combined repository is to easily accommodate the ability to directly run the (hls) firmware in ldmx-sw.  I put together a quick-and-dirty example here:
https://github.com/LDMX-Software/Trigger/blob/trunk/Algo/src/Trigger/TriggerEcalEnergySum.cxx#L16

therwig:
The "usual" ldmx-sw Ecal energy sum is calculated alongside the firmware algorithm
https://github.com/LDMX-Software/Trigger/blob/trunk/Algo/src/Trigger/TriggerEcalEnergySum.cxx#L50
which is fed by the same fixed-point object representations that will be presented to the trigger card.

tomeichlersmith:
@therwig This PR has been open for a few years now. Is the branch able to easily update to the new trunk? If it can be updated and compile without issue, it'll be nice to include it in our next minor release.

therwig:
Yes, I'd really like to have it included. Let me try to bring it up to date and run some checks so we can aim to have it in.

.gitmodules
```diff
@@ -22,6 +22,9 @@
 [submodule "TrigScint"]
 	path = TrigScint
 	url = https://github.com/LDMX-Software/TrigScint
+[submodule "Trigger"]
+	path = Trigger
+	url = https://github.com/LDMX-Software/Trigger.git
```
tomeichlersmith: Since this commit was made prior to `Tracking`, this will cause a rebase CONFLICT and will need to be resolved manually. (Easy to do, just keep both).

CMakeLists.txt
```diff
@@ -108,6 +109,9 @@ if(BUILD_ALL OR BUILD_RECON_ONLY)
     # This module contains all things trigger scintillator
     add_subdirectory(TrigScint)
 
+    # This module contains trigger algorithms.
+    add_subdirectory(Trigger/Algo)
+
```
tomeichlersmith: Since this commit was made prior to Tracking, this will cause a rebase CONFLICT and will need to be resolved manually. (Easy to do, just keep both).

tomeichlersmith (changes requested):
I did a quick `git rebase` locally and the conflicts that arose were pretty simple and easy to address (just keep both Trigger and Tracking developments). I do have some requests before we merge this in (which I'm hoping to do this week so my SWAN presentation next week is about v3.3.0 ldmx-sw).

## Requests
Mainly about the format of the Trigger repository.
- Basic README about how the repo is apart of ldmx-sw. Also could dump notes about the goals and how to use it especially since its structure is slightly different from other ldmx-sw modules (if I am interpreting it correctly, `Trigger/Algo` is the ldmx-sw module and `Trigger/Algo_HLS` is for... on chip?)
- In other ldmx-sw modules, we keep the submodules in the root directory of the repo (see Tracking and SimCore), so please `git mv submodules/* .` just to keep the same structure.
- Once these cosmetic updates have been put in, please make a release. I'm trying to keep the submodules on tagged versions so it is easier to track their development from ldmx-sw.

tomeichlersmith (commented):
I chatted with @therwig on slack and he gave me the go-ahead to make the cosmetic updates. He'll double check my updates over the next few days and give any feedback on the Trigger PR

https://github.com/LDMX-Software/Trigger/pull/1

We're getting closer!

tomeichlersmith (approved):
Alright, with the basic tests passing and @therwig approving my updates to Trigger, I am ready to merge this in.

PR #954, opened 2021-01-15
Configurable ecal rec hits collection for ecal veto

tomeichlersmith:
This was a small change, so it is already on `trunk` of Ecal. Just need to update the submodule ref here.


bryngemark (commented):
Looks good!

tomeichlersmith:
This was indirectly merged in when @omar-moreno merged in his formatting updates into Ecal and then updated the submodule ref.

Issue #946, opened 2020-12-02
HGC ROC Emulation Improvement


tomeichlersmith:
_Context_: While @jmmans and I were debugging the weird behavior of the ECal reconstruction in the v2.3 validation samples, we talked about a lot of long-term improvements that "should" be made to the chip emulator but I didn't want to slow down the already late release.

### Timing Shift
The chip emulator should "know" what its time phase is relative to the simulation global time where t=0ns occurs when the primaries cross z=0mm. In the far future, this would be a conditions table assigning a time phase to each electronic ID corresponding to each chip. Since we aren't really using electronic IDs yet, there needs to be some middleground.

*Potential Solution*: I don't know of a better solution than having the ECal/HCal digi producers do the time shift themselves. They already need to change units from MeV to mV depending on their detector conditions, so this doesn't seem like too big of an ask. In this case, these comments are just warnings for @cmantill .

### Conditions Reliance
The chip emulator currently _can_ take input from a conditions table, but it should _require_ input from a conditions table. This would make HgcrocEmulator more closely resemble HgcrocTriggerCalculations and be more forward thinking.

*Potential Solution* I just need to do this. Not much else to it...

### TOA Measurement
Currently, the simulated time (after accounting for the time phase) is used as the time of the peak of the pulse. This means that for a lot of pulses, the TOA measurement (which is calculated as the time when the pulse goes over the TOA threshold) is negative. This is not realistic (obviously).

*Potential Solution* ~~Use the sim time to set the peak time, then calculate the TOA, then shift the peak so that the calculated TOA corresponds with the actual sim time, then inject noise and actual digitize.~~ The sampling time and the simulated pulse peak time should be close to the center of the clock window. Then we subtract off a shift from the measured TOA in the reconstruction stage so that a nominal high-energy pulse has TOA ~ 0ns (lower energy pulses will "walk" to later TOAs because their pulses are thinner naturally even if they don't actually arrive at a later time).

### Digitization of Multiple Hits
*These comments were generated when looking at how the HCal is going to use the emulator.* PR #958 

Currently, the list of voltage/time pairs at the chip are simply summed/averaged before performing the digitization procedure. This isn't very realistic and requires improvement. Outline of more realistic digitization of multiple hits:

1. Sample each pulse at the sampling times and add them together to construct one pulse.
2. Does the constructed pulse go over the TOT threshold? If so, do TOT mode with our constructed pulse.
[...]

PR #958, opened 2021-02-08
Adding realism to Hcal digitization and Hcal geometry

cmantill:
# Related Submodule PR
[Hcal Pr #5](https://github.com/LDMX-Software/Hcal/pull/5)

# Summary of Changes
- Adds HcalGeometry class 
 - Adds map of hcalID and (x/y,z) position
- Adds HcalAbstractID and HcalDigiID (with end of bar/strip where pulse is collected)
- Updates Hcal Digi producer with the use of HgcrocEmulator 
 - Considers double readout for back Hcal, single readout for side Hcal
 - no readout threshold
 - considering only ADC mode for now
- New Hcal Rec producer for recHits
  - Reconstructs energy from ADC readout by taking maximum measurement

Validated recHits energies and positions w. ADC readout only, see more details of PR here: https://www.dropbox.com/s/ndaqs70eba7j8wh/CMS_Hcal_Feb9_PR.pdf?dl=0

@nhanvtran


DetDescr/include/DetDescr/HcalAbstractID.h
```diff
@@ -0,0 +1,76 @@
...
+     * @brief Parent of HcalIDs
+     */
+    class HcalAbstractID : public DetectorID {
+
+    public:
+
+        /** 
+	 * Encodes whether the side of the strips of the HCal is saved
+	 */
+        enum HcalEndType {
```
tomeichlersmith: I don't think calling this an "End" type is helpful because it will inevitably end up being more general than that. Maybe just "Bar" type? That way the `HcalBarType::Digi` has an "end" defined, but `HcalBarType::Global` does not?

DetDescr/include/DetDescr/HcalDigiID.h
```diff
@@ -0,0 +1,137 @@
...
+	 */
+	int strip() const {
+	    return (id_>>STRIP_SHIFT) & STRIP_MASK;
+	}
+
+	/**
+         * Get the value of the 'end' field from the ID.
+         * @return The value of the 'end' field.
+         */
+        int getEnd() const {
```
tomeichlersmith: Since the end can only be two options, I think it would help future users if you returned a `bool`. You've defined that `1` is negative x/y, so maybe you could have an accessor which is `isNegativeEnd()` which returns `true` if the end flag is `1` and `false` if the end flag is `0`.

DetDescr/include/DetDescr/HcalGeometry.h
```diff
@@ -0,0 +1,130 @@
...
+             *
+             * Inputs x,y,z will be set to the calculated position
+             *
+             * @sa getStripCenterAbsolute and getZPosition
+             *
+             * @param[in] id HcalID for the strip we want the position of
+             * @param[out] xy set to x/y-coordinate of strip center
+             * @param[out] z set to z-coordinate of strip center
+             */
+            void getStripAbsolutePosition( ldmx::HcalID id, double &x, double &y, double &z ) const {
```
tomeichlersmith: This is a horrible relic of an older EcalHexReadout.

(I know I told you to copy the format of EcalHexReadout, and now I'm an being a hypocrite.) Please remove this function and simply use `getStripCenterAbsolute` where-ever you would use `getStripAbsolutePosition`. Feel free to change the name if you would like to as well.

DetDescr/include/DetDescr/HcalGeometry.h
```diff
@@ -0,0 +1,130 @@
...
+  
+    /**
+     * @class HcalGeometry
+     * @brief Implementation of HCal bar readout
+     *
+     */
+    class HcalGeometry : public framework::ConditionsObject {
+
+        public:
+            static constexpr const char* CONDITIONS_OBJECT_NAME{"HcalGeometry"};
```
tomeichlersmith: Document. Specifically, mention that this name **needs to match the name used in the Python configuration exactly**.

DetDescr/include/DetDescr/HcalGeometry.h
```diff
@@ -0,0 +1,130 @@
...
+
+            /**
+             * Class constructor, for use only by the provider
+             *
+             * @param ps Parameters to configure the HcalGeometry
+             */
+            HcalGeometry(const framework::config::Parameters &ps);
+            friend class hcal::HcalGeometryProvider;
+
+            void buildStripPositionMap();
```
tomeichlersmith: This is where a lot of fun stuff happens! (in your source). Give an overview in the documentation here so that future users can understand what you've done.

DetDescr/include/DetDescr/HcalGeometry.h
```diff
@@ -0,0 +1,130 @@
...
+
+        private:
+
+            /// verbosity, not configurable but helpful if developing
+            int verbose_{2};
```
tomeichlersmith: I know you will probably want to leave the verbosity high while you are developing, but this comment is a reminder to bring it back down to zero before we merge it to `trunk`.

DetDescr/include/DetDescr/HcalGeometry.h
```diff
@@ -0,0 +1,130 @@
...
+
+	    /// thickness of scintillator 
+	    double hcalThicknessScint_{0.};
```
tomeichlersmith: Don't give a default value unless there is a good reason to. In this case (and the one below this comment), they *will* be set by the `getParameter` calls, so you don't need to define a default.

DetDescr/include/DetDescr/HcalGeometry.h
```diff
@@ -0,0 +1,130 @@
...
+	    std::vector<double> hcalLayerThickness_;
+
+	    /// Number of layers in each section 
+	    std::vector<int> hcalNLayers_;
+
+            /// Number of strips per layer in each section
+	    std::vector<int> hcalNStrips_;
+
+            /// Position of bar centers relative to world geometry (uses ID with real bar and section and layer as zero for key)
+            std::map<ldmx::HcalID, std::tuple<double,double,double>> stripPositionMap_;
```
tomeichlersmith: - Expand this comment and mention reference the function that this map is built in.
- Use `std::unordered_map` instead of `std::map`. (Faster and operates in the same way.)
- Mention that this is *not* configurable, and is calculated by this class.

DetDescr/python/HcalGeometry.py
```diff
@@ -0,0 +1,58 @@
...
+        self.hcalNLayers = []
+        self.hcalNStrips = []
+        self.hcalHalfTotalWidthBack = 0.
+        self.verbose = 0
+
+    def __str__(self) :
+        """Stringify this configuration class"""
+
+        s = 'HcalReadoutGeometry'
+        return s
```
tomeichlersmith, line 38: This `__str__` method should help a future user figure out that they have configured their run properly. Include some detail of the other parameters, maybe print `detectors_valid` as well?

DetDescr/python/HcalGeometry.py
```diff
@@ -0,0 +1,58 @@
...
+
+    """
+
+    def __init__(self):
+        self.make_v12()
+    
+    def make_v12(self) :
+        """Create the HcalGeometry with the v12 geometry parameters
+
+        Only sets parameters that must align with the gdml
```
tomeichlersmith: Please mention how you determined the numbers that are here. A first time user would be very confused, so even just a sentence or two of explanation would be helpful.

DetDescr/src/DetDescr/HcalGeometry.cxx
```diff
@@ -0,0 +1,69 @@
...
+        hcalNStrips_        = ps.getParameter<std::vector< int >>("hcalNStrips");
+        hcalHalfTotalWidthBack_ = ps.getParameter<double>("hcalHalfTotalWidthBack");
+        verbose_            = ps.getParameter<int>("verbose");
+
+	buildStripPositionMap();
+    }
+
+    void HcalGeometry::buildStripPositionMap() {
+
+      for(int section=0; section<=ldmx::HcalID::HcalSection::LEFT; section++) {
```
tomeichlersmith: This will break if some future person makes `LEFT` *not* the last section listed. I would suggest, hard-coding the number of sections here and putting in a comment in the definition of `HcalID::HcalSection` enumerator where the number of section is used.

Tools/include/Tools/HgcrocEmulator.h
```diff
@@ -156,7 +156,9 @@ class HgcrocEmulator {
   bool digitize(
       const int& channelID, const std::vector<double>& voltages,
       const std::vector<double>& times,
-      std::vector<ldmx::HgcrocDigiCollection::Sample>& digiToAdd) const;
+      std::vector<ldmx::HgcrocDigiCollection::Sample>& digiToAdd,
+      const double &attenuation=0.,                                                                                                                                                           
+      const double &shift=0.) const;
```
tomeichlersmith: These are smart additions to the emulator. Could you give a comment in the docs-comment about what they are for? Specifically, units are especially important.

Tools/python/HgcrocEmulator.py
```diff
@@ -90,6 +90,19 @@ def calculateVoltage(self, electrons) :
         """
 
         return electrons*(0.162/1000.)*(1./self.readoutPadCapacitance)
+
+    def calculateVoltageHcal(self, PE) :
```
tomeichlersmith: I'm starting to think that the `calculateVoltage`, `calculateVoltageHcal`, `setThresholdDefaults` and `setThresholdDefaultsHcal` should just be deleted in favor of moving those definitions to our
```python
def EcalHgcrocEmulator() : 
    <subsystem-specific-stuff>
```
function defintions. What do you think?

I can handle copying over that stuff to the ECal and adding that to this PR if you like that idea.

tomeichlersmith (changes requested):
# Questions from your slides

## Maintaining ADC Mode through High Energies
TOT mode kicks in when-ever the chip is saturated. The chip has a few (two or three?) settings on the charge it can handle before it is saturated; after that point, it would go into TOT readout mode.

TOT readout mode is actually simpler to modle than ADC mode since we simply calculate the amount of charge deposited and then the TOT (= Time Over Threshold) is just the time it takes for this charge to be drained off the chip in a linear fashion. The parameters that control this linear drain (the drain rate and the charge at-which the drain stops) are where the complexity lies and is not our (software people) immediate problem.

*Edit*: I have been informed that TOT mode for the HCal is far more complex than for the ECal, so including it in this already large PR may not be feasible.

## Increase readout threshold and lower gain
This is mainly a function of the HCal's purpose, so this discussion requires some input from more HCal experts. Nevertheless, I would say that you should keep the readou threshold low. My overly simplistic reasoning is that the HCal needs to "catch" anything the ECal misses, so making it extra senstive (even a little too sensitive) is worthwhile.

## (x/y) Position for Single-Ended Readout
Make it the mid-point of the strip/bar or undefined. I think it is up to the people using your rec hits downstream to handle the single-ended readout rec hits correctly.

## Implement noise on empty hits
**Yes**. This is important for the ECal and even more important for the HCal given that it has less activity. The [NoiseGenerator](https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Tools/include/Tools/NoiseGenerator.h) randomly generates a list of pure-noise"hit" amplitudes and then you are responsible for randomly generating a detector ID to go with that hit. _ comments:
1. Make sure the detector ID you generate for your noise "hit" is *not* one that already exists. The HgcrocEmulator already puts noise on top of real hits.
2. The number and amplitude of these noise hits depends on several physical factors of the detector. These physical factors are input into the NoiseGenerator and the exact values of these factors will need to be discussed with HCal experts. The NoiseGenerator does not care about units, but **all of the units of these input factors should be the same**.
3. The Ecal generates noise hits and then runs them through the emulator. This isn't *super* realistic, so don't feel required to do this if another solution presents itself.

# General Comments
- _Great Job_! This is a large amount of improvment!
- Format the C++ files to our coding format using. `find . -type f -name "*.h" -o -name "*.cxx" -exec clang-format --style=Google {} \;` (If you don't have `clang-format` installed, I can do this for you.)
- More documentation. Every detail you can possibly think of including should be included!

DetDescr/include/DetDescr/HcalGeometry.h
jmmans: Do we have a common 3d point type in ldmx-sw?  Ideally, this function would return a "Point3d" or whatever.

DetDescr/include/DetDescr/HcalGeometry.h
jmmans: I would think of this logically the opposite direction.  The Python has to match this.  The point is a good reminder however for the far future.

Tools/include/Tools/HgcrocEmulator.h
jmmans: So, I'm a little concerned here.  The HCAL bars are big and we could have hits at quite different times, when taking into account the light-propagation effects.  Is this to be called only for 'voltages' within a small region in space along the bar?  I guess these 'voltages' are really PE or something similar?

jmmans:
For the noise hits, they need to be distributed in time as well as space.  (This is true for Ecal also).  We'll need parameters to determine the time window for consideration (+/- X ns)

DetDescr/include/DetDescr/HcalGeometry.h
tomeichlersmith: We do not. The nearest option that is readily available would be using [one of ROOT's Physics classes](https://root.cern.ch/doc/master/classTVector3.html).

Tools/include/Tools/HgcrocEmulator.h
tomeichlersmith: You are correct. The 'voltages' are just unit-converted individual sim hits. I've called them 'voltages' to emphasize that the units should be mV and not fC or MeV.

tomeichlersmith: @jmmans and I talked about the best way to move forward and I believe this is in two stages.

### 1. Remove the attenuation and shift from this digitize function.
Without changing the HgcrocEmulator on `trunk`, you would get the same affect as what you have on this branch now if you factor in the attenuation and shift you calculated *onto each hit individually*.

This is important because, particularly for the HCal, hits can occur at widely-separated positions or times (but in the same bar), so their attenuations and shifts would be wildly different. 

Long story short, the voltages/times input into the `digitize` function should the voltages/times **at the chip** and should be calculated **for each sim hit individually**.

### 2. Update HgcrocEmulator to handle multiple hits better
The shortcut that is currently in the emulator works well for single hits in each cell/bar or for multiple hits that are near in time and space. This shortcut works for a majority of digitizations in both the ECal and HCal, but is not realistic.

@jmmans and I will work together on improving the emulator to handle multiple hits appropriately (Issue #946 ). Also, if the HCal calculates the attenuation and shift *for each sim-hit individually*, then the HCal would not need to change their code.

DetDescr/include/DetDescr/HcalAbstractID.h
cmantill: Sounds good. Implemented in [1589358](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/1589358813a02a43308ad6bbdb0c64ac90aa9c15)

DetDescr/include/DetDescr/HcalDigiID.h
cmantill: Sounds good. Implemented in [1589358](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/1589358813a02a43308ad6bbdb0c64ac90aa9c15)

DetDescr/include/DetDescr/HcalGeometry.h
cmantill: I switched to TVector3, seems to work well enough. See [51e1bfa](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/51e1bfab6506a789417f7ac3c85197d363f4a58c)

DetDescr/include/DetDescr/HcalGeometry.h
cmantill: Added this comment in [51e1bfa](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/51e1bfab6506a789417f7ac3c85197d363f4a58c)

DetDescr/include/DetDescr/HcalGeometry.h
cmantill: I added comments here [51e1bfa](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/51e1bfab6506a789417f7ac3c85197d363f4a58c) and also in the HcalGeometry.py (where the values are given) here [df46bdb](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/df46bdbce73f267d95019ba1dad808854a9c6449)

DetDescr/include/DetDescr/HcalGeometry.h
cmantill: Sounds good, switched to `verbose_{0}`.

DetDescr/include/DetDescr/HcalGeometry.h
cmantill: Done.

DetDescr/include/DetDescr/HcalGeometry.h
cmantill: I kept it as an `std::map` for now - was giving me problems compiling. Added reference and documentation in the function itself.

cmantill (commented):
Sorry, messed up by starting review and can't seem to go back. 

Answering here your main comments @tomeichlersmith: 
- **TOT**: Yes, I don't think I will implement this in this PR, I need further discussion with Hcal/Hgcroc experts
- **Noise hits**: I have put a first implementation basically identical to what Ecal does, but with the correct number of empty channels. It compiles and runs but will need to test/discuss this further so hold on for now.
TO DO: need to distribute them in space too.
- **X/Y position**: I have set this as the middle point in the bar. I have also extensively commented the HcalGeometry class, how the coordinates are calculated and how are the parameters obtained w. respect to the Hcal gdml constants
- **Gain/threshold**: I haven't received further feedback but it seems that for now the high gain/low threshold approach will work given that we are not using the TOT mode.

Tools/include/Tools/HgcrocEmulator.h
cmantill: Alright, I think this makes sense. I have included these modifications [here](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/49d1a6bdbe8242520b52a9782e981f100e57fc6f) and [here for Hcal](https://github.com/LDMX-Software/Hcal/commit/4c9b4b80fe20ba75b9bd81690b0039bbc6068a8c)

Tools/python/HgcrocEmulator.py
cmantill: I moved everything to the `HcalHgcrocEmulator()` and defined it as a class [here](https://github.com/LDMX-Software/Hcal/commit/deb2a21ef841041d432cc1f15989c813356996c7). Let me know if this is similar to what you were thinking.

DetDescr/python/HcalGeometry.py
cmantill: Implemented. Added in [here](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/51e1bfab6506a789417f7ac3c85197d363f4a58c)

DetDescr/python/HcalGeometry.py
cmantill: Done [here](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/51e1bfab6506a789417f7ac3c85197d363f4a58c)

DetDescr/src/DetDescr/HcalGeometry.cxx
cmantill: I switched this to the `Nsections_` parameter given by `HcalGeometry.py`. Is this in line with what you were thinking? See [here](https://github.com/LDMX-Software/ldmx-sw/blob/51e1bfab6506a789417f7ac3c85197d363f4a58c/DetDescr/src/DetDescr/HcalGeometry.cxx#L28-L30)

tomeichlersmith: Yes, I think that makes sense.

DetDescr/include/DetDescr/HcalDigiID.h
```diff
@@ -0,0 +1,131 @@
...
+   */
+  int strip() const { return (id_ >> STRIP_SHIFT) & STRIP_MASK; }
+
+  /**
+   * Get whether the 'end' field from the ID is negative.
+   * @return True if the end of the strip is negative
+   */
+  bool isNegativeEnd() const {
+    int end = (id_ >> END_SHIFT) & END_MASK;
+    if (end == 1)
```
tomeichlersmith: Avoid code repetition by moving the definition of `end()` above this function and then deleting the `int end` line and replacing `end` with `end()` in the `if` condition.

DetDescr/include/DetDescr/HcalGeometry.h
```diff
@@ -0,0 +1,152 @@
...
+  int getNumLayers(int isection) const { return NumLayers_.at(isection); }
+
+  /**
+   * Get the number of strips per layer for that section.
+   */
+  int getNumStrips(int isection) const { return NumStrips_.at(isection); }
+
+  static HcalGeometry* debugMake(const framework::config::Parameters& p) {
+    return new HcalGeometry(p);
+  }
```
tomeichlersmith: For Ecal, we have this for some drawing mechanisms inside of the Ecal module. Do you use this anywhere? If not, you should delete it so that it isn't accidentally used by a developer down the line. If you do use it, put in a documentation comment and add a warning that this shouldn't be used inside of processors! (Use conditions instead).

DetDescr/include/DetDescr/HcalGeometry.h
```diff
@@ -0,0 +1,152 @@
...
+
+  /// Number of strips per layer in each section
+  std::vector<int> NumStrips_;
+
+  /// Number of sections
+  int NumSections_;
+
+  /*
+   Map of the HcalID position of strip centers relative to world geometry.
+   The map is not configurable and is calculated by buildStripPositionMap().
```
tomeichlersmith: For doxygen to pick up this documentation comment, it needs to be formatted like
```cpp
/** <- two asterisks
 * <- asterisk beginning each line
 */
```

DetDescr/python/HcalGeometry.py
```diff
@@ -0,0 +1,106 @@
...
+
+        - NumSections:
+          Number of sections. 
+          Encoded in HcalID.
+          Back, Top, Left, Right, Bottom.
+
+        - HalfTotalWidth: 
+          For back Hcal: nstrips * scint_width / 2 (in x/y)
+          For side Hcal: (nlayers_otherside * layer_thick + Ecal_dx(y)/2)/2
+          @param gdml see: sideTB_dx, sideTB_dy
```
tomeichlersmith: Just want to say that this is :partying_face: awesome :tada: Great documentation!

tomeichlersmith (approved):
Review Round 2.

Things here just need a few more cleanup stuff. The main comments about important code is in the submodule PR.

DetDescr/python/HcalGeometry.py
cmantill: While exploring sideHcal pulse attenuation, I ended up adding Ecal dx/dy parameters and fixing a small bug on the halfwidth calculation for sideHcal: [here](https://github.com/LDMX-Software/ldmx-sw/blob/a7beaf9d9cc5a5222921dba686895006419b8437/DetDescr/python/HcalGeometry.py#L108-L114)

DetDescr/include/DetDescr/HcalGeometry.h
cmantill: Right did that in both the header file [here](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/8cd3c51eada7122290b184f04356d461949ca0fe) and the .cxx [here](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/a59c8ba4f6414cb442eb1237db0dbf1bbec8ab3f)

DetDescr/include/DetDescr/HcalGeometry.h
cmantill: I do not use it so I [deleted for now](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/8cd3c51eada7122290b184f04356d461949ca0fe).

DetDescr/include/DetDescr/HcalDigiID.h
cmantill: Good point, implemented that [here](https://github.com/LDMX-Software/ldmx-sw/pull/958/commits/8cd3c51eada7122290b184f04356d461949ca0fe0)

cmantill:
Ah, @omar-moreno, some Hcal tests will probably fail because we currently test up to 500 PEs, whereas the current gain in the ADC mode covers up to 200 PEs before saturating.

omar-moreno:
> Ah, @omar-moreno, some Hcal tests will probably fail because we currently test up to 500 PEs, whereas the current gain in the ADC mode covers up to 200 PEs before saturating.

Just realized this. Can you update the test?

cmantill:
> > Ah, @omar-moreno, some Hcal tests will probably fail because we currently test up to 500 PEs, whereas the current gain in the ADC mode covers up to 200 PEs before saturating.
> 
> Just realized this. Can you update the test?

Alright just made a PR to Hcal [here](https://github.com/LDMX-Software/Hcal/pull/6)

omar-moreno:
Thanks!

PR #965, opened 2021-02-25
Launch Display-connected programs from the container

tomeichlersmith:
A small tweak to the environment script combined with an updated container allows users (on Linux and MacOS systems) to run display-connected programs from the container.

In this PR I also clean up so that the `BUILD_EVE` option is *not* mutually exclusive with building the rest of ldmx-sw. It is still `OFF` by default, but a user could choose to build it alongside the rest of ldmx-sw if they wish.


tomeichlersmith:
While the SimCore PR linked below is not _required_ for this PR. It is helpful to merge them at the same time.
https://github.com/LDMX-Software/SimCore/pull/46

omar-moreno:
update sim core and i'll merge in.

tomeichlersmith:
@omar-moreno All ready

PR #969, opened 2021-03-12
Some long-anticipated tweaks to the overlay producer (event wrapping and start event randomization)

### LINKED ISSUE #967: Enhance overlay producer event randomization
This is part of the long foreseen development of the overlay producer's initially desired list of capabilities. I held off on it initially becasue v3.0 was in the works and I need to change EventFile a little bit.

If we were to have one unique pileup event for every interesting sim event we want to use in overlay, we'd need a large (disk space consuming) sample. It's been agreed with Physics coordination that the pool of pileup events to pull from can be 10 or 100 times smaller than e.g. the number of PN events, as long as we reuse events as randomly as possible. 

**Features needed**
- the pileup file used for a job is randomized --> this is already implemented in LDCS 
- the starting event number in the file used to pull pileup events should be randomized to maximally avoid any systematic effects when files are reused
- if we then hit the end of the pileup event tree sooner than we're out of sim events, we'd need to restart from the beginning of the pileup tree.

The event number wrapping is also a protection for the case when there is a small mismatch between the number of events in the pileup file and sim file. Rather than using a fixed starting number and making all pileup files large enough to accommodate the largest occurring number of PN events surviving filtering in a given sample, which introduces a lot of unused events, we can keep the pileup file sizes to the typically needed number of events. 

**Describe the solution**
This can be done with a simple tweak to EventFile, where for the case that the file at hand is an "overlay file", i.e. just pileup, nextEvent doesn't return when we hit the last event, but resets the event counter. I have implemented and tested this (in v2.x, so I'll just port that over to the current structure). Overlay files are identified with an additional boolean, defaulting to false, added as the last argument to the general constructor (which minimizes interference with existing overloaded functions). A specific overlay file constructor is added, with an extra boolean, which the overlay producer calls.

### LINKED ISSUE #968: Make the number of earlier/later bunches used for pileup independent
In the current implementation, out-of-time pileup is placed symmetrically in +/- _m_ bunches. This should be separately configurable (so place pileup in a random bunch from a continuum starting 3 bunches before, to 10 bunches after, or whatever). 

**Describe the solution**
I'm going to change the setup in the `OverlayProducer` such that two numbers are specified instead of one.

bryngemark:
This adapts `EventFile` as needed to have a special treatment in `nextEvent` when the file used is a pileup input file (used for overlay). All it does is to reset the event counter to start from the beginning, instead of return/crash, when we reach the end of the tree. 

In addition, the pileup event _start number_ is randomized (on an interval up to 500, a little arbitrary but I don't think we'll reuse the same pileup file more than at the very most ~100 times, and I couldn't figure out immediately how to pull the number of total events from the pileup tree).
Here I'm seeding the random number generator on time (with `(0.)`) because this happens before `produce()` is called, and until then, the RNSS is ill-defined, as far as I understand. Better suggestions are welcome. 

Finally, I have made the interval of bunches where we want to put out-of-time pileup fully configurable. It used to be symmetrical around bunch 0 (the sim event bunch). 

This is all tested and seems to work as intended. Needed for the upcoming PN overlay campaign. Closes #967 and closes #968 .


Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -358,26 +360,26 @@ void OverlayProducer::onProcessStart() {
...
   // grab the first event in the processor
-  // TODO this could also be done N random times to get a randomness in which
-  // events get matched to what sim event. noticed that shifting by a fair chunk
-  // helps remove some weak but suspicious correlations between sim and overlay
-  // particle positions. leave it hardwired until we can reset the overlay event
-  // counter in nextEvent() (implemented with the EventFile definition above)
-  int nEventsShift_ = 23;
-  for (int iShift = 0; iShift < nEventsShift_; iShift++) {
+
+  TRandom2 * rd = new TRandom2( 0.); //RNSS undefined before produce()
```
tomeichlersmith: I would really like us to figure out how you can use RNSS here so that overlay files can be faithfully re-simulated from scratch. Technically, the RNSS is not initialized until the `onNewRun` method of the conditions object provider is called.

https://github.com/LDMX-Software/Framework/blob/1d5c8aac2004444a58a746a52dcfe3b54d98c977/src/Framework/Process.cxx#L163-L170

Notice that _producers_ are called before _and_ after the conditions are initialized so that they can indirectly affect the conditions object providers via the run header. This is actually used by the simulator so that it can tell the EcalGeometryProvider what geometry to use.

For this case, I think you would be able to move the "initialization" of the overlay file to the `onNewRun` method. This is _after_ the RNSS is initialized, so I think you would be able to access it in the normal way. You can look at the Simulator for how it initializes the Geant4 random number generator.

https://github.com/LDMX-Software/SimCore/blob/61c92b0225e83a9f422558daef80f2afb4be8e35/src/SimCore/Simulator.cxx#L283-L291

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -358,26 +360,26 @@ void OverlayProducer::onProcessStart() {
...
+  // shift away a bit from the first event to avoid some weak correlations
+  // seen during initial valiadations
+  // TODO use the actual number of events in the file as upper limit
+  int startEvent = rd->Uniform(20., 500.); //assume we have at least 500 events
```
tomeichlersmith: I don't think you need to use the actual number of events in the file as the upper limit. I think you could just pick some absurbly high number (like 1M) as the upper limit because you already are wrapping around the file. It is a waste of time to go around the file again and again, but you only do it once at the start of a run.

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -358,26 +360,26 @@ void OverlayProducer::onProcessStart() {
...
+  // and if not, event number wrapping will take care of it
+  for (int iShift = 0; iShift < startEvent; iShift++) {
     if (!overlayFile_->nextEvent()) {
       std::cerr << "Couldn't read next event!";
       return;
     }
   }
```
tomeichlersmith: Now that I think about it... Since you are already making some small updates to the EventFile class, you could just add a `skip` method. In that method, you would have access to the number of events in the file, so OverlayProducer gives a randomly generated offset to the skip method and then...

```cpp
bool EventFile::skip(int off) {
  off = off % entries_; // don't want to waste time looping
  for (int i_shift{0}; i_shift < off; i_shift++) {
    if (!this->nextEvent()) return false;
  }
  return true;
}
```

Then your initialization procedure here would just need to be
```cpp
int start_event = rd->Uniform(0.,1e6); // EventFile::skip handles number of events in file
if (!overlayFile_->skip(start_event)) {
  EXCEPTION_RAISE("BadRead",
    "Couldn't read to starting offset.");
}
```

tomeichlersmith (commented):
Things look good overall. I have some ideas for cleaning up that I have put into code comments. Could you also open a PR in the Framework submodule? I didn't see anything needing comment, but that will help us get it merged in.

Recon/src/Recon/OverlayProducer.cxx
bryngemark: I'll test it a bit and see. I don't think an absurdly high number is needed though :) 10k is probably enough to cover the number of events in a pileup file. Still, I don't think the start event number needs to be pulled from a pool that is much larger than the actual times the file will be reused, and I don't think we'll want this number to be much larger than 100.

Recon/src/Recon/OverlayProducer.cxx
bryngemark: Cool, thanks, I'll give it a try!

Recon/src/Recon/OverlayProducer.cxx
bryngemark: True, but in that case, why loop at all? Since we have already established that we're within the actual number of events, can't I just implement a setter for the entry number instead? Or a better name like  
```
bool EventFile::skipToEvent(int off) {
  ientry_ = off % entries_; // don't want to waste time looping
  if (!this->nextEvent()) return false;
  return true;
} 
```

I would still use your suggested implementation in the `OverlayProducer` though and throw an exception if this event turns out to not be readable.

Recon/src/Recon/OverlayProducer.cxx
bryngemark: Alright, done, works

Recon/src/Recon/OverlayProducer.cxx
bryngemark: Implemented it without the loop, works.

omar-moreno (approved):
Once the question by Tom in the Framework PR has been addressed, the rest looks fine.

bryngemark:
This should be all done now, I implemented Tom's two suggestions

tomeichlersmith (approved):

PR #970, opened 2021-03-12
Iss966 - Update container-interacting bash functions

### LINKED ISSUE #957: singularity preserves environment variables which may cause failures
**Describe the bug**
`singularity` preserves environmental variables defined by the host system. This is an issue if the host system happens to define an environmental variable we use, for example `ROOTSYS`. `singularity` has an option to *not* pass the environment from the host, so we include that flag in our definition of the `ldmx` function.

**To Fix**
In ldmx-sw/scripts/ldmx-env.sh, change the following:
```
# in the ldmx() command...
        singularity run \
            --no-home \
            --bind ${LDMX_BASE} \
            --cleanenv \
            --env LDMX_BASE=${LDMX_BASE} \
            ${LDMX_...
```
The `--cleanenv` flag tells singularity to not copy over all the environment variables from the host computer. Then the `--env` flag tells singularity to define that environmental variable inside the container. This ensures that *only* `LDMX_BASE` is passed between the host and the container.

### LINKED ISSUE #966: Tab Completion when interacting with container
I just found out that you can [customize how your commands respond to tab-completion requests](https://iridakos.com/programming/2018/03/01/bash-programmable-completion-tutorial). This is pretty rad, and I want to do it. My idea is to have `ldmx` be **the** command with multiple sub-commands (kinda like how `docker` has multiple sub-commands). The command tree I'm thinking of implementing is

- `ldmx`
     - `list` : list tags for the input repo name
     - `clean` : remove all containers and container images on the local machine
     - `config` : print out configuration to stdout
     - `pull` : pull down input container image and set it to be used
     - `use` : use input container image and only pull down if it doesn't exist on local machine
     - `run` : Got to first argument and then run everything after it in the container (basically what the entry-point script does)
     - `fire <stuff>` : Equivalent to `run $(pwd) fire <stuff>`
     - `make <stuff>` : Equivalent to `run $(pwd) make <stuff>`
     - `cmake <stuff>` : Equivalent to `run $(pwd) cmake <stuff>`
     - (and so on with other ldmx-sw commands, I think I will be able to automate this to pick up all executables in the ldmx-sw install)

tomeichlersmith:
Since PR #965 requires some tweaks to the environment script, I have developed this updated container interface on top of those changes.

This resolves #966 by implementing what was discussed there, not much else to it.

At the request of @jmmans , I've included the **great idea** of having a `.ldmxrc` to persist the local configuration of the ldmx-sw computing environment. In the tree currently is a default configuration as an example.

:thinking: Maybe I want to write my own shell. :thinking: We could call it `ldmxsh` :laughing: 

[Why end with "rc"](https://superuser.com/questions/173165/what-does-the-rc-in-bashrc-etc-mean). I am open to other naming conventions too.

### DIFF
--- a/.ldmxrc
+++ b/.ldmxrc
@@ -1,0 +1,20 @@
+###############################################################################
+# Default Configuration File for ldmx-sw Container Environment
+#   This file is "sourced" everytime the environment script is sourced.
+#   i.e. 'source scripts/ldmx-env.sh --> ldmx source .ldmxrc'
+#
+#   Think of this file like a '.bashrc'. You can put your custom container
+#   configuration directly in this file. 
+#
+#   Each line starting with '#' is ignored. 
+#   White space is ignored.
+#   All other lines are given to the 'ldmx' command.
+#   This file is sourced in the directory that it is in.
+###############################################################################
+
+# LDMX_BASE should be parent directory of our current directory by default
+base ..
+
+# Use dev/latest container by default
+use dev latest
+
--- a/README.md
+++ b/README.md
@@ -44,10 +44,14 @@ Command | Purpose
 
 The environment script defines several other shell commands to help configure and debug the container environment.
 
-- `ldmx-container-tags repo` : List the container tags that you could use with the input repository: `dev`, `pro`, or `local`
-- `ldmx-container-pull repo tag` : Setup the environment for the container 'ldmx/repo:tag' and pull down the newest version if the repo is remote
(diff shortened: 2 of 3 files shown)


tomeichlersmith:
I've tested this on both docker and singularity systems, so please let me know if it works on your system. Let me know if you need help figuring it out, but you can look at the README or run `ldmx help` for a surface-level reference.

omar-moreno (approved):

PR #971, opened 2021-03-15
Increase HGCROC noise threshold by one ADC count

tomeichlersmith:
This moves the threshold from ~3.2 sigma noise to ~4.8 sigma noise
hopefully decreasing the number of noise hits per event

This is a patch that really helps the ECal avoid generating too many noise hits.

### DIFF
--- a/Tools/python/HgcrocEmulator.py
+++ b/Tools/python/HgcrocEmulator.py
@@ -67,7 +67,7 @@ class HgcrocEmulator() :
         self.noiseRMS         = 0. #mV - useless default
         self.setNoise( 700. , 25. ) #depends on readoutPadCapacitance
 
-        self.readoutThreshold = self.pedestal + 2. #ADC Counts
+        self.readoutThreshold = self.pedestal + 3. #ADC Counts
 
         self.toaThreshold     = 0. #mV - useless default
         self.totThreshold     = 0. #mV - useless default


omar-moreno (approved):

tomeichlersmith:
I've done 10k events and compared a readout threshold 2 ADC counts above the pedestal (what it is now) with 3 ADC counts above pedestal. The only difference I can see is in the number of linear regression tracks. This makes sense to me because the lin/reg tracks try to find lines with whatever hits are leftover. These hits are probably dis-proportionately noise hits and when there are more noise hits (2 ADC case), it is easier to happen across lines (more lin reg tracks).
[veto_vars.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6171427/veto_vars.pdf)

I haven't found an easy way to count the number of noise hits per event and I don't know why the `nReadoutHits_` distribution looks to be identical. More detail to come later today after I've had more coffee. :coffee:

tomeichlersmith:
Disregard the last message, the reason they look so similar is because I wasn't running a different digi step :+1: 

When I **actually** change the digitization, we get the following [veto variables](https://github.com/LDMX-Software/ldmx-sw/files/6172222/veto_vars.pdf). Lots of interesting stuff here, but I think it shows the expected behavior.

jmmans:
It is important that the same threshold be applied to real hits and 
noise hits -- I hope this is enforced by the code.

tomeichlersmith:
The same threshold is applied :+1:

Issue #51, opened 2016-11-10
Enclose event constants in an EventConstants class.


JeremyMcCormick:
The compiler complains about these statically defined variables that are not enclosed in a class:

/u/ey/jeremym/ldmx-dev/ldmx-sw/Event/include/Event/EventConstants.h:10:20: warning: ‘event::SIM_PARTICLE’ defined but not used [-Wunused-variable]
 static const char* SIM_PARTICLE = "event::SimParticle";
                    ^

Try putting these inside an EventConstants class to get rid of this warning.

The constants would be referenced as `event::EventConstants::SIM_PARTICLE` or `EventConstants::SIM_PARTICLE` if `using event::EventConstants;` is included.

JeremyMcCormick:
Resolved by PR #54

Issue #40, opened 2016-11-05
Change TRefArray variables in SimParticle to non-pointer.


JeremyMcCormick:
In SimParticle, the TRefArray pointers can just be allocated on the stack.  This way they do not need to be explicitly allocated and deleted.

This would break compatibility with old data so if we make the change it should be done soon.

PR #973, opened 2021-03-16
Iss972 - Remove Framework dependency on EventConstants.h 

omar-moreno:
This PR is dependent on merging in [Framework:#40](https://github.com/LDMX-Software/Framework/pull/40) and  [SimCore:#51](https://github.com/LDMX-Software/SimCore/pull/51).

### DIFF
--- a/Recon/include/Recon/OverlayProducer.h
+++ b/Recon/include/Recon/OverlayProducer.h
@@ -61,6 +61,9 @@ class OverlayProducer : public framework::Producer {
   void onProcessStart() final override;
 
  private:
+  /// The parameters used to configure this producer
+  framework::config::Parameters params_;
+
   /**
    * Pileup overlay events input file name
    */
@@ -163,6 +166,6 @@ class OverlayProducer : public framework::Producer {
   int overlayTrackID_{-1000};
   int overlayPdgCode_{0};
 };
-}  // namespace recon
+} // namespace recon
 
 #endif /* RECON_OVERLAY_OVERLAYPRODUCER_H */
--- a/Recon/include/Recon/OverlayProducer.h
+++ b/Recon/include/Recon/OverlayProducer.h
@@ -34,6 +34,13 @@ class OverlayProducer : public framework::Producer {
   void configure(framework::config::Parameters &parameters) final override;
 
   /**
+   * At the start of the run, the pileup overlay file is set up, and the
+   * starting event number is chosen, using the RNSS. 
+   */
+  void onNewRun(const ldmx::RunHeader&); // );    //
+
+  
+  /**
    * Based on the list of collections to overlay, and the desired number of
    * events, loop through all relevant collections and copy the sim event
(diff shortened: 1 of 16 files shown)


Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -395,7 +398,7 @@ void OverlayProducer::onProcessStart() {
   }
 
   // replace by this line once the corresponding tweak to EventFile is ready:
-  overlayFile_ = std::make_unique<framework::EventFile>(overlayFileName_, true);
+  overlayFile_ = std::make_unique<framework::EventFile>(params_, overlayFileName_, true);
```
tomeichlersmith, line 401: Did you try running this? I'm 90% sure that this will fail because the `EventFile` constructor is looking for the parameters `compressionSetting` and `tree_name`, but the python config class for the overlay producer doesn't define those.

Recon/src/Recon/Event/EventConstants.cxx
```diff
@@ -2,11 +2,6 @@
 
 namespace ldmx {
 
-/*
- * Default name of event tree.
- */
-const std::string EventConstants::EVENT_TREE_NAME = "LDMX_Events";
```
tomeichlersmith, line 8: I think we should delete this whole file...

Recon/include/Recon/OverlayProducer.h
```diff
@@ -65,6 +65,9 @@ class OverlayProducer : public framework::Producer {
   void onProcessStart() final override;
 
  private:
+  /// The parameters used to configure this producer
+  framework::config::Parameters params_;
```
tomeichlersmith, line 69: I _really_ want to avoid persisting the parameters class. It is meant to be the translation layer between the Python and the C++, so it is inherently non-type-specific and thus confusing in the C++ ecosphere. My dream is to have **no** use of Parameters _after_ the python configuration is done being parsed. This would allow us to maintain separation between Python and C++. (There are some complicated SimCore aspects that prevent this currently.)

tomeichlersmith (commented):
Few things to discuss, but mostly about the overall approach.

Recon/src/Recon/OverlayProducer.cxx
omar-moreno: Good catch. I forgot this was done independently of ldmxcfg. I've updated this and seems to run now.

Recon/src/Recon/Event/EventConstants.cxx
omar-moreno: We should, but that should be done in a different PR.

Recon/include/Recon/OverlayProducer.h
omar-moreno: Although I agree in principle, that will require much larger changes that are outside of the scope of this PR.  We should discuss our approach and plan to make these changes towards the end of the year once most of the recon is done and we are able to take time for the software again.

Issue #974, opened 2021-03-17
Improved realism of out-of-time pileup simulation


bryngemark:
Currently, out-of-time pileup is simulated by choosing a total number of interactions, and then a per-event time offset is calculated by uniform sampling of an interval given by `(-nEarlier, nLater)*bunchSpacing`.

**Improved method**
Given an interval of bunches to populate, the number of pileup events to plug in should be set, either as a fixed number, or, by sampling a Poisson distribution around a fixed µ, for each bunch. So we'd calculate the pileup event time shift in an outer loop over bunches to populate, rather than inside a loop over a total number of pileup events. In-time pileup is the special case when the loop reaches the sim event bunch, still respecting the same total number of interactions as in the out-of-time bunches.

**Additional features considered**
Right now we're putting pileup in bunches ` bunchNumber=[-nEarlier,-nEarlier+1,…,0,…,nLater]` (with pileup event times shifted by `bunchSpacing*bunchNumber` ns).
I'm also considering adding the flexibility to randomly shift the start and end bunch of the out-of-time bunch interval with respect to the "sim event", like so:
- sample a uniform distribution (0, _nBunches_) for a position _b_ which is the sim event (time = 0) position within the interval
- the sim event times aren't actually shifted (just adds a layer of complication and processing time)
- for all bunches with` bunchNumber=[-b, -b+1, …, 0, …, nBunches-b]`, add pileup and shift times as before 

**Additional context**
These improvements would come in really handy for early-beam simulations in time for the early-data workshop.

tomeichlersmith:
Merged with PR #975 so I'm closing this. (@bryngemark Feel free to re-open if I'm missing something.)

PR #975, opened 2021-03-18
Improved realism of OOT pileup treatment in overlay 

bryngemark:
This does the first half of what I described in #974 : changing the order of how the bunch number and overlay event number is looped through. 

The second feature (randomization of the sim event position in the "bunch train") is not in yet. I can work on that tomorrow, but prefer to get the first desired feature into the upcoming release, if there is a strict deadline. I also don't know yet if this suggested feature has that many use cases.


tomeichlersmith:
Omar just merged your looping changes into `trunk`, could you rebase or merge so that the comparison is easier?

OM: Hey I can edit your comments odd ... anyway it's done. 
LK: So can I! :D Anyway, yes I was going to but thanks. (My GitHub notifications go to spam at the moment so I'm a little late to catch up here)

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -124,199 +100,250 @@ void OverlayProducer::produce(framework::Event &event) {
...
+  // get the calo hits collections that we want to overlay, by looping over
+  // the list of collections passed to the producer : caloCollections_
+  for (const auto &collName : caloCollections_) {
+    // for now, Ecal and only Ecal uses contribs instead of multiple
+    // SimHitsCalo per channel, meaning, it requires special treatment
+    auto needsContribsAdded{collName.find("Ecal") != std::string::npos ? true
+                                                                       : false};
+
+    // start out by just copying the sim hits, unaltered.
+    simHitsCalo =
```
omar-moreno: Is there a reason for declaring this variable outside of the loop? Looks like the scope is limited to the loop.

omar-moreno: It's probably fine to just remove the declaration outside of the loop and replace by `auto SImHitsCalo{}. `

omar-moreno: The same question for the rest of the hit collection declarations.

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -124,199 +100,250 @@ void OverlayProducer::produce(framework::Event &event) {
...
-                      << timeOffset << " ns";
-      ldmx_log(debug) << "bunch position offset in event " << iEv + 1 << " is  "
-                      << bunchOffset
-                      << ", leading to an additional time offset of "
-                      << bunchTimeOffset << " ns";
+      ldmx_log(debug) << "in loop: start of collection "
+                      << trackerCollections_[iColl];
+      ldmx_log(debug) << "in loop: printing current sim event: ";
+
+      for (const ldmx::SimTrackerHit &simHit : simHitsTracker) {
```
omar-moreno: If you are using this for loop just for printout, I would surround the whole thing with the `if (verbosity_ > 2)` condition. Otherwise, you are looping through the hits every event for no reason.

omar-moreno: Actually, the calorimeter and sim hit loops seem to be doing basically the same thing. I would pull this out into a function and just run all collections through it.  It will cut down on code repeat and improve readability.

bryngemark: Nice catch, I'll take a look

bryngemark: Yeah, we've discussed the sim tracker vs sim calo hits treatment before, and initially I wanted a catch-all solution for this (more elegant, less error prone, all of that) but we came to the conclusion (in a sw dev meeting I think?) that it was not trivial to make a single function for it given the different types.

bryngemark: There was already an outer verbosity check but I've removed the redundant inner one now.

Recon/src/Recon/OverlayProducer.cxx
bryngemark: Yup, with the shuffling of loops, some of this was left in the wrong place. Fixed it, thanks

bryngemark:
Alright, think I addressed all Omar had

Recon/src/Recon/OverlayProducer.cxx
omar-moreno: > Yeah, we've discussed the sim tracker vs sim calo hits treatment before, and initially I wanted a catch-all solution for this (more elegant, less error prone, all of that) but we came to the conclusion (in a sw dev meeting I think?) that it was not trivial to make a single function for it given the different types.

Yeah, this won't hold up the merge. Just wanted to note it.

omar-moreno (approved):

PR #980, opened 2021-04-07
Iss946 hgcroc - Enhancements to Chip Emulation

tomeichlersmith:
## Related Submodules

- https://github.com/LDMX-Software/Ecal/pull/30
- https://github.com/LDMX-Software/Hcal/pull/9

## Summary of Changes
- Update emulator to handle different SOI, TOT Mode, and ADC Mode for pulses at a variety of times
- Include new executable `run-hgcroc` to run the emulator alone and compare input pulse(s) to output digi samples
- Add custom iterator to `HgcrocDigiCollection` object so we can do range-based for loops
- Transition emulator to rely on a conditions table for chip parameters

### DIFF
--- a/Tools/src/Tools/HgcrocEmulator.cxx
+++ b/Tools/src/Tools/HgcrocEmulator.cxx
@@ -79,9 +79,7 @@ bool HgcrocEmulator::digitize(
   // TODO more physical way of simulating the timing jitter
   if (noise_) timeInWindow += noiseInjector_->Gaus(0., timingJitter_);
 
-  // set time in the window to zero if noise pushed it below zero
-  // TODO better (more physical) method for handling this case?
-  //if (timeInWindow < 0.) timeInWindow = 0.;
+  timeInWindow += 13.;
 
   // setup up pulse by changing the amplitude and timing parameters
   configurePulse(signalAmplitude, timeInWindow);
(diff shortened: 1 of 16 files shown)


Tools/src/Tools/run_hgcroc.cxx
```diff
@@ -0,0 +1,133 @@
+
```
omar-moreno, line 1: Is this suppose to be a test?

Tools/src/Tools/HgcrocEmulator.cxx
```diff
@@ -54,202 +44,183 @@ void HgcrocEmulator::seedGenerator(uint64_t seed) {
...
+  double drainRate = getCondition(channelID, "DRAIN_RATE");
+  double readoutThresholdFloat = this->readoutThreshold(channelID);
+  int readoutThreshold = int(readoutThresholdFloat);
 
-  // put noise onto timing
-  // TODO more physical way of simulating the timing jitter
-  if (noise_) timeInWindow += noiseInjector_->Gaus(0., timingJitter_);
+  // sort by amplitude
+  //  ==> makes sure that puleses are merged towards higher ones
+  std::sort(
```
omar-moreno, line 71: nice to see std::sort being used.

omar-moreno (approved):

Tools/src/Tools/run_hgcroc.cxx
tomeichlersmith: No. The translation from input voltage to output ADC or TOT is heavily dependent on how the emulator functions and what the input parameters are. For this reason, there isn't a simple test that can be done, but I still think this executable is useful to maintain so that future changes to the emulator can be validated more quickly by isolating the chip emulation from the rest of the digi pipeline.

omar-moreno: OK makes sense. I'm more wondering if this can be recast as a test?

omar-moreno: Will say, this is not a blocker but something I'm curious about.

tomeichlersmith: We _could_ recast it as a test, but it would only be testing very rudimentary things like the number of hits, so I opted to leave it out of the test framework so that it is easier to run standalone.

PR #981, opened 2021-04-19
Include headers that were previously not required

tomeichlersmith:
Since the [event bus re-design in Framework](https://github.com/LDMX-Software/Framework/pull/34) does not require including `EventDef` in for the `Event` class to function, I have found several places that are missing some headers of objects they use. These are not huge issues, but does bring the Framework PR out to the wider code-base.

The details for the event bus re-design are in the Framework PR: https://github.com/LDMX-Software/Framework/pull/34

I will attach the other submodule PRs that need to be patched.

### DIFF
--- a/.github/validation_samples/it_pileup/config.py
+++ b/.github/validation_samples/it_pileup/config.py
@@ -73,10 +73,14 @@ hcalReco   =hDigi.HcalRecProducer('hcalRecon')
 hcalReco.digiPassName = thisPassName
 
 from LDMX.DQM import dqm
+ecalDigiVerify = dqm.EcalDigiVerify()
+ecalDigiVerify.ecalSimHitColl = ecalDigiVerify.ecalSimHitColl+overlayStr
+
 p.sequence.extend([
     ecalDigi, ecalReco, ecalVeto,
     hcalDigi, hcalReco,
-    tsDigisUp, tsDigisTag, tsDigisDown]+dqm.ecal_dqm+dqm.recoil_dqm)
+    tsDigisUp, tsDigisTag, tsDigisDown,
+    ecalDigiVerify,dqm.EcalShowerFeatures()]+dqm.recoil_dqm)
 
 p.inputFiles = ['ecal_pn.root']
 p.outputFiles= ['events.root']
--- a/DQM/include/DQM/PhotoNuclearDQM.h
+++ b/DQM/include/DQM/PhotoNuclearDQM.h
@@ -7,6 +7,8 @@
 #include "Framework/Configure/Parameters.h"
 #include "Framework/EventProcessor.h"
 
+#include "SimCore/Event/SimParticle.h"
+
 namespace dqm {
 
 // Forward declarations within the ldmx workspace
--- a/DQM/src/DQM/EcalDigiVerifier.cxx
+++ b/DQM/src/DQM/EcalDigiVerifier.cxx
@@ -1,6 +1,9 @@
 
 #include "DQM/EcalDigiVerifier.h"
 
+#include "SimCore/Event/SimCalorimeterHit.h"
+#include "Ecal/Event/EcalHit.h"
+
 namespace dqm {
 
 void EcalDigiVerifier::configure(framework::config::Parameters &ps) {
--- a/DQM/src/DQM/HCalDQM.cxx
+++ b/DQM/src/DQM/HCalDQM.cxx
@@ -1,6 +1,9 @@
 
 #include "DQM/HCalDQM.h"
 
+#include "Hcal/Event/HcalHit.h"
+#include "Hcal/Event/HcalVetoResult.h"
+
 namespace dqm {
 
 HCalDQM::HCalDQM(const std::string& name, framework::Process& process)
--- a/DQM/src/DQM/RecoilTrackerDQM.cxx
+++ b/DQM/src/DQM/RecoilTrackerDQM.cxx
@@ -8,6 +8,8 @@
 
 #include "TVector3.h"
 
+#include "SimCore/Event/SimTrackerHit.h"
+
 namespace dqm {
 
 RecoilTrackerDQM::RecoilTrackerDQM(const std::string &name,
--- a/DQM/src/DQM/TrigScintDQM.cxx
+++ b/DQM/src/DQM/TrigScintDQM.cxx
@@ -1,6 +1,8 @@
 
 #include "DQM/TrigScintDQM.h"
 
+#include "SimCore/Event/SimCalorimeterHit.h"
+
 namespace dqm {
 
 TrigScintDQM::TrigScintDQM(const std::string &name, framework::Process &process)
--- a/EventDisplay/CMakeLists.txt
+++ b/EventDisplay/CMakeLists.txt
@@ -23,7 +23,7 @@ include("${ROOT_DIR}/RootMacros.cmake")
 
 # Register the EventDisplay class with the ROOT dictionary
 register_event_object(module_path "EventDisplay" namespace "eventdisplay"
-                      class "Display" type "dict_only")
+                      class "Display")
 
 build_event_bus(path ${PROJECT_SOURCE_DIR}/include/EventDisplay/EventDisplayDef.h)
 
--- a/EventDisplay/src/EventDisplay/Display.cxx
+++ b/EventDisplay/src/EventDisplay/Display.cxx
@@ -183,7 +183,9 @@ Display::Display(TEveManager* manager, bool verbose)
 bool Display::SetFile(const TString file) {
 
   try {
-    the_file_ = std::make_unique<framework::EventFile>(file.Data());
+    framework::config::Parameters file_config;
+    file_config.addParameter("tree_name",std::string("LDMX_Events"));
+    the_file_ = std::make_unique<framework::EventFile>(file_config,file.Data());
     the_file_->setupEvent(&the_event_);
     if (verbose_) {
       std::cout << "[ Display ] : Input root file opened successfully."
--- a/Recon/src/Recon/OverlayProducer.cxx
+++ b/Recon/src/Recon/OverlayProducer.cxx
@@ -1,6 +1,9 @@
 #include "Recon/OverlayProducer.h"
 #include "Framework/RandomNumberSeedService.h"
 
+#include "SimCore/Event/SimTrackerHit.h"
+#include "SimCore/Event/SimCalorimeterHit.h"
+
 namespace recon {
 
 void OverlayProducer::configure(framework::config::Parameters &parameters) {
@@ -396,7 +399,7 @@ void OverlayProducer::onProcessStart() {
     ldmx_log(debug) << "onProcessStart () successful. Used input file: "
                     << overlayFile_->getFileName();
     ldmx_log(debug) << "onProcessStart () successful. Got event info: ";
-    overlayFile_->getEvent()->Print(verbosity_);
+    overlayFile_->getEvent()->Print();
   }
 
   return;
--- a/scripts/ldmx-env.sh
+++ b/scripts/ldmx-env.sh
@@ -366,9 +366,9 @@ _ldmx_clean() {
   if [[ "$_what" = "src" ]] || [[ "$_what" = "all" ]]; then
     local _old_pwd=$OLDPWD
     cd ${LDMX_BASE}/ldmx-sw
-    git clean -xi
+    git clean -xf
     rc=$?
-    git submodule foreach git clean -xi
+    git submodule foreach git clean -xf
     rc=$?
     [[ -d build ]] && rm -r build
     [[ -d install ]] && rm -r install


tomeichlersmith:
## Validation To-Do List
- [x] Generating data with the new design re-produces what we have now
- [x] Overlays of all existing DQM plots with existing data to make sure nothing has changed 
- [x] Running recon on existing files to make sure we remain backwards compatible.

tomeichlersmith:
## Related Sub-Module PRs
Several very small changes are required to basically include headers and cleanup after the Framework changes.

- https://github.com/LDMX-Software/cmake/pull/12
- https://github.com/LDMX-Software/SimCore/pull/56
- https://github.com/LDMX-Software/Hcal/pull/11
- https://github.com/LDMX-Software/Ecal/pull/32

tomeichlersmith:
I have validated that these changes have no effect on any physical distributions. In my explanation of this below, I call `ldmx:pro:edge` (the most recent commit to `trunk`) `trunk` and I call this branch with these changes on it `dev`. 

I ran the same config `sim.py` using `trunk` and `dev` generating the two files `trunk_sim.root` and `dev_sim.root`. I then ran the same config `reco.py` four times; using both `trunk` and `dev` to run recon on both sim files. I then compared these four final files to each other using the python compare script copied below. Every single one of these hundreds of distributions lie perfectly on-top of each other. All of the event files that I compared, the scripts I used to generate them, the comparison script, and all the comparison plots are in the attached archive.

[bus_redesign_validation.tar.gz](https://github.com/LDMX-Software/ldmx-sw/files/6495265/bus_redesign_validation.tar.gz)

tomeichlersmith:
@bryngemark I am seeing some differences within the TrigScint reconstruction pipleine on this PR branch.

(The IT-Pileup validation did not run because of another bug that I patched.)

The inclusive sample had 5 plots that did not match:

- [TrigScintTracks_TrigScintTracks_n_tracks.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6610231/TrigScintTracks_TrigScintTracks_n_tracks.pdf)
- [TrigScintClusterUp_TrigScintClusterUp_total_pe.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6610232/TrigScintClusterUp_TrigScintClusterUp_total_pe.pdf)
- [TrigScintClusterUp_TrigScintClusterUp_n_clusters.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6610233/TrigScintClusterUp_TrigScintClusterUp_n_clusters.pdf)
- [TrigScintClusterDn_TrigScintClusterDn_total_pe.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6610234/TrigScintClusterDn_TrigScintClusterDn_total_pe.pdf)
- [TrigScintClusterDn_TrigScintClusterDn_n_clusters.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6610235/TrigScintClusterDn_TrigScintClusterDn_n_clusters.pdf)

And the ecal pn sample had one:

- [TrigScintTracks_TrigScintTracks_n_tracks.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6610251/TrigScintTracks_TrigScintTracks_n_tracks.pdf)

It looks like the histograms have extra entries in the zero bin? The other bins look (clost to) identical, but the zero-bin for these changes are off.

The changes made in this PR are low-level Framework updates that improve the design of our event bus buffer. One bug that I caught during these changes that could be causing this discrepancy is partial collection name matching. This can be explained by an example (the bug that I patched earlier).

Before, if you requested `EcalSimHits` with no pass name, you would get `EcalSimHits` **even if** the collection `EcalSimHitsOverlay` already existed. Now, a "product ambiguous" error is thrown. I patched this bug to revert back to the old method, but I also wanted to hear if you have any insight before I run the validation again.

_Edit_:
I think that this code could be the origin.
```cpp
// ClusterProducer
  if (trigScintClusters.size() > 0)
    event.add(output_collection_, trigScintClusters);

// TrackProducer
    // done here if there were no tracks found
    if (tracks_.size() == 0) {
      if (verbose_) {
        ldmx_log(debug) << "No tracks found!";
      }
      return;
    }
```
In the old buffer design, I am not fully confident that each bus passenger was being fully cleared between events. In this new design, I am very confident that the collections are begin emptied after each event and so the default value is zero for `n_clusters`. This doesn't totally explain things (e.g. the difference in `total_pe`), but it would explain the difference in `n_tracks` and `n_clusters`.

bryngemark:
> @bryngemark I am seeing some differences within the TrigScint reconstruction pipleine on this PR branch.
> It looks like the histograms have extra entries in the zero bin? The other bins look (clost to) identical, but the zero-bin for these changes are off.

I think these plots make the case very clearly that the binning in these DQM plots is off 😄  would it be possible to remake both ref and new plots with a more reasonable binning, if I make a PR for it?

> 
> Before, if you requested `EcalSimHits` with no pass name, you would get `EcalSimHits` **even if** the collection `EcalSimHitsOverlay` already existed. Now, a "product ambiguous" error is thrown. I patched this bug to revert back to the old method, but I also wanted to hear if you have any insight before I run the validation again.

I'm not sure what you refer to here. I don't think we ever anticipate treating these two as anything but distinct collections. Meaning, if you want the overlay collection, you'd have to specify the name with `Overlay` in it, and also not assume that an existing `EcalSimHitsOverlay` collection in the current pass should be chosen by any processor requesting the `EcalSimHits` one. As for pass name clarity though, I'm all for it: no processor should have hardcoded pass ambiguity. 

> 
> _Edit_:
> I think that this code could be the origin.
> 
> ```c++
> // ClusterProducer
>   if (trigScintClusters.size() > 0)
>     event.add(output_collection_, trigScintClusters);
> 
> // TrackProducer
>     // done here if there were no tracks found
>     if (tracks_.size() == 0) {
>       if (verbose_) {
>         ldmx_log(debug) << "No tracks found!";
>       }
>       return;
>     }
> ```
> 
> In the old buffer design, I am not fully confident that each bus passenger was being fully cleared between events. In this new design, I am very confident that the collections are begin emptied after each event and so the default value is zero for `n_clusters`. This doesn't totally explain things (e.g. the difference in `total_pe`), but it would explain the difference in `n_tracks` and `n_clusters`.

This is great, I have been meaning to check a few edge cases where I've suspected that there is some residue of previous events (for instance, sometimes the distance between clusters forming a track has been seen to be larger than the max distance allowed in the algorithm). 
To help me understand if this is the cause of what you're seeing: can I track **all** the differences between reference and current PR histograms by checking code changes listed in this PR? (The reason I ask is that there have also been semi-recent changes to the `TrigScint` clustering code that lead to getting fewer clusters overall. This is the missing synchronization between `TrigScint` and `ldmx-sw` Omar mentioned in the sw-dev meeting Thursday, and I don't know if that has been fixed yet. So just want to make sure I know exactly what your reference point is.)

tomeichlersmith:
- You can change the binning and we can re-generate the reference histograms by making a new release. (stable, not pre-).
- I figured we should keep those distinct, so I patched my changes to Framework so that those are always distinct.
- The ref histograms were generated with the latest tag, so [this comparison](https://github.com/LDMX-Software/ldmx-sw/compare/v3.0.0-gamma...Framework-iss9) shows the commits between this branch and that tag. I don't think I included the patches to TrigScint you mention.

bryngemark:
Thanks @tomeichlersmith, for confirming. Most of this looks reasonable to me but I'd like to check with some more sensible DQM plots.

I just updated the DQM plot binning: PR [here](https://github.com/LDMX-Software/ldmx-sw/pull/994)

tomeichlersmith:
@bryngemark Including the patches I made to Framework did not resolve the discrepancies in the TrigScint module.

- Inclusive
  - [TrigScintTracks_TrigScintTracks_n_tracks.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6617559/TrigScintTracks_TrigScintTracks_n_tracks.pdf)
  - [TrigScintClusterUp_TrigScintClusterUp_total_pe.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6617560/TrigScintClusterUp_TrigScintClusterUp_total_pe.pdf)
  - [TrigScintClusterUp_TrigScintClusterUp_n_clusters.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6617561/TrigScintClusterUp_TrigScintClusterUp_n_clusters.pdf)
  - [TrigScintClusterDn_TrigScintClusterDn_total_pe.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6617562/TrigScintClusterDn_TrigScintClusterDn_total_pe.pdf)
  - [TrigScintClusterDn_TrigScintClusterDn_n_clusters.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6617563/TrigScintClusterDn_TrigScintClusterDn_n_clusters.pdf)
- ECal PN
  - [TrigScintTracks_TrigScintTracks_n_tracks.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6617571/TrigScintTracks_TrigScintTracks_n_tracks.pdf)

Moreover, the random segfault when running the ecal veto processor with the overlay producer has once again [reared its ugly head](https://github.com/LDMX-Software/ldmx-sw/runs/2769640301?check_suite_focus=true). I am investigating.

**New Seg Fault**:
```
   Generating stack trace...
   0x00007fec2e79cfae in TBranchElement::ReadLeavesCollection(TBuffer&) + 0x18e from /deps/cernroot/lib/libTree.so
   0x00007fec2e7901b5 in TBranch::GetEntry(long long, int) + 0xf5 from /deps/cernroot/lib/libTree.so
   0x00007fec2e7a1dbf in TBranchElement::GetEntry(long long, int) + 0x31f from /deps/cernroot/lib/libTree.so
   0x00007fec2e808728 in TTree::GetEntry(long long, int) + 0xc8 from /deps/cernroot/lib/libTree.so
   0x00007fec3185e69e in framework::EventFile::nextEvent(bool) + 0x10e from /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libFramework.so
   0x00007fec2898d5ae in recon::OverlayProducer::produce(framework::Event&) + 0x166e from /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libRecon.so
   0x00007fec31877874 in framework::Process::process(int, framework::Event&) const + 0x64 from /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libFramework.so
   0x00007fec318781a5 in framework::Process::run() + 0x6b5 from /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libFramework.so
   0x000055aaf18dec5d in main + 0x1fd from fire
   0x00007fec300e6bf7 in __libc_start_main + 0xe7 from /lib/x86_64-linux-gnu/libc.so.6
   0x000055aaf18df5ea in <unknown> from fire
```
**Old Seg Faul**:
```
 Generating stack trace...
 0x00007f4cf76b2cf0 in framework::Process::run() + 0x600 from /nfs/slac/g/ldmx/users/lene/ldmx-sw-sandbox/ldmx-sw/install/lib/libFramework.so
 0x000056020670eccd in main + 0x1fd from fire
 0x00007f4cf5e1ab97 in __libc_start_main + 0xe7 from /lib/x86_64-linux-gnu/libc.so.6
 0x000056020670f65a in <unknown> from fire
```

_Edit_: This seg-fault is different than the old one. In the old buffer, we had ROOT's TTrees handle the input objects, so resetting the event bus was enough. Now, since our buffer is handling all objects, we need to reset the TTree's addresses so they don't try to access objects that aren't there. This patch has been made and checked locally.

tomeichlersmith:
@bryngemark Your hunch about some difference between my TrigScint ref and TrigScint trunk is correct. The TrigScint commit pointed to by this branch of ldmx-sw is [substationally different from TrigScint Trunk](https://github.com/LDMX-Software/TrigScint/compare/8662abfeac25509b39e9ab604e14f686e852dac5..trunk) but this branch has the same commit for TrigScint as ldmx-sw trunk ([8662abfeac25509b39e9ab604e14f686e852dac5](https://github.com/LDMX-Software/TrigScint/tree/8662abfeac25509b39e9ab604e14f686e852dac5)). :thinking: This is getting weirder by the day...

_Edit_: Re-reading your comment above made me think that my initial hypothesis that event objects aren't getting cleared is correct. This comment shows that the TS submodule _does not_ include the updates (on either ldmx-sw/trunk or this branch). This hypothesis is supported by the fact that `TrigScintClusterProducer` and `TrigScintTrackProducer` are the _only_ producers that conditionally add objects to the event bus. (partial proof in [event_add.log](https://github.com/LDMX-Software/ldmx-sw/files/6619473/event_add.log)).

tomeichlersmith:
@omar-moreno @bryngemark 
TLDR: The changes to Framework patch a bug that is affecting the TrigScint DQM histograms. 
i.e. The discrepancy between the histograms on this branch and the reference ones are real and this branch has the correct version.

## Hypothesis Test

I want to check that `trunk` is not clearing objects the way we expect, so I went over to `trunk` and added some printouts so I could get the event numbers.

```bash
diff --git a/src/TrigScint/TrigScintClusterProducer.cxx b/src/TrigScint/TrigScintClusterProducer.cxx
index 9ac5696..8af5933 100644
--- a/src/TrigScint/TrigScintClusterProducer.cxx
+++ b/src/TrigScint/TrigScintClusterProducer.cxx
@@ -436,6 +436,8 @@ void TrigScintClusterProducer::produce(framework::Event &event) {
 
   if (trigScintClusters.size() > 0)
     event.add(output_collection_, trigScintClusters);
+  else
+    std::cout << "No clusters (" << getName() << ") " << event.getEventNumber() << std::endl;
 
   hitChannelMap_.clear();
   v_usedIndices_.resize(
diff --git a/src/TrigScint/TrigScintTrackProducer.cxx b/src/TrigScint/TrigScintTrackProducer.cxx
index fc0d19d..f189ff3 100644
--- a/src/TrigScint/TrigScintTrackProducer.cxx
+++ b/src/TrigScint/TrigScintTrackProducer.cxx
@@ -196,6 +196,7 @@ void TrigScintTrackProducer::produce(framework::Event &event) {

       if (verbose_) {
         ldmx_log(debug) << "No tracks found!";
       }
+      std::cout << "No tracks " << event.getEventNumber() << std::endl;
       return;
     }
     // now, if there are multiple seeds sharing the same downstream hits, this
```

Compiling and running `trunk` on the inclusive sample (but with only 100 events) gives
(The `grep` is there to filter out the other printouts that I don't care about.)

```bash
$ ldmx fire config.py | grep ^No
No tracks 21
No tracks 28
No tracks 41
No clusters (trigScintClustersUp) 50
No clusters (trigScintClustersDown) 50
No tracks 54
No clusters (trigScintClustersUp) 74
No clusters (trigScintClustersDown) 74
No tracks 80
No tracks 85
```

Now I can run a quick python script to look at the cluster and track objects and see if they are zero'd out for the event numbers where they weren't added.
(The `check.py` script is copied below.)

```bash
$ ldmx python3 check.py
21 tracks not cleared!
28 tracks not cleared!
41 tracks not cleared!
50 up clusters not cleared!
50 dn clusters not cleared!
54 tracks not cleared!
74 up clusters not cleared!
74 dn clusters not cleared!
80 tracks not cleared!
85 tracks not cleared!
```

**This shows that trunk is not clearing the objects that are written to the output file.**

Recompiling ldmx-sw on this branch (and keeping the TS changings printing out the event numbers) shows that this branch is clearing the objects correctly.

```bash
$ ldmx fire config.py | grep ^No
No tracks 21
No tracks 28
No tracks 41
No clusters (trigScintClustersUp) 50
No clusters (trigScintClustersDown) 50
No tracks 54
No clusters (trigScintClustersUp) 74
No clusters (trigScintClustersDown) 74
No tracks 80
No tracks 85
$ ldmx python3 check.py
$ # no printout because the collections are being cleared
```

## check.py
```python
import ROOT, sys
ROOT.gSystem.Load('libFramework.so')

rf = ROOT.TFile(sys.argv[1])
t = rf.Get('LDMX_Events')

up_clusters_not_added = [
        50,74
        ]

dn_clusters_not_added = [
        50,74
        ]

tag_clusters_not_added = [
        ]

tracks_not_added = [
        21,28,41,54,80,85
        ]

for e in t :
    eid = e.EventHeader.getEventNumber()
    
    if (eid in up_clusters_not_added
        and e.TriggerPadUpClusters_test.size() > 0):
            print(f'{eid} up clusters not cleared!')
    
    if (eid in dn_clusters_not_added
        and e.TriggerPadDownClusters_test.size() > 0):
            print(f'{eid} dn clusters not cleared!')

    if (eid in tag_clusters_not_added
        and e.TriggerPadTaggerClusters_test.size() > 0):
            print(f'{eid} tag clusters not cleared!')

    if (eid in tracks_not_added
        and e.TriggerPadTracks_test.size() > 0):
            print(f'{eid} tracks not cleared!')
```

bryngemark:
I see I never posted this, written a few hours ago (it's been a messy day):

> Hmm but wait, didn't we expect the clearing of collections to result in more empty cluster collections, and thus tracks? 
> Tracking requires a cluster in each trigger pad. Since it's enough that one of them is empty to not get a track, this effect 
> should be more pronounced for tracks than individual clusters, which is what we see. 

-- but, now I'm a little confused by your check. Are we still seeing tracks in events with no clusters? OR maybe this is taken care of already earlier, by an early exit from the tracking code if cluster collections are empty.

bryngemark:
To be clear I think your patch is good – clearing objects is the way to go – I'm just a little confused and want to make sure I understand this before I sign off 🙂

tomeichlersmith:
It's just an issue with where I put the printout statement in the track producer. In my python script, I can check how many tracks there are when there are no clusters and I don't see any.
```
50: No up/dn clusters and 0 tracks
74: No up/dn clusters and 0 tracks
```
The printout that no tracks are found is only inside the `if` statement checking if there are clusters in all the trigger pads, so I am not printing out the cases where there are no tracks because there are no clusters. Funnily enough, since the `event.add` call is _outside_ the `if` statement about clusters, these collections of tracks _are_ cleared becuase you `add` and empty track collection to the event bus (forcing the clearing of the collection).

tomeichlersmith:
### Just for Future Reference

I am merging this in despite the failure of the validation because of two reasons. 

1. The inclusive and ecal_pn samples fail because of the bug _on trunk_ that affects the TS reference histograms as described above.
2. The ECal DQM EcalDigiVerfiy uses the sim hit collection as input. The reference histograms on trunk for the it_pileup sample uses the regular sim hits and not the overlay sim hits. This isn't what we want.

After merging this in, I will create another release so we can re-generate the reference histograms after these patches.

PR #989, opened 2021-05-07
Keep Decay Products and Clean-Up SimParticle Handling

tomeichlersmith:

This resolves an issue in SimCore and adds another filter for keeping children of decays of specific particle types. This type of filter was requested by people looking at kaon decays, so that is what I looked at.

## Check List
- [x] I ran my developments. Below I have copied in the config I used to generate 11 target-kaon events and the python script I used to validate that the kaons and their children were stored in the output file.

### Related Sub-Module PRs
- https://github.com/LDMX-Software/SimCore/pull/59

## Config Script
Run Like: `ldmx fire conf.py`
```python
p=ldmxcfg.Process("sim")
p.run = 1
p.maxEvents = 1000*10

from LDMX.Biasing import target
from LDMX.Biasing import particle_filter
from LDMX.SimCore import bias_operators
from LDMX.Biasing import include as includeBiasing
from LDMX.Biasing import filters
from LDMX.Biasing import util
from LDMX.SimCore import examples

from LDMX.Ecal import digi as ecal_digi
from LDMX.Hcal import digi as hcal_digi

kaon_filter = particle_filter.PhotoNuclearProductsFilter("kaons_filter")
kaon_filter.pdg_ids = [130, # K_L^0
                       310, # K_S^0
                       311, # K^-
                       321  # K^+
]

sim = simulator.simulator("target_photonNuclear")
sim.setDetector( 'ldmx-det-v12' , True )
sim.runNumber = 0
sim.description = "ECal photo-nuclear, xsec bias 1.1e9"
sim.beamSpotSmear = [20., 80., 0.]
sim.generators.append(generators.single_4gev_e_upstream_tagger())

sim.biasing_operators = [ bias_operators.PhotoNuclear('target_region',1.1e9,2500.,only_children_of_primary=True) ]
includeBiasing.library()
sim.actions.extend([
     filters.TaggerVetoFilter(),
     filters.TargetBremFilter(),
     filters.TargetPNFilter(),
     kaon_filter,
     util.TrackProcessFilter.photo_nuclear(),
     util.DecayChildrenKeeper(kaon_filter.pdg_ids+[-1*p for p in kaon_filter.pdg_ids])
    ])
p.sequence=[ sim,
             ecal_digi.EcalDigiProducer(),
             ecal_digi.EcalRecProducer(),
             hcal_digi.HcalDigiProducer(),
             hcal_digi.HcalRecProducer(),
             ]
p.outputFiles = [ 'pn_kaonfilter_1M_events_r%04d.root'%(p.run) ]
```

## Validation Script
Ran Like: `ldmx python3 check_kaons.py pn_kaonfilter_1M_events_r0001.root`
```python
import ROOT, sys
ROOT.gSystem.Load('libFramework.so')

rf = ROOT.TFile(sys.argv[1])
t = rf.Get('LDMX_Events')

for e in t :
    print(e.EventHeader.getEventNumber())
    print(' ',e.SimParticles_sim.size())
    for id_particle in e.SimParticles_sim :
        pdg = id_particle.second.getPdgID()
        if abs(pdg) in [130,310,311,321] :
            print(f'  Found Kaon: {pdg} {id_particle.first}')

            for child in id_particle.second.getDaughters() :
                if e.SimParticles_sim.find(child) == e.SimParticles_sim.end() :
                    print(f'    Kaon child {child} not saved.')
                else :
                    print(f'    Kaon child {child} saved.')
```

### DIFF
--- a/Biasing/include/Biasing/Utility/DecayChildrenKeeper.h
+++ b/Biasing/include/Biasing/Utility/DecayChildrenKeeper.h
@@ -17,6 +17,12 @@ namespace utility {
 /**
  * Filter to keep tracks that are the products of certain
  * particle's decays.
+ *
+ * @note It is important to emphasize that this filter
+ * is only able to persist products of particles that
+ * are themselves kept. i.e. Another action needs to
+ * keep the parent particle and then this action will
+ * save that particle's children.
  */
 class DecayChildrenKeeper : public simcore::UserAction {
  public:
@@ -44,7 +50,7 @@ class DecayChildrenKeeper : public simcore::UserAction {
    * particle that has finished processing (and is therefore
    * in the stored particle map), but I focus here on "decays"
    * because that is the only one where the parent is
-   * _definitely_ kept.
+   * _definitely_ at the end of processing.
    *
    * @param[in] track Geant4 track associated with a particle.
    */
--- a/Biasing/include/Biasing/Utility/DecayChildrenKeeper.h
+++ b/Biasing/include/Biasing/Utility/DecayChildrenKeeper.h
@@ -1,0 +1,67 @@
+#ifndef BIASING_UTILITY_DECAYCHILDRENKEEPER_H
+#define BIASING_UTILITY_DECAYCHILDRENKEEPER_H
+
+/*~~~~~~~~~~~~~~~~*/
+/*   C++ StdLib   */
+/*~~~~~~~~~~~~~~~~*/
+#include <string>
+
+/*~~~~~~~~~~~~~*/
+/*   SimCore   */
+/*~~~~~~~~~~~~~*/
+#include "SimCore/UserAction.h"
+
+namespace biasing {
+namespace utility {
+
+/**
+ * Filter to keep tracks that are the products of certain
+ * particle's decays.
+ */
+class DecayChildrenKeeper : public simcore::UserAction {
+ public:
+  /**
+   * Constructor.
+   *
+   * @param[in] name the name of the instance of this UserAction.
+   * @param[in] parameters the parameters used to configure this
+   *      UserAction.
+   */
+  DecayChildrenKeeper(const std::string& name, framework::config::Parameters& parameters);
+
+  /// Destructor
+  ~DecayChildrenKeeper();
+
+  /**
+   * Method called when a track is done being processed.
+   *
+   * We get the stored particle map from the TrackMap object
+   * and look for the current track's parent in it. If the
+   * current track's parent is in it THEN we check if that
+   * parent has a PDG matching ANY of the configured PDGs.
+   *
+   * @note Techinically, we can check for children of any
+   * particle that has finished processing (and is therefore
+   * in the stored particle map), but I focus here on "decays"
+   * because that is the only one where the parent is
+   * _definitely_ kept.
+   *
+   * @param[in] track Geant4 track associated with a particle.
+   */
+  void PostUserTrackingAction(const G4Track* track) final override;
+
+  /// Retrieve the type of actions this class defines.
+  std::vector<simcore::TYPE> getTypes() final override {
+    return {simcore::TYPE::TRACKING};
+  }
+
+ private:
+  /// The PDG IDs for which to keep decay children
+  std::vector<int> parents_;
+
+};  // DecayChildrenKeeper
+
+}  // namespace utility
+}  // namespace biasing
+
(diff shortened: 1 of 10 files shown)


tomeichlersmith:
@omar-moreno The reason the `DecayChildrenKeeper` filter is needed is because sometimes the kaon decays within the `MagnetRegion` which has it's store-secondaries flag turned off. When I remove the filter and re-run, I see that the events where the kaon's children are not saved are the same events where the kaon decays inside of the MagnetRegion. I only looked at this for 10 total kaon events (about 5 of which had the kaon decaying in the MagnetRegion), but I think this makes sense.

omar-moreno:
Great! Thanks for looking into this.

tomeichlersmith:
@omar-moreno This PR passed the recon validation tests that I just merged in. You can see the DQM plots that were compared in the "artifacts" for this action: https://github.com/LDMX-Software/ldmx-sw/actions/runs/912978926

tomeichlersmith:
I have validated that this branch is ready to be merged into trunk. I did this by running all the configs in `Biasing/test` and comparing their event outputs between `ldmx/pro:edge` and a build of this branch. There are discrepancies between the branches that I found using the attached archive containing the configs I ran, the generated event files, the comparison script I used, and the plots it outputs. These discrepancies can be explained as follows.

[iss57-particle-handling-validation.tar.gz](https://github.com/LDMX-Software/ldmx-sw/files/6634286/iss57-particle-handling-validation.tar.gz)

### Vertex Kinematics (px, py, pz, time)
This update changes the time at which during the simulation we store the vertex kinematics of the track. Before, it was at the [creation of a track's `Trajectory`](https://github.com/LDMX-Software/SimCore/blob/d52d0bd5d57f03a7acf30baf3a00aa3a049686e7/src/SimCore/Trajectory.cxx#L48-L50) (and now it is during the track's first pass through `PreUserTrackingAction`). The problem with the old method is that some tracks (especially when we are using filters and using the boolean flags stored in the track info) have their trajectory created _at the end of processing_ instead of at the beginning[0]. This means their initial kinematics would be read out at the endpoint instead of the actual value. You can see this in the plots of momentum where this branch (labeled dev) takes entries from the zero bin of these distributions and spreads them out elsewhere and in the plots of `time` where this branch takes entries at later times and puts them at earlier times.

[0] This can be seen in [UserTrackingAction](https://github.com/LDMX-Software/SimCore/blob/d52d0bd5d57f03a7acf30baf3a00aa3a049686e7/src/SimCore/UserTrackingAction.cxx#L108-L112) on SimCore/trunk. `storeTrajectory` creates the Trajectory for a G4Track and sometimes it is called in PostUserTrackingAction instead of Pre.

### List of Daughter Track IDs
The old design only kept track of generational relationships between particles _that will be saved in the output event_ [1]. This is limiting in terms of understanding results and is misleading. The new design keeps track of _all_ of the descendants[2] of _each_ particle and then copies that list into the output object at the end of the event. This results in this branch having many more entries at a variety of positions when compared to trunk in the `daughters` histograms.

[1] This can be seen in [SimParticleBuilder](https://github.com/LDMX-Software/SimCore/blob/d52d0bd5d57f03a7acf30baf3a00aa3a049686e7/src/SimCore/Persist/SimParticleBuilder.cxx#L83-L88) in the loop over Trajectories (objects that are being translated to the output SimParticle objects), we add a given track ID as a child to another particle if (a) that track ID comes as one of the trajectories being persisted and (b) the parent particle is also in the output map.

[2] _edit_: By "descendants", I mean direct descendants. We don't keep track of children of children, only direct children.

### Incident ID Contribs
This is a more cosmetic change. The old design's default return value if both the current track and its parent were outside the CalorimeterRegion was to return the parent's track ID. This branch returns the current track ID.[2] Personally, I think this makes more sense, especially for the systems that aren't even in the calorimeter region, but :shrug: . This change leads to the discrepancies in the incidentIDContribs_ histograms.

[2] You can see this by comparing `findIncident` in [TrackMap](https://github.com/LDMX-Software/SimCore/compare/d52d0bd5d57f03a7acf30baf3a00aa3a049686e7...58a4f2d3fd3078348c415e6458627936637ce3b0#diff-a8a80ced8633dec8ba5c7720d95acbaaf9709568d5bfe4097f7eaac3692068cd) that function is what determines the track ID of the particle "incident" on the calorimeter region which is the most recent ancestor of the particle contributing to
 a calorimeter hit.

PR #990, opened 2021-05-14
Basic Physics Validation Action

### LINKED ISSUE #988: DQM GitHub Action
Auto-generate DQM plots for a small number of standard config files. This will just make our job of validating developments easier.

tomeichlersmith:

This resolves #988 by creating a new GitHub action that is manually triggered.

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments. No changes made to ldmx-sw.
You can see the runs of the actions on [the action page](https://github.com/LDMX-Software/ldmx-sw/actions/workflows/physics_validation.yml)
- [x] ~~I attached any sub-module related changes to this PR.~~ NA

## Explanation of Workflow
Since I expect these physics validations to take a long time to run _and_ produce lots of plots to sift through, I have set up the workflow to be triggered manually via the action page. There are pros and cons to this approach.

**Pros**:
- Only produce physics validation plots when someone is interested in looking at them (i.e. don't waste GitHub action running time).
- User can **choose** which production image to compare their branch against.

**Cons**:
- Doesn't run automatically (i.e. there is some learning curve to make the validation plots).
- The production image and the branch being compared _need_ to have the same configuration-script interface. (I could change how the action works to address this issue, but I am choosing to not do that unless there is interest.)

After being told which production image to compare against (the "base"), the action is relatively simple. We compile the branch that we launched from (the "dev" branch) [1] and then generate events using the same configuration script(s) using the currently compiled branch and the production image provided. Then there is [a comparison script](https://github.com/LDMX-Software/ldmx-sw/blob/iss988-validation-action/.github/workflows/compare.py) to plot several different outputs. All of the event generation and comparison is housed in its [own bash script](https://github.com/LDMX-Software/ldmx-sw/blob/iss988-validation-action/.github/workflows/physics_validation.sh) so that I can easily test it locally. The compiling isn't included in this bash script because I don't want to re-compile when testing this script locally.

[1] I am planning to change it so that the user can choose which branch to compile so that they don't have to have this workflow on their branch.

## Planned Improvements before Merge
- [x] Allow the "dev" branch to be an input for the action
- [x] Fold in more processors into processing chain
- [x] Include other simulation-scenarios (e.g. ECal PN) @omar-moreno Which scenarios should be prioritized?
- [x] Use DQM plots as well as plots of the raw event values. @vdutta @omar-moreno @bryngemark Can you point me to your workflows for generating validation plots? I know I will miss some if I try to do it on my own. Currently, I am just plotting each of event bus elements using the TTree::Draw method.
- [x] Attempt to split up the workflow into multiple jobs so that different simulation-scenarios can be run in parallel. This might be too big of task and would need to be its own issue moving forward. We would only need it if we are starting to push up against the 6hr job time limit.
- [x] Better [configuration of action trigger](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_review) to allow the action to run on some subset of new pushes to PRs.
- [x] Use KS test to filter out plots that contain discrepancies between the branch and the "golden" tag.


bryngemark:
hi @tomeichlersmith, thanks for getting this started! it's been in the back of my mind for quite some time to streamline the validation of code changes, by way of example from TS. i'd be curious to hear if your idea and mine could be combined to something better than our individual thinking. 

what i had thought of doing was using the DQM module. for TS, these plots are a set of basic quantities that are expected to give the broad picture of the output from our increasingly high-level algo output objects (hits->digis->clusters->tracks). the advantage of this is that these histograms (their ranges and binning for instance) are well-defined, not terribly many, and easy to compare. (correspondingly for Ecal, things like a few hit quantities but also BDT variables would be natural candidates.) using tree->Draw() histograms is generic enough but can get funky (the most annoying part being that ratios require identical binning which requires identical range).

furthermore, what i had been thinking is to ship a reference in some form; either a .root file (under version control) with reference histograms that provide the latest gold standard, or as you suggest to provide a reference tag and let an action create them on the fly. once the reference and test version of histograms exist, i would make plots with them overlaid _and_ run a KS test or similar to automatically flag discrepancies. this approach is probably an atlas data quality habit. 

i could and would like to (eventually?) see this integrated into the PR tests, because, it isn't a huge number of plots to check (only the flagged ones out of a well-defined DQM set absolutely need checking; one could also set test passing thresholds like `identical` (where e.g. an Ecal change shouldn't change TS tagger pad distributions), [some defined tolerable agreement value], and `passThrough` where we expect things to dramatically change), and i think it would help us spot problems early and habitually. if there's a way to check off/pass a test that fails in the automatic part, after human inspection of plots, that would be very neat i think. i also see this as something that would naturally evolve into real data quality monitoring (where timely and efficient warnings are crucial), done on a run-by-run basis, comparing to some set reference run.

tomeichlersmith:
Thank you for your detailed thoughts.

## Create on Fly or Store File
I do think we will be forced to "create them on the fly" because I can't think of a method that would allow us to regularly download the pre-generated '.root' file(s). Just for documentation purposes, here are the options I've considered.

- Keep the files in the repo. This gives us a _very_ hard upper limit of 100MB per file. This strongly limits the number of events we could include in a test.
- What about [Large File Storage (LFS)](https://git-lfs.github.com/)? Well, [without paying GitHub](https://docs.github.com/en/github/managing-large-files/about-storage-and-bandwidth-usage), we are only allowed 1GB of "bandwidth" (read this as downloads) _per month_. This is another upper limit that is very easy for us to reach.

The reason I went with the design in this branch is because the only "loose" requirement that GitHub has for free users is (perhaps understandably) workflow run time. [Even on free accounts](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration), _each job_ can run for up to 6 hours and _an entire workflow_ is limited to 20 parallel jobs and a total run time of 72 hours. This gives us roughly 1440 hours of computer time which is plenty for us currently.

## Plots and Testing
I'm happy to hear that the TS subsystem has used the DQM module so much. Other subsystems haven't (including my own), so for ease of development, I just ignored trying to get the DQM module working at all. I will move that up my priority list.

I like your idea of a KS test and, I will integrate that into the testing framework.

## PR Integration
The issue with having the action run automatically on a PR is that by default the action will run _on every push_ to the branch that opened the PR. For shorter (< 1hr) tests, it makes sense to check every push, but if we start having beefier tests (like a full physics validation) run on every push, we could start butting up against GitHub's (softer) limits on the number of concurrent workflows and jobs. Not to mention the fact that we only look at the tests of a PR at specific times in development (much less frequent than every push). For this reason, I made the action manually-triggered. I will look more into [configuring the action trigger](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request) better so that it could still run "on PRs" but is more limited in scope than every push.

jmmans:
Regarding the question of "Create/Store" -- Tom, I think you didn't quite hit the point which @bryngemark which was trying to raise.  I think we all agree the _events_ would need to be generated on the fly.  However, it shouldn't be an issue to have the reference _histograms_ in the release -- we should be able to keep the histogram file under 1MB, I'd expect, and that would be much-more-stable.

tomeichlersmith:
I genuinely did not think of that. Thanks @jmmans for connecting the dots.

bryngemark:
Yes, thanks @jmmans :)  

@tomeichlersmith I get your concern about the test run time. My two thoughts on this:
1. How many events do we need to quickly find out if a change broke something? For PN, simulation of 6000 filtered events typically takes 1.5h. Would 500 events (<10 minutes) be enough and still fit the <1h total? (Here I think HCal might be the hardest to reach). Inclusive should be the same time scale (it's not the filtering that takes time, it's the full detector simulation). Maybe these could be run in parallel? 
- one test comparing 1e ecal PN plots 
- yet another could be 1+1e inclusive to make sure nothing broke wrt (combining and digitizing) pileup? 
- that would be three sim jobs, and two reco jobs (ecal PN = reco + DQM, 1+1e inclusive = overlay + reco + DQM)
- some dedicated HCal sim + reco + DQM job?
2. I'm thinking these automatic tests would primarily be there to test reconstruction changes (much more common than changes to sim or detector geometry, and runs much faster than simulation); if we do want a full suite of samples to test physics/simulation changes we might need to keep that as a dedicated validation that can be requested manually the way you had envisioned. That could also be using a larger number of events than the automatic ones. Would that be a good intermediate path?

Agreed, the github actions triggers for `pull_request` look highly customizable. I think it's important that the tests have been run on the final version that is just about to get approved. Having them run when you first push is nice and helpful but less crucial. So. If it's not on every push, then on initial Opening, on Ready for review, and what more, to guarantee that we achieve this? (I suppose it's a matter of style (sometimes stress), too, if you choose to push often, or to just commit often and push once you have tested it locally and have some confidence that it works.) With all the flags to `pull_request`, hacking some ad-hoc trigger up is easy, but ideally, the less manual work from the user to remember to flag/request something, the better. (And if that is indeed required, the default should be that merge is blocked until it's done.)

tomeichlersmith:
I've generalized our compiling/running into their own GitHub actions we can call within our workflows. This makes it much simpler to write new workflows using our container. Here are just some thoughts in no particular order.

## Parallel Jobs
This is a good idea, but with one snag. Each job in a workflow is dispatched to its own runner. At first glance, this means jobs can't share files/resources unless we upload the files/resources as a "artifact". This is definitely possible and something I am going to look into.

## Separate Sim/Detector Physics Validation
I also like separating out this type of longer and in depth validation. This reduces the scope of these (more frequent) validations and thus makes them easier to write.

## Sim-Scenarios
I like your scenarios and if I can get the parallel jobs working, we can have each scenario take approximately 5 hrs (extra hour to be safe because upload/download and runners might be slow). I also want to mention that the (< 1hr) limit is just my own subjective "limit". We could run 6 hour jobs on each push if we wanted to.

bryngemark:
Hi @tomeichlersmith,

**parallel jobs**
Sorry if I'm slow: what are the shared files/resources you're thinking of when running 2 (or 3) different simulation jobs? The container itself?

**sim scenarios** 
Yeah 5h should be plenty for most typical simulations. I actually think 1h is a reasonable goal for typical dev cycle turnaround and it would be great if we can stick with it. Another option for longer tests is to schedule nightly build+run tests (which works as long as night coincides for all everyday developers) to maintain automation but keep folks from waiting too much. Suboptimal for this scale of framework and development.

**DQM running**
You asked for examples of how to run the TS DQM plots. I think you already have or would have figured it out very quickly but here: use the catch-all `trigScint_dqm` which is already defined in `dqm.py` to get sim hits, (old) reco hits, clusters, and tracks.

tomeichlersmith:
The two shared files between jobs that I could see being useful are

1. The ldmx-sw install so each job doesn't have to compile ldmx-sw on its own. This would be one step short of building a production image and pushing it to DockerHub (which I wouldn't want to do because it would clutter up DockerHub).
2. For the overlay. It would be nice for the inclusive 1e sample to be validated on its own and used for the overlay, but this is not necessary.

As you can see, neither of these really prevent the parallelization of the different jobs, so I will focus on having a job for each "scenario" to validate and if I can get either of these shared resources working that would be a bonus.

tomeichlersmith:
# Update

### Bad News
Artifacts cannot be across workflows; they seem to be designed specifically for sharing data with a human after the workflow is complete. This is frustrating because I was hoping to have one workflow that would generate new "golden histograms" ("gold" for short) whenever a new push to `trunk` was made and the same workflow would compare the current PR against this gold for each push to the PR. Alas, this cannot be the case.

### Fix
My current solution is to simply generate "gold" each time we want this recon validation to be run. Since I've separated out the compiling of ldmx-sw and can now run different scenarios in parallel, I still think we will be able to stay comfortably under the 6 hour time limit.

I'm also looking into [caching](https://github.com/actions/cache#example-workflow) which _seems_ to be what we want, but I'm struggling to think of a way to "key" the cache so that the cached gold will change when trunk changes but won't change when commits to the PR branch are made. Caching is similar to creating an artifact, but it is only accessible by the GitHub actions/workflows.

jmmans:
Honestly, automatic update of "gold" seems like a *bad idea*.

omar-moreno:
> Honestly, automatic update of "gold" seems like a _bad idea_.

My thinking is that once we create a release, we would have validated it enough to feel comfortable for the plots to be updated automatically.

jmmans:
As a counter-argument, making the gold histograms is logically _part_ of the release process.

Anyway, there are a lot of things to work on, and getting the gold histograms automagically created might not be the optimal effort...

omar-moreno:
> As a counter-argument, making the gold histograms is logically _part_ of the release process.
> 

> Anyway, there are a lot of things to work on, and getting the gold histograms automagically created might not be the optimal effort...

I agree! This is definitely something we can live without.

tomeichlersmith:
## Update

I am now committing the "golden" histogram files to the git source tree. The basic idea is that each "sample" (i.e. simulation config) has its own directory in `.github/validation_samples` with the following structure.

- `.github/validation_samples/`
  - `<sample_name>/`
    - `config.py`
    - `gold.root`

where `config.py` can be run through `fire` to generate `events.root` and `hist.root`. `gold.root` is `hist.root` as generated by our last stable release (see plans). The validation action (`.github/actions/validate`) goes into this sample directory, runs the config script and then compares the output histogram file `hist.root` to `gold.root`.

I currently have three samples operational:

- `inclusive`: unbiased single electron
- `ecal_pn` : hard brem in target, PN > 2.5GeV in ECal
- `it_pileup` : 1e ECal PN + 1e inclusive at same time

The `it_pileup` sample requires some patches in Framework and ECal in order to function. This introduces differences between `trunk` and this branch which is why the `inclusive` and `ecal_pn` validations are failing. Look at the plots (under [artifacts](https://github.com/LDMX-Software/ldmx-sw/pull/990/checks?check_run_id=2676301153)) - only the ECal digi verify plots are failing).

## Plans
- [x] Write workflow to update `gold.root` files when a release is made (not sure how to test this...)
- [x] Write ECal Shower Feature DQM processor and include it in samples
- [ ] ~~Now that the actions is functional, go back and generate `gold.root` with the `v2.3.0` release~~

tomeichlersmith:
Because of the various bugs that have been patched since v2.3.0, I can't truly generate golden histograms for the three different samples using that release. I think the best thing to do going forward is to have this PR conain the first set of golden histograms (which are generated by this branch), validate v3.0.0 manually like we have been doing previously, and then generated golden histograms for v3.0.0 when that release is made.

This is the procedure I will be following unless someone says something to the contrary.

bryngemark:
You only asked for push-back, but, yes I think this makes sense. It can never be the goal that the golden histograms is an immutable standard; you'd need to be able to merge in a change that makes substantial changes compared to the reference as long as this is known, understood, validated and intentional.

From that point of view, starting from the place we're at is as good a choice as any.

tomeichlersmith:
## Related Submodule PRs
- https://github.com/LDMX-Software/Ecal/pull/33 (otherwise segfault with overlay producer)
- https://github.com/LDMX-Software/Framework/pull/43 (not **required** for this to work but helpful)

tomeichlersmith:
Ok. After too many commits to get these workflows updated and working, I am ready for it to merge. Going to dot some i's and cross some t's and then merge it.

For more detail about the workflows, you should look at the [workflows/README](https://github.com/LDMX-Software/ldmx-sw/pull/990/files?short_path=636942e#diff-636942e0afa55e1aea027fb857aa2c6046858fb317c0b9a4f12ff3ca17e64ffb).

## Workflow Outline
- Every Push of any branch or version tag
  - Basic Tests: make sure ldmx-sw compiles and executes `ctest` successfully
    - ~20min
    - Tested here while pushing
  - (trunk only) Generate Documentation: if ldmx-sw compiles, run doxygen and sphinx and push to ldmx-software.github.io
    - ~20min
    - Not changed, only moved, will be tested post-merge automatically
  - (trunk and verstion tags) Build Production Image: update ldmx/pro repo with most recent changes
    - ~30min
    - Not changed, only moved, will be tested post-merge automatically
- PR Opened or marked "Ready for Review"
  - Recon Validation: validate that histogram files match the current gold stored in the GitHub source tree
    - ~2hrs
    - Fully tested here
- Stable Release Published (no pre-releases)
  - Generate Recon Gold: generate histogram files and commit them as the new gold stored in the GitHub source tree
    - ~2hrs
    - Compiling ldmx-sw and running configs tested here, need to test if it works when triggered by release post-merge
- Manually Triggered
  - Build Production Image: user provides their GitHub branch to compile into the image, the tag to name this version of the image, and their DockerHub repo that the image should be pushed to
    - ~30min
    - Need to manually test post-merge
 
The recon validation is done with three simulation scenarions/samples (inclusive, ecal_pn, it_pileup) and each scenario has 10k event each. This is just enough events to allow the once-per-event values to have a distribution that is faintly readable.

## Post-Merge Plans
I am planning on attempting to manually trigger the Build Production Image workflow after this is merged into trunk. (Manually dispatch is only availabe for workflows on the default branch.)

I am also planning on creating a new release to test that the Generate Recon Gold workflow functions properly. I've tested it as much as I can here, but I need to make sure that the commiting action works when the workflow was triggered by a release (rather than by a push to a branch).

tomeichlersmith:
To the moon :rocket: :moon:

Issue #993, opened 2021-06-07
Change binning in TrigScint DQM plots


bryngemark:
Some of the new TrigScint DQM plots have some pretty unreasonable binning choices. I'll fix this.

tomeichlersmith:
Resolved by PR #994

PR #994, opened 2021-06-07
Update cluster DQM binning

bryngemark:
Resolves issue #993 

This is just some Github web interface range fixes so I haven't recompiled anything.


tomeichlersmith (approved):
This looks sensible. I am curious to see how my comparison script behaves when the histograms are definitely differently binned :thinking: So I'm going to wait and merge this in after that.

tomeichlersmith:
Neat! So it finishes running but "fails" the KS test for the histograms that have changed binning. Good to know.

bryngemark:
Yeah, that's what I thought. Sorry if I didn't make this clear, but that's what I meant by asking if you could regenerate both reference and PR histograms.

PR #1002, opened 2021-06-14
More validation!

### LINKED ISSUE #992: More Validation Development
With the basic structure for validation incorporated, we can start looking into various areas of improvement.

- [x] Incorporate hcal-specific simulation sample of some kind so that it can be fully tested
- [x] Write a general SimDQM analyzer for creating a standard set of sim plots to include in validation
- [x] pre-release action to generate larger sets of histograms - verified by _humans_ so that we can choose to trust the golden histograms that will be generated by a later stable release
  - https://github.com/ncipollo/release-action will be helpful for doing this
- [x] ~~How to compare/draw 2D plots during PR validation.~~
  - The scatter plot with different colors is actually ok, so I'm going to leave it as is
- [x] ~~Post plots that failed KS test to PR as comment~~
  - Difficult, delaying this until future
- [x] ~~Re-launch PR validation without having to convert to draft -> ready for review, but also less frequent than every push. Maybe parse PR comments? Not sure if this is possible...~~
  - Not possible without _a lot_ of hassle
- [x] ~~Auto convert PR to draft if fails PR validation? Not sure if this is possible...~~
  - Don't want to do this - we want to be able to override this test if we find a bug that needs to be patched that affects the reference histograms (either directly or indirectly). 
- [x] Make sure golden histograms and label lines up with what is on trunk so developers don't have to manually merge those files in for the validation to work
- [x] ~~Register test configs with `ctest` so we can check that they run: https://github.com/LDMX-Software/cmake/pull/11~~
  - will require updates to submodule CMakeLists.txt to implement
  - not right now

## General Validation Outline
Here is my plan given some comments from others in the SWAN meeting.
I don't think we can automate everything unfortunately; however, we can automate the sample generation so the humans only have to interact by viewing the plots and commenting about their behavior.

- **Pre-Release** 
  - large samples of histograms (not compared to anything) for humans to look at before stable release
  - Done by writing an action which can be manually triggered by a developer (I can't think of a better trigger)
  - Set of histograms uploaded to pre-release in this action for humans to download and investigate
  - Should we create an issue for discussion of these plots? https://github.com/marketplace/actions/create-an-issue
- **Stable Release**
  - After software has been validated individually using the pre-releases, make a stable release. This will trigger the generation of new golden reference histograms.
  - If a pull request fixes a bug that affects the reference histograms, another stable release should be made to update them (probably just incrementing the patch number).
 - **Pull Requests**
   - Compare histograms generated by the developments to the golden reference histograms.

tomeichlersmith:
Updating our action infrasture with more workflows and better designed ones.
Documentation for the actions and workflows is scattered about in the `.github` directory.

## Issues
This resolves #992
Here we follow the outline given on the issue.

No submodules need to be updated.

---
BTW, I created this PR using gh!


tomeichlersmith:
## To-Do Before Merge
- [x] Look through HCal plots for binning - do they make sense?
- [x] Add Trigger DQM plots
- [x] Check that new PN-children histograms didn't break anything
- [x] More HCal plots
  - a 2d plot of PEs per layer and maybe per strip?: so e.g. this would be filled per hcalhits `histograms_.fill("pe_per_layer", layer, hit.getPE());` I guess w a range from 0 to 5
  - a 2d plot of the layers and strips from the hits? e.g. just `histograms_.fill("layer_vs_strip", layer, strip);` separated back vs side
  - the sum of PES only for the back hcal
  - position of all hits separated back vs side

tomeichlersmith:
Because of changes to the DQM analyzers, the PR Validation will fail, but I'm using it to double check that everything runs correctly (i.e. I'm looking to see if it fails at the correct step).

Issue #904, opened 2020-09-07
ECal Layer Shifting and incorporate layerID in ECalHexReadout


Hongyin-bug:
**Is your feature request related to a problem? Please describe.**
Layer-shift the second layer within each bi-layer of Ecal layers--- to cover cracks.
ECalHexReadout is a x-y map of cells, but does not describe how x and y changes with z.
ECalID.h has layers defined in terms of double layers (i.e. 17 layers with 12 ring modules in a layer)--- not convenient for layer shifting (need to separately define 34 layers)

**Describe the solution you'd like**
Need to add layerID and incorporate z as a function of x-y in ECalHexReadout.h and .cxx.
Add Layer shifting in ECalHexReadout.
Change ECalID so that there are 34 layers and 6 ring modules in each.

**Describe alternatives you've considered**
Link the ecal.gdml file to ECalHexReadout so that it's not necessary to change ECalHexReadout every time a new layer-configuration is tried out?

**Additional context**
![layer-shift](https://user-images.githubusercontent.com/57277920/92353112-73644a80-f094-11ea-9ae1-0dd9aff1f706.png)
![One module layer-shift](https://user-images.githubusercontent.com/57277920/92419309-9ee14680-f121-11ea-87a0-1bd427d82977.png)

jmmans:
Dear Hongyin,

EcalID doesn't know how many layers there are, it is simply a container 
for values passed in to it.  The assignment of layer numbers is done in 
EcalHexReadout as used in the simulation code.

I agree with the idea that each actual layer should be separate in the 
code, so if it's true that we have double layers in there now, it would 
make sense to change it.  This will break a lot of things, however, so 
it cannot be done in the mainline until a future major release.  It 
would be reasonable to pursue as a branch development until then.

Regarding the GDML and EcalHexReadout issue -- this is a brittle 
connection because the GDML is only loaded when an actual simulation job 
is run.  If the digis are stored and then loaded back off disk, the GDML 
isn't loaded.  Instead, EcalHexReadout needs to know what to do 
_without_ the GDML available.  Therefore, the necessary information 
would need, for example, to be put into the RunHeader.  It could be 
sourced from the GDML.  Detailed proposals of what information is 
sufficient would be valuable.

J. Mans

Hongyin-debug:
Hi Jeremiah,
I see! Yes, the ECal ID has container values of up to 17 layers.
[...]

Issue #923, opened 2020-10-08
Rename and Clean-Up EcalHexReadout


tomeichlersmith:
EcalHexReadout is a class that we are now using for almost all translation between EcalID and real-space position, not just cutting up the module hexagons into cellular hexagons. With this in mind, it makes a lot more sense to call it something more general than `EcalHexReadout`, probably `EcalGeometry` would work best, and then it would line up with the provider that provides it `EcalGeometryProvider`.

I'm also going to take this opportunity to clean-up the Ecal precision digis and the HgcrocEmulator. I need to make the collection name and pass name for the inputs configurable.

@omar-moreno I would love to put all of this Ecal-related stuff in the Ecal module. After the cmake changes you are working on, how difficult would it be to make a "Geometry" sub-module of Ecal (similar to how we are making "Event" submodules)? This way, `SimCore` and `Ecal::Ecal` could link to `Ecal::Geometry` and the source for `Ecal::Geometry` would be in the `Ecal` module.

omar-moreno:
On Thu, Oct 8, 2020 at 1:46 PM Tom Eichlersmith <notifications@github.com>
wrote:

> EcalHexReadout is a class that we are now using for almost all translation
> between EcalID and real-space position, not just cutting up the module
> hexagons into cellular hexagons. With this in mind, it makes a lot more
> sense to call it something more general than EcalHexReadout, probably
> EcalGeometry would work best, and then it would line up with the provider
> that provides it EcalGeometryProvider.
>
> I'm also going to take this opportunity to clean-up the Ecal precision
> digis and the HgcrocEmulator.
>
> @omar-moreno <https://github.com/omar-moreno> I would love to put all of
> this Ecal-related stuff in the Ecal module. After the cmake changes you are
> working on, how difficult would it be to make a "Geometry" sub-module of
> Ecal (similar to how we are making "Event" submodules)? This way, SimCore
> and Ecal::Ecal could link to Ecal::Geometry and the source for
> Ecal::Geometry would be in the Ecal module.
>
This sounds like a good idea. I have already started doing something
similar in SimCore (added SimCore/Geo) and plan to do the same in the
Tracking module.

tomeichlersmith:
> Moving future updates from #873 here.

### Known un-realistic issues to rectify:

- [ ] The cell ID with respect to the x-coordinate will flip between sensitive layers because the modules will be flipped when mounted to make the full layer "sandwich".
[...]

PR #1003, opened 2021-06-17
Implementing a rotated ECal

AminaJLi:
This partially addresses issue #904. Just the rotation (mostly done by Hongyin Liu) has been implemented. I have checked that simHits look like a rotated flower and that the scoring plane hits have adjusted to appropriately enclose the ECal.

A couple things to note: I reverted the HCal submodule and simpleTrigger to older commits (matching the original branch) to let me simulate with configs I already had. This may be causing the automatic merge issue. I was also not getting any events past EcalProcessFilter and am still looking at what is killing the events. (TaggerVetoFilter, TargetBremFilter, and TrackProcessFilter.photo_nuclear all seem to work.) 

**simHit distribution** (_blue lines are manually added by me and are where the old scoring planes should be_)
simhitX_simhitY_100k

**scoring plane hits** (_arbitrarily chosen z cutoffs to not show z plane, black lines are manually drawn old scoring planes_)
simSPhitX_simSPhitY_base


tomeichlersmith:
Whoops, I rebased onto iss992-more-validation rather than trunk. That's ok, I was planning on merging that first anyways.

tomeichlersmith:
## To-Do Before Merge
- [x] Separate rotated (v13) ecal from current v12 (unrotated) in GDML
- [x] Allow for EcalHexReadout to handle both rotated (v13) and unrotated (v12) geometries
- [x] The cellular layout on each module is slightly off center (that's why I had the 35.3 instead of the 35). Below I've put the known cellular layout, it is a 90 degree rotation away from what we are going for, but it is what we should be trying to replicate.

HG Module Layout

I will help move this along after we get v3.0 released sometime over the next week.

AminaJLi:
> * [x]  The cellular layout on each module is slightly off center (that's why I had the 35.3 instead of the 35). Below I've put the known cellular layout, it is a 90 degree rotation away from what we are going for, but it is what we should be trying to replicate.
> 
Two comments/questions with regards to this! (@tomeichlersmith but really anyone who knows about the module)
1. The known layout looks to have a center at what Hongyin and I have been calling a triple point - ie. the corner of cell 209, 210, and 233 in the cell map. With 35 cells across, the center lands on 17.5 cell radii which is this triple point. The difference between 35 and 35.3 is pretty small so I suppose a visual assessment isn't the best but please confirm or deny this particular symmetry.

**cell map with 35 cells across**
Cell_ID_Cell_Position_Map

2. Beyond nice symmetry motivations for putting the triple point in the middle, we run into "out of bounds" errors with 35.3 cells. 35 cells makes it such that the edge of the module perfectly cuts off one triangle on the cells. Below is the map with 35.3 and you can see there are gaps that cause errors in the left and upper-right side. If the answer to comment/question 1 is that 35.3 is necessary, this needs to be addressed.

**cell map with 35.3 cells across**
Cell_ID_Cell_Position_Map

jmmans:
Dear Amina,

The sensor layout as designed by CMS has a corner at the center of the 
module, not a cell-center at the middle.  This is necessary because 
otherwise it's not possible to define a consistent set of trigger sums 
within the constraints of the hexagonal geometry.

As for the rotation -- you may consider simply swaping the axes as used 
by HexReadout.  This may be less-painful than working through all the 
rotations.  Also, such a manipulation will likely be necessary /anyway/ 
to represent the left-right mirroring on the opposite sides of the 
doublelayer.

Jeremy

AminaJLi:
Thank you for the speedy reply Jeremy! I think we are saying the same thing for the center of the module. The "triple point" is a meeting of three cell corners and I believe having 35 cells for the module height (flat side to flat side) puts this "triple" point at the module center. 

The point about swapping axes is a good one and I will look into it!

AminaJLi:
> * [x]  Separate rotated (v13) ecal from current v12 (unrotated) in GDML
> * [x]  Allow for EcalHexReadout to handle both rotated (v13) and unrotated (v12) geometries
>
@tomeichlersmith I believe all the tasks are done! I just pushed the changes I made to separate out v13 and v12 plus a (slightly hacky) implementation of EcalHexReadout that accepts both. ldmx-sw compiles and just by changing the config from "ldmx-detv-12" to "ldmx-det-v13" or vice versa, I get the following plots. 

**ldmx-det-v12**
hitX_hitY_v12

**ldmx-det-v13**
hitX_hitY_v13


Hopefully this means we can incorporate the geometry into v3.0.0? Let me know any issues.

DetDescr/src/DetDescr/EcalHexReadout.cxx
```diff
@@ -20,8 +21,11 @@ EcalHexReadout::EcalHexReadout(const framework::config::Parameters& ps)
   nCellRHeight_ = ps.getParameter<double>("nCellRHeight");
   verbose_ = ps.getParameter<int>("verbose");
 
+  if (nCellRHeight_ == 35) rotated_ = true;
+  else rotated_ = false;
```
tomeichlersmith: `rotated_` should be it's own parameter that is decoupled from `nCellRHeight_`.

Also sidenote, I think "rotated" is too vague. Maybe `pointUp`? (That's not great either....) Something that says "when this is true, the module has a vertex at the top and when its false, the module has a flat side up".

DetDescr/src/DetDescr/EcalHexReadout.cxx
```diff
@@ -20,8 +21,11 @@ EcalHexReadout::EcalHexReadout(const framework::config::Parameters& ps)
...
+  
   moduleR_ = moduler_ * (2 / sqrt(3));
-  cellR_ = 2 * moduler_ / nCellRHeight_;
+  cellR_ = 2 * moduler_ / 35.;
```
tomeichlersmith: Revert this change back to `trunk` so that v12 can function properly.

tomeichlersmith (changes requested):
I opened the Ecal PR for patching the hex readout printing function:

https://github.com/LDMX-Software/Ecal/pull/35

Besides the few comments I have, I do think this solution is good enough. I have a separate issue (#923 ) for cleaning up this class anyways.

DetDescr/src/DetDescr/EcalHexReadout.cxx
AminaJLi: Tried to come up with a better name that isn't excessively long -- `cornersSideUp` at least suggests what we're going for but perhaps is not clear enough? There is of course an explanatory comment so hopefully that helps.

tomeichlersmith (approved):
Looks good! I will post the validation plots generated by the PR validation after it completes. The PR validation will "fail" because we have previously merged in changes to the reference plots without making a new release, but we can still use the plots that are generated in order to double check that these changes keep v12 stable.

tomeichlersmith:
I don't see any major issues within the validation samples. The hcal sample doesn't apply and the ecal_pn sample actually doesn't have any Ecal distributions failing the KS test against the gold generated by v3.0.0-gamma.2.

Both the inclusive and it_pileup samples have Ecal plots failing the KS test, so I've uploaded them as a slideshow that includes both the passing and failing distributions. _Remember that the solid red distribution was generated with v3.0.0-gamma.2 and has other small changes besides these ones_. The [full comparison between this branch and v3.0.0-gamma.2](https://github.com/LDMX-Software/ldmx-sw/compare/v3.0.0-gamma.2...iss904-rotation) shows the source of these changes.

- [it_pileup.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6784651/it_pileup.pdf)
- [inclusive.pdf](https://github.com/LDMX-Software/ldmx-sw/files/6784654/inclusive.pdf)

Since I don't see any major issues and we want to partially use this next release to validate v13, I'm going to merge this in.

Issue #1013, opened 2021-07-31
HCal Trigger digis


therwig:
The hcal trigger primitives, based on the HGCROC sums should be added, in a manner similar to the work done for ecal.  In the central ldmx-sw repo this requires the addition of a HcalTriggerID class.

tomeichlersmith:
Forgot to link this issue, resolved by PR #1014

PR #1014, opened 2021-07-31
Adding HcalTriggerID

therwig:
I am updating _ldmx-sw_ to include an HcalTriggerID class, resolving #1013.
This development is related to (but does not depend on) a PR to the Hcal repo adding a producer for the trigger primitive objects.

### DIFF
--- a/DetDescr/include/DetDescr/HcalDigiID.h
+++ b/DetDescr/include/DetDescr/HcalDigiID.h
@@ -33,7 +33,7 @@ class HcalDigiID : public HcalAbstractID {
     if (!null() && bar_type() != Digi) {
       EXCEPTION_RAISE(
           "DetectorIDMismatch",
-          "Attempted to create HcalID from mismatched Hcal bar_type " +
+          "Attempted to create HcalDigiID from mismatched Hcal bar_type " +
               std::to_string(bar_type()));
     }
   }
@@ -45,7 +45,7 @@ class HcalDigiID : public HcalAbstractID {
     if (!null() && bar_type() != Digi) {
       EXCEPTION_RAISE(
           "DetectorIDMismatch",
-          "Attempted to create HcalID from mismatched Hcal bar_type " +
+          "Attempted to create HcalDigiID from mismatched Hcal bar_type " +
               std::to_string(bar_type()));
     }
   }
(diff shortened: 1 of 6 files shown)


tomeichlersmith (commented):
This all looks good to me, pretty logical extension. I'm wondering if the TriggerID also needs an "End" sub-filed similar to the HcalDigiID.

https://github.com/LDMX-Software/ldmx-sw/blob/4266534065d4a4380c71239235f58648a70d328e/DetDescr/include/DetDescr/HcalDigiID.h#L101-L116

If so, then the question becomes: can we simplify the ID classes to reduce code copying? This is more of an aesthetic question right now, but would lead to easier maintainability in the future.

jmmans:
A few comments:

 (1) I strongly suggest changing STRIP to SUPERSTRIP in the constants to reduce confusion.

 (2) Will there need to be more than one level of Trigger Primitives?  For example, the HGCROCs will make TPs for superstrips on one end or the other.  At some point in the system, I'd assume you'd combine the superstrips from either end to create another "composite" superstrip.

Jeremy

jmmans:
Regarding: _If so, then the question becomes: can we simplify the ID classes to reduce code copying? This is more of an aesthetic question right now, but would lead to easier maintainability in the future._

So, from one who has had painful experience with evolution of the packing of ids, I think keeping the two separate is a GOOD IDEA.  I realize there is code reuse, but it's minor.

therwig:
Thanks to you both for the very useful comments. I've pushed commits to (1) rename strip->superstrip to avoid confusion, and (2) storing a bar end attribute, which I agree will be necessary.

The question of having 2 levels of TPs is a good one too. As long as there is still a notion of superstrip and layer, I think the current recipe is sufficient up to the 'end' attribute.   Instead of just positive and negative, I've also  tentatively added a 'composite' option, but am happy to discuss other ideas.

cmantill (approved):
hi @therwig , this looks good from my side and good to be merged! Just for curiosity, when would the "combined" superstrip be used?

therwig:
@cmantill The CMB quad and 'super trigger quad' outputs always carry correspond to a particular end, we'll want to sum both ends into a combined object in the APx.  We can identify these combined TPs with this same HcalTriggerID class if we have the 'composite' end type.

tomeichlersmith (commented):
These changes look good and they don't break things. The last request I have is formatting. If you have `clang-format` installed:

```bash
clang-format -i --style=Google
```
on the files you've added.

Let me know if you don't have it installed and I can do it.

therwig:
Thanks @tomeichlersmith, I think I've all the files should be compliant now. Let me know if anything stands out, otherwise good to go from my side :)

PR #1026, opened 2021-10-05
Changing simpleTrigger default start layer to 0

AminaJLi:

I didn't make an issue (sorry!) as I thought just making a ready-to-merge branch would be easier. This just changes the line in simpleTrigger.py to default the starting layer to 0.

## Check List

### DIFF
--- a/Recon/python/simpleTrigger.py
+++ b/Recon/python/simpleTrigger.py
@@ -42,7 +42,7 @@ class TriggerProcessor(ldmxcfg.Producer) :
         self.beamEnergy = 4000.
         self.thresholds = [ 1500.0, 1000. + self.beamEnergy, 500. + 2*self.beamEnergy, 100. + 3*self.beamEnergy ]  #toy something up 
         self.mode = 0
-        self.start_layer = 1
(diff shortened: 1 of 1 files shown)


AminaJLi:
For ease of those looking at this, all the plots that failed are either Trigger EcalESum or Trigger EcalLayercut. I'm attaching the ones from the ecal_pn sample. I believe it makes sense for the KS test to have failed when an additional layer is being included for the triggering so I don't think this is an actual problem. (Also this is all a pretty small change 🙂 )
[Trigger_Trigger_Trigger.EcalEsum.pdf](https://github.com/LDMX-Software/ldmx-sw/files/7289877/Trigger_Trigger_Trigger.EcalEsum.pdf)
[Trigger_Trigger_Trigger.EcalLayercut.pdf](https://github.com/LDMX-Software/ldmx-sw/files/7289878/Trigger_Trigger_Trigger.EcalLayercut.pdf)

tomeichlersmith (approved):
Looks good :) thank you for taking the time to do the git nonsense.

PR #1028, opened 2021-10-15
Don't sort vectors

tomeichlersmith:
Simple fix but want to run the auto tests before merging something a part of the core.


tomeichlersmith:
@bryngemark This failure is surprising to me. The plots that fail the KS test are in the TS subsystem or the Trigger. Are there parts that would depend on a sorted vector? This may be solved by simply inserting a `std::sort` at the beginning of one of the TS processors.

We may just want to push this update since the plots are really close (just slightly different) and we can debug later...

bryngemark:
This is very interesting indeed. It looks like most of these differences can be tracked down to the TS clustering. At least this is the earliest part of the chain clustering-->tracking-->electron count, which all show differences. 

It's interesting because
1. I explicitly *don't* expect anything to be sorted, but instead go through an arduous procedure of mapping/explicitly looking everything up by channel number
2. the overall effect is that efficiency improves a tiny bit (perhaps at the expense of fake rate, hard to tell with small stats; so far it looks identical), both in terms of sharper counts of things, but also "cleaner" electron tracks (higher `beamEfrac`)

I've checked the digi code too and I can't spot any place where the position in an input vector matters. 
It's hard to tell what's going on in detail without staring at the logs. I suppose the testing isn't run in debug + all processors verbose logging mode?

tomeichlersmith:
nope :/ they are not run in verbose mode

bryngemark:
OK. Perhaps running tests in verbose mode is an interesting possibility for the future. Especially if a clever diff can be run. ... it requires some thinking to make it a useful comparison in any case. 

I won't object to merging this. I see the urgency, and as you say, the difference appears marginal (with electron overcounting as my primary concern, since we are looking at relative rates at the 1e-5 level and that's untested for now). But now we have seen where that might come from, in that case. I'm so glad we have these validations 😄

Issue #1047, opened 2022-01-15
The HcalGeometry (in DetDescr) __str__ function doesn't work


EinarElen:
**Describe the bug**
Just some typos that need to be patched. 

**To Reproduce**
Try printing a HcalReadoutGeometry object :) 

**Desired behavior**
Output of all the parameters is printed without any errors.

EinarElen:
I went ahead and restructured the function as a whole to be a bit more useful for debugging things (and handling the difference between the three geometry types to some degree). Given that the function was broken, I'm going to assume that its safe to say that nobody was using it at the moment. 

Something along the lines of (from 9f2e2e6d4a48e6f3111936f445c9e1320719da3f)
```python
def __str__(self) :                                             
    """Stringify this configuration class"""                    
    s = '''                                                     
    HcalReadoutGeometry {{                                      
        Scintillator thickness: {:.1f} [mm], width: {:.1f} [mm],
        Number of sections: {{{}}},                             
        Layers: {{                                              
            Number of layers: {}                                
            Layer thickness: {} [mm]                            
            Z-position of zero-th layer: {} [mm]                
            Half total width of layers: {} [mm]                 
            Number of strips per layer: {}                      
            Location of zero-th strip per layer: {} [mm]        
        }},                                                     
        Ecal DX, DY: {}, {} [mm],                               
        Valid detector regexps: {}                              
    }}                                                          
    '''.format(self.ThicknessScint, self.WidthScint,            
            self.NumSections,                                   
            self.NumLayers, self.LayerThickness, self.ZeroLayer,
            self.HalfTotalWidth,self.NumStrips, self.ZeroStrip, 
            self.EcalDx, self.EcalDy,                           
            self.detectors_valid)                               
    return s                                                    
```

gives the following output for the different geometries
Prototype 
```
        HcalReadoutGeometry {
            Scintillator thickness: 20.0 [mm], width: 50.0 [mm],
            Number of sections: {1},
            Layers: {
                Number of layers: [19]
                Layer thickness: [49.0] [mm]
                Z-position of zero-th layer: [-438.5] [mm]
                Half total width of layers: [200.0, 200.0, 200.0, 200.0, 200.0, 200.0, 200.0, 200.0, 200.0, 300.0, 300.0, 300.0, 300.0, 300.0, 300.0, 300.0, 300.0, 300.0, 300.0] [mm]
[...]

Issue #1063, opened 2022-05-04
Updates to the hcal prototype geometry


EinarElen:
The current ldmx-hcal-prototype-v1 geometry differs from what was actually used at the test beam so we need to add a couple of new GDML descriptions (corresponding to different configurations that were used) which also brings along some necessary updates to LDMX-sw (mostly for the Hcal submodule but also in DetDescr). 

- The ordering of horizontal and vertical layers in the prototype differs from the regular Hcal  
- Some measurements were carried out with the horizontal bars shifted which means that the center of the bars cannot be assumed to be at x/y = 0

EinarElen:
@bryngemark I've pushed a version which has @PeterGy's latest work on https://github.com/LDMX-Software/ldmx-sw/tree/iss1063 in the Detectors/data/ldmx-hcal-prototype-v2.0 directory. There are some different versions of the TS available in the trigger_physvols file, but everything except plastic should be commented out by default. 

```python
from LDMX.Framework import ldmxcfg
from LDMX.SimCore import generators
from LDMX.SimCore import simulator


process = ldmxcfg.Process('test_TS')

process.maxEvents = 1000
process.run = 10
process.outputFiles = ['test_TS.root']

gun = generators.gun('particle_gun')
gun.particle = 'e-'
gun.direction = [0., 0., 1.]
gun.position = [0., 0., -1000.]
gun.energy = 4.

simulation = simulator.simulator('test_TS')
simulation.generators=[gun]
simulation.setDetector('ldmx-hcal-prototype-v2.0')

from LDMX.TrigScint.trigScint import TrigScintQIEDigiProducer
from LDMX.TrigScint.trigScint import TrigScintRecHitProducer

tsDigis = TrigScintQIEDigiProducer.up()
tsDigis.number_of_strips = 12
tsRecHits = TrigScintRecHitProducer.up()


process.sequence=[simulation,
                  tsDigis,
                  tsRecHits
                  ]
```

I stole this from what @PeterGy has been using and it does ~something~ although you and Peter are probably better than me at knowing if it is sensible :)

bryngemark:
very helpful, thanks @EinarElen! this looks reasonable.

EinarElen:
I've uploaded the basic configuration (plastic TS, no shifted bars) of the April testbeam to https://github.com/LDMX-Software/ldmx-sw/tree/iss1063 together with an updated DetDescr readout geometry. See updated test config below. 

[...]

PR #1072, opened 2022-07-19
Iss1063

EinarElen:

This resolves parts **(but not all!)** of https://github.com/LDMX-Software/ldmx-sw/issues/1063 

## Check List

I've tested the config file in the issue with both `ldmx-hcal-prototype-v1.0`, `ldmx-hcal-prototype-v1.0`, and `ldmx-det-v13` and checked that
- the position map changes looked right (using the `printPositionMap` debug function in the HcalGeometry setup. 
- that runs with a particle gun placed to the right gets both SimHit and RecHit position distributions shifted to the right 
- The HcalRecProducer doesn't crash because the HcalID's from the SimHits matches the geometry 

### Related Sub-Module PRs
- https://github.com/LDMX-Software/SimCore/pull/64 
-  https://github.com/LDMX-Software/Hcal/pull/48


tomeichlersmith (approved):
This PR is a _mega_ PR that also updates Hcal and SimCore to their initial release commits which contain a variety of bug fixes and enhancements developed via testbeam. I am going to trigger a re-run of the PR validation test to make sure that our basic simulation and basic reconstruction have not been broken, but if that passes, I'll be ready to merge.

cmantill:
@EinarElen would this be a good place to integrate your old commit on the detector description string? https://github.com/LDMX-Software/ldmx-sw/commit/9f2e2e6d4a48e6f3111936f445c9e1320719da3f

EinarElen:
Yeah that would make sense, would resolve https://github.com/LDMX-Software/ldmx-sw/issues/1047

tomeichlersmith:
I think that makes sense, go ahead and add that commit to this branch!

cmantill (approved):
looks good, @tomeichlersmith can you trigger the long checks?

tomeichlersmith:
I already did and they are [almost complete](https://github.com/LDMX-Software/ldmx-sw/runs/7430769716?check_suite_focus=true) - they don't test the string printout of the HcalGeometry, so that commit doesn't need to be included.

PR #1075, opened 2022-08-09
Update SimCore to v1.0.0

### LINKED ISSUE #1045: Examples/documentation fails to set runNumber correctly
Not entirely sure where to place this issue, it touches multiple repositories although perhaps 
primarily https://github.com/LDMX-Software/ldmx-software.github.io

When working with some of the simulation samples that were produced for the
prototype testrun, <https://github.com/PeterGy> noticed that the samples that were
supposed to have different runNumbers didn&rsquo;t. As it turns out, several parts of
the documentation/examples are telling you to set runNumbers with something like
the following

    sim = simulator.simulator('...')
    sim.runNumber = 1

I&rsquo;m assuming this was how things were done before but as far as I can tell, this
is supposed to be done with the process object instead. Since LDMX-sw currently
has no way to warn you if you provide an argument that isn&rsquo;t used, this ends up
with setting the run number to the default for every run.

I know that we aren&rsquo;t the only ones that have ran into trouble with runNumbers
not being set so it might be a good idea to have a dedicated warning in SimCore
at least as long as there is no way for LDMX-sw to warn for unused parameters as
a whole.

There are both some uses of this in some of our examples and test code, but the
primary problem is the website which most people, new or old, will look to when
they want to figure out how to configure ldmx-sw.

I think this is a rather urgent thing to fix because mistakes can cost us quite
a lot. I am willing to go through both the website and the sources for all the
occasions where we are currently doing this. Personally, I think it would also
be useful to add a dedicated warning for this in SimCore. If other people agree,
I can implement that as well.

tomeichlersmith:

This follows some big technical updates to SimCore _which are not expected to change the physics output_.

## Check List
  - I used the tree-diff executable from [root-diff](https://github.com/tomeichlersmith/root-diff) to verify that the output of the SimCore and Biasing test configs are the same as what is produced on `trunk` on a bit-by-bit level (with the exception of the timestamps and the incident ID contribs for the TriggerPad and Target sim hits).
  - Simply moving SimCore up to its latest release. If a bug is found, a patch release will be made.

### DIFF
--- a/Biasing/python/ecal.py
+++ b/Biasing/python/ecal.py
@@ -51,7 +51,6 @@ def photo_nuclear( detector, generator ) :
     sim.setDetector( detector , True )
     
     # Set run parameters
-    sim.runNumber = 0
     sim.description = "ECal photo-nuclear, xsec bias 450"
     sim.beamSpotSmear = [20., 80., 0.] #mm
     
--- a/Biasing/python/target.py
+++ b/Biasing/python/target.py
@@ -45,7 +45,6 @@ def electro_nuclear( detector, generator ) :
     sim.setDetector( detector , True )
 
     # Set run parameters
-    sim.runNumber = 0
     sim.description = "Target electron-nuclear, xsec bias 1e8"
     sim.beamSpotSmear = [20., 80., 0.] #mm
 
@@ -101,7 +100,6 @@ def photo_nuclear( detector, generator ) :
     sim.setDetector( detector , True )
 
     # Set run parameters
-    sim.runNumber = 0
     sim.description = "ECal photo-nuclear, xsec bias 450"
     sim.beamSpotSmear = [20., 80., 0.]
 
@@ -157,7 +155,6 @@ def gamma_mumu( detector, generator ) :
     sim.setDetector( detector , True )
 
     # Set run parameters
-    sim.runNumber = 0
     sim.description = "gamma -> mu+ mu-, xsec bias 10e9"
     sim.beamSpotSmear = [20., 80., 0.]
 
--- a/Biasing/src/Biasing/Utility/DecayChildrenKeeper.cxx
+++ b/Biasing/src/Biasing/Utility/DecayChildrenKeeper.cxx
@@ -10,7 +10,6 @@
 /*   SimCore   */
 /*~~~~~~~~~~~~~*/
 #include "SimCore/UserTrackInformation.h"
-#include "SimCore/UserTrackingAction.h"
 
 namespace biasing {
 namespace utility {
@@ -24,8 +23,7 @@ DecayChildrenKeeper::DecayChildrenKeeper(const std::string& name,
 DecayChildrenKeeper::~DecayChildrenKeeper() {}
 
 void DecayChildrenKeeper::PostUserTrackingAction(const G4Track* track) {
-  const auto& particle_map{
-    simcore::UserTrackingAction::getUserTrackingAction()->getTrackMap()->getParticleMap()};
+  const auto& particle_map{getCurrentParticleMap()};
   if (particle_map.find(track->GetParentID()) != particle_map.end()) {
     const int& parent_pdg{particle_map.at(track->GetParentID()).getPdgID()};
     for (const int& parent : parents_) {
--- a/Biasing/test/ecal_db.py
+++ b/Biasing/test/ecal_db.py
@@ -3,6 +3,7 @@ p = ldmxcfg.Process('ecal_db')
 from LDMX.Biasing import eat
 from LDMX.SimCore import makePath
 from LDMX.Ecal import EcalGeometry
+from LDMX.Hcal import HcalGeometry
 p.sequence = [
     eat.dark_brem( 
         10., #MeV - mass of A'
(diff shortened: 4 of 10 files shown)


tomeichlersmith:
The Basic Tests failed due to the Hcal subsystem setting the run number in-appropriately in its test configs:
```
6/7 Test #6: Hcal .............................***Failed    8.90 sec
Filters: [Hcal]

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
run_test is a Catch v2.13.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
Hcal Geometry test
-------------------------------------------------------------------------------
/export/scratch/users/eichl008/ldmx/ldmx-sw/Hcal/test/HcalGeometryTest.cxx:101
...............................................................................

/export/scratch/users/eichl008/ldmx/ldmx-sw/Hcal/test/HcalGeometryTest.cxx:108: FAILED:
  REQUIRE_NOTHROW( p = cfg.makeProcess() )
due to unexpected exception with message:
  Remove old-style of setting the simulation run number (sim.runNumber).
  Replace with using the Process object (p.run).

===============================================================================
test cases:    2 |    1 passed | 1 failed
assertions: 6002 | 6001 passed | 1 failed
```
Updating `Hcal/test/hcal_geometry_test_config.py` as below resolved the issue locally. The other PR validation tests are failing for similar reasons.

```diff
diff --git a/test/hcal_geometry_test_config.py b/test/hcal_geometry_test_config.py
index 9babf83..0541563 100644
--- a/test/hcal_geometry_test_config.py
+++ b/test/hcal_geometry_test_config.py
@@ -3,6 +3,9 @@ from LDMX.Framework import ldmxcfg
 # Create a process
 p = ldmxcfg.Process( 'test_hcal_geometry' )
 
+# run number (and random seed)
+p.run = 1
+
 # Set the maximum number of events
 p.maxEvents = 1
 
@@ -32,7 +35,6 @@ from LDMX.SimCore import generators
 
 sim = simulator.simulator("single_neutron")
 sim.setDetector( 'ldmx-det-v12' , True )
-sim.runNumber = 0
 sim.description = "HCal muon"
 sim.beamSpotSmear = [20., 80., 0.] #mm
```

EinarElen (approved):

tomeichlersmith:
The two PR validation tests failed for different reasons.

## it_pileup
The `it_pileup` test failed because the DQM tests plotting the incident IDs were different. This was expected and not (too) concerning.
```
List of Plots Failing KS Test
  sim_dqm_sim_dqm_TriggerPadDownSimHitsOverlay.incidents.pdf
  sim_dqm_sim_dqm_TriggerPadUpSimHitsOverlay.incidents.pdf
  sim_dqm_sim_dqm_TargetSimHitsOverlay.incidents.pdf
  sim_dqm_sim_dqm_TriggerPadTaggerSimHitsOverlay.incidents.pdf
```
[sim_dqm_sim_dqm_TriggerPadTaggerSimHitsOverlay.incidents.pdf](https://github.com/LDMX-Software/ldmx-sw/files/9300751/sim_dqm_sim_dqm_TriggerPadTaggerSimHitsOverlay.incidents.pdf)
[sim_dqm_sim_dqm_TargetSimHitsOverlay.incidents.pdf](https://github.com/LDMX-Software/ldmx-sw/files/9300752/sim_dqm_sim_dqm_TargetSimHitsOverlay.incidents.pdf)
[sim_dqm_sim_dqm_TriggerPadUpSimHitsOverlay.incidents.pdf](https://github.com/LDMX-Software/ldmx-sw/files/9300753/sim_dqm_sim_dqm_TriggerPadUpSimHitsOverlay.incidents.pdf)
[sim_dqm_sim_dqm_TriggerPadDownSimHitsOverlay.incidents.pdf](https://github.com/LDMX-Software/ldmx-sw/files/9300754/sim_dqm_sim_dqm_TriggerPadDownSimHitsOverlay.incidents.pdf)

## inclusive
More concerning is the `inclusive` test which failed _during the run_ due to a missing event product. More investigation to come.

```
// -- snip --
    [ ElectronCounter ] 4 : Attemping to use non-existing input collection TriggerPadTracks_ to count electrons! Exiting.
   [ fire ] 4 : [ProductNotFound] : No product found for name 'TriggerPadUpClusters'
    at /home/runner/work/ldmx-sw/ldmx-sw/Framework/include/Framework/Event.h:267 in getObject
  Stack trace: 
      0 std::vector<ldmx::TrigScintCluster, std::allocator<ldmx::TrigScintCluster> > const& framework::Event::getObject<std::vector<ldmx::TrigScintCluster, std::allocator<ldmx::TrigScintCluster> > >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const + 5890 
      1 dqm::TrigScintClusterDQM::analyze(framework::Event const&) + 63 
      2 framework::Process::process(int, framework::Event&) const + 100 
      3 framework::Process::run() + 8356 
      4 main + 509 addr2line: 'fire': No such file

// -- ROOT seg faulting --
```

tomeichlersmith:
I was able to run the `inclusive` test locally, so I am just re-running them. There could have easily been some other sort of interruption during processing that caused some funkiness, but we will see.

_Update_: The running of `inclusive` locally produced a perfect match between the golden and generated histograms. I will post another update when the GitHub tests complete.

tomeichlersmith:
With `inclusive` now passing after no changes (must have just been some weird provider stuff) and `it_pileup` failing on the same four plots as last time, I am merging this in.

PR #1076, opened 2022-08-09
Iss1048 detector id bindings

### LINKED ISSUE #1048: Python bindings for DetectorIDs
We discussed getting access to the various DetectorID classes from python with Boost.Python. I've been trying things out and I think the following is what would be needed 

- Updating the container to compile Boost.Python properly https://github.com/LDMX-Software/docker/issues/41
- Adding some `find_package` components in the CMakeLists.txt for DetDescr 
- Add a source + header file that defines the bindings that we want
- Add `${LDMX_SW_INSTALL}/lib` to the `${PYTHONPATH} environment variable 
In principle, you could actually just use a source file but that looks a bit weird. 

Doing so means that the `libDetDescr.so` shared object becomes consumable by python and you can import it with 
```python
import libDetDescr
# or 
from libDetDescr import HcalD # etc
``` 

The name of the python module has to be the same as the name of the shared object. If you wanted things to be a bit neater, you could add a separate library just for this but I don't think its worth it. 

For the various ID classes, my approach has been the following
- Export any base classes as objects only. This is the minimum required to use the derived classes. Is there a reason to export any functionality for e.g. `DetectorID`? If so, we could its just some additional boilerplate 
- For the derived classes export
  - The constructor from N pieces 
  - The constructor from a raw integer 
  - Accessor functions for the fields (in snake_case seems more pythonesque than the get* options) 
  - Any helper functions (e.g. `isNegativeEnd` for `HcalDigiID` or `idFromIndex` for `HcalElectronicsID`) 

I haven't been able to get the `__str__` function to work from the `<<` operators that we have defined. It produces a silly amount of template errors.   
  
Thoughts on this approach? I'm mostly familiar with the Hcal-related IDs so if there is something special you would want to have for the others, let me know. 

Here's an example of what the boilerplate looks like
https://github.com/LDMX-Software/ldmx-sw/blob/iss1048-detectorID-bindings/DetDescr/src/DetDescr/DetectorIDBindings.cxx

EinarElen:

This PR enables loading almost all of the DetectorID functionality from out C++ land in python-based analysis code using Boost.Python. In analysis code you can run 
```python
import libDetDescr 
``` 
to import the various DetectorIDs in one big module. As with any python module, you can also selectively import particular parts that you care about. The various IDs can be constructed either from raw ID values (like what people would find in their events) or through the corresponding pieces (e.g. section, layer, strip for an HcalID). 

The bulk of the update here is in the CMakeLists for the DetDescr directory which needs to check for the boost python libraries. And in one new source file called  DetDescr/src/DetDescr/DetectorIDBindings.cxx which needs to explicitly define everything that we make visible from the python module. 

Questions 

- I've defaulted the CMake option to enable the detectorID bindings to off since they won't compile if someone is trying to use an older version of the container that doesn't have the patch to build Boost.Python in https://github.com/LDMX-Software/docker/pull/46. I'm not entirely sure that's the right approach. 
- If we're adding this, we should probably document how to use it somewhere. I'm not sure where it would fit in best 

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1048 
## Check List
We have at least one student who has been actively using this in their analysis. 

N/A


DetDescr/include/DetDescr/DetectorIDBindings.h
```diff
@@ -0,0 +1,22 @@
...
+#include "DetDescr/EcalID.h"
+#include "DetDescr/EcalTriggerID.h"
+#include "DetDescr/HcalAbstractID.h"
+#include "DetDescr/HcalDigiID.h"
+#include "DetDescr/HcalElectronicsID.h"
+#include "DetDescr/HcalID.h"
+#include "DetDescr/HcalTriggerID.h"
+#include "DetDescr/SimSpecialID.h"
+#include "DetDescr/TrackerID.h"
+#include "DetDescr/TrigScintID.h"
```
tomeichlersmith: Is the header necessary? I feel like it would be simpler to just move these lines to the top of the source file (inside the `#if` check).

tomeichlersmith (commented):
Besides my comment about potentially removing the ID bindings header, I think this is fantastic. Thank you for documenting it thoroughly as well!

As far as your questions go:
- I think the way you have it setup is good right now. I agree that I don't want to break people's current set ups so making it default-OFF means people have to know they are changing their setup to use them.
- Put a README in DetDescr! I think that is the best option right now and will (hopefully) have a better chance of getting updated whenever the IDs themselves need to be updated.

DetDescr/include/DetDescr/DetectorIDBindings.h
EinarElen: Not really, no. It was more there for symmetry reasons than necessity.

EinarElen:
While I'm updating the, documentation + removing the header, I noticed that we are generating this Boost deprectation warning warning (I don't actually think this comes from our software, but from the header file). I know that this warning shows up elsewhere in ldmx-sw, but is there something useful to do about it? 

```
In file included from /usr/local/include/boost/python/exception_translator.hpp:10:0,
                 from /usr/local/include/boost/python.hpp:28,
                 from /home/einarelen/ldmx/ldmx-sw/DetDescr/src/DetDescr/DetectorIDBindings.cxx:2:
/usr/local/include/boost/bind.hpp:41:1: note: #pragma message: The practice of declaring the Bind placeholders (_1, _2, ...) in the global namespace is deprecated. Please use <boost/bind/bind.hpp> + using namespace boost::placeholders, or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior.
 )
```

tomeichlersmith:
We could upgrade the Boost version which [supposedly resolves this issue](https://github.com/boostorg/python/issues/359), but besides that I don't think there is anything else we could do.

EinarElen:
I also found a rather annoying bug when working with the setup now. It seems like the PyROOT machinery doesn't know what to do if you compiled `libDetDescr` with sanitizer support enabled, which crashes the Python session. I'll document this for now, since I don't think this is a priority thing to fix (most users won't be running with sanitizer support enabled).

EinarElen:
I've now updated this with some documentation. While I did that, I caught a small bug in the definition of `HcalID` which prevented you from creating one with named arguments and realized that we could document the "self" parameter which makes the built-in Python help much less weird to read.

tomeichlersmith (approved):

Issue #1085, opened 2022-09-08
trigger OR processor, needed for multielectron 


bryngemark:
**Is your feature request related to a problem? Please describe.**
I'd like to be able to OR (or AND) triggers and then trigger skim on that, for future trigger efficiency studies. 

**Describe the solution you'd like**
I think the easiest is to make a new processor, which takes a configurable list of trigger collections, a configurable logic (doOR, doAND), and given that, checks whether the event passes the trigger and satisfies the desired logic, and sets the storage hint for the event to keep or not. Then this processor is what is used to set the skim decision, not the individual ones. The individual trigger decision can still be seen in each trigger collection (`pass_`).

**Describe alternatives you've considered**
I've been thinking that the storage hint/skimming rules should change but I think it's a larger rework than I'm happy to do. A processor seems versatile enough for me. 

This should absolutely go in before the next production release so we can get going with trigger skimming the multielectron samples we already have.

mrsolt:
I support this. I was also wondering how to get a skim for the missing energy visible A' trigger for the large scale production, and I think this could be useful. The trigger logic for this visibles trigger is already in the software, and we validated it. You just select the front portion of the Ecal to sum the energy for a missing energy trigger, instead of the entire Ecal.

Unfortunately, the standard missing energy trigger is a subset of what would be the visible trigger skim, so these can't be derived directly from the standard trigger. Thus, some sort of multi-trigger framework would be very useful.

bryngemark:
Great! Yes my thought would be that this processor is just checking which of the individual specified triggers pass the event and then selects what to do based on OR/AND settings, so the actual specific trigger logic is kept separate. (Clarifying the idea here for @rodwyer100 who has volunteered to actually code it up.)

rodwyer100:
[recoSkim3.txt](https://github.com/LDMX-Software/ldmx-sw/files/9749340/recoSkim3.txt)
[recoSkim2.txt](https://github.com/LDMX-Software/ldmx-sw/files/9749341/recoSkim2.txt)

rodwyer100:
Adding validation code in comment as per LK's desires.

rodwyer100:
Here is the image generated with the other validation architecture as requested in LK's last message

![image](https://user-images.githubusercontent.com/19932569/195455249-116d6960-cb03-4df7-b7a3-e238b3059fd4.png)

These are the scripts used to run the code:

[...]

PR #1079, opened 2022-08-18
ECal Portion of DR Geometry Update

tomeichlersmith:
I am updating _ldmx-sw_, here are the details.

**This is a draft PR because I have not tested the GDML updates in simulation; however, I think I've implemented the complicated C++ side of things inside of `EcalGeometry` (formerly `EcalHexReadout`) and the visualization of the GDML detector looks correct.**

### What are the issues that this addresses?
- resolves #904
- resolves #923 
- resolves #890 

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments
- [x] I ran my developments and the following shows that they are successful.
- [x] I attached any sub-module related changes to this PR.

### Related Sub-Module PRs
- https://github.com/LDMX-Software/Ecal/pull/41
- https://github.com/LDMX-Software/SimCore/pull/71

PR #1090, opened 2022-09-23
Origin/test beam n tuppler

rodwyer100:

The pull request addresses issue #1085. 

The processor SequentialTrigger takes in mask list and a list of trigger collections and pass names, and runs through each event seeing if the combination of triggers from each collection lies in the mask list. For our purpose we have the AND mask list hard coded as a default, and in the two validation scripts (recoSkim2.py and recoSkim3.py) we have another OR mask list which can be optionally set.

## Check List

In recoSkim2.py I set the pass energy of a single electron in one trigger to 2750 and the other to 3750. Note 1 corresponds to pass, and we pass if the Ecal sum is less than the cut, so 3750 is always less harsh. This first lower threshold pass is blue (so harsher), the former red. We create a hard coded collection which is not just the keep Hint to test if this works, and for AND and or it is shown in black below.

passstuff

ORMask

image

ANDMask

image

The legend entry on this second entry is consistent with the logic of these skims (both are consistent, in the first image the legend corresponds to the latter I plotted).

Here are the Ecal Sums for the Two Triggers (GetSum()):

Trigger 1

image

Trigger 2

image

Here is the AND

image

and the OR

image

You can check the titles to see exactly how these were plotted w/ TBrowser commands.

- [X] I attached any sub-module related changes to this PR.

My modifications are on my own branch, but should not require a submodule PR because I think Recon inherits its branches from ldmx-sw proper (unlike TrigScint). Someone please educate me if that is not the case.

### DIFF
--- a/Recon/include/Recon/SequentialTrigger.h
+++ b/Recon/include/Recon/SequentialTrigger.h
@@ -1,0 +1,85 @@
+/**
+ * @file TriggerProcessor.h
+ * @brief Class that provides a trigger decision for recon using a TriggerResult
+ * object
+ * @author Josh Hiltbrand, University of Minnesota
+ * @author Lene Kristian Bryngemark, Stanford University
+ */
+
+#ifndef RECON_TRIGGER_SEQUENTIALTRIGGER_H_
+#define RECON_TRIGGER_SEQUENTIALTRIGGER_H_
+
+// LDMX
+#include "DetDescr/EcalHexReadout.h"
+#include "Ecal/Event/EcalHit.h"
+#include "Event/TriggerResult.h"
+#include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
+
+namespace recon {
+
+/**
+ * @class TriggerProcessor
+ * @brief Provides a trigger decision for recon using a TriggerResult object.
+ *
+ * @note
+ * TriggerProcessor takes in a set of parameters to be used in defining
+ * the trigger algorithm. An event is passed to the processor and the relevant
+ * algorithms are then run on the event (ECAL layer sum). A trigger decision is
+ * executed and the decision along with the algorithm name and relevant
+ * variables are stored in a TriggerResult object which is added to the
+ * collection.
+ */
+class SequentialTrigger : public framework::Producer {
+ public:
+  /**
+   * Class constructor.
+   */
+  SequentialTrigger(const std::string& name, framework::Process& process)
+      : framework::Producer(name, process) {}
+
+  /**
+   * Class destructor.
+   */
+  virtual ~SequentialTrigger() {}
+
+  /**
+   * Configure the processor using the given user specified parameters.
+   *
+   * @param parameters Set of parameters used to configure this processor.
+   */
+  void configure(framework::config::Parameters& parameters) final override;
+
+  /**
+   * Run the trigger algorithm and create a TriggerResult
+   * object to contain info about the trigger decision
+   * such as pass/fail, number of saved variables,
+   * etc.
+   * param event The event to run trigger algorithm on.
+   */
+  virtual void produce(framework::Event& event);
+
+ private:  
+  /** The name of the input collection of triggers */
+  std::vector<std::string> trigger_list_;
+
+  /**pass name of the triggers, should be uniform **/
+  std::vector<std::string> trigger_passNames_;
+
+  /** integer list corresponding to decimal values of binary masks which lead
+   * to a pass **/
+  std::vector<int> pass_masks_;
+
+  /** The name of hte input collection (events with all triggers) **/
+  std::string inputColl_;
+
+  /** The pass name of the input (the Ecal hits). */
+  std::string inputPass_;
+
+  /** The name of the output collection (the trigger decision). */
+  std::string outputColl_;
+};
+
+}  // namespace recon
+
+#endif
--- a/Recon/python/simpleSeqTrigger.py
+++ b/Recon/python/simpleSeqTrigger.py
@@ -1,0 +1,36 @@
+"""Configuration for TriggerProcessor
+
+Sets all parameters to reasonable defaults.
+
+Attributes:
+------------- 
+trigger_list
+    the collection of each trigger we perform sequential skimming with.
+
+trigger_passName
+    our input pass names for the multiple triggers. Should be the same for all
+
+pass_mask : int array
+    Collection of decimal converted binary masks which induce a pass
+
+seqtrigger_collection : string
+    Name of the output collection containing the trigger result
+
+"""
+
+from LDMX.Framework import ldmxcfg
+
+class SequentialTrigger(ldmxcfg.Producer) :
+    """Configuration for the (multi-electron aware but simple) sequential trigger on the ECal reco hits"""
+
+    def __init__(self,name) :
+        super().__init__(name,'recon::SequentialTrigger','Recon')
+
+        self.trigger_list = ["Trigger"]
+        self.trigger_passNames = ["reconSeq","reconSeq"] 
+        ANDList = [sum([2**i for i in range(len(self.trigger_list))])]
+        self.pass_mask = ANDList #[2,1];# This is OR
+        self.seqtrigger_collection = "SeqTrigger"
+
+simpleSeqTrigger = SequentialTrigger("simpleSeqTrigger")
+
--- a/Recon/src/Recon/SequentialTrigger.cxx
+++ b/Recon/src/Recon/SequentialTrigger.cxx
@@ -1,0 +1,39 @@
+
+#include "Recon/SequentialTrigger.h"
+#include "Recon/Event/TriggerResult.h"
+
+namespace recon {
+
+void SequentialTrigger::configure(framework::config::Parameters &ps) {
+  trigger_list_ = ps.getParameter<std::vector<std::string>>("trigger_list");
+  trigger_passNames_ = ps.getParameter<std::vector<std::string>>("trigger_passNames");
+  pass_masks_ = ps.getParameter<std::vector<int>>("pass_mask");
+  outputColl_ = ps.getParameter<std::string>("seqtrigger_collection");
+  return;
+}
+
+void SequentialTrigger::produce(framework::Event& event) {
+  /** Grab the Ecal hit collection for the given event */
+  bool hasPassed = false;
+  int  maskValue = 0; 
+  for (int i = 0; i<trigger_list_.size(); i++){
+    auto trigResult{event.getObject<ldmx::TriggerResult>(trigger_list_[i],
+                                                  trigger_passNames_[i])};
+    if(trigResult.passed()){
+        maskValue+=(int)pow(2,i);
+    }  
+  }  
+  for(int m : pass_masks_){
+    if(maskValue==m){hasPassed=true;}
+  }
+  event.add(outputColl_, hasPassed);
+  // mark the event
+  if (hasPassed)
+    setStorageHint(framework::hint_shouldKeep);
+  else
+    setStorageHint(framework::hint_shouldDrop);
+}
+
+}  // namespace dqm
+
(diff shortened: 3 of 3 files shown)


Recon/include/Recon/SequentialTrigger.h
```diff
@@ -0,0 +1,85 @@
...
+ * object
+ * @author Josh Hiltbrand, University of Minnesota
+ * @author Lene Kristian Bryngemark, Stanford University
+ */
+
+#ifndef RECON_TRIGGER_SEQUENTIALTRIGGER_H_
+#define RECON_TRIGGER_SEQUENTIALTRIGGER_H_
+
+// LDMX
+#include "DetDescr/EcalHexReadout.h"
```
bryngemark, line 13: It looks like this include causes a compilation error with trunk. I think this way of accessing the ecal readout is outdated. Probably, you have been compiling against a not fully updated trunk? In either case, I don't think it is needed for your processor. Same with EcalHit. I suggest removing them and pushing again.

Recon/python/simpleSeqTrigger.py
```diff
@@ -0,0 +1,36 @@
...
+
+Sets all parameters to reasonable defaults.
+
+Attributes:
+------------- 
+trigger_list
+    the collection of each trigger we perform sequential skimming with.
+
+trigger_passName
+    our input pass names for the multiple triggers. Should be the same for all
```
bryngemark, line 11: Doesn't need to be the same for all, in fact you might want to trigger skim on an OR between an existing trigger decision and one you just created and ran in the current pass. So your code should handle that. Note that if the pass name is empty, the framework will attempt to use any unique collection matching the specified collection name. If there are several with different pass names, it throws an exception, forcing you to make an explicit choice.

Recon/python/simpleSeqTrigger.py
```diff
@@ -0,0 +1,36 @@
...
+from LDMX.Framework import ldmxcfg
+
+class SequentialTrigger(ldmxcfg.Producer) :
+    """Configuration for the (multi-electron aware but simple) sequential trigger on the ECal reco hits"""
+
+    def __init__(self,name) :
+        super().__init__(name,'recon::SequentialTrigger','Recon')
+
+        self.trigger_list = ["Trigger"]
+        self.trigger_passNames = ["reconSeq","reconSeq"]
```
bryngemark, line 30: I think the collection names list and the pass name list should be the same length also in your default config. I'd use "" for path names in the default though, to have the baseline behavior of picking the default collection name ("Trigger" in this case), regardless of pass name.

Recon/include/Recon/SequentialTrigger.h
```diff
@@ -0,0 +1,85 @@
...
+   * etc.
+   * param event The event to run trigger algorithm on.
+   */
+  virtual void produce(framework::Event& event);
+
+ private:  
+  /** The name of the input collection of triggers */
+  std::vector<std::string> trigger_list_;
+
+  /**pass name of the triggers, should be uniform **/
```
bryngemark, line 66: Here too, remove the requirement that all are the same pass.
Also I'm not sure if there needs to be a space after `/**` for this to look like a comment to Doxygen but adding one wouldn't hurt. And end comments with `*/` (so a single asterisk).

Recon/include/Recon/SequentialTrigger.h
```diff
@@ -0,0 +1,85 @@
...
+  std::vector<int> pass_masks_;
+
+  /** The name of hte input collection (events with all triggers) **/
+  std::string inputColl_;
+
+  /** The pass name of the input (the Ecal hits). */
+  std::string inputPass_;
+
+  /** The name of the output collection (the trigger decision). */
+  std::string outputColl_;
```
bryngemark, line 80: I don't think you're using any of these?

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,39 @@
...
+void SequentialTrigger::configure(framework::config::Parameters &ps) {
+  trigger_list_ = ps.getParameter<std::vector<std::string>>("trigger_list");
+  trigger_passNames_ = ps.getParameter<std::vector<std::string>>("trigger_passNames");
+  pass_masks_ = ps.getParameter<std::vector<int>>("pass_mask");
+  outputColl_ = ps.getParameter<std::string>("seqtrigger_collection");
+  return;
+}
+
+void SequentialTrigger::produce(framework::Event& event) {
+  /** Grab the Ecal hit collection for the given event */
```
bryngemark, line 16: Fix comment to not mention Ecal

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,39 @@
...
+#include "Recon/SequentialTrigger.h"
+#include "Recon/Event/TriggerResult.h"
+
+namespace recon {
+
+void SequentialTrigger::configure(framework::config::Parameters &ps) {
+  trigger_list_ = ps.getParameter<std::vector<std::string>>("trigger_list");
+  trigger_passNames_ = ps.getParameter<std::vector<std::string>>("trigger_passNames");
+  pass_masks_ = ps.getParameter<std::vector<int>>("pass_mask");
+  outputColl_ = ps.getParameter<std::string>("seqtrigger_collection");
```
bryngemark, line 11: Don't think you should be using this

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,39 @@
...
+    auto trigResult{event.getObject<ldmx::TriggerResult>(trigger_list_[i],
+                                                  trigger_passNames_[i])};
+    if(trigResult.passed()){
+        maskValue+=(int)pow(2,i);
+    }  
+  }  
+  for(int m : pass_masks_){
+    if(maskValue==m){hasPassed=true;}
+  }
+  event.add(outputColl_, hasPassed);
```
bryngemark, line 29: I don't think you should have an output collection here. This is just a "skimming" code which tells us whether or not to keep the event (with all its collections). Setting the storage hint is all that is needed.

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,39 @@
...
+  int  maskValue = 0; 
+  for (int i = 0; i<trigger_list_.size(); i++){
+    auto trigResult{event.getObject<ldmx::TriggerResult>(trigger_list_[i],
+                                                  trigger_passNames_[i])};
+    if(trigResult.passed()){
+        maskValue+=(int)pow(2,i);
+    }  
+  }  
+  for(int m : pass_masks_){
+    if(maskValue==m){hasPassed=true;}
```
bryngemark, line 27: In principle you could break here after even one mask value has been satisfied.

tomeichlersmith:
The compilation for the PR is failing because it is merging your branch with `trunk` and then attempting to compile.

```
   In file included from /home/runner/work/ldmx-sw/ldmx-sw/Recon/src/Recon/SequentialTrigger.cxx:2:0:
  /home/runner/work/ldmx-sw/ldmx-sw/Recon/include/Recon/SequentialTrigger.h:13:10: fatal error: DetDescr/EcalHexReadout.h: No such file or directory
   #include "DetDescr/EcalHexReadout.h"
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~
  compilation terminated.
  make[2]: *** [Recon/CMakeFiles/Recon.dir/src/Recon/SequentialTrigger.cxx.o] Error 1
  Recon/CMakeFiles/Recon.dir/build.make:117: recipe for target 'Recon/CMakeFiles/Recon.dir/src/Recon/SequentialTrigger.cxx.o' failed
  make[1]: *** [Recon/CMakeFiles/Recon.dir/all] Error 2
  CMakeFiles/Makefile2:1035: recipe for target 'Recon/CMakeFiles/Recon.dir/all' failed
  make: *** [all] Error 2
  Makefile:145: recipe for target 'all' failed
```

In #1079 I renamed EcalHexReadout to EcalGeometry. It doesn't look like you need that class, so you can just remove that header.

tomeichlersmith:
> My modifications are on my own branch, but should not require a submodule PR because I think Recon inherits its branches from ldmx-sw proper (unlike TrigScint). Someone please educate me if that is not the case.

It does look like you _don't_ need the TrigScint module PR; however, your branch _does_ have a different commit of TrigScint than trunk so if we merged it as-is we would be changing TrigScint. Here's what you should do:

1. Make sure your local copy of ldmx-sw has an updated `trunk`
```
git fetch origin
git checkout trunk
git pull
git checkout origin/TestBeamNTuppler
```
2. Rebase your branch onto the updated trunk
    - The commit cc0d5ae45fe33b6a508b911deba52c05ae4d6655 is the one that introduced the TrigScint changes, so you can remove that commit from your rebase.
```
git rebase --interactive
# an editor will pop up, remove the "First commit" commit line so that it will be deleted, save and exit the editor
git submodule update
```
3. Double check you didn't lose anything 
4. Force-push your updates here `git push --force-with-lease`

bryngemark (changes requested):
First of all, I don't think any of the updates to TrigScint should be merged in with this pull request. But that's mostly a note to the merging person. 

Second, I like that you added plots and the event count makes sense, but I'm not sure I understand what the energy plots are showing. The distributions look identical to me. Could you overlay the energy sums in a file where you used an OR on a file where you used an AND, so we could see that the different energy sums in the trigger actually come through the expected way?

Also it would be really nice if you could add the config you used to test your code (you're referring to it but I don't see it). 

About your mask idea. I see what you're doing and I see how it is versatile. I'm not sure it's the handiest solution here though. If you look at your processor, you're looping over a list and then a list of mask values: two loops per event. Even with my suggestion to break once anything satisfies the mask, it's still a second loop. 
If you did a simpler solution with only doOR or doAND, you could do it like this (in pseudo code):
```
hasPassed=doAND (or !doOr)
in loop: 
    if doOR, if trigResult.passed() --> hasPassed=true; break
    else if doAND, if !trigResult.passed() --> hasPassed=false; break
```
and then set storage hints accordingly. 

You'd need a check that `doOR!=doAND` at config time. And, you need to deal with the case that the collection you're trying to access is not there in the event -- then what should happen? But this you should deal with in any case. 

What you gain with your mask is in principle an XOR and a NAND option. I don't see a use case for this but maybe your other reviewers do. I'm open to being convinced!

Recon/src/Recon/SequentialTrigger.cxx
rodwyer100: I was using this as an easier way from me to validate that the processor was working. I forgot to comment it out when I performed this request.

Recon/src/Recon/SequentialTrigger.cxx
rodwyer100: This will speed up OR dramatically. Thanks!

rodwyer100: Also doing it for AND it easy too.

Recon/src/Recon/SequentialTrigger.cxx
rodwyer100: Done

Recon/include/Recon/SequentialTrigger.h
rodwyer100: Good catch. Thanks.

Recon/python/simpleSeqTrigger.py
rodwyer100: It should handle it; that line was a relic of an old way I ran it. I shall include proof in my validation code (in the next push request) of two different passNames being handled correctly

Recon/python/simpleSeqTrigger.py
rodwyer100: Both items implemented.

Recon/src/Recon/SequentialTrigger.cxx
rodwyer100: Fixed

PR #1092, opened 2022-09-26
HCal Portion of DR Geometry Update

### LINKED ISSUE #1074: Design Report Geometry Update
The Design Report (DR) is a good chance for LDMX to rerun our large simulation studies with an updated geometry that is more realistic and follows the changes that have been made within those subsystems due to their own detailed studies and engineering developments. This issue is a tracking issue focused on keeping the organization of this large geometry update together.

**The DR geometry will be labeled v14 so that it can easily be distinguished from prior geometries.** Even if a component of LDMX is unchanged, it should be copied to v14 to avoid confusion. Generally, I am interested in gathering an up-to-date description of the subsystem geometries so that we know what needs to be changed. Hopefully, we can include some _rudimentary_ representations of support structures as well.

**v3.1.12 of ldmx-sw has a first copy of a v14 detector from which you should develop.**

[Indico Event with Subsystem Slides](https://indico.fnal.gov/event/55682/)

## Tracker
**Branch**: `iss1077`

- [x] remove carbon fiber copied over from HPS design
- [x] carve out more space for TS to accommodate its support/readout system (4.5mm -> ~10mm)

### Related Issues
- #1077 

## Trigger Scintillator
**Branch**: `TS_geom_edits`

- [ ] expand box holding layer-pairs (aka modules) from 4.5mm to ~10mm
- [ ] more refined LYSO target for further active-target studies
- [x] move module downstream of target to upstream of tagger tracker (try to keep within magnet bore)
- [ ] ? increase length of Scint bars to avoid shadowing recoil tracker from out-of-beam-spot tracks

### Related Issues
- #1087 

## Ecal
**Branch**: `iss1074-ecal` merged with #1079 

The ECal changes will require updates to not only the GDML but also the EcalHexReadout class (probably will be renamed to EcalGeometry) and the places that use that class. The issues below go through this To-Do list in more detail, not all of these items may make it into the DR geometry since they are so complicated.

- [x] Introduce inter-layer offsets
- [x] Update tungsten layer shape and thicknesses
- [x] Update carbon layer shape and thickness (4mm -> 5.7mm?)
- [x] Increase Motherboard spacing to 3.5mm from 2.3mm to accommodate change in design from CMS
- [x] Glue thickness to 0.1mm instead of 0.05mm
- [ ] Cell ID flips within each bi-layer due to method of construction of layer sandwich
- [ ] Mouse-bites on corners of hexagons (loss of sensitive material to leave room for mounting
- [x] Enlarge cavern according to updated engineering design
- [x] Introduce metal of support box and strong-backs of bi-layers
- [x] Motherboards are not large rectangles and instead complicated polygons
- [x] 5 ROCs per layer that overlap slightly with motherboards
- [x] #1032 
- [x] #1020 
- [x] Depth of ECal along z fixed to 600mm defined by box

### Related Issues
- #890 
- #923 
- #904 

## Hcal
**Branch**: `iss1074-hcal`

Listed roughly in order of importance.

- [x] Back HCal updated to 2x2m
- [x] Side HCal updated to one of the radial designs
- [ ] (more testbeam configurations?)
- [x] Split scintillator slabs into individual bars
  - necesary for Ralf's scintilator simulation
  - requires updating CopyNumber scheme and patching SimCore
  - should check sim run time to make sure it isn't substantially slowing the sim
- [ ] Including TiO2 coating? optical fiber? 
- [ ] Ralf's optical photon simulation-derived LUT (in digis)
- [ ] Updates to geometry condition to accommodate shifted bars in testbeam
- [ ] Clean up buggy TOA->time calculation in reco

### Related Issues
- https://github.com/LDMX-Software/ldmx-sw/issues/1349
- https://github.com/LDMX-Software/ldmx-sw/issues/1347
- https://github.com/LDMX-Software/ldmx-sw/issues/1348

## Other
- [ ] Updated drawing of magnet and re-exported to GDML via FreeCAD
- [ ] Recalculate magnetic field map and expand to make sure all of recoil tracker is covered

cmantill:

Updating Hcal Geometry to circular side-Hcal geometry. 
Note that this geometry assumes that:
- **odd** layers have scintillators oriented in z.
- **even** layers have scintillators oriented in x/y.

All the scintillator bars in this geometry have an individual volume.

- Resolves #1074 for Hcal

## Check List
- [x] I attached any sub-module (SimCore) related changes to this PR.
- [x] I introduced patch on G4 material definition.
- [x] I checked that copy numbers are not duplicated between scintillators.
- [x] I checked that there no overlaps between internal Hcal volumes.
- [x] I checked that there no overlaps between detector volumes.
- [x] I visualized the geometry.
- [x] I updated HcalGeometry python and c++ code to adapt to circular side hcal geometry.
- [x] I ran a G4 geantinos simulation across all Hcal.

### Related Sub-Module PRs
- https://github.com/LDMX-Software/SimCore/pull/73
- https://github.com/LDMX-Software/ldmx-sw/pull/1092


cmantill:
Screen Shot 2022-10-03 at 5 54 56 PM
Screen Shot 2022-10-03 at 5 55 34 PM
Screen Shot 2022-10-03 at 5 56 35 PM
Screen Shot 2022-10-03 at 5 57 46 PM

cmantill:
Used the following file to visualize side Hcal:
```
# Use this open statement to create an OpenGL view:
/vis/open OGL #900x600-0+0

# Disable auto refresh and quieten vis messages whilst scene and
# trajectories are established:
/vis/viewer/set/autoRefresh false
/vis/list all
/vis/verbose errors

#/vis/viewer/set/background white

# Draw geometry:
/vis/drawVolume
/vis/viewer/flush

# Specify view angle:
/vis/viewer/set/viewpointThetaPhi 140 160

# Specify zoom value:
/vis/viewer/zoom 3

# Specify style (surface or wireframe):
#/vis/viewer/set/style surface
/vis/viewer/set/style wireframe

# Draw coordinate axes:
/vis/scene/add/axes 0 0 0 -1 m

#/vis/geometry/set/colour back_hcal_absoVolume 0 green
#/vis/geometry/set/colour back_hcal_scintYVolume 0 blue
#/vis/geometry/set/colour back_hcal_scintXVolume 0 red

/vis/geometry/set/colour side_hcal_scintZYVolume 0 red
/vis/geometry/set/colour side_hcal_scintZXVolume 0 blue
/vis/geometry/set/colour side_hcal_absoX_0Volume 0 green
/vis/geometry/set/colour side_hcal_absoX_1Volume 0 green
/vis/geometry/set/colour side_hcal_absoX_2Volume 0 green
/vis/geometry/set/colour side_hcal_absoX_3Volume 0 green
/vis/geometry/set/colour side_hcal_absoY_0Volume 0 cyan
/vis/geometry/set/colour side_hcal_absoY_1Volume 0 cyan
/vis/geometry/set/colour side_hcal_absoY_2Volume 0 cyan
/vis/geometry/set/colour side_hcal_absoY_3Volume 0 cyan
/vis/geometry/set/colour side_hcal_scintY_0Volume 0 magenta
/vis/geometry/set/colour side_hcal_scintY_1Volume 0 magenta
/vis/geometry/set/colour side_hcal_scintY_2Volume 0 magenta
/vis/geometry/set/colour side_hcal_scintY_3Volume 0 magenta
/vis/geometry/set/colour side_hcal_scintX_0Volume 0 gray
/vis/geometry/set/colour side_hcal_scintX_1Volume 0 gray
/vis/geometry/set/colour side_hcal_scintX_2Volume 0 gray
/vis/geometry/set/colour side_hcal_scintX_3Volume 0 gray

/vis/viewer/set/autoRefresh true
/vis/verbose warnings
```

cmantill:
Tested with: 
```from LDMX.Framework import ldmxcfg

# Create a process
p = ldmxcfg.Process( 'test_hcal_geometry' )

# Set the maximum number of events
p.maxEvents = 10000

# Set the run number
p.run = 0

# Set the output file name
p.outputFiles = ['hcal_geometry_test.root']

# The histogram file name
p.histogramFile = 'hcal_geometry_test_histo.root'

# Generate muons

import LDMX.Hcal.HcalGeometry
import LDMX.Ecal.EcalGeometry

sim = simulator.simulator("single_muons")
sim.setDetector( 'ldmx-det-v14' )
sim.description = "HCal muon"
sim.beamSpotSmear = [20., 80., 0.] #mm

# shoot muon with a general Particle source
gpsCmds=[ "/gps/particle mu-",
          "/gps/pos/type Plane",
          "/gps/direction 0 0 1",
          "/gps/ene/mono 4 GeV", # fixed energy
          "/gps/pos/shape Square",
          "/gps/pos/centre 0 0 220. mm", # start at side hcal
          "/gps/pos/halfx 1500 mm",
          "/gps/pos/halfy 1500 mm",
          "/gps/number 1"
          ]
myGPS = generators.gps( 'myGPS' , gpsCmds )
sim.generators.append(myGPS)

from LDMX.Hcal import digi
from LDMX.Hcal import hcal_hardcoded_conditions
digis = digi.HcalDigiProducer()

p.sequence = [
    sim,
    digis
]
```

Screen Shot 2022-10-06 at 12 43 08 PM
Screen Shot 2022-10-06 at 12 43 31 PM
Screen Shot 2022-10-06 at 12 43 58 PM

tomeichlersmith (changes requested):
Looks good from my end, the conflict is just a difference in whitespace. Not sure how to resolve that easily.

I do have a specific request. Please make a release of Hcal so that the version of Hcal on ldmx-sw trunk is pointing to a specific version of Hcal. I have done that for Framework and SimCore already. After that, you can update this branch to point to those refs.

```
cd Hcal
git fetch origin
git checkout v0.3.0
cd ../Framework
git fetch origin
git checkout v1.1.0
cd ../SimCore
git fetch origin
git checkout v1.2.0
cd ..
git add Hcal Framework SimCore
git commit -m "put submodules onto version tags

Hcal v0.3.0
Framework v1.1.0
SimCore v1.2.0"
```

cmantill:
BTW @EinarElen - I think we need to update SimCore again since I updated the sensitive detector names in the branch that was merged.

tomeichlersmith (commented):
Looks good, I merged your patch to SimCore and bumped the patch version number, so I think all that's left is to update that reference.

tomeichlersmith (approved):

cmantill:
Thanks @tomeichlersmith. I caught some small changes to the strip numbering that were crashing the geometry test in `Hcal`. The geometry test does not check for great consistency on positions but it does verify that there are no duplicated copy numbers, that we can generate simhits and that we can access the positions of these simHits on the HcalGeometry map.

tomeichlersmith:
The Hcal Geometry test is failing due to a `Abort (abnormal termination) signal` which I don't think I've seen before. From [this issue](https://github.com/catchorg/Catch2/issues/1169) it makes me think that one of the checks or requires is failing and the framework is failing to unwind the stack properly.

EinarElen:
> The Hcal Geometry test is failing due to a `Abort (abnormal termination) signal` which I don't think I've seen before. From [this issue](https://github.com/catchorg/Catch2/issues/1169) it makes me think that one of the checks or requires is failing and the framework is failing to unwind the stack properly.

Hi everyone,

I checked this with GDB and there seems to be some issue with the GDML somewhere. The backtrace of the exception comes from within `simcore::geo::GDMLParser::read `. I'll see if I can narrow it down further.

Updates: 

1. This might be an issue with the scoring planes, the exception seems to go away if we skip 
```
sim.setDetector( 'ldmx-det-v14', True  )
```
Does give other errors though :) 
2. There seems to be a separate issue with the tagger/magnet/target/recoil gdml-files, which I'm not sure about. For now, I'm working with the tagger commented out... This then gives us `Evaluator : unknown variable` from before the exception which hopefully will narrow things down for us!
3. I'm running into one issue with the `magnet_gap_y/x` variables not being defined. Ok, this is from them being renamed to dx/dy in trunk. 
4. For the tagger/recoil issue, this seems to be caused by the auxiliary tags here https://github.com/LDMX-Software/ldmx-sw/commit/a4048f3737d608a2364b9f05090c5c2333876f44#diff-f2062242d1aa02d1d3228ec559a18f50dc4288ffe286b2bf4334ba28fdef9b27R95-R98 and here https://github.com/LDMX-Software/ldmx-sw/commit/a4048f3737d608a2364b9f05090c5c2333876f44#diff-f2062242d1aa02d1d3228ec559a18f50dc4288ffe286b2bf4334ba28fdef9b27R95-R99 being in the structure tag rather than the volume tag in trunk. Patching these two things will make the test run _if we skip the scoring planes_. Looking into them next... 
5. First issue in scoring_planes is that now we are redefining the magnet_gap_dx/y variables in there :) 
6. Second issue: Some variables like `recoil_layer_pos` have been removed but are still used in the scoring planes GDML. Removing the recoil portion of the scoring planes GDML lets the Hcal test run. Maybe this has been fixed in some other branch?
7. Next problem: The test runs, but fails
```
-------------------------------------------------------------------------------
Hcal Geometry test
-------------------------------------------------------------------------------
/home/einarelen/ldmx/ldmx-sw/Hcal/test/HcalGeometryTest.cxx:115
...............................................................................

/home/einarelen/ldmx/ldmx-sw/Hcal/test/HcalGeometryTest.cxx:42: FAILED:
  CHECK( simHits.size() > 0 )
with expansion:
  0 > 0

===============================================================================
test cases:    2 |    1 passed | 1 failed
assertions: 6012 | 6011 passed | 1 failed 
```
@cmantill, do you know if the python configuration with GPS is correct in the test repo? It doesn't look like any hits are actually produced in the Hcal, which could be if the GPS configuration is wrong.  

Ok, confirmed that it is an issue with the GPS configuration, switching it to something like a particle gun and the test produces simhits just fine!

EinarElen:
After patching/commenting out the parts that are unrelated to the Hcal subsystem, we still get a failing test for the Hcal
``` 
-------------------------------------------------------------------------------
Hcal Geometry test
-------------------------------------------------------------------------------
/home/einarelen/ldmx/ldmx-sw/Hcal/test/HcalGeometryTest.cxx:115
...............................................................................

/home/einarelen/ldmx/ldmx-sw/Hcal/test/HcalGeometryTest.cxx:42: FAILED:
  CHECK( simHits.size() > 0 )
with expansion:
  0 > 0

```

However, this is just caused by one of 10 events in here _not going through the hcal_,  so the check for any simHits in the hcal fails. I haven't really played with the GPS in Geant, so I'm not sure exactly how to guarantee that all the muons are hitting the hcal volume.

tomeichlersmith:
I believe the particles generated "outside" the Hcal are particles that are outside the volume of the back hcal are shoot straight through the absorber layer of the side hcal (or through one of the corners of the square now uncovered by the side hcal).

This can be resolved by shrinking the GPS position shape:
```python
          "/gps/pos/halfx 1500 mm",
          "/gps/pos/halfy 1500 mm",
```
Probably `1000 mm` or even less would do the trick.

Edit:
There is also this line which duplicates the smearing affect done by the GPS which can probably be removed:
```python
sim.beamSpotSmear = [20., 80., 0.] #mm
```

tomeichlersmith:
I've pushed two commits to trunk that resolve bugs 3, 4, and 5 from Einar's comment.

PR #1093, opened 2022-09-27
Issue #1085 recommit with clean trunk build

### LINKED ISSUE #1085: trigger OR processor, needed for multielectron

rodwyer100:
This is a pull request following up on a previous one for issue #1085. It now was built right off a clean trunk branch, and obtains the same plots as before. For the purposes of inspection, I will reinclude old plots. recoSkim2 and recoSkim3 are now included in this merge in Recon/python

## Old pull request plots:

In recoSkim2.py I set the pass energy of a single electron in one trigger to 2750 and the other to 3750. Note 1 corresponds to pass, and we pass if the Ecal sum is less than the cut, so 3750 is always less harsh. This first lower threshold pass is blue (so harsher), the former red. We create a hard coded collection which is not just the keep Hint to test if this works, and for AND and or it is shown in black below.

passstuff

ORMask

image

ANDMask

image

The legend entry on this second entry is consistent with the logic of these skims (both are consistent, in the first image the legend corresponds to the latter I plotted).

Here are the Ecal Sums for the Two Triggers (GetSum()):

Trigger 1

image

Trigger 2

image

Here is the AND

image

and the OR

image

You can check the titles to see exactly how these were plotted w/ TBrowser commands.

## New pull request plots:

LK wanted to see OR and AND overlayed. Here they are:

image

Blue is AND, Red OR

Essentially because one of these cuts is always more restrictive than the other, OR and AND will just choose one or the other cut.

## Check List
- [ X] I successfully compiled _ldmx-sw_ with my developments

- [X ] I attached any sub-module related changes to this PR.

Now I am more certain I don't need any submodule stuff.


tomeichlersmith:
Just so everyone is aware: The PR validations are failing because of prior changes introduced by #1079 ; these changes _are not_ causing them to fail.

rodwyer100:
Changes in response to LK's requests have been implemented. Will include plots with single electrons in next comment.

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -1,54 +1,64 @@
...
-  int  maskValue = 0; 
+  //Returns an error if some bad combination of doOR and doAND is enabled.
+  if (doAND_ and doOR_){
+    ldmx_log(fatal) << "Can't do both doAND and doOR. Exiting.";
+    return;
+  }
+  if (not(doAND_) and not(doOR_)){
+    ldmx_log(fatal) << "Must do one of doAND and doOR. Exiting.";
+    return;
+  }
```
bryngemark: if doOR == doAND...

bryngemark: but of course you get less granular error messages that way

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -1,54 +1,64 @@
...
-  if(not(doOR_)and not(doAND_)){
-    for(int m : pass_masks_){
-        if(maskValue==m){hasPassed=true;break;}
-    }
-  }
+  
   //Used to validate if code was working
-
+  if(doVAL_){
   event.add("validation", hasPassed);
```
bryngemark: Does this syntax work? I thought you needed to add a proper collection to the event bus.

bryngemark (commented):
looks good, looking forward to the validation plots too. 

could you copy one of `recoSkim2.py` or `...3.py` to a comment (formatted text) in the issue, and remove them from the repository (`git rm`), so we don't clutter the repo with example scripts but are still able to reproduce your validation?

Recon/src/Recon/SequentialTrigger.cxx
tomeichlersmith: Yes, this works. You can add anything as simple as a single bool up to a map of user classes to the event bus.

bryngemark: 🤯

Recon/src/Recon/SequentialTrigger.cxx
tomeichlersmith: also move these lines to the `configure` method - you could even simply throw an exception with `EXCEPTION_RAISE(<short-code>,<message>);` since it doesn't make sense to run this processor without choosing one of these options.

Recon/python/recoSkim3.py
```diff
@@ -0,0 +1,62 @@
...
+thisPassName="reconSeq" #overlay" 
+nElectrons=1
+p=ldmxcfg.Process( thisPassName )
+
+#import all processors
+
+# Ecal hardwired/geometry stuff
+from LDMX.Ecal import EcalGeometry
+import LDMX.Ecal.ecal_hardcoded_conditions
+egeom = EcalGeometry.EcalGeometryProvider.getInstance()
```
tomeichlersmith: If you don't need the geometry handle to change any of its paramters, this line is not needed and is just clutter.

```suggestion
```

Recon/python/recoSkim3.py
```diff
@@ -0,0 +1,62 @@
...
+
+#Hcal hardwired/geometry stuff                                                                                                                    
+from LDMX.Hcal import HcalGeometry
+import LDMX.Hcal.hcal_hardcoded_conditions
+hgeom = HcalGeometry.HcalGeometryProvider.getInstance()
```
tomeichlersmith: Same as EcalGeometry

```suggestion
```

Recon/python/recoSkim2.py
```diff
@@ -0,0 +1,88 @@
...
+simPassName="sim"
+nElectrons=1
+p=ldmxcfg.Process( thisPassName )
+
+#import all processors
+
+# Ecal hardwired/geometry stuff
+from LDMX.Ecal import EcalGeometry
+import LDMX.Ecal.ecal_hardcoded_conditions
+egeom = EcalGeometry.EcalGeometryProvider.getInstance()
```
tomeichlersmith:
```suggestion
```

Recon/python/recoSkim2.py
```diff
@@ -0,0 +1,88 @@
...
+
+#Hcal hardwired/geometry stuff                                                                                                                    
+from LDMX.Hcal import HcalGeometry
+import LDMX.Hcal.hcal_hardcoded_conditions
+hgeom = HcalGeometry.HcalGeometryProvider.getInstance()
```
tomeichlersmith:
```suggestion
```

tomeichlersmith (commented):
Besides the small patches both LK and I have suggested, I'd request that you update the documentation of this class. Some parts are well written while others seem to just be leftover from copying the structure of another class.

rodwyer100:
More plots as per LK's wishes. I have changed SequentialTrigger.cxx as per LK and Tom's request and placed the python validation plots into comments in the issue. The following plots are generated from

mc_v12-4GeV-1e-inclusive_run10725_t1636678621.root

We also implemented a cut on nTracks so that events that don't form tracks are removed. Here we have set the two trigger sum bounds as 4750 and 5550

image

I think this difference in the plot has to do with how layerEsum for TriggerProcessor and TriggerSums20Layers are defined. There are still sharp cut-offs about 750 MeV away from eachother.

Recon/include/Recon/SequentialTrigger.h
```diff
@@ -0,0 +1,81 @@
...
+ * object
+ * @author Rory O'Dwyer, Stanford University
+ * @author Lene Kristian Bryngemark, Stanford University
+ */
+
+#ifndef RECON_TRIGGER_SEQUENTIALTRIGGER_H_
+#define RECON_TRIGGER_SEQUENTIALTRIGGER_H_
+
+// LDMX
+#include "Ecal/Event/EcalHit.h"
```
bryngemark: remove this include

Recon/include/Recon/SequentialTrigger.h
```diff
@@ -0,0 +1,81 @@
...
+#include "Event/TriggerResult.h"
+#include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
+
+namespace recon {
+
+/**
+ * @class SeqeuntialTrigger
```
bryngemark: fix typo

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,71 @@
+#include "Recon/SequentialTrigger.h"
+#include "Recon/Event/TriggerResult.h"
+
+namespace recon {
+
+/**
+ *Instantiates the variables used in this processor. We take trigger_list_ and tigger_passNames_ to obtain trigger collections, a doOR_ and doAND_ option to check if one of or all events pass, and doVAL_ to output a validation collection
```
bryngemark: fix typo: tigger_passNames_ --> trigger_passNames_
I'm being really picky here because if people read the docs to get the parameters right, it has to be absolutely correct.

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,71 @@
...
+  for (int i = 0; i<trigger_list_.size(); i++){
+    //Returns an error is a trigger collection DNE
+    try{
+        auto trigResult{event.getObject<ldmx::TriggerResult>(trigger_list_[i],trigger_passNames_[i])};
+    
+    //Returns true should any trigger pass and doOR_ enabled, and sets keepTrack to true and breaks if doAND_ and any fail
+        if(trigResult.passed()){
+            if(doOR_){hasPassed=true;break;}
+        }else{
+            if(doAND_){keepTrack=true;break;}
```
bryngemark: why aren't you instead setting hasPassed=false here? is it because you initialize to false in both cases?

my suggestion would be to initialize to either doAND or !doOR (which should be equivalent by construction, so you choice). that way, if you run in AND mode, it is true until the first instance of a failed trigger decision. that setup warrants a descriptive comment though because just because it's simple doesn't mean it's something you grasp at first glance. i only advocate it because i think the extra keepTrack variable is also a little convoluted.

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,71 @@
...
+        }
+    }catch(...){
+        ldmx_log(fatal) << "Attemping to use non-existing trigger collection "
+                      << trigger_list_[i] << "_" << trigger_passNames_[i]
+                      << " to skim! Exiting.";    
+        return;
+    }
+  }
+  //In the event we have doAND_ set, this will run and only return true if all are true
+  if(doAND_){hasPassed=not(keepTrack);}
```
bryngemark: then you wouldn't need this line

bryngemark:
This is starting to be in really good shape! I'm puzzled by the energy sum (if this is events with one beam electron at 4 GeV, why is it so high??) but I'm inclined to let that go since the important part you're validating is that you can have two instances of the trigger processor and skim out the events that pass according to your selected logic. 

I also strongly suggest that you check that this works as intended in the case where you have one trigger collection in the input file, generate a second one on the fly, and use both as input to your trigger skimming processor. Similarly, try running once with two trigger processors, and use that file as input to a second reconstruction pass, where you only run the trigger skimming processor. Then you have verified that it works in the imaginable use cases.

rodwyer100:
I think the recoSkim2.py and recoSkim3.py is that second case you mention; I run once with two trigger processors in recoSkim2.py, and then separately in recoSkim3.py run again. I will attempt to do the first however.

rodwyer100:
I have verified the other requested validation architecture (generating a trigger, then a trigger on the fly with the skimmer) produces identical results. Review the comments on this issue for the validation scripts used to check that.

Recon/include/Recon/SequentialTrigger.h
rodwyer100: done

Recon/include/Recon/SequentialTrigger.h
rodwyer100: done

Recon/src/Recon/SequentialTrigger.cxx
rodwyer100: Done

Recon/src/Recon/SequentialTrigger.cxx
rodwyer100: Great idea. Done

Recon/src/Recon/SequentialTrigger.cxx
rodwyer100: Yep

Recon/python/recoSkim3.py
rodwyer100: Fixed

rodwyer100: (in comments of issue)

Recon/python/recoSkim2.py
rodwyer100: Fixed in comments of issue

Recon/python/recoSkim2.py
rodwyer100: fixed in comments of issue.

rodwyer100:
I have addressed each comment, and read over the three documents relevant to the new processor that would be added. I believe the comments are sensible (with no weird additions from copies of things). I have implemented all the changes suggested in your comments above as well. For Tom's comments on the validation code, I have put all suggested edits in a final four copies of the validation scripts along with more notes. In those validation scripts (located in the issue comments) I have also cut off any unnecessary pieces. I verified they run identically. So I believe this is ready for merger.

Recon/python/recoSkim3.py
rodwyer100: fixed

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,68 @@
...
+  trigger_list_ = ps.getParameter<std::vector<std::string>>("trigger_list");
+  trigger_passNames_ = ps.getParameter<std::vector<std::string>>("trigger_passNames");
+  doAND_ = ps.getParameter<bool>("doAND");
+  doOR_ = ps.getParameter<bool>("doOR");
+  doVAL_ = ps.getParameter<bool>("doVAL");
+  //Returns an error if some bad combination of doOR and doAND is enabled.
+  if (doAND_ == doOR_){
+    ldmx_log(fatal) << "Either tried to do both or neither of  doAND and doOR. Exiting.";
+    throw std::invalid_argument("Either tried to do both or neither of doAND and doOR. Exiting.");
+    return;
```
tomeichlersmith: Just use the framework exceptions here. This will end processing and print the exception in the same format as others.
```suggestion
    EXCEPTION_RAISE("InvalidArg","Either tried to do both or neither of doAND and doOR. Exactly one should be true.");
```

Recon/src/Recon/SequentialTrigger.cxx
```diff
@@ -0,0 +1,68 @@
...
+            if(doOR_){hasPassed=true;break;}
+        }else{
+            if(doAND_){hasPassed=false;break;}
+        }
+    }catch(...){
+        ldmx_log(fatal) << "Attemping to use non-existing trigger collection "
+                      << trigger_list_[i] << "_" << trigger_passNames_[i]
+                      << " to skim! Exiting.";    
+        return;
+    }
```
tomeichlersmith: You should allow the exception from `getObject` to propagate up and end processing. If you wish to allow for some of the requested trigger collections to be non-existent, you should use `event.exists` to check if they are in the event before attempting to `get` them.

tomeichlersmith (changes requested):
Besides a few technical comments about handling exceptions, I think this is looking really good.

Do you have `clang-format` installed? If you do, then you can also format the files according to the google-style. If not, I can do that just before merging it.
```
clang-format -i --style=Google Recon/include/Recon/SequentialTrigger.h Recon/src/Recon/SequentialTrigger.cxx
```

Recon/src/Recon/SequentialTrigger.cxx
rodwyer100: I figure if any of them fail, and we let that go through, it could yield some dangerous confusions. Wrong plots, etc. I have implemented an exception that propogates up and stops everything. Look to my latest comment on this thread.

rodwyer100:
I have addressed all of Tom's concerns, with the exception of clang. I determined that the propensity for error should we allow it to continue running if a input collection did not exist was too high. Therefore it just throws an error now that stops processing altogether. This is what that looks like:

ErrorCode

The two files requested have been clanged (atleast I submitted that code in sdf, it ran, and the format seemed to change). I there is anything else, let me know :).

tomeichlersmith (approved):
Looks good, thank you for going through my nit-picking :100:

PR #1094, opened 2022-09-29
Silicon sensors in ECal expected to be 300microns

tomeichlersmith:
There may be some other values lurking that will need to be found upon testing

Previously, we hoped to get 500micron silicon sensors for the ECal, but it looks more likely that we will be constructing with a thinner sensor (probably 300 mircons).

## Check List
- [x] Visualize detector to make sure it looks correct
- [x] Ecal-internal overlap checks
- [x] Detector-wide overlap checks
- [x] Can the simulation run?


tomeichlersmith:
Following the [same test from the last round of ecal geo developments](https://github.com/LDMX-Software/ldmx-sw/pull/1079#issuecomment-1242048722) we are looking good.

ecal_sim_hits_z
ecal_sim_hits_y
ecal_sim_hits_x

PR #1100, opened 2022-10-14
Fix recoil tracker scoring plane positions.

### LINKED ISSUE #1099: Scoring Planes around Recoil Tracker in v14
**Describe the bug**
The scoring planes around the recoil tracker use the GDML variable `recoil_layer_pos` which is not defined in constants.gdml for the v14 detector. For now, they are just commented out since their truth-level information is almost perfectly replicated by the recoil sim tracker hits themselves for subsystems outside the tracker.

**To Fix**
Update the recoil scoring plane position variable calculation so that it does not rely on this deprecated variable.

omar-moreno:

This resolves #1099. 

## Check List

### DIFF
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -1,7 +1,7 @@
 from LDMX.Framework import ldmxcfg
 p = ldmxcfg.Process('test')
 
-p.maxTriesPerEvent = 1000
+p.maxTriesPerEvent = 10000
 
 from LDMX.Biasing import ecal
 from LDMX.SimCore import generators as gen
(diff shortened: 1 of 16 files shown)


omar-moreno:
I have updated the scoring plane positions to use the updated recoil tracker variables.  Using g4-vis, the scoring_planes.gdml file is parsed correctly but it immediately seg faults
```
#5  0x00007fb66d705d77 in simcore::geo::GDMLParser::read() 
```

This is happening no matter what gdml file I load.  This is a recent development since I was able to successfully visualize the recoil and tagger trackers a couple of weeks ago.

EinarElen:
For the scoring planes gdml file, the segfault happens at https://github.com/LDMX-Software/SimCore/blob/113ff1362d4aa812b5c4916071e2ac9260246a08/src/SimCore/Geo/GDMLParser.cxx#L23 because the detector version auxinfo isn't set, so the detectorheader pointer is null. 

Regarding the other GDML files, it looks like the crashes are due to changes like
- Ecal, Hcal, Magnet: Missing region definitions like `what():  Reference region 'CalorimeterRegion' was not found!` 
- Tagger, Missing visattributes like `what():  Referenced VisAttributes 'InvisibleNoDau' was not found!`
- Target, Missing material reference `Referenced material 'Vacuum' was not found!`
- TS, Some GDML issue, `Error in expression: trig_scint_area_envelope_x`

This only happens for the individual volumes, if you load the `detector.gdml` file then everything works so it doesn't affect the regular simulation.

omar-moreno:
Thanks Einar! Let me add the version and try again.

bryngemark:
hi @omar-moreno, are these the only scoring planes changes needed for the tracker?

omar-moreno:
No, the scoring planes need to be moved.

bryngemark:
Right, but only for the recoil tracker?

omar-moreno:
I only looked at the recoil tracker.

omar-moreno:
After merging trunk into this branch.

image

omar-moreno:
Target are scoring planes need to be adjusted. Recoil tracker looks fine.

image

omar-moreno:
Hcal scoring planes need to be adjusted as well.
image

omar-moreno:
Recoil scoring plane issues are fixed. Other scoring plane issues should be resolved separately.

tomeichlersmith (approved):

tomeichlersmith:
I was able to run the configs in validation_samples and I want to get another round of validation underway, so I'm going to go ahead and merge this.

Issue #1114, opened 2022-11-03
Streamlined validation code 


bryngemark:
**Is your feature request related to a problem? Please describe.**
It would be great to have a streamlined plotting of the DQM plots produced in the GitHub validation action, as well as other standard validation plots each subsystem needs.

**Describe the solution you'd like**
@tomeichlersmith  has started making some of this for Ecal and I'd like to import that to `ldmx-sw`. His repo is already set up to run comparisons across versions, and is equipped to use with notebooks. The idea would be to have the plotting tool be general, with the possibility to run system specific simulations (e.g. what type of beam you'd like and starting from where), and list the histograms you want to plot.

**Describe alternatives you've considered**
That we start something from scratch. 

**Additional context**
This validation plot suite could also be included as the end step of our automatized GitHub validation action triggered by PRs. This way we'd make it easier to actually inspect the gold/new comparison plots.

tvami:
Does this issue just mean that we should use
https://github.com/LDMX-Software/ldmx-sw/tree/trunk/Validation
in the CI?

tomeichlersmith:
I'm not sure if that was the original intent, but I do think it would be a good idea to have the plotting code used in the CI be something that humans can use directly as well.

We would need to use NumPy's KS test rather than ROOT's which isn't a large problem.
The [PyROOT script I wrote for the CI](https://github.com/LDMX-Software/ldmx-sw/blob/trunk/.github/actions/validate/compare.py)  can be used as a potential reference although we could probably be a bit more careful and make sure the plots are prettier and maybe combine them into a slide-deck rather than a directory for easier browsing by reviewers.

bryngemark:
@tvami yes this is what i had in mind, originally, and i still think it's a good idea. as @tomeichlersmith implies, making human interactions with the CI output easier would be a good step forward. 

it would be particularly amazing if failed test plots could be attached to the PR by the action, in some non-obscure location (avoiding the whole artifact downloading, untarring, etc), perhaps even to the thread itself? we're not penalized for using large amounts of data for our discussions, right?

tomeichlersmith:
We are not penalized _directly_ for using large amounts of data within discussions, but we are prevented from doing so indirectly.

[...]

PR #1122, opened 2022-11-10
Initial commit of Validation module

bryngemark:

This will resolve issue #1114 

## Check List


bryngemark:
In order to try this out, I've run, in some appropriate validation directory of mine:

``` bash 
> source /path/to/ldmx-sw/scripts/ldmx-env.sh
> mkdir data
> ldmx fire $LDMX_BASE/ldmx-sw/Validation/simulation.py --geometry 14 --n-events 2000 --out-dir data
> ldmx fire $LDMX_BASE/ldmx-sw/Validation/simulation.py --geometry 12 --n-events 2000 --out-dir data
> ls data/
events_geometry_v12_events_2000_run_1.root histos_geometry_v12_events_2000_run_1.root
events_geometry_v14_events_2000_run_1.root histos_geometry_v14_events_2000_run_1.root
> cd $LDMX_BASE/ldmx-sw/Validation
> ln -s /path/to/validation/dir/data data
> ldmx python3 comp/__main__.py 
  File "comp/__main__.py", line 9, in <module>
    import mplhep
ModuleNotFoundError: No module named 'mplhep'
```

So the ROOT plotting lib is missing from the container. This is after doing  `docker pull ldmx/dev:latest`. `ldmx config` shows 
```
Docker Version: Docker version 20.10.12, build e91ed57
Docker Tag: ldmx/dev:latest
    Digest: ldmx/dev@sha256:89147ab75f0142f0685713067127f98696520b9ff242f96aba2aea9a0f871a39
```

tomeichlersmith:
After a discussion on slack with @bryngemark , we agreed that this `Validation` module should be designed like a pure-python package with the forward-looking idea that this module will be install-able and usable as the base for common plotting tasks and analysis plotting.

Making a set of python files into a install-able package basically amounts to following a specific structure and defining a few files in a specific way. This boiler-plate structure/code should be done now before merging so that we don't have to re-write any tutorials/usage scripts depending on how the structuring changes things. After some internet browsing, I found a [sample project](https://github.com/pypa/sampleproject) produced by the Python Packaging Authority which sounds legit. We'll probably trim some of the stuff out, but it looks like a decent start.

bryngemark:
@tomeichlersmith I've looked at the example you provided. I think this is an interesting path and I'd like to hear what other developers think, and who would be interested in working on this. I suggest we discuss this in a meeting :)

bryngemark:
Pulling in @vdutta to keep her updated on this idea

tomeichlersmith (commented):
@bryngemark I still have more to do, but I think it is ready for you to test to see how it works for you.

## To Do
- [x] More documentation on installation
- [x] Test container running
- [x] Remove systems/test.py
- [x] More examples on how to use Differ and its members
- [x] Update setup/build to support newer pythons

bryngemark:
Hi @tomeichlersmith, a basic question: how do I run this thing after your restructuring? Here's what I tried, after installing with the container:
```
> ldmx python3 -m Validation gamma/tsData
/usr/bin/python3: No module named Validation
```

tomeichlersmith:
What command did you run to install it and where are you attempting to run?

Both if these are buggy for different reasons.

bryngemark:
`ldmx python3 -m pip install $LDMX_BASE/ldmx-sw/Validation/ --target install/python/LDMX/ --no-deps --no-cache` 

attempting to run outside of `ldmx-sw`, per instruction

tomeichlersmith:
Oh! The `PYTHONPATH` looks at `install/python` so the module is `LDMX.Validation` rather than just `Validation` when installing it to `install/python/LDMX`.

```
ldmx python3 -m LDMX.Validation -h
```
should work :crossed_fingers:

bryngemark:
```
> ldmx python3 -m LDMX.Validation -h
/usr/bin/python3: No module named LDMX.Validation
```
not yet... unless I needed to pull before trying again?

tomeichlersmith:
Some cursory googling says that one cannot access a submodule with the `python3 -m` syntax, so I think you should uninstall it and re-install it as a "root" module.

```
ldmx python3 -m pip uninstall LDMX.Validation # not sure if this will work, may need to manually remove the files
ldmx python3 -m pip install Validation/ --target install/python/ --no-deps --no-cache
```

bryngemark:
Ok, this works now, but only if I install it with the container from `$LDMX_BASE/ldmx-sw` (and then run it elsewhere); installing it with that path from some other dir does not work.

bryngemark:
So, @tomeichlersmith, bear with me here, but what do i need to do to be able to develop here? Is the flag `--editable` supposed to be enough?

tomeichlersmith:
Yes, basically `--editable` "symlinks" the Validation module to the source code in `Validation/src`.

**If testing inside a notebook**, you will need to restart the python kernel when making changes so that the module is _actually_ re-imported. Just calling `import` again doesn't do anything because (from python's point of view) it is already imported and so it doesn't waste time trying to import it again.

bryngemark:
Ok. But when I do so, there is no module `Validation`, and it is not there under `install/python`, only the `egg-link` is. If I run without the `--editable` flag, it is there (along with `...-0.0.1.dist-info`). But then no local changes get picked up.

I am not running this in a notebook (yet. I might try later. This is not how I usually work and I don't want to try/debug/learn everything at once).

tomeichlersmith:
The `egg-link` file is how Python does the "symlink" so it you should still be able to do

```
ldmx python3
import Validation
```

tomeichlersmith:
I've been able to "install" and "run" Validation both inside and outside the container on a few different python versions.

bryngemark:
I somehow still can't make this run from the command line. I installed it with the container:
```
/path/to/ldmx-sw > ldmx python3 -m pip install --editable Validation/ --target install/python/ --no-deps --no-cache --upgrade
```
and then in a different directory I tried 
```
> ldmx python3 -m Validation gamma/tsData --systems trigscint.dqm 
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/lenebryngemark/work/LDMX/code/ldmx-sw/install/python/Validation/__main__.py", line 58, in <module>
    root_files = [ File.from_path(os.path.join(data,f)) for f in os.listdir(data) if f.endswith('.root') ]
  File "/Users/lenebryngemark/work/LDMX/code/ldmx-sw/install/python/Validation/__main__.py", line 58, in <listcomp>
    root_files = [ File.from_path(os.path.join(data,f)) for f in os.listdir(data) if f.endswith('.root') ]
  File "/Users/lenebryngemark/work/LDMX/code/ldmx-sw/install/python/Validation/_file.py", line 68, in from_path
    file_params =  { l[i] : l[i+1] for i in range(len(l)) if i%2 == 0 }
  File "/Users/lenebryngemark/work/LDMX/code/ldmx-sw/install/python/Validation/_file.py", line 68, in <dictcomp>
    file_params =  { l[i] : l[i+1] for i in range(len(l)) if i%2 == 0 }
IndexError: list index out of range
``` 
although I have root files in the specified path
```
> ls gamma/tsData/*.root
gamma/tsData/events_geometry_v12_events_200_run_1.root gamma/tsData/histos_geometry_v12_events_200_run_1.root
gamma/tsData/events_geometry_v14_events_200_run_1.root gamma/tsData/histos_geometry_v14_events_200_run_1.root
```

tomeichlersmith:
That's not your fault, I made some last minute changes to try to simplify things...

The `    file_params =  { l[i] : l[i+1] for i in range(len(l)) if i%2 == 0 }` line assumes that the file name has an even number of strings separated by `_` i.e. the entire filename is `key_val` pairs.

The solution is to update your filenames, I'd suggest just labeling the `events`/`histos` parameter with the `type` key by prefixing all your files with `type_`.

I can also add a more informative exception.

Edit: I also had a patch that I hadn't pushed yet :facepalm:

tomeichlersmith (approved):
I'm ready to merge this, apologies for it falling off my radar.

I think transferring the GitHub action plot generation to use this module is work for another time, getting this onto `trunk` can help make development of those actions easier as well.

EinarElen:
Reviewing this fell of my radar as well, I will be looking into it today. I want to add an Hcal component later, but similar to the action/plot generation it can be a separate piece of work (depends on #1136). The one thing I am missing right now though is a bit of documentation of 
- Typical usage 
- What file structure the setup expects (i.e. the whole thing with key1,key2)

Validation/src/Validation/__main__.py
```diff
@@ -0,0 +1,97 @@
...
+        the file name by splitting the filename into key-val pairs separated 
+        by underscores (i.e. key1_val1_key2_val2_..._keyN_valN.root). If no
+        parameter is provided, then the first key/val is used.
+        """
+        )
+
+    parser.add_argument('data',help='directory of event and histogram files')
+    parser.add_argument('--log',help='logging level',choices=['info','debug','warn','error'], default='warn')
+    parser.add_argument('--label',help='label for grouping of data, defaults to data directory name')
+    parser.add_argument('--out-dir',help='directory to which to print plots. defaults to input data directory')
```
EinarElen, line 41: I think this is a typo? The out directory is later accessed as out_dir

Validation/src/Validation/__main__.py
```diff
@@ -0,0 +1,97 @@
...
+    parser.add_argument('--systems',required=True, choices=plotter.__registry__.keys(), nargs='+',
+        help='list of plotters to run')
+    parser.add_argument('--param',help='parameter in filename to use as file labels')
```
EinarElen: I think it would be quite useful if you weren't limited to one of the parameters. Sure, for comparing v12 and v14 usually we just want those two but I've found myself wanting to compare more than two files more than once.  From what I can tell, this should be relatively straight forward to do. e.g. split  `param` by some separator (e.g. `,`) or use a parameter type directly that supports multiple arguments.  Then change the check in in _file to check each entry rather than just one and join them into one label

EinarElen (commented):
Looks good, I just have two minor things

Validation/src/Validation/__main__.py
tomeichlersmith: argparse converts `-` to `_` so that the command-line flag is easier to type and the resulting variable can be accessed in Python.

tomeichlersmith: For future reference:

[argparse docs](https://docs.python.org/dev/library/argparse.html#dest)

> Any internal - characters will be converted to _ characters to make sure the string is a valid attribute name.

Validation/src/Validation/__main__.py
tomeichlersmith: This makes sense - do you want to draft what you're thinking or should I try for it?

Validation/src/Validation/__main__.py
EinarElen: I see, then I was doing something else wrong when I ran into this :)

Validation/src/Validation/__main__.py
EinarElen: I can have a go at it, i have a hacky proof of concept implementation over here

Validation/src/Validation/__main__.py
tomeichlersmith: Were you seeing the plots not go into the selected output directory? There could still be a bug lurking

Validation/src/Validation/__main__.py
EinarElen: Pushed an implementation now. Tested with both 0, 1, and 3 parameters and all worked 
[hcal_dqm_back_hit_multiplicity.pdf](https://github.com/LDMX-Software/ldmx-sw/files/11334390/hcal_dqm_back_hit_multiplicity.pdf)

EinarElen: Now that I see that Github doesn't automatically show PDFs, maybe it would be worth making the output format configurable and default to pdf? But that can be for later

EinarElen (approved):

PR #1129, opened 2022-11-19
Add in updates to SimCore that properly include TS testbeam SDs

### LINKED ISSUE #1125: Updated TS ad names not reflected in testbeam geometry
**Describe the bug**
The testbeam geometry uses the old TS pad naming, so the SD doesn't pick up the pad, thus there are no TS sim hits. One way to do this would be to patch simcore with one more check (as we already check if we're using v12, for backwards compatibility). Another is to push a v2.1 of the testbeam geometry where `Up` --> `Pad1`.

**To Reproduce**
Run simulation with e.g. `ldmx-hcal-prototype-v2.0`, beam position (0,0,-1100) and behold the empty TS collection 

**Desired behavior**
I want to see TS hits...

bryngemark:

This resolves #1125 by adding in a specific test beam SD that maps onto the naming conventions in its gdml. It had not been updated to adhere to the new way of specifying the Sensitive Detectors.

## Check List

This plot shows simparticle _z_ (black/filled), TS simhit _z_ (red) and Hcal simhit _z_ (cyan). It all looks like expected: the beam starts at -600mm in this run.
 
z_simPartsSimHits_redTS_cyanHcal_iss1125

Config:
``` python
import sys

thisPassName="sim"
p = ldmxcfg.Process(thisPassName)

p.run = 10
p.maxEvents = 400
p.outputFiles = ['out.root']

gun = generators.gun('particle_gun')
gun.particle = 'e-'
gun.direction = [0., 0., 1.]
gun.position = [0., 0., -600] #mm
gun.energy = 4.   #gev                                                                                                                                                               

simulation = simulator.simulator('test_TS')
simulation.generators=[gun]
simulation.setDetector('ldmx-hcal-prototype-v2.0')                                                                                          
simulation.beamSpotSmear = [150, 200, 0] #mm, at start position                             

p.sequence=[simulation]
```

### Related Sub-Module PRs
https://github.com/LDMX-Software/SimCore/pull/82


tomeichlersmith (approved):
I've merged in your submodule PR (and my patch as well). My patch is orthogonal to yours since it only has to do with biasing operators, so I am going to keep them in the same release of SimCore and update them both here. I will merge this after the basic tests pass to make sure I didn't mess anything up.

PR #1144, opened 2023-01-19
production image dockerfile fixup

### LINKED ISSUE #1084: Production Image Dockerfile fixup
**Describe the bug**
The Dockerfile here assumes the dev container's entrypoint is at `/home/ldmx.sh` but in versions after `v2.0`, it was moved to `/etc/entry.sh`. This, among other things, are reasons for updating the Dockerfile.

## Other Fix-Ups
- [x] We could also avoid re-defining the entrypoint by making the environment variable `LDMX_INSTALL_PREFIX` be checked before assuming `LDMX_BASE/ldmx-sw/install` in the base image

### LINKED ISSUE #1103: GitHub Action Deprecations
**Describe the bug**
GitHub actions is deprecating some things. This issue is just helping me keep track of the changes so I can remember to update them.
- [Node.js 12 actions](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) - not sure what this is in reference to
  - referencing actions we are _using_ which need to be bumped to their next major version
- `num_jobs_per_sample` should be `num_runs_per_sample` for `generate-matrix`
- The [`set-output` and `save-state` commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
  - use `echo "{name}={value}" >> $GITHUB_OUTPUT`

tomeichlersmith:

This resolves #1084 and resolves #1103 .

## Check List
- [x] NA ~I successfully compiled _ldmx-sw_ with my developments~
  - Going to watch to make sure the actions complete like normal and the production image is built


tomeichlersmith:
The PR-triggered actions will test `basic_test.yml` and `pr_validation.yml`.
I manually triggered a build of the production image to test `build_production_image.yml` https://github.com/LDMX-Software/ldmx-sw/actions/runs/3961300359

`generate_pr_gold.yml` and `new_pre_release.yml` are the only ones that are difficult to test; however, they both are very similar to `pr_validation.yml` except for the end.

`generate_pr_gold.yml` has automatic PR generation at the end while `new_pre_release.yml` has a automatic release generation at the end both of which I've double checked that the new versions have the same parameters that we use and so they _should_ work as normal.

tomeichlersmith:
The production image build was successful. I downloaded a copy of it to make sure `fire` was being used from within the container...
```
source ldmx-sw/scripts/ldmx-env.sh
singularity build ldmx_local_new-prod-test.sif docker://tomeichlersmith/ldmx-sw:new-prod-test
ldmx use local new-prod-test
ldmx 'which fire'
ldmx fire ldmx-sw/SimCore/test/basic.py
```
which prints `/usr/local/bin/fire` and runs the basic simulation like normal.

tomeichlersmith:
The second Basic Tests run failed due to a failure to download ONNX runtime. I think this is related to the other GitHub issues reported above since we are trying to download ONNX _from_ GitHub.
```
[ ERROR ]: Could not download https://github.com/microsoft/onnxruntime/releases/download/v1.2.0/onnxruntime-linux-x64-1.2.0.tgz
```

Issue #1069, opened 2022-06-07
Photo Nuclear DQM not working with upKaons branch


LisaALoman:
I am trying to run Photo Nuclear DQM when I am doing simulations using the `upKaons` branch and a beam energy of 8GeV. I got some help to look into it and it seems like the daughter track IDs are not in the particle map.
I am getting the following error:
```
terminate called after throwing an instance of 'std::out_of_range'
  what():  map::at
/etc/entry.sh: line 60:    81 Aborted                 (core dumped) fire config.py
```
In my config file I am using the simulator function `ecal.kaon_pn`, which is in a modified version of `Biasing/python/ecal.py` .

This is the configuration file:

```
from LDMX.Framework import ldmxcfg

p=ldmxcfg.Process("test")
p.run = 1

from LDMX.SimCore import bias_operators
from LDMX.SimCore import simulator
from LDMX.Ecal import EcalGeometry
from LDMX.Hcal import HcalGeometry
import LDMX.Ecal.ecal_hardcoded_conditions as ecal_conditions
import LDMX.Hcal.hcal_hardcoded_conditions as hcal_conditions

from LDMX.Biasing import ecal  #target
from LDMX.SimCore import generators
mysim = ecal.kaon_pn('ldmx-det-v13', generators.single_8gev_e_upstream_tagger()) #changed 4 to 8
mysim.description = "ECal Kaon Test Simulation, 8GeV"
print(mysim)

# NEW to fix everything; 3.0.0 broke imports yet again...
import LDMX.Ecal.digi as ecal_digi
import LDMX.Hcal.digi as hcal_digi
import LDMX.Ecal.vetos as ecal_vetos
import LDMX.Hcal.hcal as hcal_py
from LDMX.Recon.simpleTrigger import simpleTrigger
from LDMX.Recon.electronCounter import ElectronCounter #to allow for multi-e, electron counter needed


from LDMX.TrigScint.trigScint import TrigScintDigiProducer
from LDMX.TrigScint.trigScint import TrigScintClusterProducer
from LDMX.TrigScint.trigScint import trigScintTrack

tsDigisUp   = TrigScintDigiProducer.up()
tsDigisTag  = TrigScintDigiProducer.tagger()
tsDigisDown = TrigScintDigiProducer.down()

#these lines required for trigger skimming
eCount = ElectronCounter(1, "ElectronCounter") #first argument is number of electrons in simulation
eCount.use_simulated_electron_number = True
[...]

PR #1151, opened 2023-03-05
Add ability to veto events where the primary never reaches the tagger region

### LINKED ISSUE #1146: TaggerVetoFilter inefficiency
## Description
This was initially observed when studying the signal generation using this filter. What I observed through some extra debug printouts (and using the `StepPrinter` user action) was that the _primary_ electron was not entering the target region. This did not "activate" the `TargetDarkBremFilter` which means it did not know to veto the event. The primary electron was found with $p_z < 0$ and $z < 0$ at its end point, so I think something caused the primary electron to back-scatter. We still see a near-beam-energy total reconstructed energy in the ECal however, so particles are still reaching it.

### Initial Hypothesis
Various small scatters bend the primary electron into the magnet which causes a much harder scatter (perhaps electron-nuclear or perhaps just a hard brem). The non-primary products of this hard scatter continue downstream into the ECal while the primary electron recoils back, avoiding the target and thus the `TargetDarkBremFilter`.

### Additional context
https://github.com/LDMX-Software/ldmx-sw/issues/1138#issuecomment-1376305770
https://github.com/LDMX-Software/ldmx-sw/issues/1139

## Plan
- Isolate and confirm issue: develop config where we can observe the TaggerVetoFilter failling to veto an event where the primary electron misses the target.
- Develop update: try modifications to TaggerVetoFilter to "catch" these types of events.

EinarElen:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1146 and should also be helpful for issues like https://github.com/LDMX-Software/ldmx-sw/issues/1069. I'm not sure that this is the most elegant solution out there, but this does really need to be patched. 

I opted to make the default to veto events where the primary particle never enters the tagger region if the TaggerVetoFilter is enabled, however this does technically change the behavior of existing code but arguably to a way that is less astonishing for the user. I can't think of why we would want the default to be the other way, but I don't mind switching it if needed. 

## Check List

Using the EcalPN config from the validations samples, I first ran the config before adding the step that actually tags events that enter the tagger region and confirmed that the EndOfEventAction was rejecting them. I then added the step, and ran the same config (run number 1) with a breakpoint in the EndOfEventAction to see that it was being triggered in a real scenario. 

### Related Sub-Module PRs
- https://github.com/LDMX-Software/SimCore/pull/90

### DIFF
--- a/Biasing/src/Biasing/TaggerVetoFilter.cxx
+++ b/Biasing/src/Biasing/TaggerVetoFilter.cxx
@@ -19,7 +19,14 @@ TaggerVetoFilter::TaggerVetoFilter(const std::string& name,
 
 TaggerVetoFilter::~TaggerVetoFilter() {}
 
-void TaggerVetoFilter::stepping(const G4Step* step) {
+void TaggerVetoFilter::EndOfEventAction(const G4Event *) {
+  if (reject_primaries_missing_tagger_) {
+    if (!getEventInfo()->primaryEnteredTaggerRegion()) {
+      G4RunManager::GetRunManager()->AbortEvent();
+    }
+  }
+}
+void TaggerVetoFilter::stepping(const G4Step *step) {
   // Get the track associated with this step
   auto track{step->GetTrack()};
 
--- a/Biasing/src/Biasing/TaggerVetoFilter.cxx
+++ b/Biasing/src/Biasing/TaggerVetoFilter.cxx
@@ -45,6 +45,7 @@ void TaggerVetoFilter::stepping(const G4Step *step) {
       region.compareTo("tagger") != 0)
     return;
 
+  getEventInfo()->setPrimaryEnteredTaggerRegion();
   // If the energy of the particle falls below threshold, stop
   // processing the event.
   if (auto energy{step->GetPostStepPoint()->GetTotalEnergy()};
--- a/Biasing/src/Biasing/TaggerVetoFilter.cxx
+++ b/Biasing/src/Biasing/TaggerVetoFilter.cxx
@@ -13,6 +13,8 @@ TaggerVetoFilter::TaggerVetoFilter(const std::string& name,
                                    framework::config::Parameters& parameters)
     : simcore::UserAction(name, parameters) {
   threshold_ = parameters.getParameter<double>("threshold");
+  reject_primaries_missing_tagger_ =
+      parameters.getParameter<bool>("reject_events_missing_tagger");
 }
 
 TaggerVetoFilter::~TaggerVetoFilter() {}
--- a/Biasing/src/Biasing/TaggerVetoFilter.cxx
+++ b/Biasing/src/Biasing/TaggerVetoFilter.cxx
@@ -9,8 +9,8 @@
 
 namespace biasing {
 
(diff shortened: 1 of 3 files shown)


Biasing/src/Biasing/TaggerVetoFilter.cxx
```diff
@@ -13,11 +13,20 @@ TaggerVetoFilter::TaggerVetoFilter(const std::string& name,
...
 }
 
 TaggerVetoFilter::~TaggerVetoFilter() {}
 
-void TaggerVetoFilter::stepping(const G4Step* step) {
+void TaggerVetoFilter::EndOfEventAction(const G4Event *) {
+  if (reject_primaries_missing_tagger_) {
+    if (!getEventInfo()->primaryEnteredTaggerRegion()) {
+      G4RunManager::GetRunManager()->AbortEvent();
+    }
```
tomeichlersmith: Is there a reason to have the `primaryEnteredTaggerRegion` flag be in the event info and not just a part of the `TaggerVetoFilter` class?

All `EVENT` actions come with `Begin` and `End` of event actions so you could reset the flag in `Begin` and check if the event should be aborted in `End`.

tomeichlersmith (commented):
> not sure that this is the most elegant solution

I'm not sure that a more elegant solution _exists_. Maybe, in a future ldmx-sw where we have a tracking veto set up we can let these simulations happen and then have reconstruction veto them (like the events would happen in real life), but that would probably just waste CPU cycles.

> the default to veto events

I agree with your conclusion that this should be the new default. In fact, I think most if not all users were assuming this was the case.

The `ecal_pn` PR Validation is failing because, unsurprisingly, this is affecting the distributions downstream of the simulation changed by this patch. I've unpacked the plots that were made and my reading of them is that they are all small modifications from the few events that are missing the tagger region.
[ecal_pn-pass-ks-test.pdf](https://github.com/LDMX-Software/ldmx-sw/files/10899132/ecal_pn-pass-ks-test.pdf)
[ecal_pn-fail-ks-test.pdf](https://github.com/LDMX-Software/ldmx-sw/files/10899133/ecal_pn-fail-ks-test.pdf)

Biasing/src/Biasing/TaggerVetoFilter.cxx
EinarElen: The reason for that would be that I made the addition to event info before realizing that you could have an action be multiple types :) 

Doing it with the begin action would have the benefit of only touching one module.

tomeichlersmith: I would switch to having `TaggerVetoFilter` holding the flag unless you can think of a reason for this flag to be used by other UserActions or be stored in the output (those are the two reasons I think it is useful to put something in EventInfo).

EinarElen: Yeah, I can add that instead. This would also make https://github.com/LDMX-Software/SimCore/pull/90 redundant

EinarElen: Should be there now!

tomeichlersmith (approved):

PR #1154, opened 2023-03-29
Change method for looking up recoil electron

### LINKED ISSUE #1153: Update recoil electron identification method for Ecal veto
Since a while, the recoil electron in signal generation does no longer by definition have the track ID 1. Instead, one must identify it through PDG ID and processType.

**To Reproduce**
Run the Ecal Veto as it stands and verify that the recoil electron momentum, containment variables etc are not filled.

**Desired behavior**
The correct SimParticle and track ID should be returned by the associated analysis util method. I have implemented this.

bryngemark:
It now checks for DarkBrem origin, or, returns trackID 1 <-- this last point might need some more thought before merging. I'd like to b able to use this method for getting the recoil electron also for other processes than dark brem. 

This resolves #1153 and the same issue [reported as a bug in the Ecal repo](https://github.com/LDMX-Software/Ecal/issues/46) 

## Check List

I now see non-zero entries in the recoil momentum and containment variables collections. 
This was done with both re-recoing a small number of 1 GeV signal events produced with v3.2.2, and an inclusive e sample to see that things didn't break wildly. 

- [N/A] I attached any sub-module related changes to this PR.

### DIFF
--- a/Tools/src/Tools/AnalysisUtils.cxx
+++ b/Tools/src/Tools/AnalysisUtils.cxx
@@ -29,17 +29,16 @@ namespace Analysis {
 std::tuple<int, const ldmx::SimParticle*> getRecoil(
     const std::map<int, ldmx::SimParticle>& particleMap) {
   // The recoil electron is "produced" in the dark brem geneartion 
-  for (auto const& [trackID, particle] : particleMap) {
-	if (particle.getPdgID() == 11 and
-		particle.getProcessType() ==
-		ldmx::SimParticle::ProcessType::eDarkBrem) {
-	  return {trackID, &particle};
-	}
-	else if (particle.getPdgID() == 11 and trackID == 1) {
-	  // primary electron
-	  return {trackID, &particle};
-	}
+  for (const auto& [trackID, particle] : particleMap) {
+    if (particle.getPdgID() == 11 and 
+        particle.getProcessType() == 
+        ldmx::SimParticle::ProcessType::eDarkBrem) {
+      return {trackID, &particle};
+    }
   }
+  // only get here if recoil electron was not "produced" by dark brem
+  //   in this case (bkgd), we interpret the primary electron as also the recoil electron
+  return {1, &(particleMap.at(1))};
 }
 
 const ldmx::SimParticle* getPNGamma(
--- a/Tools/src/Tools/AnalysisUtils.cxx
+++ b/Tools/src/Tools/AnalysisUtils.cxx
@@ -28,8 +28,18 @@ namespace Analysis {
 
 std::tuple<int, const ldmx::SimParticle*> getRecoil(
     const std::map<int, ldmx::SimParticle>& particleMap) {
-  // The recoil electron always has a track ID of 1.
-  return {1, &(particleMap.at(1))};
+  // The recoil electron is "produced" in the dark brem geneartion 
+  for (auto const& [trackID, particle] : particleMap) {
+	if (particle.getPdgID() == 11 and
+		particle.getProcessType() ==
+		ldmx::SimParticle::ProcessType::eDarkBrem) {
+	  return {trackID, &particle};
+	}
+	else if (particle.getPdgID() == 11 and trackID == 1) {
+	  // primary electron
+	  return {trackID, &particle};
+	}
+  }
 }
 
 const ldmx::SimParticle* getPNGamma(


Tools/src/Tools/AnalysisUtils.cxx
```diff
@@ -28,8 +28,18 @@ namespace Analysis {
...
+	if (particle.getPdgID() == 11 and
+		particle.getProcessType() ==
+		ldmx::SimParticle::ProcessType::eDarkBrem) {
+	  return {trackID, &particle};
+	}
+	else if (particle.getPdgID() == 11 and trackID == 1) {
+	  // primary electron
+	  return {trackID, &particle};
+	}
+  }
```
tomeichlersmith: I believe this will always return the primary electron since that will be "earlier" in the particleMap than secondary electrons. I think we should fall back to the primary only if secondary from dark brem is not found - this is less fast but more precise.

```suggestion
  for (const auto& [trackID, particle] : particleMap) {
    if (particle.getPdgID() == 11 and 
        particle.getProcessType() == 
        ldmx::SimParticle::ProcessType::eDarkBrem) {
      return {trackID, &particle};
    }
  }
  // only get here if recoil electron was not "produced" by dark brem
  //   in this case (bkgd), we interpret the primary electron as also the recoil electron
  return {1, &(particleMap.at(1))};
```

bryngemark: yes, of course, this is what you get for late coding after the weekly is over. thanks tom.

bryngemark:
So I made 100 inclusive events and it looks like the "recoil electron" being picked up is the right one:

``` 
root [1] LDMX_Events->Draw("sqrt(EcalScoringPlaneHits_sim.px_*EcalScoringPlaneHits_sim.px_+EcalScoringPlaneHits_sim.py_*EcalScoringPlaneHits_sim.py_)", "EcalScoringPlaneHits_sim.pdgID_==11 && EcalScoringPlaneHits_sim.trackID_==1", "")
(long long) 100
root [2] LDMX_Events->Draw("sqrt(EcalVeto_sim.recoilPx_*EcalVeto_sim.recoilPx_+EcalVeto_sim.recoilPy_*EcalVeto_sim.recoilPx_)", "EcalVeto_sim.recoilPx_ > -9999", "same")
(long long) 100
``` 

SP hits pT in blue, recoil electron pT in black/shaded. Not sure why they are not matching _exactly_, there seems to be some small shift. But overall I think this comfirms that the right electrons are picked up also for inclusive.

incl1e_recoilpT_SPhitpT

bryngemark:
also! @tomeichlersmith any idea why we are comparing to v3.2.1 gold here? it looks like current `trunk` is better e.g. at getting less 0-hit events in the TS etc but it also means i'm getting a ot of failed tests for things i can't reasonably have changed.

tomeichlersmith:
We are comparing to v3.2.1 since I didn't merge in the v3.2.2 gold histograms. https://github.com/LDMX-Software/ldmx-sw/pull/1152

Merging those and then re-running validations of interest would then use the v3.2.2 gold histos.

AminaJLi (commented):
The fix for signal looks like it'll get the correct electron. I am a little nervous about using track ID 1 as the identifier for background. We've seen some discrepancies there for PN samples, but nothing as dramatic as it was with signal. I believe one of our undergrads made a comparison plot that I can add here. I'll ask around for it!

Is there a reason we're avoiding a momentum-based recoil selection?

omar-moreno:
> The fix for signal looks like it'll get the correct electron. I am a little nervous about using track ID 1 as the identifier for background. We've seen some discrepancies there for PN samples, but nothing as dramatic as it was with signal. I believe one of our undergrads made a comparison plot that I can add here. I'll ask around for it!
> 
> Is there a reason we're avoiding a momentum-based recoil selection?

For the PN samples, the electron track is always the first one to be fired so the track ID has to equal to 1 by design. If you are seeing otherwise, it would be interesting to track what happens with the incident electron.

AminaJLi:
> For the PN samples, the electron track is always the first one to be fired so the track ID has to equal to 1 by design. If you are seeing otherwise, it would be interesting to track what happens with the incident electron.

Oop, this may point to a misunderstanding of the software then. Would the recoil electron not be a separate simParticle from the beam/incident electron? We'd been operating with the idea that the interaction produces a "new" electron with a different initial momentum that gets stepped forward.

omar-moreno:
> > For the PN samples, the electron track is always the first one to be fired so the track ID has to equal to 1 by design. If you are seeing otherwise, it would be interesting to track what happens with the incident electron.
> 
> Oop, this may point to a misunderstanding of the software then. Would the recoil electron not be a separate simParticle from the beam/incident electron? We'd been operating with the idea that the interaction produces a "new" electron with a different initial momentum that gets stepped forward.

For PN samples, a "new" electron is not produced.  The production of a "new" electron only applies to samples generated using the DarkBrem physics list i.e. signal samples.

AminaJLi (approved):
Appreciate the clarification! This all seems good to me then! And certainly better than the previous setup.

bryngemark:
As far as I can tell, none of the changes in histograms failing KS test comparisons are attributable to the changes in PR but seem to be some previous changes made between 3.2.1 and current trunk. Comparing the tags and looking at release notes, I have a hard time telling e.g. why there seems to be fewer 0-hit events in TS now in ecal PN. I know filtering has been updated but I thought this would only relate to target. If we are now also filtering out events where the beam electron never makes it to the ecal, that could explain some of the differences seen here. In any case I think this is good to go now.

tomeichlersmith:
I'm confused as well... I'm going to try to rerun all the PR validation tests by converting this to draft and then marking it as "ready" again. Yesterday, I tried asking for it to rerun the PR validations after I merged the updated gold histograms into trunk but the histograms I'm looking at today are still v3.2.1.

tomeichlersmith (approved):
Manual converting-to-draft and then re-readying-for-review allowed for the PR tests to run against v3.2.2 and now everything passes :tada:

PR #1157, opened 2023-04-24
update ldmx-sw for GCC 11 support

### LINKED ISSUE #1133: Update to Catch2 v3
**Describe the bug**
Catch2 v2 does not support GCC 11 and so we need to update if we are going to update the OS in the development container.

**Context**: https://github.com/LDMX-Software/docker/pull/54#issuecomment-1335443663

A new major version of Catch2 means that many things will need to be updated. The biggest difference with Catch2 v3 is they transitioned away from single-header so we will need to update the headers (to keep all the features while improving compile time and run time performance). I have installed Catch2 v3.2.0 into the updated OS container and will update the CMake stuff to download the same version into the previous container.

**The v3.2.0 release should not wait for this issue to be resolved.**

## CMake Changes
- Don't have a FindCatch2 cmake script since Catch2 v3 has their own CMake integration
- Don't need to write the main since Catch2 v3 provided a `Catch2::Catch2WithMain` target

## Source Changes
- `Framework/catch.hpp` should be deleted
- `catch.hpp` should be replaced by `catch2/catch_test_macros.hpp`
- If `Approx` is needed, use `catch2/catch_approx.hpp` and `using Catch::Approx;`
- If `MatcherBase` is needed, use `catch2/matchers/catch_matchers.hpp` and change `Catch::MatcherBase` to `Catch::Matchers::MatcherBase`
- If `UnorderedEqual` is needed, use `catch2/matchers/catch_matchers_vector.hpp`
- If `Contains` is needed use `catch2/matchers/catch_matchers_contains.hpp`
- The `*_THROWS_WITH` macros now need to take a StringMatcher from `catch2/matchers/catch_matchers_string.hpp`

tomeichlersmith:

This resolves #1133 as well as other, new issues.

Oh boi is this a big one. In order to upgrade our container image to Ubuntu 22.04, we essentially need to upgrade our compiler to GCC 11. This is revealing many small bugs that need to be squashed in order to make this transition seamless. This draft PR is here to keep track of these changes. In tandem with this PR is the PR in the dev container build context https://github.com/LDMX-Software/docker/pull/54 .

## Change Overview
- Catch2 v2 does not support GCC 11 so we needed to upgrade to Catch2 v3.3.1+. This is the largest change and required updating header names and the CMake infrastructure for how Catch2 is downloaded in containers without it installed.
- The newer Python in Ubuntu 22.04 required changes to how our ConfigurePython class interacts with the CPython API. (see Framework PR for details).

### Related Sub-Module PRs

- will need updates to Tracking, currently its just on the HEAD of master which is not the same as the commit on ldmx-sw trunk
- https://github.com/LDMX-Software/Framework/pull/64
- https://github.com/LDMX-Software/Conditions/pull/17
- https://github.com/LDMX-Software/Ecal/pull/47
- https://github.com/LDMX-Software/Hcal/pull/59
- https://github.com/LDMX-Software/cmake/pull/16
- https://github.com/LDMX-Software/Tracking/pull/72

## To Do
- [x] check that this updated branch can still be built by older containers[^1]
    - [x] v3.0 (without detector id bindings)
    - [x] v3.1 (without detector id bindings)
    - [x] v3.2 (without detector id bindings)
    - [x] v3.3
    - [x] v3.4.0
    - [x] v3.5.0
- [x] make detector ID python bindings ON by default
- [x] make sure `fire` in `gdb` is still operational
- [x] document compatibility between ldmx-sw versions and container versions (and notes on why)

[^1]: I am checking this by disabling the testing (commenting out `build_test()` in `CMakeLists.txt`) and then compiling, installing, and running the inclusive validation sample. **Notice**: I am _not_ checking compatibility of the ldmx-env.sh script.

### Compiling, Installing, Running Test
```
ldmx clean src
ldmx cmake -B build -s .
ldmx cmake --build build --target install
ldmx 'LDMX_RUN_NUMBER=1 LDMX_NUM_EVENTS=100 fire .github/validation_samples/inclusive/config.py'
```

### DIFF
--- a/DetDescr/CMakeLists.txt
+++ b/DetDescr/CMakeLists.txt
@@ -14,7 +14,7 @@ setup_python(package_name ${PYTHON_PACKAGE_NAME}/DetDescr)
 # Setup the test
 setup_test(dependencies DetDescr::DetDescr)
 
-option(BUILD_DETECTORID_BINDINGS "Build the python bindings for the the DetDescr/DetectorID components" OFF)
+option(BUILD_DETECTORID_BINDINGS "Build the python bindings for the the DetDescr/DetectorID components" ON)
 if(BUILD_DETECTORID_BINDINGS)
   find_package(Python COMPONENTS Development)
 
--- a/DetDescr/test/DetectorIDTest.cxx
+++ b/DetDescr/test/DetectorIDTest.cxx
@@ -2,7 +2,7 @@
  * @file DetectorIDTest.cxx
  * @brief Test the operation of DetectorID class
  */
-#include "Framework/catch.hpp"  //for TEST_CASE, REQUIRE, and other Catch2 macros
+#include <catch2/catch_test_macros.hpp>
 
 #include <sstream>
 #include "DetDescr/DetectorID.h"  //headers defining what we will be testing
--- a/Packing/test/BinaryIO.cxx
+++ b/Packing/test/BinaryIO.cxx
@@ -1,7 +1,6 @@
+#include <catch2/catch_test_macros.hpp>
 
 #include "TTree.h"
-
-#include "Framework/catch.hpp"
 
 #include "Framework/Configure/Parameters.h"
 #include "Framework/Event.h"


EinarElen:
Here's to hoping this will fix the really annoying crashes with GDB on trying to print maps that the current ubuntu version at the same time!

EinarElen:
I have one request and one questions w.r.t. to this effort that I didn't think about when we discussed in the meeting earlier today. 

Request: Could you try running fire once in gdb just to make sure that it doesn't do anything weird? 

Thought: Since we are doing a change like this, should we consider enabling https://github.com/LDMX-Software/ldmx-sw/blob/1d6239b49687bc59c1efa60ad2a1fec5c0900027/DetDescr/CMakeLists.txt#L17 by default?

tomeichlersmith:
Enabling the Detector ID bindings by default is a good idea.

I can add running `fire` through `gdb` to my list of checks to make :+1:

tomeichlersmith:
I have rebased onto `trunk` and begun testing. The rebase was rather easy since noone else has touched the testing modules since I made this branch; however, juggling git submodules during a rebase is never fun.

PR #1135, opened 2022-12-05
Iss1117

EinarElen:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
This resolves #1117. This primarily means providing the length of a scintillator bar given an HcalID from the geometry condition. Since the side hcal for v14 has different length of bars depending on the layer number, this needs to be a function of both section and layer.

Example use would be 
```C++ 
auto barLength {hcalGeometry.getScintillatorLength(id); 
```

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments
- [x] I ran my developments and the following shows that they are successful.
To check that these changes didn't change the output of the condition, I added the following to the end of configure step
```C++
  auto section_to_string{[](const ldmx::HcalID id) { // Really wish we would have reflection in C++
    switch (id.section()) {
      case ldmx::HcalID::HcalSection::BACK:
        return "BACK";
      case ldmx::HcalID::HcalSection::TOP:
        return "TOP";
      case ldmx::HcalID::HcalSection::BOTTOM:
        return "BOTTOM";
      case ldmx::HcalID::HcalSection::LEFT:
        return "LEFT";
      case ldmx::HcalID::HcalSection::RIGHT:
        return "RIGHT";
    }
  }};
  std::cout << std::boolalpha;
  std::cout << "HcalGeo->Geometry has " << getNumSections()
            << " sections, a scintillator width of " << getScintillatorWidth()
            << "mm, Ecal dx/dy of " << getEcalDx() << "/" << getEcalDy()
            << " mm" << std::endl;
  for (auto section{0}; section < getNumSections(); ++section) {
    std::cout << "\tHcalGeo->Section: " << section_to_string({section, 1, 2})
              << " has " << num_layers_[section] << " layers\n";
    for (auto layer{1}; layer <= getNumLayers(section); ++layer) {
      // Number of strips per layer
      std::cout << "\tHcalGeo-> S: " << section << ", L: " << layer << " has "
                << getNumStrips(section, layer)
                << " strips, Zero Strip: " << getZeroStrip(section, layer)
                << ", halfTotalWidth: " << getHalfTotalWidth(section, layer)
                << " mm";
      if (section == ldmx::HcalID::HcalSection::BACK) {
        std::cout << " is horizontal? " << layerIsHorizontal(layer);
      }
      std::cout << std::endl;
      for (auto strip{0}; strip < getNumStrips(section, layer); ++strip) {
        ldmx::HcalID id{section, layer, strip};
[...]

PR #1160, opened 2023-05-10
Iss1136 hcaldqm

### LINKED ISSUE #1136: Hcal DQM Updates
Cristina and I discussed what we think would be good to have in terms of DQM figures for the Hcal other than what is there already which then could be checked with the new validation module.  Since reconstruction currently isn't working, we want both figures for simhits and rechits.

EinarElen:

This resolves #1136, but it does kind of a lot of things so I'll detail them all below!

This is a major rewrite of the existing HcalDQM to address the concern that the existing DQM was hard to interpret when there were actual differences between samples (e.g. v12 vs v14). It tries to separate DQM figures for the different Hcal sections as well as by orientation of the bars when applicable. See the [May 3rd software developer meeting](https://indico.fnal.gov/event/58824/) for some example figures. I've tested it both with a neutron particle gun setup as well as the EcalPN samples we use for validation today. 

One important detail here is that I made the HcalDQM analyzer operate at one hcal section at a time. This made the code a lot easier to read but it obviously _a bit inefficient_. From my tests it is still a negligible component w.r.t. the other parts of our validation suite. 

Next, we add one separate [Hcal DQM analyzer,](https://github.com/LDMX-Software/ldmx-sw/blob/iss1136-hcaldqm/DQM/include/DQM/HcalInefficiencyDQM.h) which is focused only on things related to inefficiency/efficiency. It does a basic check for the first layer with at least one vetoable hit in each section and produces a histogram which shows what parts of the detector were involved in vetoing an event. I'm pretty sure there are more things we could be looking at, but this is a good start. In principle, this could have been a part of the other analyzer but I chose to have it separate for two reasons 
- To keep the code cleaner, putting too many things in the same bucket was making things a bit messy to read 
- This analyzer is useful on its own outside the context of validation

We then also add both an Hcal and a PNDQM component to the Validation module. When doing this, I found that I needed two additional features from the module so they are also included here. The first one just adds a command line switch that lets you choose what kind of extension you want for the figures you generate. I love PDF but google slides doesn't play nice with them. Next for categorization figures like the hcal efficiency figure or the PNDQM event category figure, it is much nicer to have labels for the bins. So I added a keyword for adding those. I am certain that these two features could be added in a better way, but this works at least. 

Finally, in the PN DQM there was a minor copy&paste issue that has been patched. This will mean that the PN DQM checks will fail :) 

Screenshot from 2023-05-03 15-15-02

I'm currently recovering from a really nasty fever so... I'm not sure that this was super clearly written, let me know if anything needs clarification.

## Check List

N/A although it relies on the results of https://github.com/LDMX-Software/ldmx-sw/pull/1135 which in turns relies on https://github.com/LDMX-Software/Hcal/pull/63 and https://github.com/LDMX-Software/SimCore/pull/92

-


bryngemark (approved):
This indeed looks like a major overhaul and an impressive amount of work that is hard to digest in code itself. It looks like you've been using most of it extensively to make plots which I take as a good sign that it works as in both runs and does what you want. 

If anything, I was hoping you'd take the opportunity to rename the hopeless event categories (`count_a` etc) to something one can actually parse. If it's too much work/too hard to come up with a scheme then I won't insist.

EinarElen:
I wouldn't mind doing that at all, they have been bothering me for some time as well :)

EinarElen:
Ok, I'm partway through rewriting the classification code and I think I've found an issue already. I'll document any other I find in this comment

The category 12 has the label  `"#pi_{0} #pi A", // 12`. It is matched as part of this block
```C++ 
  if (pi0 == 1) {
    if (count_f == 0)
      return 6;  // 1 pi0
    else if ((n == 1) && (count_i == 0))
      return 10;  // 1pi0 1 p
    else if ((n == 2) && (count_i == 0))
      return 11;  // 1pi0 1 p
    else if ((p == 1) && (count_j == 0))
      return 10;  // 1pi0 1 n
    else if ((p == 2) && (count_j == 0))
      return 11;
    else if ((n == 1) && (p == 1) && (count_o == 0))
      return 11;
    else if ((pi == 1) && ((p == 1) || (n == 1)) && (count_p == 0))
      return 12;
  }
```

If we remove the parts that are dealing with `pi0 + 1A (10)` or `pi0 + 2A (11)` 

```C++
int count_p = exotic + kaons;
// lots of code
if (pi0 == 1) {
    if ((pi == 1) && ((p == 1) || (n == 1)) && (count_p == 0))
      return 12;
  }
```
This is trying to match "1 charged pion, one nucleon, and one neutral pion". However, the `(p == 1) || (n == 1)` matches one nucleon and an arbitrary number of nucleons of the other type. So events that should go into multibody end up misclassified here.

EinarElen:
If we want to fix this bug, which I think we do, I think we should take the opportunity while we are doing backwards-incompatible things by also deciding on how to handle the kaons. The current labels in the figure for the three kaon categories are
```C++
                                     "K^{0}_{L} X",   // 16
                                     "K X",           // 17
                                     "K^{0}_{S} X",   // 18
```
and they are matched here 
https://github.com/LDMX-Software/ldmx-sw/blob/1d6239b49687bc59c1efa60ad2a1fec5c0900027/DQM/src/DQM/PhotoNuclearDQM.cxx#L388-L390

The problem here is that it isn't unlikely that you have an event with more than one hard kaon. In that case, this will assign it depending on the exact order of those if-statements, which can make for a misleading figure (more K0L events than K0s events even if they are produced with the same probability). In my view, what these categories are supposed to represent are "One hard kaon + possibly other things". The easiest way to update this would be to require only one _hard_  kaon to match one of these three. Alternatively, they could be defined similarly to single neutron/single charged pi/etc. That would be a much bigger change though.

EinarElen:
I started out adding an enumerator for the different event types so we don't have to rely returning magic numbers. Currently, I made it a unscoped enum because adding a scoped one would require some more tweaking, but I'll update that before we merge this. 

For the `count_N` part,  all of the different `count_N` parts were effectively `counts of not X`. Since we only care about low multiplicities, I think it is much more clear to work based on the multiplicity directly. In other words, instead of defining "a proton and a neutron" as 1 proton, 1 neutron, 0 pions, 0 kaons, 0 exotics, start by separating the categories into how many hard particles they have. Then "a proton and a neutron" is just 1p1n in the 2 hard particles grouping. 

I' ve added a rewrite of the counting logic as this
https://github.com/LDMX-Software/ldmx-sw/blob/545cb88a349672535ecca65bc1faaca7ae13f6ac/DQM/src/DQM/PhotoNuclearDQM.cxx#L319-L390

Note: This does not contain either of the two fixes suggested above. 

To make sure that this rewrite didn't change any behaviour, I dumped the classifications for 10k EcalPN events before and after the change and checked that the file contents were identical.

EinarElen:
I am really not understanding what the "1n_event_type" histogram is supposed to represent. The matching is done here 

https://github.com/LDMX-Software/ldmx-sw/blob/1d6239b49687bc59c1efa60ad2a1fec5c0900027/DQM/src/DQM/PhotoNuclearDQM.cxx#L260-L270

within this condition 

https://github.com/LDMX-Software/ldmx-sw/blob/545cb88a349672535ecca65bc1faaca7ae13f6ac/DQM/src/DQM/PhotoNuclearDQM.cxx#L227-L235

In other words, we are looking at either a single neutron, charged kaon, klong, kshort, or two neutron event and the `pnDaughters` vector is sorted by kinetic energy. We then sort based on the properties of the highest energy particle. But the "1" category which is labeled as (nn) will match any of the events if the neutron happens to have the highest kinetic energy. 

e.g. an event with 1 neutron and 1 klong would be considered either (nn) or (n K0) depending on which one has the highest energy.

EinarElen:
I am really not understanding what the "1n_event_type" histogram is supposed to represent. The matching is done here 

https://github.com/LDMX-Software/ldmx-sw/blob/1d6239b49687bc59c1efa60ad2a1fec5c0900027/DQM/src/DQM/PhotoNuclearDQM.cxx#L260-L270

within this condition 

https://github.com/LDMX-Software/ldmx-sw/blob/545cb88a349672535ecca65bc1faaca7ae13f6ac/DQM/src/DQM/PhotoNuclearDQM.cxx#L227-L235

In other words, we are looking at either a single neutron, charged kaon, klong, kshort, or two neutron event and the `pnDaughters` vector is sorted by kinetic energy. We then sort based on the properties of the highest energy particle. But the "1" category which is labeled as (nn) will match any of the events if the neutron happens to have the highest kinetic energy. 

e.g. an event with 1 neutron and 1 klong would be considered either (nn) or (n K0) depending on which one has the highest energy. 

> I am really not understanding what the "1n_event_type" histogram is supposed to represent. The matching is done here
> 
> https://github.com/LDMX-Software/ldmx-sw/blob/1d6239b49687bc59c1efa60ad2a1fec5c0900027/DQM/src/DQM/PhotoNuclearDQM.cxx#L260-L270
> 
> within this condition
> 
> https://github.com/LDMX-Software/ldmx-sw/blob/545cb88a349672535ecca65bc1faaca7ae13f6ac/DQM/src/DQM/PhotoNuclearDQM.cxx#L227-L235
> 
> In other words, we are looking at either a single neutron, charged kaon, klong, kshort, or two neutron event and the `pnDaughters` vector is sorted by kinetic energy. We then sort based on the properties of the highest energy particle. But the "1" category which is labeled as (nn) will match any of the events if the neutron happens to have the highest kinetic energy.
> 
> e.g. an event with 1 neutron and 1 klong would be considered either (nn) or (n K0) depending on which one has the highest energy.

After having a quick chat with Omar, I'm going to change this to only apply to the single neutron case and then check what the subleading particle is. I'm also adding a case for "other" so that there's an entry for each 1n event. I also added neutron multiplicity, total PN kinetic energy, and total kinetic energy from neutrons since I found those to be quite useful when looking at individual event categories.

EinarElen:
Ok, at commit https://github.com/LDMX-Software/ldmx-sw/pull/1160/commits/8ca6f40ce4a5748538de24e75b38d192b4d91fab I've tested again that no changes have been made to the event categories with 10k Ecal PN events. After this, I will be applying the following breaking changes. These should make validation fail for the corresponding histograms. 

- By default, count light ions as potential hard particles. This is configurable so you can still get the old behavior, but I do think it it makes the categories more realistic. An event with a 300 MeV neutron and a 1 GeV deutron probably shouldn't be classified as a single neutron event. 
- Rewrite what the `1n_event_type` histogram does. I think this was just a bug before, it is supposed to deal with categorizing the single neutron events but was looking at both 1n, kaon, and 2n events. I've changed it to only apply to 1n events. I've also added an "other" category and shifted the placement of each bin to the left. 
- Correct the 1pi01pi1A category so that it checks for exactly one nucleon, not 1 proton or 1 neutron (which could match an arbitrary number of nucleons).
- Make the kaon categories require exactly 1 hard kaon so that we don't bias for the category that happens to come first in the classification code 



On a complete sidenote, I think we should add a `getKineticEnergy` method to simparticles, the PN DQM code is littered with `energy = particle.getEnergy() - particle.getMass()` and confusing energy and kinetic energy with simparticles has tripped up multiple people.

tomeichlersmith (approved):
I like this :1st_place_medal:

EinarElen:
Failed ecal PN histograms are the expected ones
```
  PN_PN_1kp_energy_frac.pdf
  PN_PN_pn_particle_mult.pdf
  PN_PN_event_type.pdf
  PN_PN_event_type_compact_2000mev.pdf
  PN_PN_1k0_ke_2nd_h_ke.pdf
  PN_PN_1kp_ke_2nd_h_ke.pdf
  PN_PN_1k0_energy_frac.pdf
  PN_PN_1kp_energy.pdf
  PN_PN_1kp_energy_diff.pdf
  PN_PN_1n_event_type.pdf
  PN_PN_1k0_energy_diff.pdf
  ```

**Comment by EinarElen** [comment]
@tomeichlersmith do you have any idea why "This branch cannot be rebased due to conflicts" shows up here? As far as I can see there aren't any and nothing I do in the local repo seems to think there are any conflicts

**Comment by EinarElen** [comment]
https://github.com/LDMX-Software/ldmx-sw/compare/trunk...iss1136-hcaldqm 

says the branches are able to merge (┛ಠ_ಠ)┛彡┻━┻

**Comment by tomeichlersmith** [comment]
Processing...
```
> git diff --name-status trunk 
M       DQM/include/DQM/HCalDQM.h
A       DQM/include/DQM/HcalInefficiencyDQM.h
M       DQM/include/DQM/PhotoNuclearDQM.h
M       DQM/python/dqm.py
M       DQM/src/DQM/HCalDQM.cxx
A       DQM/src/DQM/HcalInefficiencyDQM.cxx
M       DQM/src/DQM/PhotoNuclearDQM.cxx
M       DetDescr/src/DetDescr/HcalGeometry.cxx
M       Hcal
M       SimCore
M       Validation/src/Validation/__init__.py
M       Validation/src/Validation/__main__.py
M       Validation/src/Validation/_differ.py
A       Validation/src/Validation/hcal.py
A       Validation/src/Validation/photonuclear.py
```
I'm guessing the inclusion of modified submodule commit refs is causing the issue. I'd try changing them to what's on trunk (which has changed compared to when you started this branch) and then see if the rebase can be done automatically.
```
git checkout trunk -- Hcal SimCore
git commit -m "align submodules with trunk"
```
I tried this and then tried to rebase `git rebase` which failed due to both `trunk` and this branch having changed the `HcalGeometry` class
```
eichl008@spa-cms017 ~/ldmx/ldmx-sw> git rebase trunk
First, rewinding head to replay your work on top of it...
Applying: Update comment to align with earlier code
Using index info to reconstruct a base tree...
M	DetDescr/src/DetDescr/HcalGeometry.cxx
Falling back to patching base and 3-way merge...
Auto-merging DetDescr/src/DetDescr/HcalGeometry.cxx
CONFLICT (content): Merge conflict in DetDescr/src/DetDescr/HcalGeometry.cxx
Failed to merge in the changes.
Patch failed at 0001 Update comment to align with earlier code
The copy of the patch that failed is found in:
   /export/scratch/users/eichl008/ldmx/ldmx-sw/.git/rebase-apply/patch

When you have resolved this problem, run "git rebase --continue".
If you prefer to skip this patch, run "git rebase --skip" instead.
To check out the original branch and stop rebasing, run "git rebase --abort".

eichl008@spa-cms017 ~/ldmx/ldmx-sw> git diff DetDescr/
diff --cc DetDescr/src/DetDescr/HcalGeometry.cxx
index 4585099,61a0665..0000000
--- a/DetDescr/src/DetDescr/HcalGeometry.cxx
+++ b/DetDescr/src/DetDescr/HcalGeometry.cxx
@@@ -150,12 -97,12 +150,21 @@@ void HcalGeometry::buildStripPositionMa
          double stripcenter = (strip + 0.5) * scint_width_;
  
          if (hcalsection == ldmx::HcalID::HcalSection::BACK) {
++<<<<<<< HEAD
 +          /**
 +             For back Hcal:
 +             - layers in z
 +             - strips occupy thickness of scintillator in z (e.g. 20mm)
 +             - strips orientation is in x(y) depending on back_horizontal parity
 +          */
++=======
+         /**
+            For back Hcal:
+            - layers in z
+            - strips occupy thickness of scintillator in z (e.g. 20mm)
+            - strips orientation is in x(y) depending on horizontal parity
+         */
++>>>>>>> Update comment to align with earlier code
            // z position: zero-layer(z) + layer_z + scint_thickness / 2
            z = zero_layer_.at(section) + layercenter;
  
eichl008@spa-cms017 ~/ldmx/ldmx-sw> 
```
So I'm guessing you have commits that are already on `trunk` from your earlier PR.

**Comment by tomeichlersmith** [comment]
I could use my unlimited admin powers to due a non-rebase merge commit locally and then push to trunk - the reason I want to avoid doing that is the same reason I disabled those types of merges for the repository in GitHub: I think it pollutes the commit history, making it more difficult to understand what changed when looking at the commits later.

All that being said, rebasing is a lot more confusing than merging so I understand if it can't be done occasionally (especially on large refactors).

**Comment by EinarElen** [comment]
What fun. Ok, lesson learned on rebasing often :)

**Comment by tomeichlersmith** [comment]
`git rebase -i` is your friend - if you can pinpoint which commits are already on trunk, then you can remove them from the rebase plan and hopefully skip the conflicts.

**Comment by tomeichlersmith** [comment]
Alright, I tried my own advice for a minute and quickly realized that this would be a big slog for not much benefit. I am going to do a non-rebase merge commit locally and push to trunk - for future admin reference:
```
git checkout <branch>
git merge trunk
git checkout trunk
git merge <branch>
git push origin trunk
```

:eyes: Now we get to see if GitHub will notice that I merged it manually or not.

:tada: GitHub automatically labeled the PR as merged :tada:

PR #1162, opened 2023-05-10
Sim core iss86

### LINKED ISSUE #1161: Event topology based PN filters + Alternative PN models
The main work here is over in https://github.com/LDMX-Software/SimCore/pull/88 and the corresponding issue https://github.com/LDMX-Software/SimCore/issues/86. However, in order for some of the models in there to be useful, you want to have some filters that veto events that don't have the kind of final state that you are interested in. This can happen for example if you have a PN reaction with a nucleus where you don't apply your model. 

To make adding such filters straight forward a generic filter ABC for looking at a PN event and based on the secondaries either accept or reject. I've included two such filters for the nothing hard/single neutron case. There isn't a need for a filter corresponding to the https://github.com/LDMX-Software/SimCore/blob/3aea21f15567975c86ee267753cf01749f6884f0/include/SimCore/PhotoNuclearModels/BertiniAtLeastNProductsModel.h model since the current PN products filter does this.

Chronologically, this should come after #1135 and #1160.

EinarElen:

This resolves #1161. Most of the details are over in SimCore, but this does add a new filter category that lets you veto events where you had a PN reaction but it didn't produce a final state that you are interested in. This includes two such filters, but adding new ones is trivial and shouldn't really require more code than e.q. 
```C++
bool SomeFilter::rejectEvent(
    const std::vector<G4Track*>& secondaries) const {
  for (const auto& secondary : secondaries) {
    // Get the PDG ID of the track
    const auto pdgID{
        std::abs(secondary->GetParticleDefinition()->GetPDGEncoding())};
    if (skipCountingParticle(pdgID)) {
      continue;
    }
    auto energy{secondary->GetKineticEnergy()};
   // Perform some bookkeeping 
  }
  // Decide on or not to reject the event
  return some_veto_condition;
}
```

## Check List

See the SimCore PR

### Related Sub-Module PRs

- https://github.com/LDMX-Software/SimCore/pull/88

### DIFF
--- a/Biasing/include/Biasing/PhotoNuclearTopologyFilters.h
+++ b/Biasing/include/Biasing/PhotoNuclearTopologyFilters.h
@@ -1,0 +1,124 @@
+#ifndef PHOTONUCLEARTOPOLOGYFILTERS_H
+#define PHOTONUCLEARTOPOLOGYFILTERS_H
+
+/*~~~~~~~~~~~~~*/
+/*   SimCore   */
+/*~~~~~~~~~~~~~*/
+#include "SimCore/UserAction.h"
+#include "SimCore/UserTrackInformation.h"
+/*~~~~~~~~~~~~~~~*/
+/*   Framework   */
+/*~~~~~~~~~~~~~~~*/
+#include <G4RunManager.hh>
+#include <G4Step.hh>
+
+#include "Framework/Configure/Parameters.h"
+// Forward declaration
+
+namespace biasing {
+
+/**
+ * Abstract base class for a user action used to filter out photo-nuclear events
+ * that don't match the topology the user is interested in studying. Derived
+ * classes implement the event rejection
+ *
+ * Similar to the PhotoNuclearProductsFilter, this user action will only process
+ * steps whose associated track has been tagged as a "PN Gamma". This tag is
+ * currently only set in ECalProcessFilter and needs to be placed in the
+ * UserAction pipeline before this class.
+ *
+ */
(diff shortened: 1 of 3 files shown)


tomeichlersmith:
The "conflicts" are due to the commits that were merged into `trunk` not via rebase and are here. You could try `git rebase -i` and remove the commits already merged, or just create a new branch.
```
git checkout trunk
git pull
git branch -D SimCore-iss86 # delete local copy of this branch
git checkout -b SimCore-iss86 # create new branch off trunk with same name
# make desired changes, probably updating SimCore
cd SimCore
git checkout trunk
git pull
cd ..
git add SimCore
git commit -m "update SimCore with new PN model infrastructure"
# forcefully push your new local branch to this one, setting this one as its upstream
git push -fu origin trunk
```

tomeichlersmith:
*Bad news*: The PR validation was run _before_ the updated gold histograms were merged into trunk so they do not faithfully check against the new histograms.

*Good news*: It looks like the exact same histograms failed as in PR #1160 which would imply that this PR does not change the PN DQM (which is expected and what we would want).

I am going to ask for the workflow to be "rerun" - although I fear that it will use the old trunk commit in order to do a "perfect" rerun.

EinarElen:
If we need to change something to trigger a rerun, we could add Lenes suggested change for the default TS collection name (I'll test it locally first of course)

tomeichlersmith (approved):
Good work, this looks great :)

For my future reference: I confirmed that SimCore is on tag v1.5.0

bryngemark (approved):

PR #1173, opened 2023-06-20
resolve memory sloppyness and solidify Python access

tomeichlersmith:

This resolves an issue seen when testing the updated OS image. https://github.com/LDMX-Software/docker/issues/59

## Check List
  - [x] v3.5.0 <- tested by CI
  - [x] v3.4.0
  - [x] v3.3
  - [x] v3.2  (with `-DBUILD_DETECTORID_BINDINGS=OFF`)
  - [x] v3.1  (with `-DBUILD_DETECTORID_BINDINGS=OFF`)
  - [x] v3.0 (with `-DBUILD_DETECTORID_BINDINGS=OFF`)
    - successful means passing the tests in this case

### Related Sub-Module PRs
- https://github.com/LDMX-Software/Framework/pull/67
- https://github.com/LDMX-Software/Framework/pull/68


EinarElen:
On the tracing feature, I feel that ends up making an already relatively hard to read part of our codebase even harder. Would doing something like what I did in https://github.com/LDMX-Software/Framework/issues/66 be an alternative?

tomeichlersmith:
That's a good point - it does make the code pretty messy looking. I can implement the update you suggested to replace the tracing option.

Issue #1168, opened 2023-06-12
format source files


tomeichlersmith:
**Describe the bug**
A lot of the C++ source files do not conform to the Google format. This is unsurprising since we don't have any auto-formatting tools _and_ we don't document how to event apply the formatting.

The soon-to-be-released v4.0 of the dev container will contain `clang-format` which means we could introduce a `ldmx` subcommand for formatting source files `ldmx format`.

I have something that works, but the problem is it changes a lot of the files across many submodules and I think a purely formatting change should be done separately.

```bash
###############################################################################
# __ldmx_format
#   format the C++ code with clang-format in the container (requires v4.0.0)
#   uses `find` to run it on all *.h and *.cxx files in ldmx-sw
###############################################################################
__ldmx_format() {
  __ldmx_run . 'if command -v clang-format &> /dev/null; then find ${LDMX_BASE}/ldmx-sw -type f -name "*.h" -o -name "*.cxx" -exec clang-format --style=Google -i {} \;; else echo "ERROR: no clang-format in container"; fi'
  return $?
}
```

We could probably update this to take some directory or file path so that it doesn't always do the entire ldmx-sw code base.

EinarElen:
Adding a list of things that have definitely been formatted by now 
- [x] SimCore https://github.com/LDMX-Software/SimCore/pull/95 
- [ ] Hcal
  - [x] Event library 
  - [x] Veto processor 
- [ ] DQM 
  - [x] Hcal 
  - [x] Photonuclear 
  - [ ] Verifiers
- [ ] TrigScint 
  - [ ] ~50% https://github.com/LDMX-Software/TrigScint/pull/54

bryngemark:
did we settle on if we wanted to apply formatting at the end of successful CI tests in a PR? and do we in that case need to check building/rerun tests afterwards? 

another option is to add code formatting to the PR checklist.

EinarElen:
I don't think we settled on anything, but since all the source files in SimCore have been formatted there is a CI check in SimCore that gives an error on unformatted source files https://github.com/LDMX-Software/SimCore/blob/trunk/.github/format-check

When it comes to rebuilding/testing after applying clang-format in a PR, I don't think we need to do this.

EinarElen:
Here is the current formatting status of the different submodules: 
I got this by
1: Copying the .clang-format file from SimCore 
```shell
find include/ -type f >  ${TMPDIR:-/tmp}/files-to-format.list
[...]

PR #1174, opened 2023-06-22
patch location of gamma->mumu conversions

horohoo:

This resolves an issue previously shown at a [sw-development meeting](https://indico.fnal.gov/event/58826/) (viewing this requires Indico access) in which muon conversion simulations create hard brem photons upstream of the target.

## Check List

gammamumu_validation_1

The start z position of the hard brem before (left) and after (right) the change is implemented.

Screenshot 2023-06-22 at 11 12 53 AM

The start z position of the daughter particles of the hard brem before (left) and after (right) the change is implemented for a target gamma --> mu+ mu- simulation.

gammamumu_validation_2

The start z position of the daughter particles of the hard brem after the change is implemented for an Ecal gamma --> mu+ mu- simulation with a biasing factor of 3e4.

A short presentation on this change will be given at a future sw-development meeting.

### DIFF
--- a/Biasing/python/ecal.py
+++ b/Biasing/python/ecal.py
@@ -74,3 +74,56 @@ def photo_nuclear( detector, generator ) :
     ])
 
     return sim
+
+def gamma_mumu(detector, generator) :
+    """Example configuration for biasing gamma to mu+ mu- conversions in the ecal.
+
+    In this particular example, 4 GeV elecrons are fired upstream of the
+    tagger tracker. The TargetBremFilter filters out all events that don't
+    produce a brem in the target with an energy greater than 2.5 GeV.
+
+    Parameters
+    ----------
+
+    detector : str
+        Path to the detector
+
+    Returns
+    -------
+    Instance of the sim configured for target gamma to muon conversions.
+
+    Example
+    -------
+
+        ecal_mumu_sim = ecal.gamma_mumu('ldmx-det-v12')
+
+    """
+
+    # Initiate the sim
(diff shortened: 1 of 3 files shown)


tomeichlersmith (approved):
The developments here look good and they are decoupled from other developments being merged into trunk so there aren't _merge_ issues but the rebase is struggling to reapply a commit here that is already present on trunk. **I am going to manually rebase onto trunk**.

For future reference, any new developments should be done on its own branch which is newly created off the most recent trunk.
```
git switch trunk
git pull
git switch -c <branch-name>
```
This will help avoid merge/rebase issues if someone else is developing something in the same area as you. As far as what `<branch-name>` should be, I think it makes sense to use the `<issnumber>-<title>` format (e.g. `1168-format-source-files` for working on https://github.com/LDMX-Software/ldmx-sw/issues/1168) so that folks know where to go for more information. If there isn't an issue related to the work, then it is especially important to use a descriptive branch name. In this case, I would name this branch something like `gamma-mumu-location-patch`.

tomeichlersmith:
I have now rebased these developments onto the current `trunk`. This has caused the commits to be re-written so you will need to update your local branch in a special way to avoid creating extra unnecessary commits.
```
git switch trunk
git branch -D horoho-dev # delete old, local branch
git fetch # download new branches from remote
git switch horoho-dev # switch to new, downloaded branch
```

_Edit_: `not` -> `now`

mrsolt (approved):

PR #1178, opened 2023-07-18
Add new CaloCluster event class

### LINKED ISSUE #1177: Common calorimeter cluster class
Currently there are parallel classes in Ecal (EcalCluster.cxx/h) and Hcal (HcalCluster.cxx/h) that unnecessarily duplicate common code and functionality.

I propose a common CaloCluster.cxx/h class that can implement the core functionality relevant for each use case and that both EcalCluster and HcalCluster can inherit from.

As a first step, the CaloCluster class should be added to Recon/Event. After this is merged, changes can be made to the Ecal and Hcal cluster objects so as to inherit from the new base class.

- [ ] Add CaloCluster to Recon/Event
- [ ] Update EcalCluster to inherit from CaloCluster in the `Ecal` submodule
- [ ] Update HcalCluster to inherit from CaloCluster the `Hcal` submodule

Developing for these two parallel collections is becoming painful; the common solution seems preferable.

therwig:
I'm adding a new CaloCluster class that EcalCluster and HcalCluster can inherit from, in order to reduce code duplication and increase usability of existing tools aiming at one class or the other.

This is the first step needed to resolve #1177 

## Check List

I re-ran clustering locally with an EcalCluster prototype based on this base class. Results were unchanged wrt the previous implementation.

I leave updates to the submodules for future work/reviews, and will follow that progress in the issue reference above.

### DIFF
--- a/Recon/CMakeLists.txt
+++ b/Recon/CMakeLists.txt
@@ -18,6 +18,8 @@ if(BUILD_EVENT_ONLY)
 
   register_event_object(module_path "Recon/Event" namespace "ldmx" 
                         class "CalorimeterHit" type "collection" )
+  register_event_object(module_path "Recon/Event" namespace "ldmx" 
+                        class "CaloCluster" type "collection" )
   register_event_object( module_path "Recon/Event" namespace "ldmx" 
                          class "TriggerResult" )
   register_event_object( module_path "Recon/Event" namespace "ldmx" 
--- a/Recon/include/Recon/Event/CaloCluster.h
+++ b/Recon/include/Recon/Event/CaloCluster.h
@@ -1,0 +1,165 @@
+/**
+ * @file CaloCluster.h
+ * @brief Class that stores calorimeter cluster information
+ */
+
+#ifndef EVENT_CALOCLUSTER_H_
+#define EVENT_CALOCLUSTER_H_
+
+// ROOT
+#include "TObject.h"  //For ClassDef
+#include "TString.h"
+
+// STL
+#include <iostream>
+#include <set>
+
+// ldmx-sw
+#include "Recon/Event/CalorimeterHit.h"
+
+namespace ldmx {
+
+/**
+ * @class CaloCluster
+ * @brief Stores cluster information from the ECal
+ */
+class CaloCluster {
+ public:
+  /**
+   * Class constructor.
+   */
+  CaloCluster();
+
+  /**
+   * Class destructor.
+   */
+  virtual ~CaloCluster();
+
+  /**
+   * Print a description of this object.
+   */
+  void Print() const;
+
+  /**
+   * Reset the CaloCluster object.
+   */
+  void Clear();
+
+  /**
+   * Take in the hits that make up the cluster.
+   * @param hit The digi hit's entry number in the events digi
+   * collection.
+   */
+  void addHits(const std::vector<const ldmx::CalorimeterHit*> hitsVec);
+
+  /**
+   * Sets total energy for the cluster.
+   * @param energy The total energy of the cluster.
+   */
+  void setEnergy(double energy) { energy_ = energy; }
+  
+  /**
+   * Sets total number of hits in the cluster.
+   * @param nHits The total number of hits.
+   */
+  void setNHits(int nHits) { nHits_ = nHits; }
+  
+  /**
+   * Sets a sorted vector for the IDs of the hits
+   * that make up the cluster.
+   * @param IDs Sorted vector of hit IDs.
+   */
+  void setIDs(std::vector<unsigned int>& hitIDs) { hitIDs_ = hitIDs; }
+
+  void setHitValsX(std::vector<float>& x) { hitX_ = x; }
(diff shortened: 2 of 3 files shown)


therwig:
FWIW even though many tests "failed" this appears to be due to the KS metric being too tight. Here's one random failing histogram from the PN sample, for example.  (I guess since the events are generated with a different random seed they aren't identical?) My changes shouldn't affect any of these plots, so I think the failures are unrelated to the PR.

[EcalShowerFeatures_EcalShowerFeatures_deepest_layer_hit.pdf](https://github.com/LDMX-Software/ldmx-sw/files/12083807/EcalShowerFeatures_EcalShowerFeatures_deepest_layer_hit.pdf)

EinarElen:
I can't comment directly on the failing ecal plot, but the seed should be identical for the two tests. But yeah it seems strange that something like the hcal sim energy would change otherwise

therwig:
@EinarElen ok interesting. Does this pipeline test usually work? The PR literally just adds a new (unrelated) event class so I really don't think the failing plots can be related to the changes here.

EinarElen:
Yeah usually it works well. But yeah the failure here can't be from the changes here (unless there's something really obscure going on), so I think there's something happening on the CI side. Maybe the most recent gold histograms were mixed up with an earlier version? A very quick check for that would be to just run a small test manually once with current trunk and once with the branch. 

If nobody has figured it out by tomorrow, I'll give that a try :)

therwig:
Hm yeah I don't see the same pipelines run for the past 2 PRs (but maybe I'm not looking in the right place?) and the last commit from @tomeichlersmith updated the reference histograms so my bet is some kind of CI issue.

tomeichlersmith:
>  (I guess since the events are generated with a different random seed they aren't identical?)

The validation samples _should_ be running with the same random seed and therefore should be generating identical events.[^1]
The reason the KS check is so tight is because these events should be identical, but we can look into loosening that cut if some other bug isn't discovered.

[^1]: there do exist some breaks in this. The TS subsystem manages its [own random seed in some processors](https://github.com/LDMX-Software/TrigScint/issues/20). Also the ECal MIP tracking SVD algorithm [depends on some randomness](https://github.com/LDMX-Software/Ecal/issues/29).

EinarElen:
Hcal as well, but the changes here aren't actually used anywhere right?

tomeichlersmith:
I was able to confirm that `v3.2.6-2-5e05b82` (trunk) is different from the gold histograms in the inclusive validation sample. This is telling me that something fishy happened with the golden histogram files as expected. Will check against v3.2.6 to make sure.


## Set Up
The `git describe --tags` is to make sure that I'm only 2 commits ahead of v3.2.6.
```
. scripts/ldmx-env.sh
ldmx pull dev latest
git describe --tags
ldmx clean src
ldmx cmake -B build -S .
ldmx cmake --build build --target install -- -j4
# set container env variables to match CI
ldmx setenv LDMX_RUN_NUMBER=1
ldmx setenv LDMX_NUM_EVENTS=10000
```

## Running
Going to do `inclusive` since its the shortest.
```
cd .github/validation_samples/inclusive
ldmx fire config.py
ldmx python3 ${LDMX_BASE}/ldmx-sw/.github/actions/validate/compare.py \
  gold.root v3.2.6 \
  hist.root $(git describe --tags)
```

tomeichlersmith:
It does appear to be an issue with how the golden histograms were generated. They do not match the manually generated histograms for the same tag.
[plots.tar.gz](https://github.com/LDMX-Software/ldmx-sw/files/12088355/plots.tar.gz)

tomeichlersmith:
The run which produced the golden histograms appears to have cloned the correct commit of ldmx-sw (deduced from looking at the logs).

https://github.com/LDMX-Software/ldmx-sw/actions/runs/5349505801

I fear that this is revealing an actual code bug but I have no idea what it could be...

tomeichlersmith:
Alright, I found a difference between the environment that produced the histograms and the environment producing them now.

The run producing the histograms is using `ldmx/dev:v3.5.0`. I was able to look this up by copying out the digest from the run logs
```
Digest: sha256:1bf3f3aa946eb91f3b1bf118ee7f5e74a61e64b59b32b11eb1878cd026ec68ab
```
and then looking at `docker image ls --digests ldmx/dev` on my computer to see which tag had this digest.

I am going to run the inclusive sample on `trunk` but in `ldmx/dev:v3.5.0` and if they pass the cuts then I will make a new patch release to recreate the histograms with `ldmx/dev:v4.0.0`.

I forget where @EinarElen pointed this out, but I vaguely remember you mentioning that changing the compiler version may affect how the RNG behaves which seems to be the case.

EinarElen:
Yep, that can happen :) Great for reproducibility

tomeichlersmith:
Rerunning `trunk` with `ldmx/dev:v3.5.0` did match the golden histograms, so I made a new patch release and we can mark this as "Ready for Review" after those new histograms are generated and merged so that these tests can pass.

tomeichlersmith (approved):

Issue #1166, opened 2023-05-31
Memory bug patching/ASAN


EinarElen:
List of memory issues caught by ASAN/UBSAN and their fix-status
- [x] PN DQM missing &
- [x] Ecal veto buffer overflow missing `.size()` 
- [x] https://github.com/LDMX-Software/Hcal/issues/58 
- [x] Ecal veto mapsx/mapsy
- [x] Undefined `noise` entry in calorimeter hits
- [x] Missing case in hcal scintillator orientation
- [x] SimSpecialID UB? Probably not worth dealing with

**Describe the bug**
A missing `&` in one of the helper functions in the PN DQM code meant that the particle map was copied rather than referenced which made any pointers to simparticles therein invalid after the scope of the function. This bug hasn't had any impact on the DQM output, if it did we would have seen some difference. Regardless, we should of course patch it. 

The culprit is here
https://github.com/LDMX-Software/ldmx-sw/blob/0079ce00ff851fd427f2191c45c72d7857c37c17/DQM/include/DQM/PhotoNuclearDQM.h#L96-L110 

Where the map is supposed to be passed in by const ref. 

I caught this while trying to diagnose a completely different issue by running with address sanitizer enabled. I'm wondering if it would be useful to include these kinds of tools directly in our DQM jobs. At least ASAN has a minimal performance overhead. UBSAN could be interesting, but it doesn't by default kill a job if it catches something (it has some false positives) so you would only benefit from it if you read the logs.

EinarElen:
Caught another one in the Ecal veto/geometry code. When we are trying to fill the isolated hit map, we assume the list of neighbouring cells that we get from the Ecal geometry object to always have 6 entries. However, I'm seeing some events where you get a smaller number of neighbours (e.g. 4). This makes 
` cellNbrIds[k] = ldmx::EcalID(id.layer(), cellNbrIds[k].module(),  cellNbrIds[k].cell());` do a buffer overflow. 

I'm not sure if the assumption that there are always 6 neighbours is correct, if so there's another bug in some other place (e.g. the ecal geometry code). If not, the solution is to just loop over the size of the vector. @tomeichlersmith maybe you know?

https://github.com/LDMX-Software/Ecal/blob/bb668db60852b86d85eab1bce910c64be02b16e2/src/Ecal/EcalVetoProcessor.cxx#L907-L922

```C++
    // Skip hits that have a readout neighbor
[...]

PR #1182, opened 2023-08-02
Iss1181 - Hcal position reconstruction + geometry verifier

### LINKED ISSUE #1181: Side Hcal position reconstruction
**Describe the bug**
As Bertrand reported from his work on the v14 PN samples, the side Hcal position reconstruction isn't working right. The side hcal position reconstruction is taken directly from the Hcal geometry so the correct location to look for issues will be in building the side hcal center position map. 

I've walked through all of the code leading up to this part, including the GDML and python configuration and I believe those are correct. I have also checked that the conversion from copy number to HcalID is correct on the simulation level. 

Here are two sketches I made that I think is helpful to have as reference. Obviously, nothing here is to scale. A layer here corresponds to one absorber plate, two airgaps and one scintillator layer. 
Overview of a single section
image

Overview of the v14 side hcal from the front. Note the orientation of the X-axis.
image

Based on this sketch, I've outlined a small set of sanity checks that I'm going to check. Checked are tested and pass, checked and strikethrough are tested and fail at least once.

### General features
- [x] All odd bars are oriented along the Z axis 
- [x] All even bars are oriented along the X axis
- [x] All bars have have negative Y centers for the bottom section
- [x] All bars have have positive  Y centers for the top section
### Features for odd bars (along Z)
- [x] ~~Bars oriented along the Z axis all have the same Z center~~
Bars within different modules have different Z center, e.g. 
- [x] ~~Bars oriented along the Z axis all have different X center within a given layer (offset by the width of a scintillator bar), but same as bars with the same strip number~~ 
All bars within a given module currently have the same X center
- [x] Bars oriented along the Z axis all have the same Y center within a given layer, but different between different layers (offset by two times the layer thickness, taking into account the horizontal layers between) 
### Features for even bars (along X)
- [x] ~~Bars oriented along the X axis all have the same X center within a given module, but different between modules.~~
Example bars in the same module with different X
```
# Top
Hcal(1,6,6) -> (565, 554, 565) # Both in first module, horizontal bar 
Hcal(1,2,9) -> (715, 378, 715) # Should have same X center
# Bottom
Hcal(2,8,5) -> (-515, -642, 515) # Same
Hcal(2,4,3) -> (-415, -466, 415)
```
Example bars in different modules with the same X
``` 
# Top
Hcal(1,10,7) -> (615, 730, 615) # Second module, horizontal bar
Hcal(1,2,7) -> (615, 378, 615) # First module, horizontal bar
# Bottom
Hcal(2,14,10) -> (-765, -906, 765) # Second module
Hcal(2,4,10) -> (-765, -466, 765) # First module

```
- [x] Bars oriented along the X axis all have the same Y center within a given layer, but different between different layers (same offset as above) 
- [x] Bars oriented along the X axis all have different Z center within a layer but have the same Z center as bars with the same strip number in the other layers
- [x] ~~All bars along the X axis all have (positive) X centers for the (bottom) section~~
Fails on e.g. 
```
Hcal(2,8,10) -> (-765, -642, 765)
```
- [x] ~~All bars along the X axis all have negative X centers for the top section~~
Fails on e.g. 
```
Hcal(1,8,8) -> (665, 642, 665)
```
I think the above two issues might just be that we assume the X-axis is oriented to the right when looking from the front. But you can check either via a Geant4 geometry visualization or by looking at some hits in the HcalSD that the X-axis is oriented to the left. This is because we set the Z-direction along the beam.

This... might affect other geometries as well and possibly also the back hcal  :|

##  Left/Right (Same offsets etc, just being very explicit here) 
```
- [ ] All odd bars are oriented along the Z axis 
- [ ] All even bars are oriented along the Y axis 
- [ ] Bars oriented along the Z axis all have the same Z center
- [ ] Bars oriented along the Z axis all have different Y center within a given layer, but same as bars with the same strip number
- [ ] Bars oriented along the Z axis all have the same X center within a given layer, but different between different layers 
- [ ] Bars oriented along the Y axis all have the same Y center within a given module, but different between modules. 
- [ ] Bars oriented along the Y axis all have the same X center within a given layer, but different between different layers (same offset as above) 
- [ ] Bars oriented along the Y axis all have different Z center within a layer but have the same Z center as bars with the same strip number in the other layers
- [ ] All bars along the Y axis all have negative (positive) Y centers for the left (right) section [Haven't checked if this holds for module 4
- [ ] All bars have have negative (positive) X centers for the left (right) section
```
@cmantill let me know if you think I've missed anything here, its a bit complicated to keep in your head so I wouldn't be too surprised if there's something slightly off :)

EinarElen:

This resolves #1181 and patches one part of #1166 (missing case statement in `getScintillatorOrientation`. 

This updates the hcal geometry condition so that the map of strip centers aligns with the GDML. I will say that there almost certainly are more elegant expressions than what I ended up using, so if anyone has a suggestion then feel more than free to suggest them. 

One thing which caused me quite a lot of headache before I caught it was that _the entire Hcal geometry is offset by 19.05 mm in y_. I'm not entirely sure why this is, but I've added a configuration parameter that handles this. 

Since this was quite fiddly, I wanted to make sure we really got this right so I also added an analyzer in the DQM module which goes through all simulated and reconstructed hits in the hcal and checks if the position is within the bounding box of the corresponding scintillator bar. For the v14 geometry this is 100% correct, excluding reco issues in https://github.com/LDMX-Software/ldmx-sw/issues/1348. 

I have not gone through a similar exercise for v12.

## Check List

See analyzer above.

N/A

### Related Sub-Module PRs
N/A

### DIFF
--- a/DQM/include/DQM/HcalGeometryVerifier.h
+++ b/DQM/include/DQM/HcalGeometryVerifier.h
@@ -1,0 +1,59 @@
+#ifndef HCALGEOMETRYVERIFIER_H
+#define HCALGEOMETRYVERIFIER_H
+#include <TCanvas.h>
+
+#include <iomanip>
+#include <sstream>
+
+#include "DetDescr/HcalGeometry.h"
+#include "DetDescr/HcalID.h"
+#include "Framework/Configure/Parameters.h"
+#include "Framework/Event.h"
+#include "Framework/EventProcessor.h"
+#include "Hcal/Event/HcalHit.h"
+#include "SimCore/Event/SimCalorimeterHit.h"
+
+namespace dqm {
+
+class HcalGeometryVerifier : public framework::Analyzer {
+ public:
+  HcalGeometryVerifier(const std::string &name, framework::Process &process)
+      : framework::Analyzer{name, process} {}
+  void configure(framework::config::Parameters &parameters) override;
+  /*
+   *
+   * Determine which of x/y/z corresponds to the direction along, across, and
+   * through the bar respectively. Along corresponding to the length of the bar,
+   * across to the width of the bar, and through to the thickness of the bar.
+   *
+   *
+   */
+  std::array<int, 3> determine_indices(const ldmx::HcalID id);
+
+  /*
+   *
+   * Check if the hit at position `position` is within the bounds of the
+   * scintillator bar with HcalID `id`.
+   *
+   * @note: On error, this function will raise an exception if `stop_on_error`
+   * is set to true or log the issue if false.
+   *
+   */
+  bool hit_ok(const ldmx::HcalID id, const std::array<double, 3> &position);
+
+  void analyze(const framework::Event &event) override;
+
+ private:
+  std::string hcalSimHitsCollection_{"HcalSimHits"};
+  std::string hcalRecHitsCollection_{"HcalRecHits"};
+  std::string hcalSimHitsPassName_{""};
+  std::string hcalRecHitsPassName_{""};
+
+  // Maximum difference between position and bounds of the scintillator bar that
+  // will be accepted [mm]
+  double tolerance{};
+  bool stop_on_error{};
+};
+}  // namespace dqm
+
+#endif /* HCALGEOMETRYVERIFIER_H */
--- a/DQM/python/dqm.py
+++ b/DQM/python/dqm.py
@@ -1,6 +1,44 @@
 """Configuration for DQM analyzers"""
 
 from LDMX.Framework import ldmxcfg
+
+class HCalGeometryVerifier(ldmxcfg.Analyzer) :
+    """Configured HCalGeometryVerifier python object
+
+    Contains an instance of the verifier that has already been configured.
+
+    This analyzer verifies that all simhits and rechits for the hcal are within
+    the bounds of the scintillator strips as set in the geometry condition.
+
+    If the analyzer encounters an error and `stop_on_error` is true, raises an
+    exception with details about the issue. Otherwise, the error is logged and
+    histograms for each section is produced.
+
+    Examples
+    --------
+        from LDMX.DQM import dqm
+        p.sequence.append( dqm.HcalGeometryVerifier() )
+
+    """
+
+    def __init__(self,name="hcal_geometry_verifier", stop_on_error=False) :
+        section_names = ['back', 'top', 'bottom', 'right', 'left']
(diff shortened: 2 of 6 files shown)


EinarElen:
I would expect this to fail some of the validation figures, anything that has to do with Hcal positions are going to change.

cmantill:
Oh dang - is it because: 
- https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Detectors/data/ldmx-det-v14/detector.gdml#L49
- https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Detectors/data/ldmx-det-v14/constants.gdml#L427C44-L427C51
?

tomeichlersmith:
> the entire Hcal geometry is offset by 19.05 mm in y

@cmantill is correct - I implemented this offset because the ECal support box is slightly asymmetric in the vertical direction (it has to have a base to rest on but it is open at the top for the layers to be slid in like a filing cabinet). I had to slide the ECal volume up so that the sensitive layers were still centered on the beam and thus I had to do the same for the HCal. I apologize for wasting your time from my lack of documenting this.

> I would expect this to fail some of the validation figures, anything that has to do with Hcal positions are going to change.

That makes sense - do you want to include your geometry verifier in the validation histograms? Maybe in the `hcal` validation sample? If we need to recreate the golden histograms anyways, now is a good time to introduce new ones.

EinarElen:
That would explain it :sweat_smile:, I've obviously been staring at the stuff for too long because I completely missed that when looking for offsets in the GDML... 

Regarding the verifier, I'm not 100% convinced that it is going to be super useful at the moment but yeah maybe on the hcal one. Ideally, if we didn't have https://github.com/LDMX-Software/ldmx-sw/issues/1348 I think it would be quite nice to have it run with quitting on mismatch, that would catch position reco bugs early but before that has been patched that's not an option.

EinarElen:
All the failing tests are HcalDQM tests related to the side Hcal :)

tomeichlersmith (approved):
Looks good and code is understandable. I'm approving this since I think it is ready to be merged in. If you want to add the geo verifier to the CI now on this branch, I'm happy to re-review. But we can also wait to pull that verifier into the CI until the HCal reco bug you mentioned is resolved.

EinarElen:
I also think we can merge it. We can return to the CI question later if we want.

cmantill (approved):

PR #1185, opened 2023-08-16
an 8 gev version of v14 detector geometry

### LINKED ISSUE #1155: An 8 GeV version of v14 detector geometry
It would be great to get going on some 8 GeV samples that we can compare apples-to-apples to the latest and greatest 4 GeV samples.

The tagger tracker and TS upstream pads need to move to be aligned with the 8 GeV beam which has different rigidity.

ehh

**Additional context**
If there is a simple calculation to fix this, whoever has it, please add it to the issue :)

tomeichlersmith:

This resolves #1155 primarily but it also implements other changes.
- write up and enforce a naming convention for our detectors
- patch the primary generators for the electron beam at different energies (SimCore update)
  - **Note**: I expect this change to cause the PR validation plots to fail since it moves the location of the beam and therefore will cause the hits in the detector to be modified slightly.

### To Do
- [x] update Detectors cmake code so it can deduce when it should rewrite the installed GDMLs from the source GDMLs being edited
    - the solution I've implemented does pollute the build log with a single line for _each_ of the GDML files in the Detectors/data directory, but it allows the build system to only rewrite and reinstall individual files as they are updated

## Check List

### Related Sub-Module PRs
- https://github.com/LDMX-Software/SimCore/pull/98 which was merged and included in SimCore v1.6.1

## Validation of New Geometry
I copied the v14 detector into a new directory and halved the x positions of the tagger and trigger pad layers. This properly centers the beam on them. The pad/layer centers were copied out from the GDML and not deduced from the hit information so we can be pretty confident that the sim hits are being centered on the detector layers.

### Tagger
tagx

### Trigger Pads
tpx

The notebook that made these plots and the config used to generate the events is uploaded within this tar ball: 
[1155-8gev-det.tar.gz](https://github.com/LDMX-Software/ldmx-sw/files/12360383/1155-8gev-det.tar.gz)


Detectors/data/README.md
```diff
@@ -0,0 +1,35 @@
...
+  energy changes.
+  - If no beam energy suffix is provided, the beam energy is assumed
+    to either be 4GeV or that the detector would not change structure
+    with a different beam.
+
+### Categories
+This defines what the different category names mean so that future
+collaborators understand what the detector is meant to describe.
+- `det`: the full LDMX detector
+- `ti-target`: the full detector but using a Ti target instead of W
```
bryngemark: maybe nitpicky but why this different naming convention (including 'target' vs not) between Ti and LYSO targets?

Detectors/data/README.md
```diff
@@ -0,0 +1,35 @@
...
+- `lyso`: the full detector but using a LySO target instead of W
+- `hcal-prototype`: the HCal prototype detector used during testbeam
+
+### Examples
+- `ldmx-det-v14` : version 14 of the full LDMX detector for the 4GeV beam
+- `ldmx-hcal-prototype-v2.0` : version 2.0 the HCal prototype detector used during testbeam
+- `ldmx-ti-target-v9` : version 9 of the full LDMX detector for the 4GeV beam but with a Ti target (instead of W)
+- `ldmx-det-v14-8GeV` : version 14 of the full LDMX detector updated for an 8GeV beam
```
bryngemark: why capital G and V in GeV when capitalization is not used anywhere else?

bryngemark (commented):
Thanks for this! It looks good, I just have some minor comments. 

I have had my doubts about the beam alignment with the TS in `v14` and I think your updated beam configuration probably helps, but could you also just check that the (_x_-pad center) distributions look ok for 4 GeV now?

Detectors/data/README.md
tomeichlersmith: not nitpicky at all, a strict and understandable naming convention can be very helpful, I'll remove the `-target` since I think it is overkill.

Detectors/data/README.md
tomeichlersmith: good point - I'll lowercase them to unify it with the python generators.

tomeichlersmith:
@bryngemark The 4GeV detector is alright, it looks like Pad2 is slightly misaligned but only by ~1mm.

tpx-4gev

tomeichlersmith:
As expected, the [PR Validations](https://github.com/LDMX-Software/ldmx-sw/actions/runs/5880153002) failed for the runs that depend on the beam generator I edited. Scrolling through the distributions that failed, I didn't see any wild changes in distribution shape, just small statistical effects from the slight change in input beam.

bryngemark (approved):

EinarElen:
>  update Detectors cmake code so it can deduce when it should rewrite the installed GDMLs from the source GDMLs being edited

This is great, especially for people that aren't super used to how build systems work

EinarElen (approved):

PR #1187, opened 2023-08-17
bump ldmx/pro base image to ldmx/dev:v4.0.0

tomeichlersmith:
The OS update will eventually cause breaking changes and so I'm pre-emptively updating the base image for the production container to the new version.

I am going to merge #1186 first and if dependabot does its job. This PR will be overshadowed by a auto-generated one by the bot.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 #   for the development image, look at the LDMX-Software/docker repo
 ###############################################################################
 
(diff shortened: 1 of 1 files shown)


tomeichlersmith:
superseded by #1189 which also was able to notice that the Docker meta action dropped the `v` prefix in the docker repo which is fine, I just didn't notice it yet.

PR #1190, opened 2023-08-18
ECal as Target Signal Simulation Update

tomeichlersmith:

This contains several (small-ish) patches that were discovered while studying larger EaT signal samples. In summary,
- Store the atomic Z of the element on which the dark brem occurred in the event header. This is helpful for later validation that the signal sample is properly distributed in space and material. It also helps inform any library generation of materials that need to be created.
- Update to G4DarkBreM v2.0.0 which supports dark brem libraries from elements of varying Z. This is helpful for the ECal where we want to load a single library and have it used within all the volumes of differing materials.
- **Patch ECal Biasing Attachment** to include the ECal motherboards. Their names did not align with the assumptions made by the previous function and so they were not being included in the biasing. This leads to an extremely minimal change since these PCBs are very light, but I expect it to cause a failure of the ECal PN - related validation samples.
- I studied the biasing factor for EaT which I had previously lowered because I was worried about over-biasing. Raising it for the higher signal masses helpful speeds up the run without significantly distorting the position distribution.

## Check List

### Related Sub-Module PRs
- https://github.com/LDMX-Software/SimCore/pull/99

## Atomic Z Storage
Just an interesting plot, helpful to confirm that we can select elements from multi-element materials within the ECal.

image6

## ECal Biasing Patch
I generated `LDMX.Biasing.eat.midshower_nuclear` samples.

Without the biasing attachment update, I was seeing a majority of the events passing the filter without any affect on the event weight due to biasing originating within the ECal motherboard volumes.

image21

With the patch, the events passing the filter without weighting is reduced and more naturally spread out.

image17

## Signal Sample Bias Factor
Just showing one of the plots I looked at but it is the physical variable probably closely tied to over-biasing. "Low" and "High" refer to the biasing factor I was testing ("Low" is what is currently on trunk, "High" is this PR). We don't see a distortion of this (or any others I looked at).

image18

### DIFF
--- a/Biasing/python/eat.py
+++ b/Biasing/python/eat.py
@@ -130,8 +130,11 @@ def dark_brem( ap_mass , lhe, detector ) :
 
     #Biasing dark brem up inside of the ecal volumes
     from LDMX.SimCore import bias_operators
+    from math import log10
     sim.biasing_operators = [ 
-            bias_operators.DarkBrem.ecal(sim.dark_brem.ap_mass**2 / db_model.epsilon**2)
+            bias_operators.DarkBrem.ecal(
+                sim.dark_brem.ap_mass**max(2, log10(sim.dark_brem.ap_mass)) / db_model.epsilon**2
+              )
             ]
     
     sim.actions = [


tomeichlersmith:
Looking at the plots manually, the failure comes from the (slight) change in PN products observed during the simulation which is as expected.

[ecal_pn-fail.pdf](https://github.com/LDMX-Software/ldmx-sw/files/12381666/fail.pdf)

EinarElen:
The Hcal layer number (back) plot (page 31) I guess is just a bit unlucky

Do you have an idea of why this changes the PN category plots?

EinarElen (commented):
The Hcal layer number (back) plot (page 31) I guess is just a bit unlucky

Do you have an idea of why this changes the PN category plots?

tomeichlersmith:
Even a slight change in the biasing code will lead to different events being produced. It also should be emphasized that - for many particles - the motherboard is the first piece of ECal material seen.

To me, the category plots look like single-event migrations arising from this slightly different sampling and so it did not cause me concern.

EinarElen (approved):

PR #1191, opened 2023-08-18
Change default ElectronCounter input collection name

### LINKED ISSUE #1163: Wrong collection by default in electron counter?
**Describe the bug**
I was looking a bit at the output of the PR jobs and half of the output is just 
```sh
 [ ElectronCounter ] 4 : Attemping to use non-existing input collection TriggerPadTracks_ to count electrons! Exiting.
```

Seems to be 
https://github.com/LDMX-Software/ldmx-sw/blob/1d6239b49687bc59c1efa60ad2a1fec5c0900027/Recon/python/electronCounter.py#L19-L23 

In LDCS jobs, the collection is called 
https://github.com/LDMX-Software/LDCS/blob/50570eab9b24100fd9be66221adc2a720a1260ae/productionScripts/v3.2/ldmxjob-1e-3.2.0.py.tpl#L119-L122
```python
eCount.input_collection="TriggerPadTracksY"
```

I'm not super familiar with the details of the TS/electron counting components so it isn't obvious what to do to deal with it

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1163

The validation config https://github.com/LDMX-Software/ldmx-sw/blob/trunk/.github/validation_samples/inclusive/config.py currently gives the following msg per event:
```
[ ElectronCounter ] 4 : Attemping to use non-existing input collection TriggerPadTracks_ to count electrons! Exiting.
```

## Check List

The print-out statements about the wrong collection are gone.

N/A

### DIFF
--- a/Recon/python/electronCounter.py
+++ b/Recon/python/electronCounter.py
@@ -16,7 +16,7 @@ class ElectronCounter(ldmxcfg.Producer) :
     def __init__(self,nSimBeamElectrons,name="ElectronCounter") :
         super().__init__(name,'recon::ElectronCounter','Recon')
 
-        self.input_collection = "TriggerPadTracks"
+        self.input_collection = "TriggerPadTracksY"
         self.input_pass_name = "truth"
         self.output_collection = "BeamElectronCount"
         self.simulated_electron_number = nSimBeamElectrons


tomeichlersmith (approved):
Thank you @tvami !

I am very interested to see if using the correct collection actually changes any of the validation histograms. This will be interesting! Even if they fail, it will be ok - I will need to make another patch release for the changes in #1190 anyways where I know some of the validations will fail there.

As far as if you have to attach submodule PRs, the answer is no: since many subsystems share the Recon module, it is kept within this repository and not stored in a separate `git` submodule.

tvami:
Indeed I'm very interested to see the validation plots, too!

Great thanks! I'll edit the PR desc to reflect that there is no need for the submodule PR.

EinarElen (approved):

tvami:
> Indeed I'm very interested to see the validation plots, too!

How can I access the artifacts created (containing the [failed] validation plots)?

bryngemark:
@tvami if you click on the details of the merge into trunk, you expand the list of CI tests and can click on `Details` of e.g. the inclusive sample. That takes you to a step-by-step view of the CI tests but there is also a `Summary` on the top left from where you can access artifacts. The pileup one is probably the most interesting since we're now using the electron count rather than the simulated electron number of 1.

PR #1196, opened 2023-09-04
rework doc site

### LINKED ISSUE #1194: rework documentation website
Instead of pushing the generated manuals to a different repo which is then published on GitHub pages, I think it makes more sense (and is less complicated) to simply publish this repo's manual on its own GitHub pages page. This will have its root GitHub page be the same as the doc website (`ldmx-software.github.io`) but it will exist in its own sub-route `ldmx-sw`. We can then link between the two sites with that information.

This will also give me an opportunity to fix up the themes a bit and avoid including submodules that have their own documentation published elsewhere (e.g. acts and G4DarkBreM).

tomeichlersmith:

This resolves #1194 

I've updated the doxygen+sphinx infrastructure to
1. Actually work. `sphinx-apidoc` needs the `__init__.py` files which we removed awhile ago since they were all empty. Adding in a `find` command to create empty ones in our module allowed it to refind them easily.
2. Change sphinx and doxygen themes to be lighter with higher contrast. I know some folks like dark, hacker themes but I find them difficult to read (and I don't even have glasses!).
3. Update docs workflow to have these manuals hosted simply by GitHub pages here rather than push to the other pages repo. This will naturally partition them and allow them to evolve asynchronously while keeping them linked together. There isn't a good way to test this workflow so I'll just have to make sure it works after merging. I'm not too worried since I've done a few page-sites like this before.
4. Add a doxygen alias `\pythonpackage` which will insert a hyperlink to the python manual. This is helpful if you want to point the reader to the python documentation from the C++ documentation.


tomeichlersmith (commented):
Alright, since this is focused on the documentation and the PR tests passed cleanly, I am going to use my :sparkle: unlimited admin powers :sparkle: to just merge this and see if the new workflow works.

PR #1197, opened 2023-09-06
Bump to v1.2

pbutti:

Bump tracking to stable release v1.2

## Check List

- [X] I attached any sub-module related changes to this PR.


bryngemark (commented):
This looks good to me -- it's hard to really comment. I was wondering though, in the example config/joboptions, are you mostly using parameter settings that are already the default? If you want to make it a great example, indicating only those settings you'd typically want to play with or change would make it more useful to the beginner. Also, I have no idea if this is the case, but if there are typical sets of parameter changes that would go together, they could be bundled up in a predefined function (like the different simulation types are, for example) to minimize the risk that a user picks bad parameter combinations.

pbutti:
@bryngemark 
Hi Lene Kristian,
I'm using parameters that are default, although many of those would need to be properly tuned for ensuring a reasonable reconstruction: I'm not there yet. 
The parameters are separated by processor, so in principle one would be able to to change each processor separately to act on different stages of the reconstructions. 
In principle one could start define some quality cuts marked by strings, i.e. loose/tight tracks that would bundle together a set of parameters. 
I can not this down and try to include it in the new library version.

tomeichlersmith:
@pbutti Can you post a config (or configs if you did multiple steps) you used to run the tracking? I tried the following and got an error about using a pre-existing product. These errors usually crop up from running processors in the wrong order (or omitting ones that are necessary).

## Process
#### Commands
```
# after building and install this branch
ldmx fire sim.py
ldmx fire trk.py justSim_10_events.root
```
#### Error
```
# after GDML parsing printouts
 [ fire ] 4 : [ProductExists] : A product named 'TaggerTruthTracks' already exists in the event (has been loaded by a previous producer in this process).
  at /export/scratch/users/eichl008/ldmx/ldmx-sw/Framework/include/Framework/Event.h:160 in add
Stack trace: 
    0 void framework::Event::add<std::vector<ldmx::TruthTrack, std::allocator<ldmx::TruthTrack> > >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::vector<ldmx::TruthTrack, std::allocator<ldmx:
:TruthTrack> >&) + 2898 
    1 tracking::reco::TruthSeedProcessor::produce(framework::Event&) + 7777 
    2 framework::Process::process(int, framework::Event&) const + 113 
    3 framework::Process::run() + 693 
    4 main + 461 addr2line: 'fire': No such file
```
#### sim.py
This is just a copy of `SimCore/test/basic.py`
```python
# create my process object
p = ldmxcfg.Process( "test" )
# how many events to process?
import sys
p.maxEvents = 10
if len(sys.argv) > 1 :
# we want to see every event
p.logFrequency = 1
p.termLogLevel = 0
# Set a run number
p.run = 9001
# we also only have an output file
p.outputFiles = [ "justSim_" + str(p.maxEvents) + "_events.root" ]
import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry
mySim.setDetector( 'ldmx-det-v14' , True )
# Get a pre-written generator
mySim.generators.append( gen.single_4gev_e_upstream_tagger() )
p.sequence.append( mySim )
```
#### trk.py
```python
import argparse
from pathlib import Path

parser = argparse.ArgumentParser()
parser.add_argument('sim_file', type=Path)
args = parser.parse_args()

from LDMX.Tracking import examples

p = examples.single_e_track_recon_10um()

from LDMX.Tracking import geo

p.inputFiles = [ str(args.sim_file) ]
p.outputFiles = [ args.sim_file.stem + '_tracked.root' ]
```

tomeichlersmith:
It looks like the `TruthSeedProcessor` should only be run once. I'm basing this on the fact that it adds the truth track collections for both the recoil and tagger.

https://github.com/LDMX-Software/Tracking/blob/18f443683a92d5015d0c128033d59ad42e827f36/src/Tracking/Reco/TruthSeedProcessor.cxx#L360-L362

But it only takes either the RecoilSimHits or the TaggerSimHits as input? We might need to update Tracking and make a new release of it before it can be merged.

pbutti:
The `p = examples.single_e_track_recon_10um()`  is not up-to-date. The example that should be used to build a tracking job is `reco.py` inside `Tracking/python`. In there you will find all the single producers configuration. 

The truth seeder runs only once. The  hit inputs have been superseded by the usage of scoring plane information.

tomeichlersmith (approved):
Alrightie, after conferring with @pbutti , I made some usability patches and then bumped the version patch number. Now I am happy with merging this in and start the work of integrating it into the CI.

PR #1198, opened 2023-09-12
use logs as part of the validation

### LINKED ISSUE #1192: Use logs as part of the validation

Connected to https://github.com/LDMX-Software/ldmx-sw/pull/1191, the validation config shows a lot of warnings, which triggered the discussion on what's the best way to notice these kind of changes

The logs of the validation could be compared to the output logs before the PR, if the character count is increased significantly that could be a sign of new warning msgs.

tomeichlersmith:

This resolves #1192 by updating the CI scripts to store and compare the terminal output when running `fire`.

## Check List


tvami (approved):
Looks great, thanks Tom!

bryngemark:
Just noticed that the logs for, was it inclusive and maybe pileup/hcal/more? spit out a configuration error where some set parameter is wrongly typed as float or int (don’t remember which way). A good time to fix this?

EinarElen:
Ive been wondering what that warning comes from, if it is a parameter that is wrong wouldn't the whole thing halt? So I would have thought that it comes from somewhere else

tomeichlersmith:
I first noticed that warning/error when updating to the newer compiler. I am unsure where it is originating - if its in the python or C++ side of configuration. I agree with Einar, that it would halt if the parameter was wrong, but we could easily be misinterpreting some of the Python C-API functions that we are using.

For now, I think it is best to leave it since I do not know what kind of rabbit-hole tracing it will lead to. If I have time, I can open an issue with a first look at it.

tvami:
I made https://github.com/LDMX-Software/ldmx-sw/issues/1199 as the possible source for the `TypeError: 'float' object cannot be interpreted as an integer`. Although maybe it's unrelated...

tomeichlersmith:
Alright, I generated the gold logs with an intentionally slightly old commit of ldmx-sw which produced slightly different logs. This means we get the following report from GitHub.

warning-eg

I don't think this is enough since the checks are still labeled as "passing" here, so I am going to move the character-count checking from warning to error and put it in a different job step.

tomeichlersmith:
Alright, the new infrastructure is now failing at the log check (as intended). The next step is to merge this and make a new patch release to make sure that I can replace the logs with newly-generated ones.

log-comp-fail

PR #1201, opened 2023-09-15
add check on actions we use

tomeichlersmith:
this will check and bump actions versions we use in our workflows when they are released, it checks once a week which is plenty of time given our pace of development

This also is a double check that the PR Validation action updates in #1198 were successful since this change obviously does not change the code and so should pass the validations.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,3 +9,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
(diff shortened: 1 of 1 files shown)


tomeichlersmith:
Looks like the tests are working well, we get warnings that the text is different but no errors since the character count is the same:
image

And the diff is printed to the run log (just timestamp differences as expected):
image

PR #1216, opened 2023-10-06
Sample scripts for validating future samples

horohoo:
I am updating _ldmx-sw_ with sample validation scripts discussed in the [9/25/23 SWAN meeting](https://indico.fnal.gov/event/61438/). Here are the details.

## Check List

- [x] I ran my developments and the plots shown in 9/25/23 SWAN meeting show that they are successful.

I have included a sample configuration file for running the code (in .txt form that needs converted to .py, since GitHub doesn't support .py attachments to PRs):

[val_config.txt](https://github.com/LDMX-Software/ldmx-sw/files/12833366/val_config.txt)

It can be executed with the syntax
`ldmx fire val_config.py <input file to validate> <output file name, with .root extension>`

### DIFF
--- a/DQM/include/DQM/SampleValidation.h
+++ b/DQM/include/DQM/SampleValidation.h
@@ -1,0 +1,38 @@
+#ifndef DQM_SAMPLEVALIDATION_H
+#define DQM_SAMPLEVALIDATION_H
+
+//LDMX Framework
+#include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
+
+namespace dqm {
+
+    /**
+     * @class SampleValidation
+     * @brief
+     */
+
+    class SampleValidation : public framework::Analyzer {
+        public:
+
+            SampleValidation(const std::string& name, framework::Process& process) : Analyzer(name, process) {}
+
+            virtual void configure(framework::config::Parameters& ps);
+
+            virtual void analyze(const framework::Event& event);
+
+            virtual void onFileOpen(framework::EventFile&);
+
+            virtual void onFileClose(framework::EventFile&);
+
+            virtual void onProcessStart();
+
+            virtual void onProcessEnd();
+
+        private:
+
+    };
+
+}
+
+#endif
--- a/DQM/include/DQM/SampleValidation.h
+++ b/DQM/include/DQM/SampleValidation.h
@@ -20,6 +20,8 @@ namespace dqm {
             virtual void configure(framework::config::Parameters& ps);
 
             virtual void analyze(const framework::Event& event);
+
+	    int pdgid_label(const int pdgid);
         private:
 
     };
--- a/DQM/include/DQM/SampleValidation.h
+++ b/DQM/include/DQM/SampleValidation.h
@@ -20,15 +20,6 @@ namespace dqm {
             virtual void configure(framework::config::Parameters& ps);
 
             virtual void analyze(const framework::Event& event);
-
-            virtual void onFileOpen(framework::EventFile&);
-
-            virtual void onFileClose(framework::EventFile&);
-
-            virtual void onProcessStart();
-
-            virtual void onProcessEnd();
-
         private:
 
     };
--- a/DQM/src/DQM/SampleValidation.cxx
+++ b/DQM/src/DQM/SampleValidation.cxx
@@ -1,0 +1,97 @@
+#include "DQM/SampleValidation.h"
+#include "SimCore/Event/SimParticle.h"
+#include "Framework/NtupleManager.h"
+#include <iostream>
+#include <fstream>
+#include <algorithm>
+
+namespace dqm {
+
+  void SampleValidation::configure(framework::config::Parameters& ps) {
+
+    return;
+  }
+
+  void SampleValidation::analyze(const framework::Event& event) {
+
+    //Grab the SimParticle Map                                                        
+    auto particle_map{event.getMap<int, ldmx::SimParticle>("SimParticles")};
+
+    std::vector<int> primary_daughters;
+
+    //Loop over all SimParticles                                                      
+    for (auto const& it : particle_map) {
+      ldmx::SimParticle p = it.second;
+      int pdgid = p.getPdgID();
+      std::vector<double> vertex = p.getVertex();
+      double energy = p.getEnergy();
+      std::vector<int> parents_track_ids = p.getParents();
+      std::vector<int> daughters = p.getDaughters();
+
+      for (auto const& parent_track_id: parents_track_ids) {
+	if (parent_track_id == 0) {
+	  histograms_.fill("beam_smear", vertex[0], vertex[1]);
+	  histograms_.fill("pdgid_primaries", pdgid);
+	  histograms_.fill("energy_primaries", energy);
+	  primary_daughters = daughters;
+	}
+      }
+    }
+
+    std::vector<std::vector<int>> hardbrem_daughters;
+
+    for (auto const& it : particle_map) {
+      int trackid = it.first;
+      ldmx::SimParticle p = it.second;
+      for (auto const& primary_daughter : primary_daughters) {
+	if (trackid == primary_daughter) {
+	  histograms_.fill("pdgid_primarydaughters", p.getPdgID());
+	  if (p.getEnergy() >= 2500) {
+	    histograms_.fill("startZ_hardbrem", p.getVertex()[2]);
+	    histograms_.fill("endZ_hardbrem", p.getEndPoint()[2]);
+	    histograms_.fill("energy_hardbrem", p.getEnergy());
+	    hardbrem_daughters.push_back(p.getDaughters());
+	  }
+	}
+      }
+    }
+
+    for (auto const& it : particle_map) {
+      int trackid = it.first;
+      ldmx::SimParticle p = it.second;
+      for (const std::vector<int> &daughter_track_id : hardbrem_daughters){
+	for (const int &daughter_id : daughter_track_id) {
+	  if (trackid == daughter_id) {
+	    histograms_.fill("pdgid_hardbremdaughters", p.getPdgID());
+	    histograms_.fill("startZ_hardbremdaughters", p.getVertex()[2]);
+	  }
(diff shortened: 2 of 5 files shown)


bryngemark:
hi @tylerhoroho-UVA , thanks for this! did you try using the validation module for plotting at all?

DQM/src/DQM/SampleValidation.cxx
```diff
@@ -0,0 +1,97 @@
...
+
+  void SampleValidation::onProcessStart() {
+
+    return;
+  }
+
+  void SampleValidation::onProcessEnd() {
+
+    return;
+  }
```
tomeichlersmith:
```suggestion
```

Trimming undefined functions.

DQM/include/DQM/SampleValidation.h
```diff
@@ -0,0 +1,38 @@
...
+            virtual void analyze(const framework::Event& event);
+
+            virtual void onFileOpen(framework::EventFile&);
+
+            virtual void onFileClose(framework::EventFile&);
+
+            virtual void onProcessStart();
+
+            virtual void onProcessEnd();
+
```
tomeichlersmith:
```suggestion
```

Trimming undefined functions

tomeichlersmith (commented):
I concur with @bryngemark , this is great and thank you for sharing :)

My comments are 100% cleaning comments.

1. Remove the undefined functions (I added some suggestions).
2. Format the code according to our style guide so its easier for other folks to collaborate.
3. Make the hard brem energy configurable? I'm not sure if this is desirable since we often set-it and forget-it, but its something to think about.

### Formatting
In container images v4 and newer, `clang-format` is already installed so you can.
```
ldmx clang-format -i --style=Google DQM/include/DQM/SampleValidation.h DQM/src/DQM/SampleValidation.cxx
# commit these changes
```

horohoo:
Hi @bryngemark, I made some effort to use the Validation module, but integration with this module would be more difficult than keeping everything contained in DQM. Since Einar developed the Validation module, maybe I could work with them to integrate the python code into Validation if that's where it should be kept?

bryngemark:
Right, so what I think we ideally want is a combination of DQM for doing all the ldmx-sw processing/analyzing (like histogram booking and filling) and then a mirroring validation script that is just used to find those histograms, draw them, compare them, etc etc. So keeping what you have right now but just adding the Validation bit. 

If you want inspiration you could look at Tom's recent PR here: https://github.com/LDMX-Software/ldmx-sw/pull/1215 
I have a vision that we start using this combination to help streamline validation in the future. If it's too much work then just say no :)

horohoo:
I can put in some more effort to integrating Validation into the workflow before deciding it's too much work. Thanks for linking Tom's PR; I think it will be helpful!

DQM/src/DQM/SampleValidation.cxx
```diff
@@ -117,6 +121,8 @@ namespace dqm {
     if (pdgid > 2300) label = 16; //exotic (e.g., baryon with strangeness)
```
bryngemark: Have you considered what should happen if some PDGID doesn't match any of your defined criteria?

Validation/src/Validation/simparticles.py
```diff
@@ -22,6 +22,14 @@ def beamenergy_comp(d: Differ, out_dir=None):
              tick_labels=pdgid_labels,
              out_dir=out_dir,
              density=True)
+    d.plot1d("SampleValidation/SampleValidation_energy_daughterphoton", "Energy spectrum of all photons from primary",
+             out_dir=out_dir,
+             density=True)
+
+    d.plot1d("SampleValidation/SampleValidation_pdgid_harddaughters" "PDG ID of hard primary daughter",
+             tick_labels=pdgid_labels,
```
bryngemark, line 30: This is really great. Makes it SO much easier to read and grasp the plot!

DQM/src/DQM/SampleValidation.cxx
horohoo: I have considered this and plan to make changes to account for this. The current setup was put together quickly so that Jessica could proceed with validation plots. I think I will fork "exotic" into "strange baryons" and "something else" (which would then require looking at the event level to see the mystery particle). I can also expand the A' label to be any dark mediator.

bryngemark: Sounds great. Just a catch-all in the end for anything unexpected can turn out to be very useful.

DQM/src/DQM/SampleValidation.cxx
```diff
@@ -89,7 +87,7 @@ namespace dqm {
   }
 
   int SampleValidation::pdgid_label(const int pdgid) {
-    int label = 0;
+    int label = 18;
```
bryngemark: Cool! Just add a comment that this is a catch-all/overflow bin for code readability

bryngemark (approved):

DQM/src/DQM/SampleValidation.cxx
```diff
@@ -0,0 +1,130 @@
...
+    
+    if (pdgid == -321) label = 12; // K-
+    
+    if (pdgid == 130) label = 13; // K-Long
+    
+    if (pdgid == 310) label = 14; // K-Short
+    
+    if (pdgid == 3122 || pdgid == 3222 || pdgid == 3212 || pdgid == 3112 || pdgid == 3322 || pdgid == 3312) label = 16; // strange baryon
+    
+    if (pdgid > 10000) label = 15; //nuclei
```
EinarElen: Is it worth differentiating between general nuclei and light ions here? The production of alpha, deuterium, and tritium from PN interactions is non-trivial.

DQM/src/DQM/SampleValidation.cxx
```diff
@@ -0,0 +1,130 @@
...
+    //Grab the SimParticle Map and Target Scoring Plane Hits
+    auto targetSPHits(event.getCollection<ldmx::SimTrackerHit>("TargetScoringPlaneHits"));
+    auto particle_map{event.getMap<int, ldmx::SimParticle>("SimParticles")};
+
+    std::vector<int> primary_daughters;
+
+    double hard_thresh = 2500;
+
+    //Loop over all SimParticles
+    for (auto const& it : particle_map) {
```
EinarElen, line 27: Not something that needs to be fixed but something that is useful is that when iterating over a map you can do 

```C++
for (auto const& [id, particle] : particle_map) { ... }
```
To define the id and particle variables directly without having to do it.second :)

DQM/src/DQM/SampleValidation.cxx
```diff
@@ -0,0 +1,130 @@
...
+
+  void SampleValidation::analyze(const framework::Event& event) {
+
+    //Grab the SimParticle Map and Target Scoring Plane Hits
+    auto targetSPHits(event.getCollection<ldmx::SimTrackerHit>("TargetScoringPlaneHits"));
+    auto particle_map{event.getMap<int, ldmx::SimParticle>("SimParticles")};
+
+    std::vector<int> primary_daughters;
+
+    double hard_thresh = 2500;
```
EinarElen: Why are we defining hard_thresh down here and redefining it later?

Validation/src/Validation/simparticles.py
EinarElen: +1

DQM/src/DQM/SampleValidation.cxx
horohoo: I can include this distinction since I meant for the label to mainly catch remnants of a tungsten explosion. Do you think "light"/"heavy" nuclei is granular enough, with "light" being only hydrogen or helium nuclides?

DQM/src/DQM/SampleValidation.cxx
horohoo: I needed to declare hard_thresh outside the first for loop since it is used in later loops. But in principle I should only declare it here without assigning it a value yet. This was likely a remnant from when hard_thresh was a hard-coded value. I can remove the value definition but will still keep the variable declared here.

DQM/src/DQM/SampleValidation.cxx
EinarElen: The bertini cascade will attempt to take nucleons being ejected from the cascade that are "similar" and bunch them together into deuterium, tritium, alpha (called "Coalescence") so you can get these even for quite ordinary interactions. 

There is a version of a test for what counts as a light ion here (that probably should live somewhere else...)
https://github.com/LDMX-Software/ldmx-sw/blob/8e6e9e195eb71ad78f7aa94648a179342a12e1f6/Biasing/include/Biasing/PhotoNuclearTopologyFilters.h#L62C1-L78

EinarElen (approved):
All good, assuming all of the changed files have been formatted before merged (they might already be)

bryngemark (approved):

bryngemark:
@tomeichlersmith are the failed build tests a consequence of needing to rebase, or something else? if all is good then i think we can merge this

tomeichlersmith:
The failed tests are due to UMN losing its network connection on Monday and thus the tests were unable to download an example CSV table condition.
```
  1:   Wget error 4 retreiving URL 'http://webusers.physics.umn.edu/~jmmans/ldmx/
  1:   condtest//testbeam22_pedestals.csv'
```

This is the third time this has happened... we may want to move our URL test somewhere else (or stop testing it since we don't use the internet retrieval of conditions right now).

EinarElen:
I'm not super familiar with our test setup, but is there a way to mark that test as not fatal of it fails?

EinarElen:
Is the plan to include this in some of the CI DQM configs btw?

jmmans:
Dear all,

I think a great solution (if possible) would be to put copies of the 
document in two locations (e.g. UMN and SLAC or Lund) and then fail only 
if /both/ fail.

Jeremy

bryngemark:
i think @EinarElen and @jmmans both make some excellent points here. we could punt on (as in go ahead and merge now) copying stuff into the CI DQM, maybe think a little more about what we'd like to include. copying the conditions to both slac and Lund could work.

PR #1219, opened 2023-10-31
Iss1172 - Memory bugs and typos in DQM code 

### LINKED ISSUE #1172: Bug and typo in HcalDQM
**Describe the bug**
This was caught thanks to the improved GCC warnings with the new container :) 

In https://github.com/LDMX-Software/ldmx-sw/blob/ad86bf811769fa94411712f7ef82ba6e2efdea00/DQM/src/DQM/HcalInefficiencyDQM.cxx#L30-L37 

only the back check actually tests against the failure condition, the others are doing an implicit conversion bewteen bool and int. 

Also, there as a minor typo in the python config 
https://github.com/LDMX-Software/ldmx-sw/blob/ad86bf811769fa94411712f7ef82ba6e2efdea00/DQM/python/dqm.py#L132-L137
should of course have been an f-string and not just a string that starts with f :)

EinarElen:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1172 and one of the entries in #1166  (missing & in PN DQM). 

I expect this to fail the validation step for the Hcal but not for the PN DQM, we have changed the histograms in the HcalDQM. 

The following warning/static-analysis/cleanup was also done 
- Add missing virtual destructors 
- override final -> override 
- Remove tautological check (`pdgCode < 10000000000) `) 
- Fallthrough and default cases in switches 
- Removing unused variables

### DIFF
--- a/DQM/include/DQM/HCalDQM.h
+++ b/DQM/include/DQM/HCalDQM.h
@@ -1,6 +1,5 @@
 #ifndef _DQM_HCAL_DQM_H_
 #define _DQM_HCAL_DQM_H_
-
 //----------//
 //   STL    //
 //----------//
@@ -14,6 +13,8 @@
 /*~~~~~~~~~~~~~~~*/
 /*   Framework   */
 /*~~~~~~~~~~~~~~~*/
+#include <map>
+
 #include "DetDescr/HcalGeometry.h"
 #include "DetDescr/HcalID.h"
 #include "Framework/Configure/Parameters.h"
@@ -24,11 +25,10 @@
 #include "Hcal/Event/HcalVetoResult.h"
 #include "SimCore/Event/SimCalorimeterHit.h"
 #include "Tools/AnalysisUtils.h"
-#include <map>
 namespace dqm {
 
 class HCalDQM : public framework::Analyzer {
-public:
+ public:
   /** Constructor */
   HCalDQM(const std::string &name, framework::Process &process);
 
@@ -40,7 +40,7 @@ public:
    *
    * @param parameters Set of parameters used to configure this processor.
    */
-  void configure(framework::config::Parameters &parameters) final override;
+  void configure(framework::config::Parameters &parameters) override;
 
   /**
    * Process the event and make histograms ro summaries.
@@ -66,7 +66,7 @@ public:
     return false;
   }
 
-private:
+ private:
   /// Hcal Sim Hits collection name
   std::string sim_coll_name_;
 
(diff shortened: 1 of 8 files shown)


tomeichlersmith (commented):
This looks good but I am curious about where histogram _content_ is being changed - I don't see anything affecting `fill` or binning calls outside of white-space formatting.

EinarElen:
This bugfix (it's subtle, thanks implicit conversion warnings!)

https://github.com/LDMX-Software/ldmx-sw/commit/9305a03fb01a3c295031799d0b97cca13a044e4e

EinarElen:
The plots that are failing are the expected ones. The Hcal one is not failing because that is the only one where we arent running the particular analyzer :)

tomeichlersmith:
Since only one plot fails for each of the three failing validation tests, I am posting it here.[^1] @EinarElen do these make sense? It seems like a physical change (i.e. I don't see any screaming bugs), but I'm unsure how to interpret the change.

[^1]: I got this plot by downloading the artifacts for the failing PR Validation jobs. Unzipping and unpacking and then copying out the plot in the "fail" subdirectory of the unpacked directory. I then used pdftoppm to convert the PDF plots to PNG so GitHub will render them.

## Signal
signal-HcalInefficiencyAnalyzer_HcalInefficiencyAnalyzer_efficiency

## Inclusive
inclusive-HcalInefficiencyAnalyzer_HcalInefficiencyAnalyzer_efficiency


## Ecal PN
ecal_pn-HcalInefficiencyAnalyzer_HcalInefficiencyAnalyzer_efficiency

_Edit_: Cropped PNGs so there's less ugly white-space.

EinarElen:
It is a physical change in that the previous version was just straight up invalid. What the analyzer is trying to do is to check if any hit has been above threshold in the different hcal sections and if so, what was the earliest layer that had such a hit. By default we have each section set to a negative "failed to veto" value. The histogram is supposed to contain different categories like "the hit was vetoed by one of the side hcal modules" or "the hit was only vetoed by the back hcal"

```

  bool vetoedByBack{firstLayersHit[ldmx::HcalID::HcalSection::BACK] !=
                    failedVeto};
  bool vetoedByTop{firstLayersHit[ldmx::HcalID::HcalSection::TOP]};
  bool vetoedByBottom{firstLayersHit[ldmx::HcalID::HcalSection::BOTTOM]};
  bool vetoedByRight{firstLayersHit[ldmx::HcalID::HcalSection::RIGHT]};
  bool vetoedByLeft{firstLayersHit[ldmx::HcalID::HcalSection::LEFT]};
  bool vetoedBySide{vetoedByTop || vetoedByBottom || vetoedByRight ||
                    vetoedByLeft};
```

Problem is right here, only the first (vetoedByBack) condition is correct, all the others have missed the != failedVeto part.

tomeichlersmith (approved):
Thanks for the extra explanation! This looks good and I can make a tag release after merging.

PR #1223, opened 2023-11-02
1220 filter and handle missing entries in the validation module

### LINKED ISSUE #1220: Filter and handle missing entries in the validation module

I've been making use of the validation module quite a lot when developing ldmx-sw. It is really nice to just run the validation code on what you just produced and get some legible plots without having to run a whole ROOT-style browsing job. However there are two things that tend to make the process slower for me when doing this. I suspect one thing might be harder than the others to add.
- The validation code trips over if a histogram is missing in one of the files being used 
An example where this would happen is if you try using the HcalDQM code with the TB geometry where you won't produce histograms for the side hcal modules. I think this should be logged but letting the process continue gracefully. 
- There is no way to restrict what files are used 
When developing you might be producing a bunch of slightly different versions of the same event files and histograms. Most of the time you only want to make comparisons between two of them at a time but you probably don't want to delete the others. Having some optional way to restrict which files are used would be great.

(Maybe this is already a feature and I've just missed it)

For the first I suspect that we just need a try/catch somewhere. For the second, a command line argument? Either a list or some kind of regex?

What I'm doing right now: Editing the validation code directly to only deal with the histograms that I care about and having my ldmx-sw configs manage a whole directory structure to deal with moving files around. Not very pleasant.

EinarElen:

I've added two new command line parameters for the validation module, one which lets you specify exactly which files you want to compare (as opposed to all files in the input data directory). You can either specify the name of a file in the input directory or give the entire path to the file.  The other parameter lets you filter the list of root-files (a use-case for this in my case would be filtering on the number of events in a pile of files). 

Also added an exception handling step in Differ to handle missing histograms gracefully. 

Example usage: 
Assuming you have three files in the hists/ folder 
- ecal_trunk_10000_events.root 
- ecal_trunk_100_events.root 
- ecal_PR_10000_events.root
- hcal_trunk_10000_events.root 

Running 
```
ldmx python3 -m Validation hists --input-files ecal_trunk_10000_events.root hists/ecal_PR_10000_events.root  ...
```
Would only plot the two ecal files with 10k events 
Running 
```
ldmx python3 -m Validation hists --input-file-filter="ecal" ...
```
Would use the three files including "ecal" in the name 

etc 
This resolves #1220  
## Check List

Tested both the new parameters and handling of missing histograms. Worked like a charm :) 

N/A

### Related Sub-Module PRs
N/A

### DIFF
--- a/Validation/src/Validation/_file.py
+++ b/Validation/src/Validation/_file.py
@@ -28,7 +28,6 @@ class File :
         self.__file = uproot.open(filepath, **open_kwargs)
         self.__colmod = colmod
         self.__df = None
-        self.path = filepath
 
         if 'histtype' not in hist_kwargs :
             hist_kwargs['histtype'] = 'step'
@@ -105,7 +104,12 @@ class File :
         Otherwise, we assume that it is a histogram file.
         """
         return 'LDMX_Events' in self.__file
-    
+
+    @property
+    def path(self):
+        """Retrun the path to the file on disk"""
+        return self.__file.file_path
+
     def events(self, **kwargs) :
         """Callback for retrieving a full in-memory data frame of the events
         
--- a/Validation/src/Validation/_file.py
+++ b/Validation/src/Validation/_file.py
@@ -28,6 +28,7 @@ class File :
         self.__file = uproot.open(filepath, **open_kwargs)
         self.__colmod = colmod
         self.__df = None
+        self.path = filepath
 
         if 'histtype' not in hist_kwargs :
             hist_kwargs['histtype'] = 'step'
--- a/Validation/src/Validation/__main__.py
+++ b/Validation/src/Validation/__main__.py
@@ -45,6 +45,9 @@ if __name__ == '__main__' :
     parser.add_argument('--param',
                         nargs='+',
                         help='parameter(s) in filename to use as file labels')
+    parser.add_argument('--input-files', nargs='+', type=str,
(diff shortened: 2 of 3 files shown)


Validation/src/Validation/_file.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, filepath, colmod = None, hist_kwargs = dict(), **open_kwargs)
         self.__file = uproot.open(filepath, **open_kwargs)
         self.__colmod = colmod
         self.__df = None
+        self.path = filepath
```
tomeichlersmith: Instead of storing our own copy of the filepath, we could just grab uproot's

```python
@property
def path(self):
    return self.__file.file_path
```
would enable the same `f.path` syntax via the `property` decorator.

tomeichlersmith (approved):
I have one suggestion but it is minor and may not be worth it.

In general, I like this feature improvement and I'm glad to see that folks are using the Validation module!

Validation/src/Validation/_file.py
EinarElen: I was thinking there probably was a smarter way to do this :)

EinarElen: Something like this?
https://github.com/LDMX-Software/ldmx-sw/pull/1223/commits/13b008ca9723fab73d6b84a869c713524649a490

tomeichlersmith: Exactly :+1:

PR #1235, opened 2023-11-22
Update SimCore to include recent features

### LINKED ISSUE #1234: Update LDMX-sw to include the new Kaon physics in SimCore and improved resimulation
See https://github.com/LDMX-Software/SimCore/pull/105, https://github.com/LDMX-Software/SimCore/issues/106, https://github.com/LDMX-Software/SimCore/pull/108, and https://github.com/LDMX-Software/SimCore/issues/107, https://github.com/LDMX-Software/SimCore/pull/109. 

These PRs introduce a couple of new and improved features for LDMX-sw that are relevant for upcoming analysis efforts. In particular 
- We let kaon properties be configurable at runtime (lifetimes and branching ratios) 
- Handle multiple input files during resim efforts 
- Increase the number of process types that can be assigned to sim particles so that less entries are marked as "Unknown" 

The only change that should make a difference is the last one, which I would expect to fail one figure in the SimObjects DQM. Everything else **should** be the same

EinarElen:

This resolves #1234 

## Check List

See PRs in SimCore

### Related Sub-Module PRs

- https://github.com/LDMX-Software/SimCore/pull/109 
- https://github.com/LDMX-Software/SimCore/pull/108
- https://github.com/LDMX-Software/SimCore/pull/105
- https://github.com/LDMX-Software/SimCore/pull/102
- https://github.com/LDMX-Software/SimCore/pull/101

Of these I expect one of the SimObjects DQM plots to fail (because primary particles will be counted as "Primary" rather than "Unknown"). W.r.t. the signal simulation, I leave it to @tomeichlersmith  to say if anything is expected to change

### DIFF
--- a/DQM/src/DQM/SimObjects.cxx
+++ b/DQM/src/DQM/SimObjects.cxx
@@ -2,34 +2,19 @@
 #include "DQM/SimObjects.h"
 
 #include "Framework/ProductTag.h"
-
+#include "SimCore/Event/SimCalorimeterHit.h"
 #include "SimCore/Event/SimParticle.h"
 #include "SimCore/Event/SimTrackerHit.h"
-#include "SimCore/Event/SimCalorimeterHit.h"
 
 namespace dqm {
 
-void SimObjects::configure(framework::config::Parameters &ps) {
+void SimObjects::configure(framework::config::Parameters& ps) {
   sim_pass_ = ps.getParameter<std::string>("sim_pass");
   return;
 }
 
 void SimObjects::onProcessStart() {
   getHistoDirectory();
-
-  // create sim particles histograms 
-  histograms_.create("SimParticles.E","Vertex Total Energy [MeV]",800,0.,8000.);
-  histograms_.create("SimParticles.px","Vertex Momentum in x-Direction [MeV]",50,0.,500.);
-  histograms_.create("SimParticles.py","Vertex Momentum in y-Direction [MeV]",50,0.,500.);
-  histograms_.create("SimParticles.pz","Vertex Momentum in z-Direction [MeV]",400,0.,4000.);
-  histograms_.create("SimParticles.time","Global Time of Creation [ns]",50,0.,10.);
-  histograms_.create("SimParticles.pdg","PDG ID of Particle",201,-100,100);
-  histograms_.create("SimParticles.x","Vertex x-Position [mm]",401,-200,200);
-  histograms_.create("SimParticles.y","Vertex y-Position [mm]",401,-200,200);
-  histograms_.create("SimParticles.z","Vertex z-Position [mm]",171,-700,1000.);
-  histograms_.create("SimParticles.track_id","Track ID of Particle",100,0,1000);
-  histograms_.create("SimParticles.parent","Track ID of Parent",100,0,1000);
-  histograms_.create("SimParticles.children","Track IDs of Children",100,0,1000);
   std::vector<std::string> labels = {
       "Unknown",              // 0: Unknown
       "e+ e- ->  γγ",         // 1: annihil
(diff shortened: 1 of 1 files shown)


EinarElen:
Right, so something was changed by all of this. I'm not sure why but TS and Hcal features seem to have changed but Ecal/other features have remained the same. 

[ecal_pn_fail.pdf](https://github.com/LDMX-Software/ldmx-sw/files/13448752/ecal_pn_fail.pdf)

EinarElen:
The logs are different as well so something has changed w.r.t. the simulation 
```
Gold 
[ EcalProcessFilter ]: 10000 Brem photon produced 12 particle via biasWrapper(photonNuclear) process.
[ Simulator ] : Started 823126 events to produce 10000 events.
PR 
[ EcalProcessFilter ]: 10000 Brem photon produced 36 particle via biasWrapper(photonNuclear) process.
[ Simulator ] : Started 824435 events to produce 10000 events.

**Comment by EinarElen** [comment]
The first difference occurs at event 150
```
Gold 
[ EcalProcessFilter ]: 150 Brem photon produced 8 particle via biasWrapper(photonNuclear) process.
[ EcalProcessFilter ]: 150 Brem photon produced 10 particle via biasWrapper(photonNuclear) process.
PR 
[ EcalProcessFilter ]: 150 Brem photon produced 8 particle via biasWrapper(photonNuclear) process.
[ EcalProcessFilter ]: 150 Brem photon produced 9 particle via biasWrapper(photonNuclear) process.
```
So here something changed. As far as I can tell, this was the first time that there was a second PN attempt.

**Comment by EinarElen** [comment]
What fun, I can't reproduce either the gold or the PR log results on my machine....

**Comment by EinarElen** [comment]
On my machine with trunk vs PR branches for a very small ecal pn sample and using https://github.com/tomeichlersmith/root-diff/
```shell
LDMX_Events mismatched between files
== Branches with different content ==
EventHeader.timestamp_.fSec
EventHeader.timestamp_.fNanoSec
SimParticles_test.second.processType_
``` 

So this looks like what we would expect

EinarElen:
Ok, I've confirmed that the difference comes from the introduction of the KaonPhysics, I will have to figure out what causes it.

EinarElen:
I'm retracting my previous confirmation that the issue was caused by the kaon changes... There was an event seed mismatch between the two branches I was comparing so it doesn't tell us anything. The seed for the previous event however was the same (!?)

EinarElen:
Found the culprit in a comment in a Geant4 header file:
```C++
// G4DecayTable.hh
// Insert a decay channel at proper position 
// (i.e. sorted by using branching ratio ) 
```

EinarElen:
From the G4DecayTableMessenger, the accepted way to change the branching ratio of a process is 
```C++
void G4DecayTableMessenger::SetNewValue(G4UIcommand * command,G4String newValue)
{
    if (command == brCmd) {
      //Command   /particle/property/decay/br
      G4double  br = brCmd->GetNewDoubleValue(newValue);
      if( (br<0.0) || (br>1.0) ) { 
	G4cout << "Invalid brancing ratio. Command ignored." << G4endl;
      } else {
        currentChannel->SetBR(br);
      }
    }
  }
}
```

This doesn't change the location of the channel, so according to G4 we should be ok in doing this as well. To be sure, here is the algorithm used to pick decay channels 
```C++
while(true){
    G4double sumBR = 0.0;
    G4double r= G4UniformRand();
    // select decay channel
    for (auto channel : channels) {
      sumBR += channel->GetBR();
      if ( // Not allowed due to parent mass) { continue} 
      if (r < sumBR) return channel;
    }
  }
```

This should not depend on the order that the change is made. So we can use the _actual_ Geant4 positions safely.

EinarElen:
Hey whaddya know only the expected to fail plots failed :)

tomeichlersmith:
The only failing plot is the expected proceess type which makes me think we should probably add more bins since the bin max right now is 12

sim_dqm_sim_dqm_SimParticles process png

I am going to merge the SimCore patch and make a new SimCore release to be merged in.

tomeichlersmith (changes requested):
Everything looks good here, I'm thankful the CI helped us find this very funky bug :100:

I have two house-keeping requests before I'm ready to merge.
1. Update SimCore to the v1.8.0 tag (I try to refer to the tag name in the commit message).
2. Update the SimParticle DQM to align with the increase in the number of process types. (maybe add the names to the bins as well?)

EinarElen:
Updated process type histogram
Screenshot 2023-11-27 at 17 51 58

Checked that I didn't off-by-one the bin labels, there are 1k events and 1k in the "Primary" bin :)

tomeichlersmith:
Were you running a special/nuclear config? I'm just surprised there are so many "unknowns"

EinarElen:
It was a very soft kaon particle gun, so most of the unknown are ionization from hadrons

EinarElen:
But that was what made me catch the pi-inelastic mapping being wrong :)

EinarElen:
Ok, SimCore bumped!

tomeichlersmith (approved):

PR #1240, opened 2023-12-05
Issue #1239 bug fix

### LINKED ISSUE #1239: Strange bug in SampleValidation.cxx
**Describe the bug**
The `hard_thresh` variable in **SampleValidation.cxx** of the DQM submodule is always printing out as 0. This happens both when it is calculated with non-zero values then immediately printed out, and when it is defined outside of a loop as a non-zero value and then printed out. 

**To Reproduce**
1. Ensure that your local version of `ldmx-sw` is up-to-date by doing `git pull` and then `git submodule update`.
2. Create an analysis config file, `test_config.py.` A screenshot of the config file I use for this analysis is below.
3. You need to have a ROOT file to input to your analysis config file - for the purposes of this bug, it might be easiest to produce a small sample using an electron particle gun. If you don't want to do that, reach out to me (@jpasc27 ) and I can send you the ROOT file that I use when doing this analysis.
4. Add a few print statements to `$LDMX_BASE/ldmx-sw/DQM/src/DQM/SampleValidation.cxx` to see the value(s) of the `hard_thresh` variable at different points in the code. For example, I added three print statements - one right after the variable is defined on L24, another right after it is calculated in the first `for` loop on L39, and a third in the second `for` loop where it is used as a filter (just under L54).
5. After saving your changes, navigate to `$LDMX_BASE/ldmx-sw/build/`and type `ldmx make install` to rebuild `ldmx-sw` and incorporate your changes.
6. Run your new config file outside of `ldmx-sw` with `ldmx fire test_config.py <output_file.root> <input_file.root>`. Your print statements should appear in your terminal window.

**Desired behavior**
The `hard_thresh` variable should print out a non-zero rescaled value of the energy.

**Screenshots**
Example Config

**Environment:**
Output of `ldmx config`:
```
LDMX base directory: /sfs/qumulo/qhome/jlp4td/LDMX
uname: Linux udc-ba34-36 3.10.0-1160.59.1.el7.x86_64 #1 SMP Wed Feb 23 16:47:03 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
OSTYPE: linux-gnu
Bash version: 4.2.46(2)-release
Display Port: 
Container Mounts: /sfs/qumulo/qhome/jlp4td/LDMX
Container Environments: 
Singularity Version: 3.7.1
Singularity File: /sfs/qumulo/qhome/jlp4td/LDMX/ldmx_dev_latest.sif
```

**Additional context**

horohoo:

This resolves issue #1239 which was causing the variable hard_thresh to be assigned 0 instead of an appropriate fraction of the beam energy. The fraction (2500/4000) was replaced with (2500./4000.) so that the variable does not get converted to int.

## Check List

- [x] I ran my developments and the following shows that they are successful. I have attached two plots of the start/end z position of the "hard" primary daughter. See #1239 for how these plots look before bug fix.

startZ_hardbrem
endZ_hardbrem

We are now seeing hard primary daughters starting in the target and ending in the ecal, as expected for an ecal PN simulation. Print statements of the variable hard_thresh also confirm it is being assigned a value 2500 MeV on a 4 GeV beam simulation instead of 0 MeV.

Sorry for the inconvenience this oversight has caused!

### DIFF
--- a/DQM/src/DQM/SampleValidation.cxx
+++ b/DQM/src/DQM/SampleValidation.cxx
@@ -36,7 +36,7 @@ namespace dqm {
 	if (parent_track_id == 0) {
 	  histograms_.fill("pdgid_primaries", pdgid_label(pdgid));
 	  histograms_.fill("energy_primaries", energy);
-	  hard_thresh = (2500/4000)*energy;
+	  hard_thresh = (2500./4000.)*energy;
 	  primary_daughters = daughters;
 	  for (const ldmx::SimTrackerHit &sphit : targetSPHits) {
 	    if (sphit.getTrackID() == it.first && sphit.getPosition()[2] < 0) {


tomeichlersmith (approved):
No apology necessary! This is why we are having folks test and review each others code - I'm glad we were able to collaborate and find such a pesky bug.

I am approving this although I suspect that the the DQM histograms will fail to match for one of two reasons
1. This PR was opened _before_ the new golden histograms were merged #1238 meaning this PR will be comparing against the v3.3.2 histograms.
2. This threshold change will change how the histograms are filled themselves.

Either way, I still think this is a good change to make!

bryngemark (approved):
Ah, yes. Nice catch.

EinarElen:
I am a little bit surprised that we wouldn't get a warning about narrowing conversions from this... I would have expected the compiler to look at that and say "are you sure you want this fraction to always be 0?"

therwig (approved):

Issue #1229, opened 2023-11-15
Parameters hardwired for 4 GeV (or 8 GeV) should be configurable/deducible 


bryngemark:
**Is your feature request related to a problem? Please describe.**
Whenever we change the beam energy, a whole set of filter and biasing thresholds need updating. This is a lot of work going deep into the simulation workings of ldmx-sw, which is not very user friendly, and which is error prone. If we invest some effort now when we are anyway identifying and changing all these settings (as we move defaults from 4 to 8 GeV), we might be grateful later.

**Describe the solution you'd like**
When we have chosen some beam energy, any beam energy dependent setting should automatically change. Since most of our thresholds etc are using fractions of the beam energy, this might as well be how they are defined in the configs. (One exception is trigger thresholds which are studied and tuned, but ironically, there the possibility of using thresholds that are a function of the beam energy is already implemented. Using non-default thresholds is also easily set up as it's a parameter directly in the trigger processor itself, and not buried in calls to anything deeper.) 

**Describe alternatives you've considered**
Another option is to have a different version of each filter, biasing etc, just like we have different functions prepared for different particle guns and whatnot. This looks like an unnecessary multiplication to me.

bryngemark:
@jpasc27 if you have an idea of where to start making these changes, and are willing to have a go, please feel free to pick this one up! If not, just listing where you have found necessary changes would be very helpful too.

@tomeichlersmith since you had some examples of how to go about this, please add them here :)

tomeichlersmith:
The examples I've implemented are all in the `eat.py` module and I think the best example is the one for `midshower_nuclear`, below is the relevant code where I configure the various simulation user actions that do the filtering.

https://github.com/LDMX-Software/ldmx-sw/blob/3c1058e3a2fdbf0826fb61c61a3d028fee3f748b/Biasing/python/eat.py#L83-L90

I keep some of the more relevant actions to have their parameters given to the function as arguments (`bias_threshold` and `min_nuc_energy`), but the other threshold simply scales with the energy of the passed generator (For EaT, I require the primary electron to reach the ECal with 87.5% of its original beam energy.) The one complexity is that the `ParticleGun` stores its beam energy in GeV whereas most (if not all?) filters use MeV so there is an extra factor of 1000.

[...]

PR #1245, opened 2023-12-20
Update beam energy and thresholds in simpleTrigger.py

bryngemark:
Added in some changes to the trigger processor defaults that I came across when setting up for Ecal PN production. Adding this to the list of changes needed for the filters and biasing etc, tracked in #1229. This will cause changes to gold histograms insofar as there are any for the trigger processor (less strict threshold when properly taking the higher beam energy into account --> more will pass)

Partial solution to #1229.

### DIFF
--- a/Recon/python/simpleTrigger.py
+++ b/Recon/python/simpleTrigger.py
@@ -36,11 +36,11 @@ from LDMX.Framework import ldmxcfg
 class TriggerProcessor(ldmxcfg.Producer) :
     """Configuration for the (multi-electron aware but simple) trigger on the ECal reco hits"""
 
-    def __init__(self,name) :
+    def __init__(self, name, beamEnergy) :
         super().__init__(name,'recon::TriggerProcessor','Recon')
-
-        self.beamEnergy = 8000.
-        self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]  #toy something up 
+          
+        self.beamEnergy = beamEnergy
+        self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
         self.mode = 0
         self.start_layer = 0
         self.end_layer = 20
@@ -48,5 +48,5 @@ class TriggerProcessor(ldmxcfg.Producer) :
         self.input_pass = ''
         self.trigger_collection = "Trigger"
 
-simpleTrigger = TriggerProcessor("simpleTrigger")
+simpleTrigger = TriggerProcessor("simpleTrigger", 8000.)
 
--- a/Recon/python/simpleTrigger.py
+++ b/Recon/python/simpleTrigger.py
@@ -39,8 +39,8 @@ class TriggerProcessor(ldmxcfg.Producer) :
     def __init__(self,name) :
         super().__init__(name,'recon::TriggerProcessor','Recon')
 
-        self.beamEnergy = 4000.
-        self.thresholds = [ 1500.0, 1000. + self.beamEnergy, 500. + 2*self.beamEnergy, 100. + 3*self.beamEnergy ]  #toy something up 
+        self.beamEnergy = 8000.
+        self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]  #toy something up 
         self.mode = 0
         self.start_layer = 0
         self.end_layer = 20
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -5,7 +5,8 @@ p.maxTriesPerEvent = 10000
 
 from LDMX.Biasing import ecal
 from LDMX.SimCore import generators as gen
-mySim = ecal.photo_nuclear('ldmx-det-v14',gen.single_4gev_e_upstream_tagger())
+det = 'ldmx-det-v14'
+mySim = ecal.photo_nuclear(det,gen.single_4gev_e_upstream_tagger())
 mySim.beamSpotSmear = [20.,80.,0.]
 mySim.description = 'ECal PN Test Simulation'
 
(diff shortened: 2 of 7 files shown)


tomeichlersmith (approved):
The failing tests are failing because of the change in trigger thresholds. All of them only have two histograms changing:
```
  Trigger_Trigger_Trigger.pass.pdf
  Trigger_Trigger_Trigger.EcalEcut.pdf
```

I downloaded the artifacts and manually merged the failed plots from the different validation samples into a document to browse. The changes look as expected to my eye.
- The Ecal Cut distributions scale with 8GeV and the electron count rather than 4GeV and the electron count.
- There is an increase in trigger-passing for all distributions following the more appropriate (looser) trigger cut being applied.

[changing-histograms.pdf](https://github.com/LDMX-Software/ldmx-sw/files/13729516/changing-histograms.pdf)

jpasc27 (approved):
I agree with Tom's interpretation of the plots he shared - they all seem reasonable! Looks good.

bryngemark:
Thanks @tomeichlersmith for reviewing the CI artifacts and verifying that we see only the expected changes. @jpasc27 will you have a chance to go over the changes you did to your config and make sure they are plugged into the defaults for the relevant processors, as we discussed in #1229 ? I don't want to close this issue before there is a full solution to that issue.

bryngemark:
hey @jpasc27 , just a reminder about my question to you in this issue. we're trying to close all open PRs within about a week, before we pull all the submodules directly into `ldmx-sw` so i figured we might as well get this done.

jpasc27:
Sorry, completely forgot about this (as I'm sure you can tell)! I will check my code to answer your question on Monday, that way this can be resolved before your Thursday deadline.

bryngemark:
awesome, thanks @jpasc27.

Recon/python/simpleTrigger.py
```diff
@@ -39,8 +39,8 @@ class TriggerProcessor(ldmxcfg.Producer) :
     def __init__(self,name) :
         super().__init__(name,'recon::TriggerProcessor','Recon')
 
-        self.beamEnergy = 4000.
-        self.thresholds = [ 1500.0, 1000. + self.beamEnergy, 500. + 2*self.beamEnergy, 100. + 3*self.beamEnergy ]  #toy something up 
+        self.beamEnergy = 8000.
```
tvami: I was wondering if we should deduce this number based on the detector geometry name, and not do any hardcoded number? i.e. `ldmx-det-v14-8gev` will be 8000, otherwise 4000? I guess what it really should be is that if the det name contains the `8gev` string than 8000, otherwise 4000

bryngemark: sure, if you want to code that up that would be very convenient for the user. since it's configurable and i assumed most trigger studies would target 8 GeV i haven't gone to any lengths to harmonize it, but it would for sure be safer to do it that way. if you have the time and energy?

tvami: OK I did it in https://github.com/LDMX-Software/ldmx-sw/pull/1245/commits/cffc5ba7722c6042c06d6cc6f1bf7bd3396de7bf

Please note the ECAL pn validation config still uses the 4 GeV geom + beam, so updating the trigger to 8GeV would have lead to some inconsistencies (although it's just a test file and I understand we'll move that to 8GeV soon, so maybe fine?)

tvami: @bryngemark does this look ok for you?

bryngemark: yeah originally #1229 should deal with this. i'll put in a catch-all issue for whatever we don't get to now.

bryngemark: i hadn't realised your change would affect every existing use of the trigger processor. does it have to be made a mandatory argument? then why not just ship the beam energy?

bryngemark: > yeah originally #1229 should deal with this. i'll put in a catch-all issue for whatever we don't get to now.

i think the way to deal with this is to not let this PR close issue #1229.

tvami: I think having it a mandatory argument is actually helpful, that will enforce the user to think about it, while if there is a default then that can lead to a situation where the beam energy is changed but the trigger settings were forgotten to be changed. Shipping the beam energy as a mandatory argument would lead to the same situation indeed. But then I just like the idea that we govern everything with the `det` string.

bryngemark: ok, i suppose we can be forced to remember this :) 
but the more i think of it, the trigger decision is informed by beam energy more than it needs to know detector version, so i think this is the more logical thing to explicitly pass to it. tbh i thought you would implement reading the detector under-the-hood, from something else than an argument (not that i had a clear picture of how you'd achieve it). and i am probably biased by my assumption that in the long run, there will be no need to include the beam energy in the detector name, as we will only have 8 GeV beam. 

if your argument is that more people explicitly set the detector than the beam energy, then i might disagree but i'll fold, now you have my arguments, i trust your judgement.

tvami: >  tbh i thought you would implement reading the detector under-the-hood,

yes, I tried to do that first but it seems the simple trigger class does not see that info directly

tvami: ok so then let's pass the beam energy explicitly, I also had a quick chat with @tomeichlersmith and this method seems to be OK to everybody IIUC @bryngemark 's take as well. I'll make the commit later today if everybody is ok with that solution

tvami: Done in https://github.com/LDMX-Software/ldmx-sw/commit/c99cdcbf7bc342911c0d9239a6d5183a34c24d17

tvami:
> i think the way to deal with this is to not let this PR close issue 1229.

with this and if people are happy with https://github.com/LDMX-Software/ldmx-sw/commit/c99cdcbf7bc342911c0d9239a6d5183a34c24d17 I think this PR can be marked as "Ready for review" and doesnt have to stay Draft, right?

tomeichlersmith (approved):
I'm happy with these changes and (as before) the plots that are failing the tests are related to the trigger and how it is being applied. only the validation samples that have changed to 8GeV are failing now which is to be expected.

[failing.pdf](https://github.com/LDMX-Software/ldmx-sw/files/15138558/failing.pdf)

<details>

  <summary>TeX which made the above PDF</summary>

```tex
\documentclass{beamer}

\usepackage{graphicx}

\begin{document}

\begin{frame}
  \centering
  Failing Plots
\end{frame}

\begin{frame}
  \centering
  Inclusive
\end{frame}

\begin{frame}
  \includegraphics[width=\textwidth]{inclusive/Trigger_Trigger_Trigger.EcalEcut.pdf}
\end{frame}

\begin{frame}
  \centering
  Signal
\end{frame}

\begin{frame}
  \includegraphics[width=\textwidth]{signal/Trigger_Trigger_Trigger.EcalEcut.pdf}
\end{frame}

\begin{frame}
  \includegraphics[width=\textwidth]{signal/Trigger_Trigger_Trigger.pass.pdf}
\end{frame}

\end{document}
```

</details>

tvami (approved):

PR #1255, opened 2024-01-29
start performance plotting within Validation py module

tomeichlersmith:
I wanted to show the time separation between 4GeV and 8GeV and I figured
this is a general enough plot that it should go into Validation.

This just plots the `__ALL__` duration since I couldn't find an obvious
way to deduce what the processor names were given the current
infrastructure. The `__ALL__` duration is just timing how long the entire
event takes to process (including all processors). The plot I show in my
release slides only has the simulation processor so it is comparing the
simulation times.

### DIFF
--- a/Validation/src/Validation/__init__.py
+++ b/Validation/src/Validation/__init__.py
@@ -8,3 +8,4 @@ from . import hcal
 from . import photonuclear
 from . import dark_brem
 from . import simparticles
+from . import performance
--- a/Validation/src/Validation/performance.py
+++ b/Validation/src/Validation/performance.py
@@ -1,0 +1,25 @@
+"""Plotting of performance plots"""
+
+from ._differ import Differ
+from ._plotter import plotter
+import logging
+
+log = logging.getLogger('performance')
+
+@plotter(hist=True,event=False)
+def event_timing(d : Differ, out_dir = None) :
+    """Plot time it took to process events
+
+    Parameters
+    ----------
+    d : Differ
+        Differ containing files that are not event files (presumably histogram files)
+    """
+
+    # just plotting the __ALL__ branch which represents all processors in the sequence,
+    # however, if we introduce some introspection into the Differ class, we could deduce
+    # the other branch names and plot the duration of different processors specifically
+    for processor in ['__ALL__']:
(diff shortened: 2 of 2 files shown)


tomeichlersmith:
Both failures are happening when trying to run the `compare.py` script

```
   TClass::Init:0: RuntimeWarning: no dictionary for class ldmx::RunHeader is available
    File "/home/runner/work/ldmx-sw/ldmx-sw/./.github/actions/validate/compare.py", line 195, in <module>
      compare(sys.argv[1], sys.argv[2], sys.argv[3], sys.argv[4])
    File "/home/runner/work/ldmx-sw/ldmx-sw/./.github/actions/validate/compare.py", line 162, in compare
      for key in gold.list_histograms() :
    File "/home/runner/work/ldmx-sw/ldmx-sw/./.github/actions/validate/compare.py", line 89, in list_histograms
      return flatten(self.__file.GetListOfKeys())
    File "/home/runner/work/ldmx-sw/ldmx-sw/./.github/actions/validate/compare.py", line 43, in flatten
      flat_l.extend(flatten(d.GetListOfKeys()))
    File "/home/runner/work/ldmx-sw/ldmx-sw/./.github/actions/validate/compare.py", line 43, in flatten
      flat_l.extend(flatten(d.GetListOfKeys()))
  ReferenceError: attempt to access a null-pointer
```

not sure why this is happening right now (maybe my branch doesn't have the right base?)

tomeichlersmith:
I did some investigating spurred by the line
```
   TClass::Init:0: RuntimeWarning: no dictionary for class ldmx::RunHeader is available
```
since there shouldn't be a run header inside the histogram files we are comparing.

Indeed, there is a bug I introduced into Framework when shuffling the order of writing output objects that puts the `LDMX_Run` tree into the histogram file if a processor changes the current ROOT directory to one within the histogram file during processing. The SimObjects DQM processor does exactly this so it can dynamically create histograms during processing. We want to support this functionality and so I'm working on patching Framework. https://github.com/LDMX-Software/Framework/pull/93

### In the Mean Time
This should not affect production especially since we don't have **any** histogram fillers within our production configs meaning the run tree is ending up in the output ROOT file.

PR #1256, opened 2024-01-29
add a script to print markdown list of submodule states

tomeichlersmith:
This script mainly uses awk to look through the submodule states and
retrieve their versions. If a submodule is not on a clean version tag,
then we fall back to just linking to the commit its on.

Besides awk, this script also requires a version of git that has the
`-C` flag (>= 1.8.5, very easy on any newer systems, I've only run into issues
with this requirement on CentOS7 systems.) and all submodules to have
their remote named `origin` which is the `git` default so I expect only
super-users to see issues.

## Use Case
This is helpful for anyone writing release notes. You can run this script and then copy the output into the release notes markdown to give an itemized list of the submodules linked to the commit/release that is included in that release.

In the future, we could look at including this script as part of a larger automatic release notes creation function. The current release notes generator just lists the PRs that were merged and then other notes are written manually. This script could be apart of a system that would itemize the list of PRs _and_ the submodule versions and then leave a space for manually written notes at the top.

### DIFF
--- a/scripts/state
+++ b/scripts/state
@@ -1,0 +1,41 @@
+#!/bin/sh
+# scripts/state
+#   display the state of ldmx-sw
+#
+# REQUIREMENTS
+# - awk availabe
+# - git has the '-C' flag (>= 1.8.5)
+# - all the submodules have their remotes named 'origin'
+
+set -o errexit
+set -o nounset
+
+git submodule foreach --recursive git fetch --tags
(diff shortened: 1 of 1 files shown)


tomeichlersmith:
Same as https://github.com/LDMX-Software/ldmx-sw/pull/1255#issuecomment-1915215856

PR #1266, opened 2024-02-23
Iss1265

### LINKED ISSUE #1265: Set up a processor to identify the beam electron positions at the target
There are many many performance studies that would benefit from truth info on where the beam electron hits the target and what momentum it had. 

I want a processor that takes the simhit collection made by the existing truth hit processor and figures out what the beam electron info is (by grouping them etc). 

Scoring plane info? But that has the same issue, that there are several hits along z per incoming electron, which means grouping is needed. Furthermore, currently pileup is not added to scoring planes, meaning that multielectron studies can't rely on this.

bryngemark:

This resolves #1265 , with the caveat that I don't have a way to get the particle momentum from the TargetSimHits but this might be added later from some other collection if people want to add this info. So I punt on that. 

## Check List

I ran the inclusive validation config once with 100 events, then ran it again with overlay generating 30 1e inclusive events on the fly, adding 2 pileup events to each. 

I ran `ldmx fire config.py` with the following config (first time `nEle=1`, second time `3`):
``` python

nEle=3
overlayString=""
thisPassName='sim'
if nEle > 1 :  #use overlay collection
    overlayString="Overlay"
    thisPassName='overlay'
pileupFilePassName='sim'  #to pull pileup from 
simPassName=thisPassName   #to pull "sim" event from -- this one; we run sim on the fly

p = ldmxcfg.Process(thisPassName)

mySim.setDetector( 'ldmx-det-v14' )
mySim.generators.append( gen.single_4gev_e_upstream_tagger() )

p.sequence = [ mySim ]


##################################################################
# Below should be the same for all sim scenarios

import os
import sys

p.run = 1+nEle
p.maxEvents = 30


p.outputFiles = [f'events{nEle}e.root']


from LDMX.Recon.overlay import OverlayProducer
overlay=OverlayProducer('events1e.root')  #use an earlier iteration of this setup. with runNb=1
overlay.passName = simPassName                  #sim input event pass name
overlay.overlayPassName = pileupFilePassName    #pileup input event pass name
overlay.totalNumberOfInteractions = float(nEle)
overlay.overlayCaloHitCollections=[ 
        "TriggerPad1SimHits", "TriggerPad2SimHits", "TriggerPad3SimHits", "TargetSimHits" ]
overlay.verbosity=0

if nEle > 1 : 
    p.sequence.extend([overlay])

#usual imports
import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry

#TS stuff can be used for comparisons but not shown/done here 
from LDMX.TrigScint.trigScint import TrigScintClusterProducer
from LDMX.TrigScint.trigScint import trigScintTrack
ts_digis = [
        TrigScintDigiProducer.pad1(),
        TrigScintDigiProducer.pad2(),
        TrigScintDigiProducer.pad3(),
        ]
for d in ts_digis :
    d.randomSeed = p.run
    d.input_collection = d.input_collection+overlayString

#run the truthhitprdocer on the target as this isolates truth beam electron hits and puts them in a new output collection
from LDMX.TrigScint.truthHits import TruthHitProducer
truthHits=TruthHitProducer('targetBeamElectrons')
truthHits.input_collection="TargetSimHits"+overlayString
truthHits.input_pass_name=thisPassName
truthHits.output_collection="truthBeamElectronsTarget"

#use the truthhit collection as input for this processor which picks out one location per beam electron
from LDMX.Recon.beamElectronLocator import BeamElectronLocator
beamLoc=BeamElectronLocator('BeamElectronLocator')
beamLoc.input_collection=truthHits.output_collection
beamLoc.input_pass_name = p.passName
beamLoc.granularity_X_mm = float(20/8)  # this is the envisioned TS granularity with vertical bars 
beamLoc.granularity_Y_mm = float(80/48) # this is the TS granularity in y
beamLoc.verbose = True

from LDMX.DQM import dqm

from LDMX.Recon.electronCounter import ElectronCounter
from LDMX.Recon.simpleTrigger import TriggerProcessor

count = ElectronCounter(nEle,'ElectronCounter')
count.input_pass_name = p.passName

p.termLogLevel=1

p.sequence.extend([
        *ts_digis,
        TrigScintClusterProducer.pad1(),
        TrigScintClusterProducer.pad2(),
        TrigScintClusterProducer.pad3(),
        trigScintTrack,
        count,
    truthHits,
    beamLoc
        ])


```

First plot shows, for one example 3e event (box option, size is proportional to counts), the (_x_,_y_) coordinates of TargetSimHits in black, the subset belonging to beam electrons in magenta, the locations found by the beam electron locator in dashed green, and the corresponding binned coordinates of the latter given a TS granularity in cyan. This all looks good. 

3e_targetHitXYblack_beamEhitsXYmagenta_beamEectronTruthCollXYgreen_beamElectronTruthCollBinnedXYcyan

In the next plot, the binned and "true" (within some minimum matching requirement window) _x_ coordinates from the beam locator are compared. If the actual _x_ coordinate is outside of the beam spot, its bin number is -1, corresponding to the "shifted" hits in the lower left corner. 
3e_truthX_vs_binnedX_30events
3e_truthY_vs_binnedY_30events

And the same distribution in _y_.
- [ ] I attached any sub-module related changes to this PR.
N/A

### DIFF
--- a/Recon/include/Recon/BeamElectronLocator.h
+++ b/Recon/include/Recon/BeamElectronLocator.h
@@ -15,9 +15,13 @@ namespace recon {
 /**
  * Electron counting processor.
  *
- * This processor will use objects reconstructed in the trigger scintillators,
- * or truth info on the number of electrons, and set the electron count in the
- * event.
+ * This processor uses calo simhits in e.g. the target to get truth info
+ * about beam electrons. The hits associated with beam electrons can already
+ * be isolated by the truth hit collection producer. However, the simhits
+ * have approximately infinite resolution. This processor's raison d'être 
+ * is to run some sort of  grouping, to ensure we have at most one 
+ * reconstructed truth information object per electron.
+ *
  */
 class BeamElectronLocator : public framework::Producer {
  public:
@@ -37,13 +41,18 @@ class BeamElectronLocator : public framework::Producer {
    * Configure the processor using the given user specified parameters.
    *
    * The user specified parameters that are availabed are defined
-   * in the python configuration class. Look at the my_processor.py
-   * module of the EventProc python for the python structure.
+   * in the python configuration class. Look at the beamElecronLocator.py
+   * in Recon/python for the python structure.
    *
    * @param parameters Set of parameters used to configure this processor.
    */
   void configure(framework::config::Parameters &parameters) final override;
 
+  /**
+   * Prints the configuration to log in debug mode
+   */
+  void onProcessStart() ;
+  
   /**
    * Process the event and put new data products into it.
    *
@@ -67,13 +76,31 @@ class BeamElectronLocator : public framework::Producer {
    * variables
    */
   std::string outputColl_;
-
+  
   /**
-   * The granularity of the TS in X, in mm 
+   * The min value measured by the system (edge) in X, in mm 
+   **/
+  double minXmm_;
+  /**
+   * The max value measured by the system (edge) in X, in mm 
+   **/
+  double maxXmm_;
+  
+  /**
+   * The min value measured by the system (edge) in Y, in mm 
+   **/
+  double minYmm_;
+  /**
+   * The max value measured by the system (edge) in Y, in mm 
+   **/
+  double maxYmm_;
+  
+  /**
+   * The granularity of the detector (e.g. TS) in X, in mm 
    **/
   double granularityXmm_;
   /**
-   * The granularity of the TS in Y, in mm 
+   * The granularity of the detector (e.g. TS) in Y, in mm 
    **/
   double granularityYmm_;
 
@@ -83,8 +110,19 @@ class BeamElectronLocator : public framework::Producer {
    **/
   double tolerance_;
 
+  /** 
+   * Indicate verbose printout to log according to log level. 
+   */
   bool verbose_{false};
-  
+
+  /**
+   * Bins coordinates according to some given granularity (passed as argument).
+   * Returns the lower bin edge. -1 for underflow (coordinate < min_in_mm); 
+   * with a system of N bins, returns n = N if coordinate > max_in_mm.
+   * 
+   * TODO also implement a function that returns the grid of non-empty hit coordinates,
+   * which accounts for that we don't know the multiplicity at a location
+   */
   int bin(float coordinate, float binWidth, float min, float max);
   
 };  // BeamElectronLocator
--- a/Recon/include/Recon/BeamElectronLocator.h
+++ b/Recon/include/Recon/BeamElectronLocator.h
@@ -8,6 +8,7 @@
 
 //---< Recon >---//
 #include "Recon/Event/BeamElectronTruth.h"
+#include "SimCore/Event/SimCalorimeterHit.h"
 
 namespace recon {
 
@@ -70,11 +71,11 @@ class BeamElectronLocator : public framework::Producer {
   /**
    * The granularity of the TS in X, in mm 
    **/
-  float TSgranularityXmm_;
+  float granularityXmm_;
   /**
    * The granularity of the TS in Y, in mm 
    **/
-  float TSgranularityYmm_;
+  float granularityYmm_;
 
   /**
    * The tolerance within which simhits are considered to belong to 
@@ -82,6 +83,7 @@ class BeamElectronLocator : public framework::Producer {
    **/
   float tolerance_;
 
+  int bin(float coordinate, float binWidth, float min, float max);
   
 };  // BeamElectronLocator
 }  // namespace recon
--- a/Recon/include/Recon/BeamElectronLocator.h
+++ b/Recon/include/Recon/BeamElectronLocator.h
@@ -18,8 +18,8 @@ namespace recon {
  * This processor uses calo simhits in e.g. the target to get truth info
  * about beam electrons. The hits associated with beam electrons can already
  * be isolated by the truth hit collection producer. However, the simhits
- * have approximately infinite resolution. This processor's raison d'être 
- * is to run some sort of  grouping, to ensure we have at most one 
+ * have approximately infinite resolution. This processor's raison d'être
+ * is to run some sort of  grouping, to ensure we have at most one
  * reconstructed truth information object per electron.
  *
  */
@@ -51,8 +51,8 @@ class BeamElectronLocator : public framework::Producer {
   /**
    * Prints the configuration to log in debug mode
    */
-  void onProcessStart() ;
-  
+  void onProcessStart();
+
   /**
    * Process the event and put new data products into it.
    *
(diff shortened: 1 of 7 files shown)


Recon/src/Recon/BeamElectronLocator.cxx
```diff
@@ -0,0 +1,123 @@
...
+  outputColl_ = parameters.getParameter<std::string>("output_collection");
+  granularityXmm_ =
+      parameters.getParameter<double>("granularity_X_mm");
+  granularityYmm_ =
+      parameters.getParameter<double>("granularity_Y_mm");
+  tolerance_ =
+      parameters.getParameter<double>("min_granularity_mm");
+  verbose_ =  
+    parameters.getParameter<bool>("verbose");
+
```
tomeichlersmith: Move the log message to the first callback after the logging has been set up.

```suggestion
}

void BeamElectronLocator::onProcessStart() {
```

This would also require declaring `onProcessStart` in the header.

Recon/src/Recon/BeamElectronLocator.cxx
```diff
@@ -0,0 +1,123 @@
...
+	  if  (min + n*binWidth > max) {
+		//n--; // this moves the point back, but, perhaps better to see that we go out of bounds on return 
+		break;
+	  }
+	}
+	// the n we have is the first bin which is too large. 
+	// the resulting coordinate is between n which is the upper edge, and the previous
+	// return the lower edge 
+	return n-1;
+  }
```
tomeichlersmith: Move some of these comments to doxygen-style in the header.

My general strategy is to put explanation of why this function exists, what it does, and any long term notes in the header and then implementation details in the source. Doxygen will still pick up on doxygen-style (`/** <notes> */`) in the source and so it is a good idea to write them that way so they are put into the manual. I'm especially think of the last three lines which would be helpful for a future developer using this function to see in the docs.

Recon/src/Recon/BeamElectronLocator.cxx
```diff
@@ -0,0 +1,123 @@
...
+	if (!event.exists(inputColl_, inputPassName_)) {
+	  ldmx_log(fatal) << "Attemping to use non-existing input collection "
+					  << inputColl_ << "_" << inputPassName_
+					  << " to locate electrons! Exiting.";
+	  return;
+	}
+
+	
+	// the simhits have approximately infinite resolution.
+	// this processor's raison d'être is to run some sort of grouping.
```
tomeichlersmith: Put this in the class description as well - right now it mentions recon or truth but the implementation only uses truth.

tomeichlersmith (commented):
Looks good, just a few documentation/cosmetic comments. The only overall thing I'd ask is to run a clang format:
```
ldmx clang-format -i --style=Google \
  Recon/include/Recon/BeamElectronLocator.h \
  Recon/include/Recon/Event/BeamElectronTruth.h \
  Recon/src/Recon/BeamElectronLocator.cxx \
  Recon/src/Recon/Event/BeamElectronTruth.cxx
```

tomeichlersmith (approved):

Recon/include/Recon/BeamElectronLocator.h
```diff
@@ -0,0 +1,132 @@
...
+   * in Recon/python for the python structure.
+   *
+   * @param parameters Set of parameters used to configure this processor.
+   */
+  void configure(framework::config::Parameters &parameters) final override;
+
+  /**
+   * Prints the configuration to log in debug mode
+   */
+  void onProcessStart();
```
EinarElen: Missing override here!

Recon/src/Recon/BeamElectronLocator.cxx
```diff
@@ -0,0 +1,122 @@
...
+                                            granularityYmm_);
+
+      beamElectronInfo.push_back(electronInfo);
+    }
+  }  // over simhits in the collection
+
+  event.add(outputColl_, beamElectronInfo);
+}
+
+int BeamElectronLocator::bin(float coordinate, float binWidth, float min,
```
EinarElen: Is there a reason for why we are doing single precision things when most of the data seems to be double precision?

Recon/include/Recon/BeamElectronLocator.h
```diff
@@ -0,0 +1,132 @@
...
+  BeamElectronLocator(const std::string &name, framework::Process &process);
+
+  /// Destructor
+  virtual ~BeamElectronLocator();
+
+  /**
+   * Configure the processor using the given user specified parameters.
+   *
+   * The user specified parameters that are availabed are defined
+   * in the python configuration class. Look at the beamElecronLocator.py
```
EinarElen, line 44: Really appreciate this!

Recon/src/Recon/BeamElectronLocator.cxx
```diff
@@ -0,0 +1,122 @@
...
+
+  for (const auto &simHit : simHits) {
+    // check if we already caught this position, else, add it
+    bool isMatched = false;
+    std::vector<float> pos = simHit.getPosition();
+    for (auto foundElectrons : beamElectronInfo) {
+      // this check makes it square rather than a dR circle
+      if (fabs(pos[0] - foundElectrons.getX()) < tolerance_ &&
+          fabs(pos[1] - foundElectrons.getY()) < tolerance_) {
+        if (verbose_)
```
EinarElen: Can we do curly-braces for single statement if-conditions? It is so easy to accidentally break otherwise and especially when combined with a bunch of stream operations (like <<) can produce compilation errors that are really hard for new people to decipher

Recon/src/Recon/BeamElectronLocator.cxx
```diff
@@ -0,0 +1,122 @@
...
+                          << foundElectrons.getX()
+                          << ", y = " << foundElectrons.getY()
+                          << "); skip this simhit";
+        isMatched = true;
+        break;  // finding a match means Move on
+      }         // if coordinates match something we already found
+    }           // over found electrons
+    if (!isMatched) {
+      if (verbose_) {
+        ldmx_log(info) << "\tHit at (x = " << pos[0] << ", y = " << pos[1]
```
EinarElen, line 79: Like this :)

EinarElen (commented):
Looks great, just had a couple of minor comments. I ran the build through both GCC/Clang with extended warnings as well as static analysis. Other than the missing override, everything looks good. 

Since we are touching the Recon module, could we take the opportunity to patch this 
```shell
/Users/einarelen/tmp/ldmx-sw/Recon/include/Recon/Event/HgcrocDigiCollection.h:447:11: warning: field 'coll_' will be initialized after field 'digi_index_' [-Wreorder-ctor]
        : coll_{c}, digi_index_{index} {}
          ^~~~~~~~  ~~~~~~~~~~~~~~~~~~
          digi_index_{index} coll_{c}
```
Basically, C++ members are always initialized in the order that they are declared in the class even if you list them in a different order. Not a huge issue here but it is one of those warnings that can catch some real nightmare bugs if we aren't used to ignoring it. Fix is literally just to swap the order in the list :)

Recon/src/Recon/BeamElectronLocator.cxx
bryngemark: Hmmm, no, I think it's even the other way around. When I had trouble compiling and running etc, for a while there I got configuration complaints about `d`s and `f`s (for the record, solved when recompiling _everything_, not just `Recon`) and having double precision data is a remnant from trying to understand this runtime complaint from different angles. For sure in a precision reducing processor most numbers really don't need to be double precision.

Recon/include/Recon/BeamElectronLocator.h
bryngemark: Haha thanks it's a tweaked comment from Tom :)

Recon/include/Recon/BeamElectronLocator.h
bryngemark: Ohh yes, thanks

Recon/src/Recon/BeamElectronLocator.cxx
EinarElen: Ok, I would probably advocate for doing double precision unless we have a good reason not to. Just a lot easier to avoid getting into trouble

Recon/src/Recon/BeamElectronLocator.cxx
bryngemark: Absolutely.

Recon/src/Recon/BeamElectronLocator.cxx
bryngemark: Ok, I verified that these configurable parameters have to be double precision. So, adjusting this function to respect precision of the different parameters.

bryngemark:
Ok, I made all the changes I wanted to, thanks both. I also swapped the order of `coll` and `digi _index` listed in the mentioned function in `HgcrocDigiCollection.h` to read  
```
iterator(HgcrocDigiCollection& c, long index = 0)
      : digi_index_{index}, coll_{c} {}
```
as Einar suggested. This builds but I haven't tested it since it's not really my domain.

bryngemark:
> This builds but I haven't tested it since it's not really my domain.

@EinarElen perhaps you have ways to test this, or confidence that it's fine.

EinarElen:
I'm confident that 
- if this changes something, our CI will notice 
- if this changes something, we had a bug that well need to look at 

i.e I'm perfectly happy to just go ahead

EinarElen (approved):

PR #1272, opened 2024-03-20
8GeV config added for inclusive sample

pritampalit:

## Check List

Link to [presentation](https://indico.fnal.gov/event/63853/#6-8-gev-inclusive-sample-valid)

- [ ] I attached any sub-module related changes to this PR.

To run this config : 
```
ldmx fire .github/validation_samples/inclusive/config_inclusive_8gev.py [run_number] [max_events] [max_TriesPerEvent]
```


bryngemark (commented):
I think we don't want a new file called `..._8gev`but actually an update to the existing file. 8 GeV is our new default and this should be reflected in our CI configs. Changing the original file also makes it easier to review your exact changes in the pull request.

bryngemark:
by the way i expect the associated test to break if you rename the existing file, since two different beam energies are compared. that is ok. it would be good to just take a quick look at the resulting histograms ("artifacts") that the diffrences are expected, just as you have done for validation.

.github/validation_samples/inclusive/config.py
```diff
@@ -20,8 +20,8 @@
 p.run = int(os.environ['LDMX_RUN_NUMBER'])
 p.maxEvents = int(os.environ['LDMX_NUM_EVENTS'])
 
-p.histogramFile = 'hist.root'
-p.outputFiles = ['events.root']
+p.histogramFile = 'hist_8gev.root'
+p.outputFiles = ['events_8gev.root']
```
tomeichlersmith: The output histogram and event files should keep the same names as before so that they are picked up by the CI.

```suggestion
p.histogramFile = 'hist.root'
p.outputFiles = ['events.root']
```

tomeichlersmith (changes requested):
Besides a niche output-filename change that is required for the CI, I can't think of anything else to change. As @bryngemark said, this will fail the PR Validation histograms but those histograms will be interesting to look at.

.github/validation_samples/inclusive/config.py
pritampalit: I changed the output file name, because probably the validation tool takes a key_val pair as the file name. I have again changed it to default one.

tomeichlersmith: That makes sense, the CI script pre-dates the Validation tool and we are hoping to unify them one day.

bryngemark (approved):

tvami (approved):

tvami:
I think this needs to go in before we change the CI configs to add the debug level info as discussed in https://github.com/LDMX-Software/ldmx-sw/pull/1277#issuecomment-2045344641

tomeichlersmith (approved):

PR #1273, opened 2024-03-31
1237 add an option to print steps of the daughters of the particle of interest

### LINKED ISSUE #1236: Step printer crashes when next volume is undefined
The step printer code assumes that the track has both a current and a next volume. When this isn't the case, we get a nullpointer dereference

### LINKED ISSUE #1237: Add an option to print steps of the daughters of the particle of interest
A typical use case for the step printer would be to look at the results of a PN-interaction. To do this, we would be interested not just in what happens to the photon but what happens to its daughter particles. This is straight-forward to add to the step printer by adding a check if either the trackID or parentID matches the track we asked for.

EinarElen:

This resolves #1237 and patches a rare crash with the stepprinter.
- Add an option to require a minimum energy for the PN products filter 
- Patch a crash in stepprinter when the track is in the world volume and the next step is outside the geometry 
- Add an option to the stepprinter that lets you track any particle produced by a particular process (e.g. "I want to see what happened to all particles from PN interactions")
- Add an option to the stepprinter to also print descendents of the track of interest up to an arbitrary depth

## Check List

```python
#!/usr/bin/env python3


p = ldmxcfg.Process('test')

p.maxTriesPerEvent = 10000

from LDMX.Biasing import ecal, util
mySim = ecal.photo_nuclear('ldmx-det-v14',gen.single_4gev_e_upstream_tagger())
mySim.description = 'ECal PN Test Simulation'
# Test how the step printer handles printing an arbitrary proces
# step = util.StepPrinter(process_name='photonNuclear')
# Regular tracking of a particular particle
step = util.StepPrinter(track_id=1)
# Track the secondaries of the first particle
step = util.StepPrinter(track_id=1, depth=2)
# step.depth=3
mySim.actions.extend([step])

p.sequence = [ mySim ]

##################################################################
# Below should be the same for all sim scenarios

import os
import sys

if 'LDMX_NUM_EVENTS' in os.environ:
    p.maxEvents = int(os.environ['LDMX_NUM_EVENTS'])
else:
if 'LDMX_RUN_NUMBER' in os.environ:
else:
    p.run = 1

if len(sys.argv) > 2:
    output_base = sys.argv[2]


# p.histogramFile = f'{output_base}_kind_hist.root'
p.outputFiles = ['output.root']

import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry
```

## Related Submodule PRs
- https://github.com/LDMX-Software/SimCore/pull/116

### DIFF
--- a/Biasing/include/Biasing/Utility/StepPrinter.h
+++ b/Biasing/include/Biasing/Utility/StepPrinter.h
@@ -45,6 +45,12 @@ class StepPrinter : public simcore::UserAction {
   int trackID_{-9999};
   std::string processName_{"UNDEFINED"};
   int depth_{0};
+  std::unordered_map<int,int> trackParents_{};
+
+
+  /// Check if the given track is a descendent of the track we are interested in
+  /// up to a certain depth
+  bool isDescendent(int trackID) const;
 
 };  // StepPrinter
 
--- a/Biasing/include/Biasing/Utility/StepPrinter.h
+++ b/Biasing/include/Biasing/Utility/StepPrinter.h
@@ -44,7 +44,7 @@ class StepPrinter : public simcore::UserAction {
   /// The track ID to filter on
   int trackID_{-9999};
   std::string processName_{"UNDEFINED"};
-  int depth{1};
+  int depth{0};
 
 };  // StepPrinter
 
--- a/Biasing/include/Biasing/Utility/StepPrinter.h
+++ b/Biasing/include/Biasing/Utility/StepPrinter.h
@@ -21,10 +21,11 @@ class StepPrinter : public simcore::UserAction {
    * @param[in] parameters the parameters used to configure this
    *      UserAction.
    */
-  StepPrinter(const std::string& name, framework::config::Parameters& parameters);
+  StepPrinter(const std::string& name,
+              framework::config::Parameters& parameters);
 
   /// Destructor
-  ~StepPrinter();
+  virtual ~StepPrinter() = default;
 
   /**
    * Stepping action called when a step is taken during tracking of
@@ -32,16 +33,17 @@ class StepPrinter : public simcore::UserAction {
    *
    * @param[in] step Geant4 step
    */
-  void stepping(const G4Step* step) final override;
+  void stepping(const G4Step* step) override;
 
   /// Retrieve the type of actions this class defines
-  std::vector<simcore::TYPE> getTypes() final override {
+  std::vector<simcore::TYPE> getTypes() override {
     return {simcore::TYPE::STEPPING};
   }
 
  private:
   /// The track ID to filter on
   int trackID_{-9999};
+  std::string processName_{"UNDEFINED"};
 
 };  // StepPrinter
 
--- a/Biasing/include/Biasing/Utility/StepPrinter.h
+++ b/Biasing/include/Biasing/Utility/StepPrinter.h
@@ -50,7 +50,7 @@ class StepPrinter : public simcore::UserAction {
 
   /// Check if the given track is a descendent of the track we are interested in
   /// up to a certain depth
-  bool isDescendent(int trackID) const;
+  bool isDescendent(const G4Track* track) const;
 
 };  // StepPrinter
 
--- a/Biasing/include/Biasing/Utility/StepPrinter.h
+++ b/Biasing/include/Biasing/Utility/StepPrinter.h
@@ -44,7 +44,7 @@ class StepPrinter : public simcore::UserAction {
   /// The track ID to filter on
   int trackID_{-9999};
   std::string processName_{"UNDEFINED"};
-  int depth{0};
+  int depth_{0};
 
 };  // StepPrinter
 
--- a/Biasing/include/Biasing/Utility/StepPrinter.h
+++ b/Biasing/include/Biasing/Utility/StepPrinter.h
@@ -5,6 +5,7 @@
 /*   SimCore   */
 /*~~~~~~~~~~~~~*/
 #include "SimCore/UserAction.h"
+#include "SimCore/G4User/TrackingAction.h"
 
 namespace biasing {
 namespace utility {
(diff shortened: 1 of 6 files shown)


Biasing/src/Biasing/Utility/StepPrinter.cxx
```diff
@@ -9,43 +9,94 @@
...
-  auto volume{track->GetVolume()->GetName()};
+  auto volume{track->GetVolume()};
+  auto volumeName{volume->GetName()};
 
-  // Get the next volume
-  auto nextVolume{track->GetNextVolume()->GetName()};
+  // Get the next volume (can fail if current volume is WorldPV and next is
+  // outside the world)
+  auto nextVolume{track->GetNextVolume() ? track->GetNextVolume()->GetName()
+                                         : "undefined"};
```
tomeichlersmith, line 66: Doesn't this resolve #1236 as well?

Biasing/include/Biasing/Utility/StepPrinter.h
```diff
@@ -21,27 +21,36 @@ class StepPrinter : public simcore::UserAction {
...
   /// The track ID to filter on
   int trackID_{-9999};
+  std::string processName_{"UNDEFINED"};
+  int depth_{0};
+  std::unordered_map<int,int> trackParents_{};
+
+
+  /// Check if the given track is a descendent of the track we are interested in
+  /// up to a certain depth
+  bool isDescendent(const G4Track* track) const;
```
tomeichlersmith: This functionality is already present (although inaccessible) within the `TrackMap`.

https://github.com/LDMX-Software/SimCore/blob/191f2eccfefb20f293f9c2e8ab5e89ec3da1d9aa/include/SimCore/TrackMap.h#L128-L129

Perhaps we add a function to `TrackMap` so that we don't have two copies of the ancestry tree?
```c++
bool TrackMap::isDescendent(int potential_ancestor_id, int track_id, int max_depth);
```

`TrackMap::findIncident` does something similar although instead of depth, it looks for first ancestor that was born outside of the calorimeter region. One can retrieve the `TrackMap` via
```cpp
g4user::TrackingAction::get()->getTrackMap()
```
We could add this retreival as a protected member of the `UserAction` class like [a few other helpers](https://github.com/LDMX-Software/SimCore/blob/191f2eccfefb20f293f9c2e8ab5e89ec3da1d9aa/include/SimCore/UserAction.h#L158-L177).

**Moving the implementation into `TrackMap` is not necessary**, but I am slightly concerned that this copying of the ancestry tree could leak into code that is more performance critical.

tomeichlersmith (commented):
It seems like the changes to PhotoNuclearProductsFilter are unrelated? I'm not opposed to including them, just wanted to know if this was related to another issue being resolved.

Looks good to me, I haven't tested it yet because I want to check your thoughts on my `TrackMap` comment before moving forward.

Biasing/src/Biasing/Utility/StepPrinter.cxx
EinarElen: Yep, that's the one that got me started on this... I had just forgotten about the issue

EinarElen:
> It seems like the changes to PhotoNuclearProductsFilter are unrelated? I'm not opposed to including them, just wanted to know if this was related to another issue being resolved.
> 
> Looks good to me, I haven't tested it yet because I want to check your thoughts on my `TrackMap` comment before moving forward.

They were developed together, but I don't think we made an issue for the pn change

EinarElen:
I didn't know that the trackmap stored a complete track/parent history (if it does, I would be a bit concerned about perf 💀)

Biasing/src/Biasing/Utility/StepPrinter.cxx
```diff
@@ -9,43 +9,94 @@
...
+                         framework::config::Parameters& parameters)
     : simcore::UserAction(name, parameters) {
   trackID_ = parameters.getParameter<int>("track_id");
+  processName_ = parameters.getParameter<std::string>("process_name");
+  depth_ = parameters.getParameter<int>("depth");
 }
 
-StepPrinter::~StepPrinter() {}
-
+bool StepPrinter::isDescendent(const G4Track* track) const {
```
bryngemark: i vote for the spelling isDescendant, regardless of where this ends up being implemented

bryngemark (approved):
i approve this pending resolution of tom's suggestion

Biasing/src/Biasing/Utility/StepPrinter.cxx
jpasc27: Agreed with the spelling!

jpasc27 (approved):
Looks good! I think these changes will be very useful.

Biasing/src/Biasing/Utility/StepPrinter.cxx
EinarElen: Aren't both of those valid in this context? I'm by no means about to argue about this though, happy to change it

EinarElen:
If we ever wanted an argument for moving away from submodules, I was about to say no to Tom's suggestions just because it would require a separate PR in SimCore :|

Biasing/src/Biasing/Utility/StepPrinter.cxx
jpasc27: I'm sure there's some semantics on why one over the other - I'm just used to seeing _descendant_, which is why I agreed with LK's comment. At the end of the day, not sure how much it matters, but maybe LK has a more robust argument that me (:

bryngemark: my take was that descendant is a noun and descendent is an adjective. and from the discussion and culturally enforced old habits i suppose i was led to think of descendant on equal footing as ancestor (parent, daughter), and just thought it was a typo. but like you, einar, i won't argue about it 😄 i see your point now.

PR #1274, opened 2024-04-02
Bump cmake and MagFieldMap

tomeichlersmith:
Just bumping some submodules to their most recent releases.

- [MagFieldMap v0.2.0](https://github.com/LDMX-Software/MagFieldMap/releases/tag/v0.2.0) unpacks the map at installation time rather than configuration time. This is desirable to avoid writing to the install directory unless the user actually wants to install. In addition, this allows users to change the install directory without needing to re-configure or re-compile (e.g. by using `DESTDIR`).
- [cmake v0.4.1](https://github.com/LDMX-Software/cmake/releases/tag/v0.4.1) adds an ABI warning to the list of compiler errors to ignore since there isn't much we can do about it (and we don't super care).


tvami:
hi @tomeichlersmith 
regarding the magnetic field bump: we plan to add another file that has no mag field (*) for the reduced LDMX running next year. So maybe let's bump that after that?

(*) maybe there is actually a better way to do this than adding a file with 0T everywhere, please let me know if you have another proposal. The advantage of doing the file is that we can just load that in the GDML that's being introduced in https://github.com/LDMX-Software/ldmx-sw/pull/1268

tomeichlersmith:
There is an easier way! You can simply omit the GDML lines defining the magnetic field and then there will not be any magnetic field in the simulated detector model. The lines I am talking about are these

https://github.com/LDMX-Software/ldmx-sw/blob/4460d5aee856c88b0b172ccde4d7a0008eac0923/Detectors/data/ldmx-det-v14/detector.gdml#L114-L121

tvami:
> There is an easier way! 

Perfect, thanks Tom!

SanjitMasanam:
Hello @tomeichlersmith,
we tried doing as you said and omitting those lines of the detector.gdml file but it resulted in the simulation producing 0 events.
`[ XsecBiasingOperator ]: Biasing particles of type gamma`
`[ Simulator ] : Started 1000000 events to produce 0 events.`
We tried a few different variations such as removing part of that gdml block of code or removing the reference to the magnet.gdml and all of these changes either resulted in configuration errors or the same issue of 0 events being produced. If you know why this is or have any other suggestions, please let us know

tomeichlersmith:
Without the magnetic field, the normal beam electron will be shot at such a wide angle that it will entirely miss the detector. I say this because if you simply removed the above lines from a current detector (e.g. `ldmx-det-v14-8gev`) and then ran a current standard config (e.g. `.github/validation_samples/ecal_pn/config.py`), you would observe this behavior. What config and detector did you use? If you want to run without magnetic field, you will need to update the primary generator for the simulation so that the electron will enter the detector as intended. I apologize for not mentioning this earlier, but the reduced detector (with its different beam configuration and detector volumes) will at least require different configuration scripts if not changes to C++ simulation filters to generate the samples of interest.

tomeichlersmith:
Marking this as ready-for-review after merging #1281 so that I can double check that the merged gold histograms/logs are re-created by a PR that didn't previously change the gold histograms/logs.

tvami (approved):

tvami:
> Without the magnetic field, the normal beam electron will be shot at such a wide angle that it will entirely miss the detector

I followed up in https://github.com/LDMX-Software/ldmx-sw/pull/1268

PR #1277, opened 2024-04-08
Move msgs in EcalProcessFilter to logging system

### LINKED ISSUE #1250: EcalPN is too verbose
Currently, there is no verbosity setting governing the rate of [this printout](https://github.com/LDMX-Software/ldmx-sw/blob/a0a28a311ca2da622af809ff9b6522fb4f793105/Biasing/src/Biasing/EcalProcessFilter.cxx#L169-L174): 
`[ EcalProcessFilter ]: (event nb) Brem photon produced (some number of) particle via biasWrapper(photonNuclear) process.`

While it looks harmless, this unnecessarily bloats the logs in large-scale production which in unlucky circumstances can cause the submission machine disk to fill up, which in turn halts production.

My preferred solution would be to demote this from always going to stdout, to debug level in our logger stream.

### LINKED ISSUE #1319: Put SimCore logging messages into logger format
This isn't very complicated, just cumbersome.

Now that Framework has the logging enabled and configurable from the python, we start updating classes to use the logger. My plan is to leave "verbosity" settings in each of the classes that have them so that a user can have specific channels be very verbose while keeping others quiet for debugging. But the messages can still be sent through the logger framework.

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1250

## Check List

Running 
```
ldmx fire ldmx-sw/.github/validation_samples/ecal_pn/config.py
```

Now the 
```
[ EcalProcessFilter ]: (event nb) Brem photon produced (some number of) particle via biasWrapper(photonNuclear) process.
```
 are not printed out by default.

N/A

### DIFF
--- a/Biasing/include/Biasing/EcalProcessFilter.h
+++ b/Biasing/include/Biasing/EcalProcessFilter.h
@@ -15,6 +15,7 @@
 /*   Framework   */
 /*~~~~~~~~~~~~~~~*/
 #include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
 
 // Forward declaration
 class G4Step;
@@ -62,6 +63,9 @@ class EcalProcessFilter : public simcore::UserAction {
 
   /// Process to filter
   std::string process_{""};
+  
+  /// Enable logging
+  enableLogging("EcalProcessFilter")
 
 };  // EcalProcessFilter
 }  // namespace biasing
--- a/Biasing/src/Biasing/EcalProcessFilter.cxx
+++ b/Biasing/src/Biasing/EcalProcessFilter.cxx
@@ -166,12 +166,12 @@ void EcalProcessFilter::stepping(const G4Step* step) {
       return;
     }
 
-    std::cout << "[ EcalProcessFilter ]: "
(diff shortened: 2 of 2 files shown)


tvami:
I also expect the log diff tests to fail with this

EinarElen (approved):

bryngemark (approved):
Really great, thanks for following up on this @tvami!

bryngemark:
indeed the log diff is (except a few time stamps from configuration printouts) an ecal printout for each event.  @tomeichlersmith @EinarElen thoughts on this, should the CI use logs at debug level to catch any differences?

EinarElen:
Seems reasonable to me!

tvami:
> should the CI use logs at debug level to catch any differences?

I agree with this! It should be for all the tests, so beyond this PR, maybe we should just made its own issue for this?

tomeichlersmith (approved):

tvami:
> > should the CI use logs at debug level to catch any differences?
> 
> I agree with this! It should be for all the tests, so beyond this PR, maybe we should just made its own issue for this?

For reference: https://github.com/LDMX-Software/ldmx-sw/issues/1280

PR #1282, opened 2024-04-10
Add scripts to compile, and recompile and fire a config (intro of `ldmx compile` and `ldmx recompFire`)

### LINKED ISSUE #1278: Simplify (re)compilation and use a script instead

Not really a problem but would be nice if we didn't need to run
```
ldmx cmake -B build -S .
ldmx cmake --build build --target install
```
all the time we need to (re)compile.

```
ldmx ./scripts/ldmx-compile.sh
```
where `ldmx-compile.sh` contains the commands above using `LDMX_BASE`

Copy-pasting every time when compilation is needed.

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1278

## Check List

The way I ran this was to copy the pn validation config to my location and named it config.py, then run
```
############################# Usage #############################
# ldmx ./scripts/ldmx-recompileAndFire.sh
# for changing the defaults
# ldmx setenv LDMX_COMPILE_CORES=<num cores to be used>, i.e.
# ldmx setenv LDMX_COMPILE_CORES=16
# or
# ldmx setenv LDMX_COMPILE_BUILD=<build location>
```
also ran
```
ldmx compile
```
and
```
ldmx recompFire config.py
```

N/A

### DIFF
--- a/scripts/ldmx-compile.sh
+++ b/scripts/ldmx-compile.sh
@@ -4,7 +4,7 @@
 # ldmx ./scripts/ldmx-compile.sh
 # for changing the defaults
 # ldmx setenv LDMX_COMPILE_CORES=<num cores to be used>, i.e.
-# ldmx setenv LDMX_COMPILE_CORES=8
+# ldmx setenv LDMX_COMPILE_CORES=16
 # or
 # ldmx setenv LDMX_COMPILE_BUILD=<build location>
 
@@ -19,5 +19,5 @@ fi
 echo "-- Compiling ldmx-sw in ${LDMX_COMPILE_BUILD} with ${LDMX_COMPILE_CORES} cores"
 
 # Compile ldmx-sw
-cmake -B ${LDMX_COMPILE_BUILD}/build -S .
+cmake -B ${LDMX_COMPILE_BUILD}/build -S ${LDMX_COMPILE_BUILD}
 cmake --build ${LDMX_COMPILE_BUILD}/build --target install -j=${LDMX_COMPILE_CORES}
--- a/scripts/ldmx-compile.sh
+++ b/scripts/ldmx-compile.sh
@@ -1,33 +1,23 @@
+#!/bin/bash
+
 ############################# Usage #############################
-# source scripts/ldmx-compile.sh.sh [-j <numCores>] [-b <buildLocation>]
-# e.g. source scripts/ldmx-compile.sh
-# or   source scripts/ldmx-compile.sh -j 16 -b ${LDMX_BASE}/ldmx-sw
+# ldmx ./scripts/ldmx-compile.sh
+# for changing the defaults
+# ldmx setenv LDMX_COMPILE_CORES=<num cores to be used>, i.e.
+# ldmx setenv LDMX_COMPILE_CORES=8
+# or
+# ldmx setenv LDMX_COMPILE_BUILD=<build location>
 
-# set number of cores to be used, build location
-while [[ $# -gt 0 ]]; do
-  # number of cores
-  if [[ $1 == "-j"* ]]; then
-      if [[ $1 =~ ^-j\s*[=]?\s*([0-9]+)$ ]]; then
-          CORES=${BASH_REMATCH[1]}
-      fi
-  else
-    CORES=$(nproc)
-  fi
-  # build directory
-  if [[ $1 == "-b"* ]]; then
-      if [[ $1 =~ ^-b\s*[=]?\s*([a-zA-Z0-9]+)$ ]]; then
-          BUILD=${BASH_REMATCH[1]}
-      fi
-  else
-      BUILD=${LDMX_BASE}/ldmx-sw
-  fi
-  shift
-done
+if [ -z "$LDMX_COMPILE_CORES" ]; then
+    LDMX_COMPILE_CORES=$(nproc)
+fi
 
-echo "-- Compiling ldmx-sw in ${BUILD} with ${CORES} cores"
+if [ -z "$LDMX_COMPILE_BUILD" ]; then
+    LDMX_COMPILE_BUILD=${LDMX_BASE}/ldmx-sw/
+fi
+
+echo "-- Compiling ldmx-sw in ${LDMX_COMPILE_BUILD} with ${LDMX_COMPILE_CORES} cores"
 
 # Compile ldmx-sw
-cd ${BUILD}
-ldmx cmake -B ${BUILD}/build -S .
-ldmx cmake --build ${BUILD}/build --target install -j=$CORES
-cd -
+cmake -B ${LDMX_COMPILE_BUILD}/build -S .
+cmake --build ${LDMX_COMPILE_BUILD}/build --target install -j=${LDMX_COMPILE_CORES}
--- a/scripts/ldmx-compile.sh
+++ b/scripts/ldmx-compile.sh
@@ -19,5 +19,5 @@ fi
 echo "-- Compiling ldmx-sw in ${LDMX_COMPILE_BUILD} with ${LDMX_COMPILE_CORES} cores"
 
 # Compile ldmx-sw
-cmake -B ${LDMX_COMPILE_BUILD}/build -S ${LDMX_COMPILE_BUILD}
+cmake -B ${LDMX_COMPILE_BUILD}/build -S ${LDMX_COMPILE_BUILD} $@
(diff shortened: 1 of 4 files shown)


tomeichlersmith (changes requested):
This review is mainly focused on the overall design choices. These scripts are functional and so I don't mean to block any merging - if I'm overruled by other devs, I'd be happy to merge this anyways. With that caveat, I am going to go through my general design-choice comments with their motivations in no particular order.

1. **Don't have these scripts be `source`d**: This could pollute the host environment (e.g. with the variables `CORES` and `BUILD`) which could confuse other programs. I would imagine these scripts as shell programs (i.e. with a shebang `#!/bin/bash` and executable permissions `chmod +x`) so that they are run in their own process.
2. **Have this run inside the container**: Every time you call `ldmx`, you trigger a container open/close cycle. This is not a huge performance issue on most systems, but it can be helpful to remember that and avoid more open/closes by putting an entire script inside the container. This is simply removing the `ldmx` prefixes in the script and forcing the user to run the script with the `ldmx ` prefix by checking for the existence of container-only files (`/.dockerenv` or `/singularity`).
3. **Avoid directory movement**: `cmake`, when given full paths for `-B`, `-S`, and `--build` handles the movement for us, so we can drop the `cd`. This has the added benefit of users not ending up in a different directory if the kill the process with ctrl+C[^1].
4. **Pass script arguments to commands and have script-specific args be environment variables**: I think this is a more natural way to partition the arguments supported by the script and the (larger) set of arguments supported by the command(s) being run in the script. Combined with (2), this would also naturally lead to `ldmx setenv LDMX_COMPILE_LOCATION=/full/path/to/ldmx-sw; ldmx ./scripts/ldmx-compile.sh` or something. Moving the script-specific args to environment variables also supports passing the entire command line to `fire` which allows users to have a config script that parses the command line to tune its configuration (something I use quite a lot especially to define input files).
5. **shellcheck**: https://www.shellcheck.net/ Install it or use the browser version. It just helps so much with catching tricky shell bugs and helps teach better shell coding practices. I have a [`.shellcheckrc`](https://github.com/LDMX-Software/ldmx-sw/blob/698cafa81f3bedd253f9e2027888e8e293e7aaf4/scripts/.shellcheckrc) that I think would help tune shellcheck for our scripts which gives you the options I would use and the warnings/errors I would disable.

[^1]: Probably? The `ldmx` suite of bash functions does do some directory movement and so I've noticed that I sometimes end up in `LDMX_BASE` after ctrl+C if I'm on a host running singularity. This aesthetic confusion would be resolved by using a program to launch the container (e.g. `denv`) which puts any directory movement into its own process rather than the current one.

bryngemark:
to add to @tomeichlersmith's ideas, i was thinking, wouldn't it be neat if the syntax to run these helper scripts were just `ldmx recompile` and similar? i.e. making them a function like many of the others we already have in `ldmx-env.sh`?

tvami:
Thanks for the suggestions! I made it work on my Mac+`docker`, now I'm going to test it out on a linux too and if it works, I can change it back from draft

tvami:
Works nicely on UCSB's cluster too, which uses `apptainer` (and is a linux)

scripts/ldmx-compile.sh
```diff
@@ -0,0 +1,23 @@
...
+############################# Usage #############################
+# ldmx ./scripts/ldmx-compile.sh
+# for changing the defaults
+# ldmx setenv LDMX_COMPILE_CORES=<num cores to be used>, i.e.
+# ldmx setenv LDMX_COMPILE_CORES=16
+# or
+# ldmx setenv LDMX_COMPILE_BUILD=<build location>
+
+if [ -z "$LDMX_COMPILE_CORES" ]; then
+    LDMX_COMPILE_CORES=$(nproc)
```
tomeichlersmith, line 12: I'm not super committed to this, but maybe we should have the default be one core. I say this because https://github.com/LDMX-Software/Tracking/issues/71 but I'm not sure how isolated that issue is.

scripts/ldmx-compile.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+if [ -z "$LDMX_COMPILE_BUILD" ]; then
+    LDMX_COMPILE_BUILD=${LDMX_BASE}/ldmx-sw/
+fi
+
+echo "-- Compiling ldmx-sw in ${LDMX_COMPILE_BUILD} with ${LDMX_COMPILE_CORES} cores"
+
+# Compile ldmx-sw
+cmake -B ${LDMX_COMPILE_BUILD}/build -S ${LDMX_COMPILE_BUILD}
```
tomeichlersmith: I like the idea of being able to pass arguments to cmake before the build.
```suggestion
cmake -B ${LDMX_COMPILE_BUILD}/build -S ${LDMX_COMPILE_BUILD} $@
```

scripts/ldmx-env.sh
```diff
@@ -537,6 +539,24 @@ __ldmx_checkout() {
...
+  ldmx . ${LDMX_BASE}/ldmx-sw/scripts/ldmx-compile.sh
+}
+
+###############################################################################
+# __ldmx_recompFire
+#   Recompile ldmx-sw and run fire on the first argument
+###############################################################################
+
+__ldmx_recompFire() {
+  ldmx . ${LDMX_BASE}/ldmx-sw/scripts/ldmx-recompileAndFire.sh $1
```
tomeichlersmith: Instead of just the first command line argument, we should pass all of them in case the user is passing arguments to their config on the command line (.e.g `ldmx fire config.py input.root` is one of the more common examples)

```suggestion
  ldmx . ${LDMX_BASE}/ldmx-sw/scripts/ldmx-recompileAndFire.sh $@
```

scripts/ldmx-recompileAndFire.sh
```diff
@@ -0,0 +1,27 @@
...
+
+############################# Usage #############################
+# ldmx ./scripts/ldmx-recompileAndFire.sh
+# for changing the defaults
+# ldmx setenv LDMX_COMPILE_CORES=<num cores to be used>, i.e.
+# ldmx setenv LDMX_COMPILE_CORES=16
+# or
+# ldmx setenv LDMX_COMPILE_BUILD=<build location>
+
+FIREINPUT=$1
```
tomeichlersmith: Instead of storing only the first argument, we should pass the entire command line to `fire`. For this reason, I'm just removing this local variable and giving `$@` to `fire` below.

```suggestion
```

scripts/ldmx-recompileAndFire.sh
```diff
@@ -0,0 +1,27 @@
...
+    LDMX_COMPILE_BUILD=${LDMX_BASE}/ldmx-sw/
+fi
+
+echo "-- Compiling ldmx-sw in ${LDMX_COMPILE_BUILD} with ${LDMX_COMPILE_CORES} cores, and running "
+
+# Compile ldmx-sw
+cmake -B ${LDMX_COMPILE_BUILD}/build -S ${LDMX_COMPILE_BUILD}
+cmake --build ${LDMX_COMPILE_BUILD}/build --target install -j=${LDMX_COMPILE_CORES}
+# Run fire on the input config
+fire $FIREINPUT
```
tomeichlersmith: Instead of storing only the first argument, we should pass the entire command line to fire. For this reason, I'm just giving `$@` to `fire` directly (no double quotes because we want `fire` to see the split command line and not just one string).

```suggestion
fire $@
```

tomeichlersmith (changes requested):
The only functionality change I have is about passing arguments so that these scripts work with configs that need additional command line arguments. I have other comments that I'm okay with dropping if folks don't like them.

scripts/ldmx-compile.sh
tvami: I'd prefer to keep it on `$(nproc)`, the issue you pointed to seems isolated and it takes quite some time to compile on a single core. If somebody runs into issues with the multi-core compilation they can still recompile with a single core, but given that is not what happens most of the time, I'd say let's keep the default on `$(nproc)`.

tvami:
I tested with a config that takes the number of events from an argument, works nicely!
```
ldmx recompFire yetAnotherConf.py 100
```

scripts/ldmx-compile.sh
tomeichlersmith: Sounds reasonable to me :+1:

tomeichlersmith (approved):

bryngemark (approved):

PR #1285, opened 2024-04-17
Nonfiducial filter for ECAL pn

tvami:

This work was done years ago at UCSB (by David Jiang). Given that we'd like to get rid of stale branches I rebased this and cleaned it a bit.
I'm opening the PR more of a discussion if this is still useful.

As for testing I ran `Biasing/test/ecal_pn_nonfiducial.py`
looks all good, and produces less events than the equivalent `Biasing/test/ecal_pn.py` which makes me thing that all the events are non-fiducial

## Check List

### Related Sub-Module PRs

https://github.com/LDMX-Software/SimCore/pull/117

### DIFF
--- a/Biasing/python/ecal.py
+++ b/Biasing/python/ecal.py
@@ -75,6 +75,42 @@ def photo_nuclear( detector, generator ) :
 
     return sim
 
+def nonfiducial_photo_nuclear( detector, generator ) :
+
+    # Instantiate the simulator. 
+    sim = simulator.simulator("photo-nuclear")
+    
+    # Set the path to the detector to use.
+    #   the second parameter says we want to include scoring planes
+    sim.setDetector( detector , True )
+    
+    # Set run parameters
+    sim.description = "ECal photo-nuclear, xsec bias 450"
+    sim.beamSpotSmear = [20., 80., 0.] #mm
+    
+    sim.generators.append( generator )
+    
+    # Enable and configure the biasing
+    sim.biasing_operators = [ bias_operators.PhotoNuclear('ecal',450.,2500.,only_children_of_primary = True) ]
+
+    # the following filters are in a library that needs to be included
+    includeBiasing.library()
+
+    # Configure the sequence in which user actions should be called.
+    sim.actions.extend([
+            filters.TaggerVetoFilter(),
+            # Only consider events where a hard brem occurs
+            filters.TargetBremFilter(),
+            # Only considers events that are Non-Fiducial (Doesn't enter an ECal volume)
+            filters.NonFiducialFilter(),
+            # Only consider events where a PN reaction happens in the ECal
+            filters.EcalProcessFilter(),     
+            # Tag all photo-nuclear tracks to persist them to the event.
+            util.TrackProcessFilter.photo_nuclear()
+    ])
+
+    return sim
+
 def gamma_mumu(detector, generator) :
     """Example configuration for biasing gamma to mu+ mu- conversions in the ecal.
 
--- a/Biasing/python/filters.py
+++ b/Biasing/python/filters.py
@@ -39,6 +39,26 @@ class TargetBremFilter(simcfg.UserAction):
         self.brem_min_energy_threshold = brem_min_e
         self.kill_recoil_track = False
 
+class NonFiducialFilter(simcfg.UserAction):
+    """ Configuration for rejecting events that are fiducial.
+
+    Parameters
+    ----------
+    recoil_max_p : float
+        Maximum momentum the recoil electron can have [MeV]
+    abort_fiducial: bool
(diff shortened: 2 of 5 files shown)


Biasing/python/ecal.py
```diff
@@ -75,6 +75,45 @@ def photo_nuclear( detector, generator ) :
...
+    sim.description = "ECal photo-nuclear, xsec bias 450"
+    sim.beamSpotSmear = [20., 80., 0.] #mm
+    
+    sim.generators.append( generator )
+    
+    # Enable and configure the biasing
+    #   In order to conform to older samples,
+    #   we can only attach to some of the volumes in the ecal
+    #   so we ask for this operator to be attached to the 'old_ecal'
+    sim.biasing_operators = [ bias_operators.PhotoNuclear('old_ecal',450.,2500.,only_children_of_primary = True) ]
```
tvami: Not sure what this `old_ecal` is actually doing, or if it's correct at all, @tomeichlersmith ?

Biasing/src/Biasing/TargetBremFilter.cxx
```diff
@@ -85,6 +85,8 @@ void TargetBremFilter::stepping(const G4Step* step) {
    */
   if (auto volume{track->GetNextVolume()->GetName()};
       volume.compareTo("recoil_PV") == 0
+//      volume.compareTo("recoil_PV") == 0)
+//      volume.compareTo("recoil") == 0)
```
tvami: similarly here I dont know how `volume.compareTo("recoil")` would differ from `olume.compareTo("recoil_PV")`, but I think these two lines dont need to be included anymore

Biasing/python/ecal.py
tomeichlersmith: https://github.com/LDMX-Software/SimCore/blob/191f2eccfefb20f293f9c2e8ab5e89ec3da1d9aa/src/SimCore/DetectorConstruction.cxx#L54 from three years ago lmao

I caught a very minor bug where an old method of attaching biasing operators was only biasing within the tungsten and silicon and not the other material. I suspect we can update this to `'ecal'` but that might end up changing distributions slightly since now PN is biased in more volumes.

Biasing/src/Biasing/TargetBremFilter.cxx
tomeichlersmith: The Geant4 GDML parser adds the `_PV` suffix and `compareTo` requires a full match (not just a substring), so using the full name `recoil_PV` is necessary here. You are correct, we could remove the extraneous comments for code cleanliness.

tvami:
I'll do the cleanup mentioned in https://github.com/LDMX-Software/ldmx-sw/pull/1285#discussion_r1569389393 
and also compare how moving from `old_ecal` to `ecal` changes. 

But generally, @tomeichlersmith do you agree that this PR adds something useful to ldmx-sw, or somehow since this branch was opened there are better ways to achieve the same thing?

tomeichlersmith:
Yea I think this is useful, I could definitely imagine someone wanting to generate a similar sample and this is the only way currently I know of to do that.

Biasing/src/Biasing/TargetBremFilter.cxx
tvami: Resolved in https://github.com/LDMX-Software/ldmx-sw/pull/1285/commits/20a7ceb54936afaf433a5e6b2ccfda606e99d4b3

Biasing/python/filters.py
```diff
@@ -39,6 +39,25 @@ def __init__(self,recoil_max_p = 1500.,brem_min_e = 2500.) :
...
 
+class NonFiducialFilter(simcfg.UserAction):
+    """ Configuration for rejecting events that are fiducial.
+
+    Parameters
+    ----------
+    recoil_max_p : float
+        Maximum momentum the recoil electron can have [MeV]
+    abort_fid_e: bool
+        If turned on, this aborts fiducial events. Else, this tags fiducial events.
```
tomeichlersmith: This is not a currently implemented option (although it sounds cool?)

In the interest of speed, I would just remove this comment.
```suggestion
```

Although if you can think of an easy/quick way to implement it, then go ahead.

tomeichlersmith (commented):
The SimCore changes have been included in SimCore v1.8.4 which is now on trunk. I tested a simple rebase locally which ran smoothly, so I would request another rebase onto trunk before merging:
```
git switch trunk
git pull
git switch -
git rebase trunk
```

tvami:
@tomeichlersmith I think you'll have to do the `merge trunk` method again :/ following 
```
git switch trunk
git pull
git switch -
git rebase trunk
```
resulted in a conflict

tomeichlersmith (approved):

PR #1289, opened 2024-04-27
Introduce DeepEcalProcessFilter

### LINKED ISSUE #1288: Create DeepEcalProcessFilter

We currently dont have a filter that would concentrate on having a brem in the deeper layers of ECAL

Make `DeepEcalProcessFilter` that  keeps events that undergo a conversion process in the Ecal-back (definition of that should be configurable)

**Additional context**

See @tylerhoroho-UVA 's talk at the SWAN meeting today:
https://indico.fnal.gov/event/64332/timetable/#16-late-showering-photon-backg

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1288

## Check List

Produced events with the depth of 400:
```
Started 2962289 events to produce 10000 events.
```

@tylerhoroho-UVA is testings them now

N/A

### DIFF
--- a/Biasing/include/Biasing/DeepEcalProcessFilter.h
+++ b/Biasing/include/Biasing/DeepEcalProcessFilter.h
@@ -71,6 +71,8 @@ class DeepEcalProcessFilter : public simcore::UserAction {
   std::vector<std::string> processes_{"conv","phot"};
   /// Minimum Z location where the deep process should happen
   double ecal_min_Z_{400.};
+  /// Require that the hard brem photon originates from the target
+  bool require_photon_fromTarget_{false};
   /// Enable logging
   enableLogging("DeepEcalProcessFilter")
 
--- a/Biasing/include/Biasing/DeepEcalProcessFilter.h
+++ b/Biasing/include/Biasing/DeepEcalProcessFilter.h
@@ -1,0 +1,80 @@
+/**
+ * @file DeepEcalProcessFilter.h
+ * @class DeepEcalProcessFilter
+ * @brief Class defining a UserActionPlugin that allows a user to filter out
+ *        events where the interaction happened deep in the ECAL
+ * @author Tamas Almos Vami (UCSB)
+ * @author Anmol Sandhu, Tyler Horoho (UVA)
+ */
+
+#ifndef BIASING_DEEPECALFILTER_H
+#define BIASING_DEEPECALFILTER_H
+
+//----------------//
+//   C++ StdLib   //
+//----------------//
+#include <algorithm>
+
+/*~~~~~~~~~~~~~*/
+/*   SimCore   */
+/*~~~~~~~~~~~~~*/
+#include "SimCore/UserAction.h"
+
+/*~~~~~~~~~~~~~~~*/
+/*   Framework   */
+/*~~~~~~~~~~~~~~~*/
+#include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
+
+namespace biasing {
+
+/**
+ * User action that allows a user to filter out events where the interaction happened deep in the ECAL
+ */
+class DeepEcalProcessFilter : public simcore::UserAction {
+ public:
+  /// Constructor
+  DeepEcalProcessFilter(const std::string& name,
+                    framework::config::Parameters& parameters);
+
+  /// Destructor
+  virtual ~DeepEcalProcessFilter() = default;
+  
+  /// Method to set flags in the beginning of the event
+  void BeginOfEventAction(const G4Event* event) final override;
+
+  /**
+   * Implement the stepping action which performs the target volume biasing.
+   * @param step The Geant4 step.
+   */
+  void stepping(const G4Step* step) final override;
+
+//  /**
+//   * Method called at the end of every event.
+//   * @param event Geant4 event object.
+//   */
+//  void EndOfEventAction(const G4Event*) final override;
+  
+  void NewStage() final override;
+
+  /// Retrieve the type of actions this class defines
+  std::vector<simcore::TYPE> getTypes() final override {
+    return {simcore::TYPE::STACKING,
+            simcore::TYPE::STEPPING,simcore::TYPE::EVENT};
+  }
+
+ private:
+  /// Minimal energy the products  should have
+  double bias_threshold_{1500.};
+  ///  The allowed processes that can happen deep inside the ECAL, default is
+  ///  conversion (conv) and photoelectron (photo)
+  std::vector<std::string> processes_{"conv","phot"};
+  /// Minimum Z location where the deep process should happen
+  double ecal_min_Z_{400.};
+  /// Enable logging
+  enableLogging("DeepEcalProcessFilter")
+
+};  // DeepEcalProcessFilter
+}  // namespace biasing
+
+#endif  // BIASING_DEEPECALFILTER_H
--- a/Biasing/include/Biasing/DeepEcalProcessFilter.h
+++ b/Biasing/include/Biasing/DeepEcalProcessFilter.h
@@ -75,7 +75,10 @@ class DeepEcalProcessFilter : public simcore::UserAction {
   bool require_photon_fromTarget_{false};
   /// Enable logging
   enableLogging("DeepEcalProcessFilter")
-
+  /// member used to help tag events where the photon comes from the target
+  bool photonFromTarget_{false};
+  /// member used to help tag events that have a deep-ecal process ocurr
+  bool hasDeepEcalProcess_{false};
 };  // DeepEcalProcessFilter
 }  // namespace biasing
 
--- a/Biasing/src/Biasing/DeepEcalProcessFilter.cxx
+++ b/Biasing/src/Biasing/DeepEcalProcessFilter.cxx
@@ -17,9 +17,6 @@
 
 namespace biasing {
 
-bool hasProcessNeeded_{false};
-bool photonFromTarget_{false};
-bool hasDeepEcalProcess_{false};
 
 DeepEcalProcessFilter::DeepEcalProcessFilter(const std::string& name,
                                      framework::config::Parameters& parameters)
@@ -37,8 +34,6 @@ void DeepEcalProcessFilter::BeginOfEventAction(const G4Event* event) {
 }
 
 void DeepEcalProcessFilter::stepping(const G4Step* step) {
-  hasProcessNeeded_ = false;
-  
   // Get the track associated with this step.
   auto track{step->GetTrack()};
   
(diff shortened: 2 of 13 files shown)


tvami:
As a validation I ran the config at 400, the hard brem seem to stop there and that's where the daughters are starting: 
Screenshot 2024-05-02 at 19 04 53

@tylerhoroho-UVA I put the file that contains 276 events (I meant to have 1000, but from 10000000 tries it could only go upto 276 events [also meaning that the z=500 will take a long time to produce stats])
```
/u/gu/tamasvami/events_pn_deep_test_276events.root
```

tvami:
In https://github.com/LDMX-Software/ldmx-sw/pull/1289/commits/24f9e0ba0a0b7a15b4ca80ae642c25c9a8447974?diff=split&w=1 I updated the DQM plots to show the particle types, now it looks like this:
Screenshot 2024-05-03 at 15 25 26

tvami:
OK I'm done with the filter that doesnt require the photon to come from the target.
The changes for the file size: the brem photon is tagged, no other brem is saved; tagging the relevant conversion vs saving all conversion has a ~15% increase in size (152KB vs 130KB per event) so I am more comfortable to save all of them, this makes the code easier to read anyway (I can say more if needed why the code gets complicated otherwise) and makes sure we dont miss any of the relevant physics.

Tyler in the meanwhile is taking an event from my event.root and resimulates it with the target filter on to see what's happening. We may want to have that part decoupled from this PR, although we are not in a rush.

tvami:
OK in the last push I introduced `require_photon_fromTarget_` that governs if we want to force the photon to come from the target. Worked nicely on one event, I'll make a bit more and then move the PR to "ready to review"

There is also a new config that runs the from target photon case:
```
Biasing/test/ecal_conv_deep_fromTarget.py
```

tvami:
Interesting, the `volume.contains("target")` actually brings in much more than just the target. 
Screenshot 2024-05-19 at 09 13 53

@tylerhoroho-UVA I assume this is not good. Maybe we could enforce instead that it's coming from around Z=0 instead of going thru the volume definition. Alternatively if these are called `TS_trgt` or something like that I could enforce that it should be target but not any TS related. But since the target will always be at Z=0(ish) I think that's a less involved solution. What do you think?

horohoo:
Interesting. From looking at the .gdml files I don't see any objects (other than the target) that contain a `"target"` substring in their volume name. Maybe it's something on the G4 level that's assigning the name?
If z values can be used instead of substrings in volumes, I think -0.5 < z < 0.5 would fully contain everything in the target plus a little extra room.

tvami:
Yes, I think everything the the target.gdml, i.e. these https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Detectors/data/ldmx-det-v14-8gev/target.gdml#L72-L91 are going to be called target by G4. 

OK, let's do it with the `-0.5 < z < 0.5`, I'll push the commit when I have it

tvami:
Talked to Tyler about the issue above, and we concluded that it's better to keep this as it is (in real data we wont know if the interaction happened on the TS or the target), and we can always just remove these at the analysis level. It's also a small effect

horohoo (approved):

DQM/include/DQM/SampleValidation.h
```diff
@@ -7,25 +7,22 @@
...
+class SampleValidation : public framework::Analyzer {
+  public:
+    SampleValidation(const std::string& name, framework::Process& process) : Analyzer(name, process) {}
+    virtual void configure(framework::config::Parameters& ps);
+    virtual void analyze(const framework::Event& event);
+    int pdgid_label(const int pdgid);
+    /// Method executed before processing of events begins.
+    void onProcessStart() override;
+    private:
+
```
tomeichlersmith:
```suggestion
```

Biasing/test/resim_ecal_conv_deep.py
```diff
@@ -0,0 +1,46 @@
...
+from LDMX.Hcal import HcalGeometry
+import LDMX.Ecal.ecal_hardcoded_conditions as ecal_conditions
+import LDMX.Hcal.hcal_hardcoded_conditions as hcal_conditions
+from LDMX.Biasing import ecal
+from LDMX.SimCore import generators
+from LDMX.Biasing import filters
+
+det = 'ldmx-det-v14-8gev'
+#mysim = ecal.deep_photo_nuclear(det, generators.single_8gev_e_upstream_tagger(), bias_threshold = 5010., processes=['conv','phot)'], ecal_min_Z = 300.)
+mysim = ecal.deep_photo_nuclear(det, generators.single_8gev_e_upstream_tagger(), bias_threshold = 3000., processes=['conv','phot)'], ecal_min_Z = 400.)
```
tomeichlersmith, line 15: This definition of a simulator does not match the one in the original simulation which can lead to funky results when re-simulating. Maybe just add a comment explaining that the user should align these two?

Biasing/src/Biasing/DeepEcalProcessFilter.cxx
```diff
@@ -0,0 +1,128 @@
...
+/*   SimCore   */
+/*~~~~~~~~~~~~~*/
+#include "SimCore/UserEventInformation.h"
+#include "SimCore/UserTrackInformation.h"
+
+namespace biasing {
+
+bool hasProcessNeeded_{false};
+bool photonFromTarget_{false};
+bool hasDeepEcalProcess_{false};
```
tomeichlersmith:
```suggestion
```

Biasing/include/Biasing/DeepEcalProcessFilter.h
```diff
@@ -0,0 +1,82 @@
...
+  ///  The allowed processes that can happen deep inside the ECAL, default is
+  ///  conversion (conv) and photoelectron (photo)
+  std::vector<std::string> processes_{"conv","phot"};
+  /// Minimum Z location where the deep process should happen
+  double ecal_min_Z_{400.};
+  /// Require that the hard brem photon originates from the target
+  bool require_photon_fromTarget_{false};
+  /// Enable logging
+  enableLogging("DeepEcalProcessFilter")
+
```
tomeichlersmith:
```suggestion
  /// member used to help tag events where the photon comes from the target
  bool photonFromTarget_{false};
  /// member used to help tag events that have a deep-ecal process ocurr
  bool hasDeepEcalProcess_{false};
```

Biasing/src/Biasing/DeepEcalProcessFilter.cxx
```diff
@@ -0,0 +1,128 @@
...
+
+
+void DeepEcalProcessFilter::BeginOfEventAction(const G4Event* event) {
+  hasDeepEcalProcess_ = false;
+  photonFromTarget_ = false;
+}
+
+void DeepEcalProcessFilter::stepping(const G4Step* step) {
+  hasProcessNeeded_ = false;
+
```
tomeichlersmith:
```suggestion
```

Biasing/src/Biasing/DeepEcalProcessFilter.cxx
```diff
@@ -0,0 +1,128 @@
...
+    }
+  }
+  
+  // If we require that the photon comes from the target and
+  // and if it does not, let's skip the event
+  if (require_photon_fromTarget_ and !photonFromTarget_) {
+    return;
+  }
+
+  // Tag if the event has the processes we are looking for
```
tomeichlersmith, line 75:
```suggestion
  // Tag if the event has the processes we are looking for
  bool hasProcessNeeded{false};
```

Biasing/src/Biasing/DeepEcalProcessFilter.cxx
```diff
@@ -0,0 +1,128 @@
...
+  for (auto& process : processes_) {
+    // ldmx_log(debug) << "Allowed processed " << process << " now we have " << processName;
+    if (processName.contains(process)) {
+      hasProcessNeeded_ = true;
```
tomeichlersmith:
```suggestion
      hasProcessNeeded = true;
```

Biasing/src/Biasing/DeepEcalProcessFilter.cxx
```diff
@@ -0,0 +1,128 @@
...
+      break;
+    }
+  }
+
```
tomeichlersmith:
```suggestion
  // skip this step if it does not have any of the processes needed
  if (not hasProcessNeeded) return;

```

The goal in moving this check earlier is two fold:
1. Keep a value's definition and usage close together so folks can see how it is used.
2. Avoid costly string comparisons below if they aren't needed.

Biasing/src/Biasing/DeepEcalProcessFilter.cxx
```diff
@@ -0,0 +1,128 @@
...
+  auto isInEcal{((volume.contains("Si") || volume.contains("W") ||
+                  volume.contains("PCB") || volume.contains("strongback") ||
+                  volume.contains("Glue") || volume.contains("CFMix") ||
+                  volume.contains("Al") || volume.contains("C")) &&
+                 volume.contains("volume")) ||
+    (volume.contains("nohole_motherboard"))};
+  
+  // Skip this step if it does not have the processes needed
+  // or if it's not in the ECAL
+  if (!hasProcessNeeded_ or !isInEcal) return;
```
tomeichlersmith:
```suggestion
  if (not isInEcal) return;
```

tomeichlersmith (changes requested):
I have some suggested refactors hoping to insure ourselves against future confusion. Mainly moving variable declarations, definitions, and usage "close" to each other in the code.

## Larger Comment
This filter makes the decision on whether to abort an event within `NewStage` which makes sense and allows for the simulation to avoid costly shower simulations that are unneeded. Nevertheless, this choice ties the filter closely to the PartialEnergySorter. I would encourage you to survey values of the biasing threshold and the sorting threshold to see how high you can raise it before losing access to the events you want to study. Generally, the higher the thresholds are, the less of the shower the simulation needs to do before it has the opportunity to abort the event.

Biasing/test/resim_ecal_conv_deep.py
```diff
@@ -0,0 +1,46 @@
...
+from LDMX.Ecal import EcalGeometry
+from LDMX.Hcal import HcalGeometry
+import LDMX.Ecal.ecal_hardcoded_conditions as ecal_conditions
+import LDMX.Hcal.hcal_hardcoded_conditions as hcal_conditions
+from LDMX.Biasing import ecal
+from LDMX.SimCore import generators
+from LDMX.Biasing import filters
+
+det = 'ldmx-det-v14-8gev'
+#mysim = ecal.deep_photo_nuclear(det, generators.single_8gev_e_upstream_tagger(), bias_threshold = 5010., processes=['conv','phot)'], ecal_min_Z = 300.)
```
tvami, line 14:
```suggestion
# Please make sure that the simulator here matches the simulator in the original config
#mysim = ecal.deep_photo_nuclear(det, generators.single_8gev_e_upstream_tagger(), bias_threshold = 5010., processes=['conv','phot)'], ecal_min_Z = 300.)
```

tomeichlersmith (approved):

tvami:
> I would encourage you to survey values of the biasing threshold and the sorting threshold to see how high you can raise it before losing access to the events you want to study. 

Maybe this could be a good study for Anmol, what do you think @tylerhoroho-UVA ?
(This certainly can be done after the PR is merged since it's configuration variable)

I ran the test file again after the changes (with a start Z of 240 mm), the DQM histograms look good
Screenshot 2024-06-03 at 10 17 29

Issue #1275, opened 2024-04-05
[Discussion] Move back from the submodule approach to a monorepo


EinarElen:
I'm making this issue so we can have the discussions regarding whether or not to move away from submodules in a persistent place (as opposed to slack 😄).

Summarizing the discussion on Slack briefly

From @tomeichlersmith 

- Proposed timeline 
  - T: the time when the submodule->monorepo conversion is done
  - T - 1 week: announce last call for pre-conversion contributions.
  - T - 3 days: write docs page about conversion (e.g. how to update your branch) and post reminder announcement
  - T - 1 day: last merges and final releases of submodules pre-conversion, make pre-conversion ldmx-sw release
  - T - 2 hours: convert submodules into subdirectories and post PR
  - T: merge conversion PR
  - T + 1 hour: new ldmx-sw release marking change after double checking functionality locally
  - T + 1 day: transfer issues from submodules to ldmx-sw repo, tag with relevant subdirectory
  - T + 1 week: add comment to submodule README, archive submodules
  - T + 2 weeks: test run past ldmx-sw version to make sure archived submodules are still accessible as promised

Questions:
- What should `T` be?
- How to deal with stale branches (Tom favours removing them)
- Should this be a major release? 
- What parts can/should remain submodules?
  - G4DarkBreM (separate paper)
  - MagFieldMap (Large and stable, needs to be checked if it can be in a monorepo)
  - Framework (Up for discussion, but has utility in being separate + quite stable)
  - Projects owned by others (ACTS, ruckus, doxygen-awesome-css, HLS_arbitrary_Precision_Types)

From @EinarElen:
- T should be _before_ we start getting summer students 

From @tvami 
- Keeping G4DarkBreM separate makes sense. Does MagFieldMap and Framework need to be?

tvami:
My main point is that maybe the measure should be "how ldmx specific the package is" instead of stability. And given that both `Framework` and `MagFieldMap` are very ldmx specific, I'd argue that they should also be merged into the mono-repo.

tomeichlersmith:
Okay, after doing some extremely simple estimates, I am not worried about pulling `MagFieldMap` into `ldmx-sw` anymore.
```
tom@nixos:~/code/ldmx/monorepo-ification/ldmx-sw$ du -sh \
  --exclude 'Tracking/acts*' \
  --exclude '*G4DarkBreM*' \
  --exclude '*doxygen-awesome-css*' \
[...]

PR #1290, opened 2024-04-29
Conditions on v0.2.0

tomeichlersmith:

This is helpful preparation for #1275 by closing the last PR in Conditions and then making the last release (hopefully if this doesn't break anything which it shouldn't).

## Check List

### Related Sub-Module PRs

- https://github.com/LDMX-Software/Conditions/pull/19


tomeichlersmith:
The failing plots are the same as from #1245

PR #1295, opened 2024-05-01
add config for studying ecal resolution

tomeichlersmith:
This is an re-opening of a submodule PR (https://github.com/LDMX-Software/Ecal/pull/50) where I followed [the instructions](https://ldmx-software.github.io/transition-to-ldmx-sw-v4.html) to re-apply my changes in one commit onto the subdirectory.

---

I've copied this config around for awhile so it is time to put it into the repository.

This config is focused on firing electrons _directly_ into the ECal and then running the various processors that emulate electronics and reconstruct stuff within the ECal so users can then analyze the output.

## To Do
- [x] Compare module-level sum between trigger primitives and precision rec hits
- [x] Compare trigger primitives to trigger-group-sum of precision rec hits
- [ ] Write a processor that re-calculates the trigger primitives using the precision rec hits. This allows us to see how imprecise the trigger primitive "condensation" algorithm is.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,8 @@ EventDisplay/include/EventDisplay/EventDisplayLinkDef.h
 *.DS_Store
 *.vscode/*
 *__*
+
+# data files from running configs
+*.root
+# except special files storing test results
+!.github/validation_samples/**.root
--- a/Ecal/exampleConfigs/ecal-resolution-study-with-trig-prim-monoenergetic-beam.py
+++ b/Ecal/exampleConfigs/ecal-resolution-study-with-trig-prim-monoenergetic-beam.py
@@ -1,0 +1,85 @@
+"""Ecal-focused config script
+
+This config script allows the user to choose the energy and angle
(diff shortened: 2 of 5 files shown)


tvami:
The `denv` changes seem to have been also added to this PR.

If you need help with the TODO, there is somebody at UCSB who is looking for a project. This also means that it wont be done very quickly, so if it's urgent then maybe not the best thing to give him.

tomeichlersmith:
Thanks :+1: I was being too quick with creating new branches!

We already have a grad student here at UMN working on this (not sure on his GitHub username right now), so no need to give it to the UCSB student.

tomeichlersmith (commented):
## Clean-Up To Dos Before Merge
- [x] Rebase excluding ROOT files erroneously commited in earlier commits
- [x] Add `*.root` files to `.gitignore`? Would need to check this doesn't break CI which relies on golden histogram ROOT files.
- [x] Documentation pass
- [x] Promote compile-time constants to configuration parameters that copy from existing reco

PR #1322, opened 2024-05-06
Introduce variable for total events with no tries limit

### LINKED ISSUE #1299: Add an ability to ask for N total events to process in addition to the maximum number of events
This comes from https://github.com/LDMX-Software/LDCS/pull/21 

When you don't know roughly how many events you need to try to achieve a certain number of final events, it would be useful if you could (optionally) specify how many events you want in total, e.g. I want to produce 10000 kaon events regardless of how many tries that would take

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1299

## Check List

I ran the pn code and set 
```
p.run = 2
p.maxEvents = 1
```
 * In default I get
```
Started 23 events to produce 1 events.
```

 * then
```
p.maxEvents = 1
p.maxTriesPerEvent = 10
```
leads to
```
Started 10 events to produce 1 events.
```

 * then
```
p.maxEvents = 1
p.maxTriesPerEvent = 10
p.totalEvents = 1
```
leads again to 
```
[ Process ] 2 : The totalEvents was set, so maxEvents and maxTriesPerEvent will be ignored!
Started 23 events to produce 1 events.
```
as expected

### DIFF
--- a/Framework/src/Framework/Process.cxx
+++ b/Framework/src/Framework/Process.cxx
@@ -30,6 +30,7 @@ Process::Process(const framework::config::Parameters &configuration)
 
   maxTries_ = configuration.getParameter<int>("maxTriesPerEvent", 1);
   eventLimit_ = configuration.getParameter<int>("maxEvents", -1);
+  totalEvents_ = configuration.getParameter<int>("totalEvents", -1);
   logFrequency_ = configuration.getParameter<int>("logFrequency", -1);
   compressionSetting_ =
       configuration.getParameter<int>("compressionSetting", 9);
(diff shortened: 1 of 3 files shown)


Framework/src/Framework/Process.cxx
```diff
@@ -237,6 +245,12 @@ void Process::run() {
     runHeader.setNumTries(totalTries);
     ldmx_log(info) << runHeader;
     outFile.writeRunTree();
+    
+    // starting at line 240 of Process.cxx on trunk
```
tomeichlersmith:
```suggestion
```

tomeichlersmith (changes requested):
This looks good, I would just ask for the `totalEvents` config option to be added to the python config class documentation and include reference to it in the `max*` descriptions since they will be ignored if it is used.

https://github.com/LDMX-Software/ldmx-sw/blob/11d6a8e89250b357885602075c88b423dc04cdeb/Framework/python/ldmxcfg.py.in#L330-L334

tvami:
Thanks Tom, I've changed those points

tomeichlersmith (approved):

danyi211 (approved):

horohoo (approved):

PR #1362, opened 2024-06-07
Clang format ldmx-sw

### LINKED ISSUE #1294: Clang Format the Source Code
A few of the submodules were making sure that the source code is clang-formatted in the Google style. Now that these submodules are not their own repos, the actions checking their style are not being run. Implementing an ldmx-sw wide format check would require everyone to format their code. `clang-format` is installed within the dev images, so formatting (and testing the format) can be done via `ldmx clang-format` both locally and in CI.

## For Example, SimCore
- https://github.com/LDMX-Software/ldmx-sw/blob/trunk/SimCore/.clang-format
- https://github.com/LDMX-Software/ldmx-sw/blob/trunk/SimCore/.github/workflows/format-check.yml
- https://github.com/LDMX-Software/ldmx-sw/blob/trunk/SimCore/.github/workflows/format-check.yml

Once format testing is done in the ldmx-sw actions (`.github`), then the directories `SimCore/.github`, `Hcal/.github`, `Framework/.github` can be removed. Other dangling `.clang-format` files can be removed as well.

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1294

## Check List

Running 
```
git ls-tree -r HEAD --name-only | egrep '(\.h|\.cxx)$' > format.list
clang-format --style=Google -Werror --dry-run $(cat format.list)
```
returns nothing now!
There is absolutely nothing expected to change, this is a cosmetics only PR.

After this is merged, we should change the CI to make the clang format checks automatically.


danyi211 (approved):

tomeichlersmith (changes requested):
I'm thinking we could introduce the clang-format testing in this PR as well and I don't think it would be too hard given we've done it for other submodules in the past.

- [x] Copy `.clang-format` from `Framework/.clang-format` (it's just saying using the Google style right now, but having a file will make it easier in the future to define ldmx-sw wide format definitions)
- [x] Copy over action from Framework or SimCore (under `.github/workflows/format-check.yml` in either of those directories). To avoid submodules, probably would want to update the script that either of those runs `.github/format`. I'd be happy with either writing the few lines of code into the action yaml or putting a `format` script into our `scripts/` directory which we can call from CI.
- [x] Remove the formatting infra from Framework/SimCore. `git rm -r Framework/.clang-format Framework/.github SimCore/.clang-format SimCore/.github` (just cleaning up)

tvami:
> I'd be happy with either writing the few lines of code into the action yaml 

I'm not sure how to skip the submodules. For now I did the rest in https://github.com/LDMX-Software/ldmx-sw/pull/1362/commits/7c988daa60c2862cfbd295ba583f3140ade27709

tvami:
oh I think it works out of the box: https://github.com/LDMX-Software/ldmx-sw/actions/runs/9421674449/job/25956257560
Screenshot 2024-06-07 at 11 43 58

tomeichlersmith:
For skipping submodules, it would mainly entail changing the `find ...` command to the `git ls-tree ...` command to avoid files in another repo. It looks like the files in the other repos either follow the Google style as well or have a different extension so we don't check them.

https://github.com/LDMX-Software/ldmx-sw/blob/7c988daa60c2862cfbd295ba583f3140ade27709/.github/format-check#L23

That is what I _meant_, but luckily there is another way! Just have the action not even clone the submodules :) then we are all good.

https://github.com/LDMX-Software/ldmx-sw/blob/7c988daa60c2862cfbd295ba583f3140ade27709/.github/workflows/format-check.yml#L12

tomeichlersmith (approved):

tvami:
> Just have the action not even clone the submodules :) 

ahh that's why it worked, because I didnt have the `recursive` option setup. Gotcha!

PR #1372, opened 2024-07-09
Make all truth tracking collections even with 0 SimHits

### LINKED ISSUE #1371: Truth seeding and tracking not making collections when there are no Tagger OR Recoil sim hits
**Describe the bug**
The truth seeding/tracking returns without making **_any_** collections when either the tagger or recoil sim hits collection has 0 hits.  Downstream processes that try to use Truth seeds or tracks will cause program to crash.  

This makes problems for MC that's generated at the target where only recoil tracker hits are of interest anyway.  Similarly, if  we only wanted to study tagger tracking, MC generated with just tagger and no recoil tracker would fail.  

The[ code that checks for sim hits is here ](https://github.com/LDMX-Software/ldmx-sw/blob/8b4650fea34453bd12ccfb7dc03a3645f8616877/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx).  

**Desired behavior**
We'd like for the collections to be made no matter what, even if empty.  This code actually runs just fine as it is now and makes the (empty) collections if there are no Sim hits if we just remove that return statement.  However, it also spits out an error message every time it sees an empty Tagger/Recoil sim hits collection...which is a lot of messages if the e.g. Tagger wasn't even simulated and you don't care about that.  

So in order to accommodate this,  I've added a couple of bool switches to TruthSeedProcessor,  "skip_tagger" & "skip_recoil", which tell the code to skip over the appropriate code blocks (and warning messages) for the truth tracking of that tracker.  The code no longer returns if a SimHits collection is empty, but just prints an error message as long as "skip_XXXX" is false, which is the default.

bloodyyugo:
I've modified TruthSeedProcessor so that it no longer skips MC events with 0 SimHits in either recoil or tracker collections (issue 1371).  This caused a crash downstream when one of the truth collections was (attempted) to be used.  

In the update, an error-level message will be printed when it encounters an empty SimHits collection but (empty) truth seed and track collections will be made.  

The error messages can be turned off (for instance if the tagger wasn't simulated) by setting the new config variable skip_XXXX to true, for example:   
`truth_tracking.skip_tagger       = True`
a 'skip_recoil' variable exists as well.  Note that these also turn off the relevant truth-track-making code as well.  These config variables are set to 'False' by default. 

I've also changed some log entries from 'info' to 'debug' level.

### DIFF
--- a/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
+++ b/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
@@ -60,6 +60,9 @@ void TruthSeedProcessor::configure(framework::config::Parameters& parameters) {
   // In tracking frame
   beamOrigin_ = parameters.getParameter<std::vector<double>>(
       "beamOrigin", {-880.1, -44., 0.});
+  skip_tagger_=parameters.getParameter<bool>("skip_tagger", false);
+  skip_recoil_=parameters.getParameter<bool>("skip_recoil", false);
+
 }
 
 void TruthSeedProcessor::createTruthTrack(
@@ -318,7 +321,7 @@ ldmx::Track TruthSeedProcessor::seedFromTruth(const ldmx::Track& tt,
   Acts::BoundVector stddev;
 
   if (seed_smearing) {
-    ldmx_log(info) << "Smear track and inflate covariance" << std::endl;
+    ldmx_log(debug) << "Smear track and inflate covariance" << std::endl;
 
     /*
       double sigma_d0     = rel_smearfactors_[Acts::eBoundLoc0]   * tt.getD0();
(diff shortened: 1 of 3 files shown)


tomeichlersmith (changes requested):
- The clang-format check is failing, please format the source files you've changed.[^1] You can view the log showing why the format checked failed [on GitHub](https://github.com/LDMX-Software/ldmx-sw/actions/runs/9862720152/job/27233935839?pr=1372) for the next 30ish days before GitHub deletes it.
- Please put the new parameters into the [associated python config class](https://github.com/LDMX-Software/ldmx-sw/blob/8b4650fea34453bd12ccfb7dc03a3645f8616877/Tracking/python/tracking.py#L241-L297). While this isn't necessary (as you've discovered), it is helpful for discoverability of the available run-time parameters.

[^1]: This program is available within the container so you can run `ldmx clang-format -i Tracking/include/Tracking/Reco/TruthSeedProcessor.h Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx` in order to do the formatting. Please let me know if this doesn't work so I don't keep telling people the wrong thing.

bloodyyugo:
Ok, I've fixed the formatting and added the variables to tracking.py.  It looks like tracking.py needs some other updating, but I'll add that to my list.

tomeichlersmith:
Yea, it looks like you ran the clang-format over the python file. This generally mangles the results unfortunately.

bloodyyugo:
> Yea, it looks like you ran the clang-format over the python file. This generally mangles the results unfortunately.

Damn it.   I fixed this up

sophiemiddleton (approved):
this looks good, thanks for looking at this

Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
```diff
@@ -57,9 +57,13 @@ void TruthSeedProcessor::configure(framework::config::Parameters& parameters) {
   inflate_factors_ = parameters.getParameter<std::vector<double>>(
       "inflate_factors", {10., 10., 10., 10., 10., 10.});
 
-  // In tracking frame
+  // In tracking frame:  MG where do these numbers come from?
```
tomeichlersmith: These numbers come from an approximation of where the beam generator places the beam (without smearing) just upstream of all detector components. `-880.1` is just upstream of TriggerPad1 in `z` and `-44` is approximately the bend in `x` imposed on the beam by the magnet (I think 4GeV but could be 8GeV). Then they are re-ordered to be in tracking coordinates.

tomeichlersmith (approved):
Looking good, thanks for cleaning things up!

PR #1374, opened 2024-07-09
Added multi-electron tagger truth tracking capability

### LINKED ISSUE #1373: Tagger truth seeding/tracking not producing multiple tracks for multi-electron events
**Describe the bug**
Truth seeding/tracking for the tagger returns only a single track, even for events with multiple primary electrons. The current implementation in [TruthSeedProcessor](https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx) is constrained to only look for a single beam electron, meaning that additional tracks are disregarded. 

**Desired behavior**
We'd like the tagger truth information to accomodate multi-electron events, in the same manner as the recoil truth information.

EBerzin:
I've modified TruthSeedProcessor so that truth seeding/tracking for the tagger is able to handle multi-electron events. This resolves #1373. Tagger truth tracking now operates the same way as recoil truth tracking, and was tested on simulated samples up to four electrons. Similar to recoil truth tracking, the tagger truth hits are selected according to a momentum cut in the appropriate target scoring plane. 

Tracks corresponding specifically to beam electron hits are stored in the `beamElectrons` collection. This selection is done according to a new parameter called `max_track_id`, which defines the maximum track ID corresponding to a primary electron, as well as a pdgID selection. 

## Check List

This plots shows the number of tracks identified in the TaggerTruthTracks collections, for a 2e- sample. We see that we are now able to recover 2e- events. 
taggerTruth_Ntracks

### DIFF
--- a/Tracking/python/tracking.py
+++ b/Tracking/python/tracking.py
@@ -278,11 +278,11 @@ class TruthSeedProcessor(Producer):
     p_cut_ecal : double 
         Minimum seed track momentum(MeV) at the ECAL scoring plane 
     skip_tagger : bool 
-    max_track_id : double
-        Maximum track ID for a hit to be selected in the target scoring plane.
         Ignore the tagger tracker(makes empty collections). 
     skip_recoil : bool 
-        Ignore the recoil tracker(makes empty collections)
+        Ignore the recoil tracker(makes empty collections).
+    max_track_id : double
+        Maximum track ID for a hit to be selected in the target scoring plane.
     """
     def __init__(self, instance_name = "TruthSeedProcessor"):
         super().__init__(instance_name, 'tracking::reco::TruthSeedProcessor',
--- a/Tracking/python/tracking.py
+++ b/Tracking/python/tracking.py
@@ -278,6 +278,8 @@ class TruthSeedProcessor(Producer):
     p_cut_ecal : double 
         Minimum seed track momentum(MeV) at the ECAL scoring plane 
     skip_tagger : bool 
+    max_track_id : double
+        Maximum track ID for a hit to be selected in the target scoring plane.
         Ignore the tagger tracker(makes empty collections). 
     skip_recoil : bool 
         Ignore the recoil tracker(makes empty collections)
@@ -299,3 +301,4 @@ class TruthSeedProcessor(Producer):
         self.p_cut_ecal = -1. #MeV
         self.skip_tagger = False
         self.skip_recoil = False
+        self.max_track_id = 5
--- a/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
+++ b/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
@@ -574,10 +574,10 @@ void TruthSeedProcessor::produce(framework::Event& event) {
     if (zhit < 0.) {
       // Tagger selection cuts
       // Negative scoring plane hit, with momentum > p_cut
-      if (p_vec(2) < 0. || p_vec.norm() < p_cut_)
-      {
-        continue;
-      }
+      if (p_vec(2) < 0. || p_vec.norm() < p_cut_) continue;
+      
+      // Check that the hit was left by a charged particle
+      if (abs(particleMap[hit.getTrackID()].getCharge()) < 1e-8) continue;
 
       if (p_vec.norm() > tagger_p_max) {
         tagger_sh_count_map[hit.getTrackID()].push_back(i_sh);
--- a/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
+++ b/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
@@ -615,7 +615,7 @@ void TruthSeedProcessor::produce(framework::Event& event) {
   }
 
   // Sort tagger hits.
-  for (auto& [ _track_id, hit_indices ] : tagger_sh_count_map) {
+  for (auto& [_track_id, hit_indices] : tagger_sh_count_map) {
     std::sort(
         hit_indices.begin(), hit_indices.end(),
         [&](const int idx1, int idx2) -> bool {
@@ -654,29 +654,27 @@ void TruthSeedProcessor::produce(framework::Event& event) {
   auto beamOriginSurface{Acts::Surface::makeShared<Acts::PerigeeSurface>(
       Acts::Vector3(beamOrigin_[0], beamOrigin_[1], beamOrigin_[2]))};
 
-    if (!skip_tagger_) {
-  for (const auto& [track_id, hit_indices] : tagger_sh_count_map) {
-    const ldmx::SimTrackerHit& hit = scoring_hits.at(hit_indices.at(0));
-    const ldmx::SimParticle& phit = particleMap[hit.getTrackID()];
+  if (!skip_tagger_) {
+    for (const auto& [track_id, hit_indices] : tagger_sh_count_map) {
+      const ldmx::SimTrackerHit& hit = scoring_hits.at(hit_indices.at(0));
+      const ldmx::SimParticle& phit = particleMap[hit.getTrackID()];
 
-    if (hit_count_map_tagger[hit.getTrackID()].size() > n_min_hits_tagger_) {
+      if (hit_count_map_tagger[hit.getTrackID()].size() > n_min_hits_tagger_) {
+        ldmx::Track truth_tagger_track;
+        createTruthTrack(phit, hit, truth_tagger_track, targetSurface);
+        truth_tagger_track.setNhits(
+            hit_count_map_tagger[hit.getTrackID()].size());
+        tagger_truth_tracks.push_back(truth_tagger_track);
 
-      ldmx::Track truth_tagger_track;
-      createTruthTrack(phit, hit, truth_tagger_track, targetSurface);
-      truth_tagger_track.setNhits(
-          hit_count_map_tagger[hit.getTrackID()].size());
-      tagger_truth_tracks.push_back(truth_tagger_track);
-
-
-      if (hit.getPdgID() == 11 && hit.getTrackID() < max_track_id_) {
-        ldmx::Track beamETruthSeed = TaggerFullSeed(
-          particleMap[hit.getTrackID()], hit.getTrackID(), hit,
-          hit_count_map_tagger, beamOriginSurface, targetUnboundSurface);
-        beam_electrons.push_back(beamETruthSeed);
+        if (hit.getPdgID() == 11 && hit.getTrackID() < max_track_id_) {
+          ldmx::Track beamETruthSeed = TaggerFullSeed(
+              particleMap[hit.getTrackID()], hit.getTrackID(), hit,
+              hit_count_map_tagger, beamOriginSurface, targetUnboundSurface);
+          beam_electrons.push_back(beamETruthSeed);
+        }
       }
     }
   }
-    }
 
   // Recover the EcalScoring hits
   std::vector<ldmx::SimTrackerHit> ecal_spHits =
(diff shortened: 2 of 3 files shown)


bryngemark:
Hi @EBerzin, this might be a good time to reraise the point that when we create multielectron samples with overlay, the scoring plane collections aren’t included (and contain no pileup). Using the multiparticle gun, all information is there in the scoring planes too. 

Would it be hard to use simhits for truth seeding instead? In that case we should consider expanding the overlay procedure to scoring planes too.

pbutti:
@bryngemark 
The truth tracks / seeds already use the tagger sim hits. The recoil use the target scoring plane hits, but can be changed to use the sim hits. 
However, the scoring plane hits are used to store the truth states at target / ecal.
Those can be dropped but you won't be able directly to compare reconstruction to truth at those locations, unless one propagates the truth track to those surfaces by hand from the closest sim hit, which is possible using the track extrapolator tool

Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
```diff
@@ -40,6 +40,7 @@ void TruthSeedProcessor::configure(framework::config::Parameters& parameters) {
   recoil_sp_ = parameters.getParameter<double>("recoil_sp", true);
   target_sp_ = parameters.getParameter<double>("tagger_sp", true);
   seedSmearing_ = parameters.getParameter<bool>("seedSmearing", false);
+  max_track_id_ = parameters.getParameter<int>("seedSmearing", 5);
```
tomeichlersmith: Looks like there is a typo made here.

```suggestion
  max_track_id_ = parameters.getParameter<int>("max_track_id", 5);
```

Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
```diff
@@ -610,6 +608,23 @@ void TruthSeedProcessor::produce(framework::Event& event) {
         });
   }
 
+  // Sort tagger hits. TO DO: simplify/merge with above
```
tomeichlersmith: I'm not sure if we can merge this with the above. Generally, insertion needs to happen before sorting unless you are willing to change the target data structure (the `std::vector<int>` in this case) to be one that is sorted (like a `std::set`) but since you want the sort to be done by momentum and not index, this could get confusing.

Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
```diff
@@ -610,6 +608,23 @@ void TruthSeedProcessor::produce(framework::Event& event) {
...
+  for (std::pair<int, std::vector<int>> element : tagger_sh_count_map) {
+    std::sort(
+        element.second.begin(), element.second.end(),
```
tomeichlersmith: We can use [range-based for loops](https://en.cppreference.com/w/cpp/language/range-for) in order to make this a bit more readable. (I put an underscore in front of `track_id` to point out that it won't be used but we know what it is.)

```suggestion
  for (auto& [ _track_id, hit_indices ] : tagger_sh_count_map) {
    std::sort(
        hit_indices.begin(), hit_indices.end(),
```

Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
```diff
@@ -633,34 +648,24 @@ void TruthSeedProcessor::produce(framework::Event& event) {
   auto beamOriginSurface{Acts::Surface::makeShared<Acts::PerigeeSurface>(
       Acts::Vector3(beamOrigin_[0], beamOrigin_[1], beamOrigin_[2]))};
 
-  // I found a tagger scoring plane hit
-  if (idx_taggerhit != -1) {
-    const ldmx::SimTrackerHit& hit = scoring_hits.at(idx_taggerhit);
+  // For tagger scoring plane hits
+  for (std::pair<int, std::vector<int>> element : tagger_sh_count_map) {
+    const ldmx::SimTrackerHit& hit = scoring_hits.at(element.second.at(0));
```
tomeichlersmith: Same comment as before

```suggestion
  for (const auto& [track_id, hit_indices] : tagger_sh_count_map) {
    const ldmx::SimTrackerHit& hit = scoring_hits.at(hit_indices.at(0));
```

tomeichlersmith (changes requested):
https://github.com/LDMX-Software/ldmx-sw/pull/1372 might slighty conflict with these changes since you both are editing the same files; however, since the edits are focused on different goals, the conflicts (as reported by `git`) can be resolved by accepting both set of changes.

---

Besides this, I only found a few technical changes I'd request for readabilities sake and one typo which may just be uncommitted (or not have a large effect).

EBerzin:
If I’m understanding the current implementation correctly, it creates truth tracks for the tagger only for SimParticles that leave hits in the target scoring plane (and similarly for the recoil truth tracking). The z-positions of the target scoring planes are also used to distinguish between tagger vs. recoil hits. I think this should be doable with target sim hits or hits in the nearest tracker modules.

pbutti:
@EBerzin 
The usage of scoring plane hits is there to have the possibility to check tracking performance at the target, for example.
One could even use directly the SimParticle container to make truth seeds. The issue is how / where you check the resolution with reconstructed tracks. In colliders is common practice to check resolution at the interaction point, with coincides with the generation location of primary tracks. For LDMX the situation is different as we are interested in checking the matching between the tagger and recoil at the target, while simulated particles are generated upstream before the TS. 

The usage of scoring plane hits is there to ensure that reconstructed tracks can then be checked against truth in points of interest, such as Target or face of the Ecal. 
You can always drop this requirement, but you should make sure that when extrapolations are compared to Target or ECAL the right truth value is used.

bryngemark:
so what I hear is that we could use simhits at the target and the first layer of Ecal as the truth reference? 

yeah I don't know, @tomeichlersmith, what do you think, will it be cleaner to implement overlay of scoring planes (and simparticles? ... that will take up a lot of space... ) to allow for all kinds of performance studies? instead of forcing all producers to stop using scoring planes. maybe overlay of scoring planes and simparticles could be optional. and for large sample production, I would anyway really like to know what the baseline reconstruction I should run is, so I can drop most truth info afterwards; if we double down on that approach then it doesn't matter too much if the intermediate collections are large.

tomeichlersmith:
I think using the Target and ECal Sim Hits as the first-pass performance check is the way to go (however, _matching_ these truth sim hits with truth tracks in an overlay sample would be difficult due to clashes in track ID). Overlaying Scoring Planes would be very heavy (as you mention) and overlaying Sim Particles would be technically challenging since there _will_ be clashes in track ID that we would need to figure out how to deal with.

Overlaying scoring planes is already feasible since they are the same type as the sim tracker hits (just add the scoring planes you want to overlay to the `overlayTrackerHitCollections` list and it should :tm: work) so folks could still do that if they wanted to. The issue then again becomes one of remembering that there will be track ID clashes and this will naturally be difficult to understand and disentangle during any analysis.

## :warning: Big Picture Ramblings
I think this discussion naturally reveals that we can't have our cake and eat it too. We can't skip fully simulating N > 1 electrons _and_ have their track reconstruction be easily/fully comparable to truth-level information.[^1] Therefore, I think the sample needs to be defined by the goals of the analysis.
- If you want to study the tracking of multiple electrons and compare the reconstruction to truth information, you should use a sample that is generated with multiple electrons at the Geant4 level. (We have [a generator](https://github.com/LDMX-Software/ldmx-sw/blob/97b6dbd1ea01ff9363459b01c25cecef97dc94df/SimCore/python/generators.py#L51-L81) for this.)
- If you want to study multiple-electron situations where only one of the electrons undergoes a specific interesting process, then you can overlay but analyzers _must_ know that truth positions of specific particles will be difficult (if not impossible) to find.

With these ramblings in mind, I think the best course of action is to keep using the scoring planes as the truth reference for the track state at the target and the ECal (so that folks using a fully-simulated multi-electron sample can do tracker performance studies) and then drop these truth-track-states if the scoring planes are not available (so that they don't provide erroneous / confusing information in the context of overlay samples).

[^1]: We _could_ attach an "event id" to _each_ sim hit we overlay in some fashion but that would require making the sim information even heavier and would only be useful in this specific scenario. Maybe its useful enough? I'm not sure.

bryngemark:
thanks @tomeichlersmith for your ramblings 😄 I find them very useful. let's leave the overlay producer as is, and assume that the relevant people have now been well informed that their producer might break. 

this does leave me in the conundrum, that I'm not sure I can usefully run tracking in e.g. central 2e ecal PN sample production, and then I have to leave simhits in for later tracking performance studies, etc, which I'm not all that keen on considering storage constraints. @EBerzin @tvami do we need to solve this issue for any of your studies?

EBerzin:
@bloodyyugo pointed out that in my original changes, the tagger truth parameters were defined relative to the beam generation surface, rather than the target. This is because I had been using `TaggerFullSeed()`, which assumes that everything handed to it is a beam electron, to directly create `TaggerTruthTracks`. I have now made a change such that `TaggerFullSeed()` is used only on tracks in the `beamElectrons` collection, while `createTruthTracks()` is used to define parameters at the target surface for the `TaggerTruthTracks` collection. The plot below shows the effect of this change—in blue are parameters from iss1373 before this change, and in purple after (now matching trunk, in green).

iss1373_val

@bryngemark , for the moment I have been using a smaller custom-generated set of samples including all sim hit information, and this should be sufficient for initial tracking performance studies and optimizing parameters. In general though I think it would be nice to be able to rerun tracking on central samples (maybe just for the recoil), which requires RecoilSimHits.

tomeichlersmith (commented):
Thank you for making my non-algorithmic changes!

There is still conflict with trunk since this PR makes changes to some of the same lines as https://github.com/LDMX-Software/ldmx-sw/pull/1372 -- I've pushed for using `git rebase` before merging in ldmx-sw so that the commit history is as linear as possible so please try to rebase onto `trunk` and update this branch. If we can't get this to work, that is okay, I can still manually merge this branch without the rebase.

### My Rebase Procedure
```
# make sure `trunk` is up to date so I don't have to do this again
git fetch
git switch trunk
git pull
git switch iss1373
```
To first order, `rebase` _re-applies_ the changes from commits onto the new "base" (Here, your current branch is "based" on an old commit of trunk and we are updating the base to be the latest commit on trunk.) Often times, there are commits that don't need to be re-applied (you tried them and then later undid them because they didn't work or whatever); thus, I start an `--interactive` rebase.
```
git rebase --interactive
```
Now a file is open listing the different commits that would be re-applied. The syntax for how to interact with these commits is at the bottom of this file. Most often, I just remove the lines with commits that I don't want to keep. Once you save and exit this file the rebase begins.

If there is a CONFLICT, make sure to resolve it. There are many tools available to more quickly resolve conflicts, but the simplest is just to open the file and look for the area with `<<<<<` denoting the beginning of the conflicting section. This corresponds to the two changes and - as mentioned in the previous comment - I suspect that we can keep both in which case, you would just remove the lines with `<<<<<`, `=====`, and `>>>>>`.

Once you've resolved the conflict, you `git add <file>` and `git rebase --continue`.

This cycle continues until all commits have been re-applied. You may have some commits that don't have any conflicts. These commits will just be automatically applied and the rebase continues unless you changed the command associated with that commit in the file opened earlier.

Now you have a rebased branch on your _local_ computer. Double check that things work as intended (e.g. at least compiles) and that the changes you want to make are included. I often do this by `git diff trunk` to make sure my changes (and only my changes) are different compared to the latest trunk.

Once you are certain that everything is good and rebased, you can push your branch `git push`. Rebasing _rewrites_ commits so you may need to `git push --force` to _overwrite_ the copy of your branch on GitHub.

Tracking/python/tracking.py
```diff
@@ -278,6 +278,8 @@ class TruthSeedProcessor(Producer):
     p_cut_ecal : double 
         Minimum seed track momentum(MeV) at the ECAL scoring plane 
     skip_tagger : bool 
+    max_track_id : double
+        Maximum track ID for a hit to be selected in the target scoring plane.
```
tomeichlersmith: It looks like these two documentation lines got put in between two other lines during the rebase, can you move them up/down?

tomeichlersmith (approved):
This looks great! I just have one more change related to the documentation and then I'm happy to merge :)

Thank you for spending the time to rebase :1st_place_medal: I really appreciate it.

bloodyyugo (approved):
This looks like it's working and doesn't change other stuff.  I approve.

PR #1380, opened 2024-07-31
just adopt a task runner

### LINKED ISSUE #1248: add `apptainer` explicitly to the `ldmx-env.sh` script

### LINKED ISSUE #1377: just adopt a task runner
After I [started looking into adopting `denv` within ldmx-sw](https://github.com/LDMX-Software/ldmx-sw/pull/1269) I realized that I was just writing custom shell scripts to run various common commands that ldmx-sw developers run. In the software biz, this is called "task running" and there are a plethora of solutions already available. After some browsing, I can across the task runner [`just`](https://just.systems/) which I find to be beneficial for a few reasons:

- Very simple syntax associating a task name and the command(s) that the task runs
- All tasks located in a single file which could be reference if the command itself is not available
- Natural and easy self-documentation of tasks via comments in the file defining them
- (meme) Written in :fire: blazing fast :rocket: Rust (but in all seriousness, this makes it incredibly cross platform and easy to install)
- I like the name and the mechanic of `just <task>` e.g. `just build`

On the branch associated with this issue, I've included a justfile that I haven't thoroughly tested but shows what I'm imagining could be used. The terminal output of running `just` without any arguments (and the default task being `just --list`) gives the help message.
```
tom@appa:~/code/ldmx/ldmx-sw$ just
just --list
Available recipes:
    build             # compile and install ldmx-sw
    check             # check that the necessary programs for running ldmx-sw are present
    configure *CONF   # configure how ldmx-sw will be built
    fire config *ARGS # run ldmx-sw with the input configuration script
    init              # initialize a containerized development environment
    mount *DIR        # mount a directory into the denv
    pull *IMAGE       # make sure the image is pulled down
    setenv *ENVVAR    # pass an environment variable into the denv
    use *IMAGE        # change which image is used for the denv
```
I also want to include an image of this because it is :sparkles: pretty :sparkles: 

image

In the [SW Dev Meeting where I first brought up `just`](https://indico.fnal.gov/event/64896/), the request was made to include the possibility of still have `ldmx` be the root command. I also like this and so the solution I'm imagining is to simply have `ldmx` be an alias for `just`.

```
tom@appa:~/code/ldmx/ldmx-sw$ alias ldmx=just
tom@appa:~/code/ldmx/ldmx-sw$ ldmx
just --list
Available recipes:
    build             # compile and install ldmx-sw
    check             # check that the necessary programs for running ldmx-sw are present
    configure *CONF   # configure how ldmx-sw will be built
    fire config *ARGS # run ldmx-sw with the input configuration script
    init              # initialize a containerized development environment
    mount *DIR        # mount a directory into the denv
    pull *IMAGE       # make sure the image is pulled down
    setenv *ENVVAR    # pass an environment variable into the denv
    use *IMAGE        # change which image is used for the denv
```

## Related Discussions for Background
- https://github.com/LDMX-Software/ldmx-sw/issues/1232
- https://github.com/LDMX-Software/ldmx-sw/pull/1269

tomeichlersmith:
I am updating _ldmx-sw_ by including a justfile and rewriting the README to point folks to `denv` and `just`. The ldmx-env.sh script is left in place and is only changed to include a deprecation notice if a user sources it and does not have both `denv` and `just` installed. (If both are available, then `ldmx` is just an alias for `just` and the justfile recipes are used instead of the ldmx-env.sh bash functions.)

The best way to show why I chose this workflow is probably to show some examples assuming `denv` and `just` are installed.

**New Clone**
```
git clone --recursive git@github.com:LDMX-Software/ldmx-sw.git
cd ldmx-sw
just init
```

**Default Build**
```
just configure build
```

**From somewhere else**
```
just path/to/ldmx-sw/build test
```

**Pretty Output**
image

This resolves #1377 directly and resolves #1248 indirectly (via `denv`).

## Check List
- [x] Add `rootbrowse` recipe and `root`
- [x] Update doc website https://github.com/LDMX-Software/ldmx-software.github.io/pull/35
  - As the re-written README points out, I think using `denv` and the production images allows many folks who are just using ldmx-sw to do studies to avoid compiling ldmx-sw themselves. I think this is a nice workflow since it both pins the version of ldmx-sw of the study _and_ allows users to jump right in to a physics study without needing to run the compilation. This is a workflow change though and so I'd want to update the doc website to reflect this suggestion.
- [x] Stabilize `denv` and create LDMX-Software fork
  - Before wide adoption, I'd like to make sure `denv` has GUI support on MacOS (https://github.com/tomeichlersmith/denv/issues/80 -> https://github.com/tomeichlersmith/denv/pull/120), create a major stable version (v1), and create an LDMX-Software fork of `denv` so that it can hang around with the collaboration separate from my personal account.

After @tvami 's first review...

- [x] Have an `ldmx` symlink to `denv` to allow folks used to the `ldmx` program to continue using it after transition
- [x] Flesh out README (and doc website) with example snippets of installing and enabling tab complete
- [x] Connect tab complete of `just` to `ldmx` alias within ldmx-env.sh
- [x] Undo chmod on TrigScint script

## :stop_sign: Won't Do :stop_sign: 
- #1252 could be done within these recipes but I think it is too complicated to try to do on this first pass
- implement a `from-ldmx-env` recipe. Thought about it, but `denv` pretty much requires the ldmx/dev image to be one of the newest available and other updates to ldmx-sw deprecate non-v4 images, so encouraging updating is a good way to go.
- put in python formatting right now. This is a bigger discussion that we can add later.
- fully test all command orderings. Further bugs can be patched as we encounter them, going to focus on making sure _new_ developers can understand how to get started.


README.md
```diff
@@ -16,50 +16,61 @@
...
+In order to use ldmx-sw, no more dependencies are required!
+Simply choose the version of ldmx-sw you wish to use with your project.
+```
+mkdir my-project
+cd my-project
+denv init ldmx/pro:v4.0.1 # or some other ldmx-sw version
+```
+And then you can run ldmx-sw with a configuration script of your choice.
+```
+denv fire my-config.py
```
tvami, line 61: it would be nice if this also could become `ldmx fire`

README.md
```diff
@@ -16,50 +16,61 @@
...
-- `ldmx clean all` : Reset environment to a blank state
+> [!NOTE]
+> You need to [setup an SSH-key with your GitHub account](https://docs.github.com/en/authentication/connecting-to-github-with-ssh) on the computer you are using.
 
-Use `ldmx help` for a full listing of these commands.
-If we don't define a command outside of the container,
-then the command is given to the container to run inside the current directory.
+```
+cd ldmx-sw
+just # no arguments prints out the possible options
```
tvami, line 111: uhm so no sourcing here? I just cloned, moved to this branch, and then did `just` (on a linux), I get `-bash: just: command not found`

README.md
tvami: Since with `just`, I understand we will use `ldmx fire` thru an alias, so if this stays like this, that means running fire will be with two different options depending if we are using the image or the cloned repo, right?

tvami:
```
TrigScint/exampleConfigs/runTestBeamHitReco.py
```
this only changed the `chmod`, I think you can use `git restore TrigScint/exampleConfigs/runTestBeamHitReco.py` to remove that file from being modified and thus in this PR

README.md
tomeichlersmith: No sourcing. The point of moving to `denv`+`just` is to avoid the `bash` requirement and sourcing the environment script everytime. Once `denv` and `just` are installed, developing ldmx-sw can be done via those tools without additional environment scripts.

The `ldmx-env.sh` script is still there which provides `alias ldmx=just` so that developers who are simply updating their clone and not reading the README can still run `ldmx ....` commands, but I try to advise new folks to use `just` directly so there is less redirection.

README.md
tomeichlersmith: The only way for a person using ldmx-sw without cloning it to get an `ldmx` alias for `denv` (or an `ldmx` wrapper for `denv`) would be to write this alias/wrapper into their terminal profile. This felt too advanced for new folks using ldmx-sw and so I omitted it and instead pointed them towards using `denv` directly.

tvami: OK this might be too much to ask :D but if I understand you plan to have a fork of `denv` under ldmx-sw, can we have it with the name `ldmx` instead? Like that the transition is very smooth, people have been calling `ldmx` in the past to get the environment, and they'd do the same from now on too, it's just the under the hood that's changed.

tomeichlersmith: It's not too much to ask, its just a stylistic choice. The main reasons I'm resistant to propagating an alias/wrapper/symlink of `denv` to `ldmx` are

1. While it maintains some backwards compatibility, there is little self-discovery. i.e. Users of `ldmx` (which simply redirects to `denv`) would not be directed towards denv's documentation and therefore potentially miss out on helpful information.
2. The implication is restrictive. `ldmx` implies that it is special to LDMX and ldmx-sw but `denv` does not. It (correctly in my opinion) implies that other projects can use the same software interaction workflow in order to facilitate easier development. I particularly am fond of this reason - I use `denv` to enable using identical and modern Python versions on the UMN cluster, SLAC cluster, and my personal computers which has made it so much easier to share my analysis and plotting code. This is something that I would forsee being useful for everyone - containers can be helpful for portability and reproducibility not just in ldmx-sw but in all projects that write any kind of code (even LaTeX lmao).

These reasons being stated, I could imagine the LDMX-Software fork of `denv` including a symlink to `denv` named `ldmx` within the `install` script so that users could do `ldmx ...` in place of `denv ...` if they desire to out of habit.

tvami:
Another feature I'm missing is the autocomplete, with earlier `ldmx comp<tab>` autocomplete to `ldmx compile`, now that's gone, can we add it?

README.md
tvami: Yes, I understand `denv` is bigger than `ldmx`, I think supporting both would be nice, indeed at this point it's more about my habits than anything else

tvami: That being said I'm more likely to use the `just` approach anyway since I seem to be forever developing something. Given that we have an alias there for `ldmx` this might lead to a conflict? So maybe the symlink should be `ldmxenv` for `denv` and we keep the alias of `ldmx` for `just`?

README.md
```diff
@@ -16,50 +16,61 @@
...
-- Setup the environment (in bash): `source ldmx-sw/scripts/ldmx-env.sh`
-  - **Note**: If you are working with ldmx-sw at SLAC's SDF, you will need to set the `TMPDIR` environment variable so that program running the container has more than ~5GB of space to write intermediate files. The default temporary space (`/tmp`) is often full of other files already. A decent replacement is `TMPDIR=${SCRATCH}` which gives the program plenty of room for the files it needs to manipulate.
-- Configure and compile: `ldmx compile`
-- Now you can run any processor in _ldmx-sw_ through `ldmx fire myconfig.py`
-- If you are developing and need to recompile and run `ldmx fire`, you can use `ldmx recompFire myconfig.py`
+  - Only necessary on personal computers. Shared computing clusters should have `apptainer` installed.
+  - (on Linux personal computers) [Manage docker as non-root user](https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user)
+- [Install `denv`](https://tomeichlersmith.github.io/denv/getting_started.html#installation)
```
tvami, line 28: So I think having direct instructions here, and then adding a link to the details (like what you have on the OSes) would be better, just so it's one less click to have everything installed. And then this issue with the PATH for Macs should be either mentioned here or in the link

README.md
```diff
@@ -16,50 +16,61 @@
...
+> [!WARNING]
+> If you are on Windows, make sure to install these tools _inside_ WSL where `docker`
+> will be run and ldmx-sw will be developed. Since WSL is often a virtual Ubuntu machine,
+> following the instructions for Ubuntu or Linux can be appropriate.
 
-### Other Container Configuration Commands
+- Make sure `git` is installed.
+  - `git` is a very common tool used by software developers and so it may already be available.
+  - (on MacOS systems) Make sure `git lfs` is installed. (Test: `git lfs` prints out a help message instead of an error about `lfs` not being found.) The default installation of `git` that is included with Apple's developer tools does not include `git lfs` which is required by acts to download and unpack one of its own submodules. [GitHub has a nice tutorial](https://docs.github.com/en/repositories/working-with-files/managing-large-files/installing-git-large-file-storage?platform=mac) on how to install `git lfs` on MacOS.
+- [Install `just`](https://just.systems/man/en/chapter_5.html)
```
tvami, line 78: Again having some specific example here would be needed

README.md
tvami, line 61: As we discussed above, I'd like to have a simlink when we have a fork with the name of `ldmxenv`

TrigScint/exampleConfigs/runTestBeamHitReco.py
tvami, line 1: Restore the original

README.md
```diff
@@ -16,50 +16,61 @@
...
+> You need to [setup an SSH-key with your GitHub account](https://docs.github.com/en/authentication/connecting-to-github-with-ssh) on the computer you are using.
 
-Use `ldmx help` for a full listing of these commands.
-If we don't define a command outside of the container,
-then the command is given to the container to run inside the current directory.
+```
+cd ldmx-sw
+just # no arguments prints out the possible options
+just init # initialize a new development environment (once per clone)
+just configure build test # configure ldmx-sw, build it, then test it
```
tvami, line 113: I'll add it here for a lack of better point to comment on, I'd like to have autocomplete back.

justfile
```diff
@@ -0,0 +1,176 @@
...
+# we could think about removing them if folks are happy with committing to the
+# just-style commands above
+
+# open the ROOT shell within the software environment
+root *ARGS="":
+    denv root {{ ARGS }}
+
+# open a ROOT file with a graphical browser
+rootbrowse FILE:
+    denv rootbrowse {{ FILE }}
```
tvami, line 154: So I did
```
xhost + 127.0.0.1
denv rootbrowse
```
on my Mac and I get 
```
 can't open display ":0"
```
I thought we fixed this, I'm confused.

tvami (changes requested):
I tested most of the features, both on Linux and Mac. My most important ask is the autocomplete, the rest I could imagine life w/o....

tomeichlersmith:
You can enable `just`'s autocomplete https://just.systems/man/en/chapter_70.html

and then you need to attach these autocomplete functions to the `ldmx` alias as well https://just.systems/man/en/chapter_69.html?highlight=.j#shell-alias

I can put the second one into the `ldmx-env.sh` but, similar to the installation of `just` itself, tab-completion depends on the shell being used and where the user would like to install things. (What I do on my clusters which I could put into the README as an example as well is
```
just --completions bash >> ~/.local/share/bash-completion/completions/just
```
so that they are auto-loaded by `bash` when needed.)

justfile
tomeichlersmith: I believe you need to either do `xhost +` (no hostname) or `xhost + localhost` (since localhost isn't always 127.0.0.1). I also thought this was fixed and am uncertain about what could be changing in your environment.

README.md
tomeichlersmith: Reasonable feature request, it is close to available and I can include links and information in the README. `just` tab complete is handled at install time (either via the package manager or by the user manually). Tab complete of a `just` alias requires me to update the `ldmx-env.sh` script.

https://github.com/LDMX-Software/ldmx-sw/pull/1380#issuecomment-2322164988

tvami:
> What I do on my clusters which I could put into the README as an example as well is

OK this works for me, I'm happy now!

README.md
tomeichlersmith: There would not be a conflict in most[^1] shells. The alias would overwrite the symlink (which acts as a program). The symlink would be accessible with `command ldmx` (`command` avoids shell functions/aliases).

[^1]: I believe its a POSIX sh requirement that aliases/functions can have the same names as programs.

justfile
tvami: If doesnt work with `xhost +` either. But ok, I think we can come back to this in another PR

tvami (approved):
Major changes requested are done (TS config file is still there, but I guess that's ok)

Edit: All good now! Thanks Tom!

PR #1383, opened 2024-08-09
Introduction of SegMIP BDT + other improvement on ECAL veto

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1287
Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1305

We have shown the results of the software changes already at a SWAN meeting. The latest BDT that's now trained on the full E14 trigger skimmed data is included now too. This is to be shown at a SWAN meeting. 

Future improvements in another PR will include factorizing the MIP tracking out of this class. 

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments
- [x] I ran my developments and the following shows that they are successful.

PR #1385, opened 2024-08-12
Move validation configs to 8 GeV, intro reduced LDMX validation, add tracking to ecal pn 

### LINKED ISSUE #1384: Move validation config to 8 GeV + add tracking + intro reduced LDMX config to validation

Right now the ecal pn config's settings are incompatible with the ecal veto's setting. Instead of fixing the 4 GeV setting let's move the config to 8 GeV instead.

- move ecal pn to 8 GeV
- given that the tracking is in reasonable shape, I'd like to have it included in the validation suit
- add a new validation config for the reduced geometry (which is still at 4 GeV)

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1384

- Moved ecal pn to 8 GeV
- Given that the tracking is in reasonable shape, I included in the validation suit
- Added a new validation config for the reduced geometry (which is still at 4 GeV)

## Check List

This needs https://github.com/LDMX-Software/ldmx-sw/pull/1383/ to be merged first. After that's done, I'll rebase


tvami:
@tomeichlersmith is there anything I can do to generate gold for the reduced ldmx?

> This needs https://github.com/LDMX-Software/ldmx-sw/pull/1383 to be merged first. After that's done, I'll rebase

I realized this can be merged since the gold doesnt exists.

tomeichlersmith:
The current design (which I'm not commited to but it is functional) is to have files follow specific naming conventions and then the CI detects what validation samples exist based off those assumptions.

https://github.com/LDMX-Software/ldmx-sw/tree/trunk/.github/actions/validate#assumptions

Both the gold-generation and the pr-validation use this action to run the samples so that they are run in the same way and they get which samples to include in generation/validation by simply using all directories within the `.github/validation_samples` directory.

https://github.com/LDMX-Software/ldmx-sw/tree/trunk/.github/actions/generate-matrix

So, to add a new sample to be included in validation, the general procedure would be

1. Develop a config script that creates the sample you want _and fills the histograms you want_
2. Put this script in `.github/validation_samples/my-sample/config.py`
3. Update `config.py` to take the run number and number of events from the environment and have the output histogram file be named `hist.root`. ([for example](https://github.com/LDMX-Software/ldmx-sw/blob/613a9d10b87235eb4ad3a6d3fc8229752692ad98/.github/validation_samples/inclusive/config.py#L18-L22))
4. Run `config.py` [locally](https://github.com/LDMX-Software/ldmx-sw/tree/trunk/.github/workflows#local-equivalence) to create a first copy of `hist.root` and `gen.log`.
5. Rename the created ROOT and log files to be `gold.root` and `gold.log`
6. Commit those gold files

This process is arduous but I couldn't think of anything better at the time. We can probably improve this but I'm not sure how many more validation samples we will want to include. I believe ee will quickly approach the computing limit of our free-tier GitHub workflows which is undesirable.

tvami:
Thanks Tom! I only miss steps 4-6 then, I'll commit those soon

tvami:
>  I'll commit those soon

done now! also about the new version, maybe let's do it after I'm done with iss1387 (which I could do tomorrow [Tue])

tomeichlersmith (approved):
Looks good to me :)

I'm happy to add more tests and let them tell us where things are wonky.

tvami:
@tomeichlersmith when I run the CI here
https://github.com/LDMX-Software/ldmx-sw/actions/runs/10372637362/job/28715827843#step:4:9
it says
```
error: pathspec '.github/validation_samples/reduced_ldmx/gold.root' did not match any file(s) known to git
```
I'm a little bit confused since that file is added

tomeichlersmith:
Ah yes I forgot, the golden files are retrieved _from a checkout of trunk_, so attempting to get a sample that is not on trunk yet will fail.

A few paths forward
- Leave the CI as is. We merge without the tests passing because this acknowledges we are introducing new tests via a new sample.
- Update CI to skip samples that weren't found. I think this is preferable, we could issue a warning in the CI as well, but this development would not be simple. (Both the step that failed using `git checkout` and the `generate-matrix` action would need to be updated to allow for a mismatch between the trunk and dev-branch sample sets.) Again, similar to what I've said elsewhere, not sure if this development is worthwhile since I'm not sure how often we will be adding/subtracting entire samples from validation.

_Edit_: The `generate-matrix` action is the one that would need to be aware of trunk/dev sample set separation. The `validate` action is told which sample to run and so it will work fine as long as we get the right list of samples.

tvami:
OK makes sense thanks!

danyi211 (approved):

PR #1389, opened 2024-08-13
Introduce EcalPreselectionSkimmer

### LINKED ISSUE #1387: Add a producer that can be used to make an ECAL preselection skim

I'd like to have a dataset that's trigger skimmed + has a minimal preselection but is before the BDT. 
The use case I have in mind is to run tracking on this preselection and modify the ecal veto to include the tracking information (instead of scoring plane stuff)

Make a new producer that does the preselection. I propose the location in 
https://github.com/LDMX-Software/ldmx-sw/tree/trunk/Recon/src/Recon/Skims

One could run tracking on the whole trigger skimmed sample. That probably has some advantages, too. 

**Additional context**

Some notes to self:
The trick is 
```
  if (passedPreselection) {
    setStorageHint(framework::hint_shouldDrop);
  } else {
    setStorageHint(framework::hint_shouldKeep);
  }
```
and in the config to have
```
p.skimDefaultIsDrop()
p.skimConsider("ecal_preselection")
```

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1387

## Check List

I used the validation ecal pn config to produce 10k events. Then I run  `Recon/test/ecal_preselection_skim.py` on it. The current settings lead that 10k to 53 events, and looking at the distributions indeed they are cut off at the values set in the config. I made everything configurable and kept the defaults so high that it doesnt do anything in default.

### DIFF
--- a/Recon/include/Recon/Skims/EcalPreselectionSkimmer.h
+++ b/Recon/include/Recon/Skims/EcalPreselectionSkimmer.h
@@ -12,27 +12,26 @@
 //----------//
 #include "Ecal/Event/EcalVetoResult.h"
 #include "Framework/EventProcessor.h"
-#include "Tools/AnalysisUtils.h"
 
 namespace recon {
 
 class EcalPreselectionSkimmer : public framework::Producer {
  public:
-  // @param parameters Set of parameters used to configure this processor.
-  virtual void configure(framework::config::Parameters &);
-
   /** Constructor */
   EcalPreselectionSkimmer(const std::string &name, framework::Process &process);
 
   /** Destructor */
   ~EcalPreselectionSkimmer() = default;
 
+  // Configure this processor with a set of parameters passed
+  virtual void configure(framework::config::Parameters &) final;
+
   /**
    * Run the processor and select events that pass pre-selection in ECAL
    *
    * @param event The event to process.
    */
-  void produce(framework::Event &event);
+  virtual void produce(framework::Event &event) final;
 
  private:
   /// Collection Name for veto object
--- a/Recon/include/Recon/Skims/EcalPreselectionSkimmer.h
+++ b/Recon/include/Recon/Skims/EcalPreselectionSkimmer.h
@@ -1,0 +1,66 @@
+/**
+ * @file EcalPreselectionSkimmer.h
+ * @brief Processor used to pre-select events for the ECAL studies
+ * @author Tamas Almos Vami (UCSB)
+ */
+
+#ifndef RECON_SKIMS_EcalPreselectionSkimmer_H_
+#define RECON_SKIMS_EcalPreselectionSkimmer_H_
+
+//----------//
+//   LDMX   //
+//----------//
+#include "Ecal/Event/EcalVetoResult.h"
+#include "Framework/EventProcessor.h"
+#include "Tools/AnalysisUtils.h"
+
+namespace recon {
+
+class EcalPreselectionSkimmer : public framework::Producer {
+ public:
+  // @param parameters Set of parameters used to configure this processor.
+  virtual void configure(framework::config::Parameters &);
+
+  /** Constructor */
+  EcalPreselectionSkimmer(const std::string &name, framework::Process &process);
+
+  /** Destructor */
+  ~EcalPreselectionSkimmer() = default;
+
+  /**
+   * Run the processor and select events that pass pre-selection in ECAL
+   *
+   * @param event The event to process.
+   */
+  void produce(framework::Event &event);
+
+ private:
+  /// Collection Name for veto object
+  std::string ecal_veto_name_;
+  /// Pass Name for veto object
+  std::string ecal_veto_pass_;
+  /// Max value for summed det
+  double summed_det_max_;
+  /// Max value for summed tigh iso
+  double summed_tight_iso_max_;
+  /// Max value for ecal back energy
+  double ecal_back_energy_max_;
+  /// Max value for num readout hits
+  int n_readout_hits_max_;
+  /// Max value for shower rms
+  int shower_rms_max_;
+  /// Max value for shower rms in Y
+  int shower_y_std_max_;
+  /// Max value for shower rms in X
+  int shower_x_std_max_;
+  /// Max value for maximal cell deposition
+  double max_cell_dep_max_;
+  /// Max value for std layer hits
+  int std_layer_hit_max_;
+  /// Max value for num straight tracks
+  int n_straight_tracks_max_;
+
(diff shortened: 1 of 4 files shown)


Recon/include/Recon/Skims/EcalPreselectionSkimmer.h
```diff
@@ -0,0 +1,66 @@
...
+
+  /** Destructor */
+  ~EcalPreselectionSkimmer() = default;
+
+  /**
+   * Run the processor and select events that pass pre-selection in ECAL
+   *
+   * @param event The event to process.
+   */
+  void produce(framework::Event &event);
```
tomeichlersmith: Just some restructuring (I like putting the functions in order that they are called except the constructor/desturcor) and adding `final`.

```suggestion
  /** Constructor */
  EcalPreselectionSkimmer(const std::string &name, framework::Process &process);

  /** Destructor */
  ~EcalPreselectionSkimmer() = default;

   // @param parameters Set of parameters used to configure this processor.
  virtual void configure(framework::config::Parameters &) final;

  /**
   * Run the processor and select events that pass pre-selection in ECAL
   *
   * @param event The event to process.
   */
  virtual void produce(framework::Event &event) final;
```

Recon/include/Recon/Skims/EcalPreselectionSkimmer.h
```diff
@@ -0,0 +1,66 @@
...
+
+#ifndef RECON_SKIMS_EcalPreselectionSkimmer_H_
+#define RECON_SKIMS_EcalPreselectionSkimmer_H_
+
+//----------//
+//   LDMX   //
+//----------//
+#include "Ecal/Event/EcalVetoResult.h"
+#include "Framework/EventProcessor.h"
+#include "Tools/AnalysisUtils.h"
```
tomeichlersmith: Not used

```suggestion
```

Recon/python/ecalPreselectionSkimmer.py
```diff
@@ -0,0 +1,63 @@
...
+        self.shower_rms_max = 9999
+        self.shower_y_std_max = 9999
+        self.shower_x_std_max = 9999
+        self.max_cell_dep_max = 9999.
+        self.std_layer_hit_max = 9999
+        self.n_straight_tracks_max = 9999
+
+
+ecalPreselectionSkimmer = EcalPreselectionSkimmer("ecalPreselectionSkimmer")
+
```
tomeichlersmith: Folks should be creating their own instance of the config class.

```suggestion
```

If you want there to be a default name, then update `__init__` to use `name = "defaultName"`.

Recon/src/Recon/Skims/EcalPreselectionSkimmer.cxx
```diff
@@ -0,0 +1,62 @@
...
+                       (ecalVeto.getNStraightTracks() < n_straight_tracks_max_);
+
+  // Tell the skimmer to keep or drop the event based on whether preselection
+  // passed
+  if (passedPreselection) {
+    ldmx_log(debug) << "This event passed preselection!";
+    setStorageHint(framework::hint_shouldKeep);
+  } else {
+    setStorageHint(framework::hint_shouldDrop);
+  }
```
tomeichlersmith, line 58: I think it would be good to also add the decision to the event. This will mean folks could also have this branch of the output tree at the analysis stage in order to compare distributions before and after pre-selection if they wanted.

```suggestion
  }
  
  event.add("EcalPreselectionDecision", passedPreselection);
```

Recon/src/Recon/Skims/EcalPreselectionSkimmer.cxx
```diff
@@ -0,0 +1,62 @@
...
+  max_cell_dep_max_ = ps.getParameter<double>("max_cell_dep_max");  // MeV
+  std_layer_hit_max_ = ps.getParameter<int>("std_layer_hit_max");
+  n_straight_tracks_max_ = ps.getParameter<int>("n_straight_tracks_max");
+
+  return;
+}
+
+void EcalPreselectionSkimmer::produce(framework::Event &event) {
+  bool passedPreselection{false};
+  auto ecalVeto{
```
tomeichlersmith:
```suggestion
  const auto& ecalVeto{
```

Recon/test/ecal_preselection_skim.py
```diff
@@ -0,0 +1,17 @@
+import os import sys
```
tomeichlersmith: Oops! looks like the `clang-format` was run on this Python file. Reverting this file to the commit before formatting is probably what you want to do.

It looks like the only change I'd suggest is to avoid using the pre-created skimmer (that I suggest to avoid creating) and instead create your own.

```python
from LDMX.Recon.ecalPreselectionSkimmer import EcalPreselectionSkimmer
eps = EcalPreselectionSkimmer("ecal-preselection")
eps.summed_tight_iso_mas = 1100.
eps.n_readout_hits_max = 90
p.sequence = [ eps ]
```

tomeichlersmith (changes requested):
Just a pile of technical stuff, no real issues with how this is implemented.

Recon/include/Recon/Skims/EcalPreselectionSkimmer.h
tvami: done in next commit

Recon/python/ecalPreselectionSkimmer.py
tvami: done in the next commit

Recon/src/Recon/Skims/EcalPreselectionSkimmer.cxx
tvami: done now

Recon/src/Recon/Skims/EcalPreselectionSkimmer.cxx
tvami: done now. Also FYI clang like to put the `&` exactly next to the variable, i.e.`const auto &ecalVeto{`

Recon/test/ecal_preselection_skim.py
tvami: oh god haha, ok fixing this now

Recon/include/Recon/Skims/EcalPreselectionSkimmer.h
tvami: fixed

tomeichlersmith (approved):

danyi211 (approved):

PR #1396, opened 2024-08-20
copy in positron-deduced positions and directions from the TS scoring planes

### LINKED ISSUE #1395: Change upstream beam location to come from the SP at the upstream TS

Tracking points out that
```
 [ TruthSeedProcessor ] 3 : Linear extrapolation to a far away surface in B field.  This will cause inaccuracies in track parameters  Distance extrapolated = 5119.9
```
This is because now we shoot the beam from outside the `World`

Let's shoot it from the upstream TS's SP

Defined another SP, but I think we agreed it's more physical to have it at the TS

tomeichlersmith:
moving the z-position up to the upstream face of the furthest upstream trigger pad so that it is more feasibly understood by the Tracking algorithms

This resolves #1395 

## Check List

position-and-momentum-trunk-1395-comp

The tar-ball below has the configs I used and the python scripts I used to extract the beam positions and make the plot above.
[1395-testing.tar.gz](https://github.com/user-attachments/files/16676057/1395-testing.tar.gz)

### DIFF
--- a/SimCore/exampleConfigs/backwards-positron.py
+++ b/SimCore/exampleConfigs/backwards-positron.py
@@ -1,0 +1,40 @@
+"""fire a positron with the input energy backwards from the target
+
+This is helpful for the situation where the beam spot needs to be
+determined. Instead of doing complicated math to figure out how much
+an electron of a given energy curves, let's just use a positron and Geant4
+to do the math for us!
+
+Note
+----
+This only functions well if the detector components upstream of the target
+are not included in the simulation. (If they are, then the positron interacts
+with them easily spoiling the measurement.) This can be done by commenting
+out their inclusion in the detector.gdml with `<!-- ... -->`.
+"""
+
+from LDMX.Framework import ldmxcfg
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--n-events', type=int, default=10, help='number of events to simulate')
+parser.add_argument('--beam', type=float, required=True, help='beam energy in GeV')
+
+args = parser.parse_args()
+
+p = ldmxcfg.Process("beam")
+p.maxEvents = args.n_events
+p.termLogLevel = 1
+p.run = 1
+p.outputFiles = [ f'backwards_positron_beam_{args.beam}.root' ]
+
+from LDMX.SimCore import generators
+from LDMX.SimCore import simulator
+import LDMX.Ecal.EcalGeometry
+import LDMX.Hcal.HcalGeometry
+
+mySim = simulator.simulator( "mySim" )
+mySim.setDetector( 'ldmx-det-v14' , True )
+mySim.generators = [ generators.single_backwards_positron(args.beam) ]
+mySim.description = 'Basic test Simulation'
+p.sequence = [ mySim ]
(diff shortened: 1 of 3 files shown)


tvami:
I loved that you attached the tar file in the PR. I'm actually wondering if it would make sense to add the `backwards-positron.py` file in the repo, let's say under `SimCore/test/` with a bit more comment in the file about the use case. Like when this `single_backwards_positron` was mentioned I did a search in the repo and nothing uses it, so if we added that it would be a little bit more self documenting. What do you think?

tvami:
Oh and one question:

> mySim.setDetector( 'ldmx-det-v14' , True )

This is fine in `backwards-positron.py` for all energies because you anyway comment out the trackers? and with that the `ldmx-det-v14` and `ldmx-det-v14-8gev` are the "same"?

tomeichlersmith:
> This is fine in backwards-positron.py for all energies because you anyway comment out the trackers? and with that the ldmx-det-v14 and ldmx-det-v14-8gev are the "same"?

Correct :+1:

Edit: For more detail and for future readers. We are merely using Geant4 to numerically move a charged particle through a magnetic field. In our situation, the easiest way to get this trajectory is just to remove the upstream components since we know that is where the positron will be passing through. We just need the magnetic field map and the target to be in the same locations and coordinates systems as our detector for this to work. I could've used `ldmx-det-v9` and commented out the upstream stuff because v9 has the same mag field map, coordinate system, and target location and v14.

tvami (approved):

tvami:
So now our tracking msg is down to 2.9
```
[ TruthSeedProcessor ] 3 : Linear extrapolation to a far away surface in B field.  This will cause inaccuracies in track parameters  Distance extrapolated = 2.9
```

@bloodyyugo do you think we could simply add a range where this is not an issue? like if the distance extrapolated is less than 5, that's ok?

bloodyyugo:
> So now our tracking msg is down to
> 
> ```
> [ TruthSeedProcessor ] 3 : Linear extrapolation to a far away surface in B field.  This will cause inaccuracies in track parameters  Distance extrapolated = 2.9
> ```
> 
> @bloodyyugo do you think we could simply add a range where this is not an issue? like if the distance extrapolated is less than 5, that's ok?

I do think +/- 5mm is fine, but what we should do is set the "beamOrigin_" to (-882.5055, -21.745876,0) as default in TruthSeedTracker here: 

https://github.com/LDMX-Software/ldmx-sw/blob/c25fab362f9f478ad55518c20278645b9334f797/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx#L67-L68

...this is also settable on the config.  

I picked the value for 8 GeV but really only the z-position (x in tracking coordinates) matters so this should work for any energy.

tomeichlersmith (commented):
I looked through the failing histograms in the inclusive sample and they appear to just be statistical fluctuations. This makes sense because the generator has been moved up, less steps are taken by the primary, so the sequence of the RNG has been shifted and thus different parts of the distributions are being sampled.

tvami:
What was the final `Distance extrapolated` after you changed the seeder? Is the msg gone?

tomeichlersmith:
Sorry I rushed it, I did not check.

I am running the `ecal_pn` validation sample right now and not seeing any message. (Which makes sense? since the Distance extrapolated seemed to always be the difference between the true beam source z and the setting in the seeder?)

tvami:
>  and not seeing any message

very good! I'm happy

PR #1401, opened 2024-08-21
Including ratio subplot within differ when comparing

### LINKED ISSUE #1382: Ratio inset for plots in DQM Validation module

Right now the `Validation` module offers to overlay 1D histograms.

I believe it would be super useful to automatically generate a ratio plot in the bottom of the overlaid plot to show the differences.

One could do this outside of the `Validation` module in a private script, but wouldnt it be easier to do this once and have it in ldmx-sw?

tomeichlersmith:
The denominator for the ratio is whatever histogram is first in the list. This could lead to problems in two situations:
1. The first file in found in the directory (or given on the command line) does not have the histogram requested and so the denominator just becomes the next histogram that is found.
2. There are zero values in the denominator. This will lead to annoying divide-by-zero warnings which we can silence if there are a lot of them.

We try to inherit the color from the histogram in the ratio plot so that the legend applies. This may fail if I misunderstood the return type of the matplotlib `hist` function.

If we want to include error bars, we should probably switch to using [`hist` ](https://hist.readthedocs.io/en/latest/) directly since they have better error bar handling and have ironed out a lot of the common bugs. `uproot` can cast ROOT histograms into `hist.Hist` objects so the loading of already filled histograms is as seamless as it is now.

This resolves #1382 .

## Check List

### DIFF
--- a/Validation/src/Validation/_differ.py
+++ b/Validation/src/Validation/_differ.py
@@ -108,30 +108,46 @@ class Differ :
             All other key-word arguments are passed into each File.plot1d
         """
         fig = plt.figure('differ',figsize=(11,8))
-        ax = fig.subplots()
+        raw_ax, ratio_ax = fig.subplots(
+            nrows = 2,
+            sharex = 'col',
+            height_ratios = [2, 1],
+            gridspec_kw = dict(
+                hspace = 0.05
+            )
+        )
 
+        raw_histograms = []
         for f in self.files :
             try:
-                weights, bins, patches = f.plot1d(ax, column, **hist_kwargs)
+                raw_histograms.append(f.plot1d(raw_ax, column, **hist_kwargs))
             except uproot.KeyInFileError:
                 f.log.warn(f"Key {column} doesn't exist in {self}, skipping")
                 continue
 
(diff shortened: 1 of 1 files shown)


tvami:
It looks promising, here is an example:

Screenshot 2024-08-27 at 14 03 26

I do think tho that we should not connect the points in the ratio inset!

tvami:
Screenshot 2024-08-27 at 14 06 20

In this example, we have zero bins for the first histo, and the following histos too --> I think the current behavior is great as it is for that

tvami (changes requested):
The only thing I'd like to change is the connecting the dots in the ratio inset

tvami:
Looks very nice when the binning size is reasonable
Screenshot 2024-08-28 at 13 08 01

tvami:
when the binning is tiny, it's a bit messy:
Screenshot 2024-08-28 at 13 08 58

tvami (approved):
I think this looks good now, my comment for the binning is beyond this PR

tvami:
Plots were made by Gregorio @ohghvstt

PR #1407, opened 2024-08-25
Patching to ECAL lin-reg tracking

### LINKED ISSUE #1405: Error in <TDecompSVD::Diagonalize>: no convergence after XX steps
**Describe the bug**

This msg from the EcalVeto has been there since forever, and is probably not a big deal
```
Error in <TDecompSVD::Diagonalize>: no convergence after XX steps
```

**To Reproduce**

Just run the ECAL veto.
Specifically, if you run the test ecal pn config with seed 56 and one event, it will give you a problematic matrix already.

**Desired behavior**

Not to have this error.
My thinking is that it's having a matrix that has repeated elements, and the fix is probably just about conditioning what goes into the `TDecompSVD`. I'll have a look at it in details and report back.

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1405
Several approaches were tried, in the end I'm cleaning up the input to the matrix to be decomposed, and making sure that the determinant is non-zero.
While digging into this I denote many potential improvements in the code.

## Check List

### DIFF
--- a/Ecal/src/Ecal/EcalVetoProcessor.cxx
+++ b/Ecal/src/Ecal/EcalVetoProcessor.cxx
@@ -85,6 +85,7 @@ void EcalVetoProcessor::buildBDTFeatureVector(
 }
 
 void EcalVetoProcessor::configure(framework::config::Parameters &parameters) {
+  verbose_ = parameters.getParameter<bool>("verbose");
   doBdt_ = parameters.getParameter<bool>("do_bdt");
   featureListName_ = parameters.getParameter<std::string>("feature_list_name");
   if (doBdt_) {
@@ -824,14 +825,12 @@ void EcalVetoProcessor::produce(framework::Event &event) {
 
   float cellWidth = 8.7;
   for (int iHit = 0; iHit < trackingHitList.size(); iHit++) {
-    int track[34];  // list of hit numbers in track (34 = maximum theoretical
-                    // length)
-    int currenthit;
-    int trackLen;
+    // list of hit numbers in track (34 = maximum theoretical length)
+    int track[34];
+    int currenthit{iHit};
+    int trackLen{1};
 
     track[0] = iHit;
-    currenthit = iHit;
-    trackLen = 1;
 
     // Search for hits to add to the track:
     // repeatedly find hits in the front two layers with same x & y positions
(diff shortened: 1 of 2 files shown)


tvami:
OK, I'll move this back to "draft", the signal (right) looks way to much like background (left) now.
So I'll have to look into this in more details
Screenshot 2024-08-25 at 16 34 27

tvami:
I'm getting there, still not the best
Screenshot 2024-08-26 at 08 35 34

tomeichlersmith:
How are you defining best? If you are cutting out singular matrices, could there be some lin-reg tracks "found" that were just "found" due to numerical instability of a singular matrix? I could see this effect being very distinct in the signal case where the ECal as a whole is fairly quiet.

tvami:
Yes but we have more tracks now in signal! (blue is new red is old, right?)
So the changes seeing here are a bit counter-intuitive, how does removing singular matrices lead to more tracks?
Also I want to measure how it affect the other masses, I have a student Jihoon looking into that part now.

tomeichlersmith:
ah yes, you're right - that is suspicious.

Ecal/src/Ecal/EcalVetoProcessor.cxx
```diff
@@ -1042,13 +1044,21 @@ void EcalVetoProcessor::produce(framework::Event &event) {
...
         }
 
-        // Perform "linreg" on selected points:
-        TDecompSVD svdObj = TDecompSVD(hdt);
+        // Perform "linreg" on selected points
+        // Calculate the determinant of the matrix
+        double determinant =
+            hdt(0, 0) * (hdt(1, 1) * hdt(2, 2) - hdt(1, 2) * hdt(2, 1)) -
+            hdt(0, 1) * (hdt(1, 0) * hdt(2, 2) - hdt(1, 2) * hdt(2, 0)) +
+            hdt(0, 2) * (hdt(1, 0) * hdt(2, 1) - hdt(1, 1) * hdt(2, 0));
```
tomeichlersmith, line 1052: `TMatrixD` has its own determinant function which is more clear and (hopefully) less error prone.

```suggestion
double determinant = hdt.Determinant();
```

tvami:
Gosh... I'm ready for review, but there is waaay to many things to improve on this part of the code. My approach is to make something minimally sensible now, and add `// TODO` to the point that needs to be improved. This can be a project for an undergrad (I'm sure we'll find somebody at UCSB).

As for the physics in the version pushed now:
Screenshot 2024-08-29 at 10 55 09

Looks pretty good: we loose tracks (that were anyway nonsensical [hits from the same layer, exactly the same hits, inverting singular matrices....]), but we still have move tracks in the background case.

tvami:
> This can be a project for an undergrad

I'll make a reminder issue about this after we merged this

Ecal/src/Ecal/EcalVetoProcessor.cxx
```diff
@@ -1042,13 +1044,21 @@ void EcalVetoProcessor::produce(framework::Event &event) {
...
+        // Exit early if the matrix is singular (i.e. det = 0)
+        if (determinant == 0) continue;
```
tomeichlersmith, line 1054: Any concerns on numerical precision? (Should we check `abs(determinant) < 1e-6` or something instead?)

tomeichlersmith (changes requested):
Looks good to me, I just have a cleanup :broom: and a technical question :thinking:

Ecal/src/Ecal/EcalVetoProcessor.cxx
tvami: Haha I know! It's ridiculous... Calling this `.Determinant();` function will create a similar error msg that I wanted to avoid in the first place, i.e. ROOT internal and something we cannot silence. There is also a way to check the `.Condition()` of the matrix, which could serve a similar purpose, you wont guess... that will also lead to a ROOT internal msg...

Ecal/src/Ecal/EcalVetoProcessor.cxx
tvami: I was thinking about this too, but looking at some specific cases, the problematic events are always zero. And since in the matrix we have this subtraction of the mean, it can lead to small numbers, which then means the determinant can be a small number. And then in a few lines down, we are calling ROOT's decompose that will kill the events like that, so I'd prefer to have this as is, just to kill the diabolical events.

tvami: (i.e. mathematically it's possible to divide with  1e-6, while it's not with 0)

Ecal/src/Ecal/EcalVetoProcessor.cxx
tomeichlersmith: Well, we _could_ switch to Eigen which is already installed in the development image for acts. It has an [SVD decomposition](https://eigen.tuxfamily.org/dox/group__LeastSquares.html), [a determinant member function](https://eigen.tuxfamily.org/dox/classEigen_1_1MatrixBase.html#a7ad8f77004bb956b603bb43fd2e3c061), and compile-time dimensions which might help improve performance.

tvami: Sure, but I'd do that in another PR tho.

tvami:
Funny thing is that the KS score is not big enough for the signal to even flag the change:

Screenshot 2024-08-29 at 12 22 20

tomeichlersmith:
Yea, KS testing is somewhat weighted by the bin content and so the fact that 0 lin-reg tracks are very close for a vast majority of events means that the KS test value will be above our arbitrary threshold.

(In a perfect world, we would do a full comparison of the events with something like [root-diff](https://github.com/tomeichlersmith/root-diff/) and then make histograms if any change cropped up there. The limiting thing in this case is storing the golden event files or implementing the complex workflow to re-generate them everytime.)

Edit: The N Lin-Reg Tracks distribution is also closer in the PR validation sample. Not sure why this is but could be from using different samples.
EcalMipTrackingFeatures_EcalMipTrackingFeatures_n_linreg_tracks

tomeichlersmith (approved):

danyi211 (approved):

tvami:
> The N Lin-Reg Tracks distribution is also closer in the PR validation sample. 

Ahh no sorry, I think I introduced a mistake in the last push regarding the initialization of the r2

tvami:
> > The N Lin-Reg Tracks distribution is also closer in the PR validation sample.
> 
> Ahh no sorry, I think I introduced a mistake in the last push regarding the initialization of the r2

False alarm
https://github.com/LDMX-Software/ldmx-sw/blob/117acb437eabca116030524eee056eefe1385228/Ecal/src/Ecal/EcalVetoProcessor.cxx#L1075C15-L1079
the requiremenet on `r_corr > .6` makes it ok

PR #1410, opened 2024-08-26
Fix underflow/overflow in EcalVeto's closest cell calc

tvami:

Part of the https://github.com/LDMX-Software/ldmx-sw/issues/1166

As for the source of the issue, the `recoilPos` is already outside the ECAL, and that's calculated at the SP. However the SP is bigger then the ECAL, so we can have hits that are on the ECAL SP, but are not on the real ECAL. This PR deals with those hits, in a reasonable way with the X position. But to be extra sure we dont reach the underflow and overflow in the `mapsx`/`mapsy`, I also edited the `while` loop

## Check List

Compiling with ASAN build
```
ldmx cmake -B build -S . -DENABLE_SANITIZER_ADDRESS=ON
ldmx cmake --build build --target install -- -j$(nproc)
<run the ecal pn validation config with seed=1 and at event 142 you'll run into one issue>
```
the problem is gone.

### DIFF
--- a/Ecal/src/Ecal/EcalVetoProcessor.cxx
+++ b/Ecal/src/Ecal/EcalVetoProcessor.cxx
@@ -659,23 +659,32 @@ void EcalVetoProcessor::produce(framework::Event &event) {
     faceXY[1] = -9999.0;
   }
 
-  int inside = 0;
-  int up = 0;
-  int step = 0;
-  int index;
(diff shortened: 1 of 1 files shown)


tomeichlersmith (approved):
:broom: Thank you :tada:

danyi211 (approved):

PR #1421, opened 2024-08-30
Add kaon validation + fix tracking seeder settings in ecal validation

### LINKED ISSUE #1418: Add kaon to the CI validation suite

The kaons seems to be super important for the experiment, but we dont have a CI workflow that would run that simulation.

I'd like to add one.

tvami:

 - Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1418
- Furthermore, I edited the tracking (seeding) part of the ecal pn validation with more realistic parameters with the help of @bloodyyugo 

## Check List


.github/validation_samples/kaon_enhanced/config.py
```diff
@@ -0,0 +1,162 @@
...
+
+from LDMX.Framework import ldmxcfg
+
+thisPassName = 'test'
+p=ldmxcfg.Process(thisPassName)
+
+p.termLogLevel = 0
+# This need to be set but has no meaning given p.totalEvents
+p.maxEvents = 1
+p.totalEvents = int(0.5*int(os.environ['LDMX_NUM_EVENTS']))
```
tvami: It takes about double time to have the kaons done, i.e. with taking 1/2 of the events it will take about the same time for validation

tvami:
@tomeichlersmith I think now there is enough change to cut a new patch + golds after this? Maybe with the `denv` PR in too (feels like it's close, if not that's ok)?

.github/validation_samples/kaon_enhanced/config.py
tomeichlersmith: I would add this comment into the file itself to help any future folks looking at this.
```suggestion
# kaon-focused sample takes about twice as long as normal PN,
# so we ask for half as many events such that validation jobs stay
# time-limited by the PN
p.totalEvents = int(os.environ['LDMX_NUM_EVENTS']) // 2
```

.github/validation_samples/kaon_enhanced/config.py
```diff
@@ -0,0 +1,162 @@
...
+if beamEnergy == 4 :
+     detector = 'ldmx-det-v14'
+     generator=gen.single_4gev_e_upstream_tagger()
+     bias_factor = 450.
+     bias_treshold = 2500.
+else :
+     detector = 'ldmx-det-v14-8gev'
+     generator=gen.single_8gev_e_upstream_tagger()
+     bias_factor = 550.
+     bias_treshold = 5000.
```
tomeichlersmith: The thresholds in the filters below are hard-coded to the 8GeV values so I'm removing the 4GeV branch here to avoid hinting to users that they could just change the `beamEnergy` above and it would work.

```suggestion
detector = 'ldmx-det-v14-8gev'
generator=gen.single_8gev_e_upstream_tagger()
bias_factor = 550.
bias_treshold = 5000.
```

.github/validation_samples/kaon_enhanced/config.py
```diff
@@ -0,0 +1,162 @@
...
+        util.TrackProcessFilter.photo_nuclear()
+])
+
+# set up upKaon parameters
+mySim.kaon_parameters = kaon_physics.KaonPhysics.upKaons()
+
+# Alternative pn models
+myModel = pn.BertiniAtLeastNProductsModel.kaon() 
+# These are the default values of the parameters
+myModel.hard_particle_threshold=0. # Count particles with >= 200 MeV as "hard"
```
tomeichlersmith:
```suggestion
myModel.hard_particle_threshold=0. # Count all (not stopped) particles as "hard"
```

tomeichlersmith (changes requested):
Just some detail changes since these CI configs tend to last a long time and become references.

tomeichlersmith:
Yea, I think I can get the `just`+`denv` PR ready in a few days. I mainly just want to finish the updates to the documentation website and do some final testing on `denv` before making a v1 release.

.github/validation_samples/kaon_enhanced/config.py
EinarElen: What is the reason for having no threshold? Or, i guess, what would the concern be with a 30 MeV kaon?

tvami: This is the config used in the official production, 
https://github.com/LDMX-Software/LDCS/blob/main/productionScripts/v3.3/ldmxjob_kaon_v3.3.3.py.tpl#L66
Maybe @danyi211 remembers why zero was put here

danyi211: I think the reason is we want to accept all the kaons, including the soft ones that are more difficult to reject. According to [here](https://github.com/LDMX-Software/ldmx-sw/blob/9088c6798683f0aaaec727ba3b9e5dc69e07c7db/SimCore/src/SimCore/PhotoNuclearModels/BertiniAtLeastNProductsModel.cxx#L14), the kaons with energy less than this threshold would not be accepted.

EinarElen: Soft kaons would be hard to reject yeah, but wouldn't you expect such events to be vetoed by the rest of the particles from the reaction? The rest of the photon energy has to go somewhere. Maybe I'm missing something!

danyi211: I think "difficult" events will include the ones whose final-state particles only have kaons and invisible particles...?

tvami: ok so my take is that this region deserves some extra attention, and it's better to be more inclusive now (i.e. no threshold) and cut those out at the analysis level. And then given that this was the config used in production, I'd say we keep it as is, figure these out, and if needed, modify in the future (also meaning a future production). Does that sounds reasonable @EinarElen ?

EinarElen: Yep, definitely not something to resolve here. I was mostly wondering what I was missing :)

.github/validation_samples/kaon_enhanced/config.py
tvami: done in next push

.github/validation_samples/kaon_enhanced/config.py
tvami: done in the next push

.github/validation_samples/kaon_enhanced/config.py
tvami: Done in the next push

.github/validation_samples/kaon_enhanced/config.py
```diff
@@ -0,0 +1,158 @@
...
+        filters.TaggerVetoFilter(thresh=2*3800.),
+        # Only consider events where a hard brem occurs
+        filters.TargetBremFilter(recoil_max_p = 2*1500.,brem_min_e = 2*2500.),
+        # Only consider events where a PN reaction happnes in the ECal
+        filters.EcalProcessFilter(),
+        # Tag all photo-nuclear tracks to persist them to the event.
+        util.TrackProcessFilter.photo_nuclear()
+])
+
+# set up upKaon parameters
```
EinarElen: I guess this also would be worth a comment, I'm guessing that the name upKaons doesn't immediately tell a new user what we are doing.

EinarElen (approved):

.github/validation_samples/kaon_enhanced/config.py
tvami: Any suggestions for the text to use?

EinarElen: Sure, we can steal from the docs directly :) 

Reduces the charged kaon lifetimes by a factor 1/50 and forces decays to be into one of the leptonic decay modes.

tvami: Sure, done now!

tomeichlersmith (approved):

PR #1424, opened 2024-08-30
Add TaggerHitFilter and no-cals detectors

### LINKED ISSUE #1423: Add filter on the number of sim hits in the tagger and add detectors with just tracker & scintillator
We need to study the effect of off-energy beam electrons in our detector, namely how often they are identified as full-energy electrons.   The probability for this is very small so we need to add a filter after simulation to reject events that clearly wouldn't be mistaken for full energy, namely they don't leave enough hits in the tracker and no hits in layers closest to target.  

Omar had already done a study for 4gev electrons,[ see his presentation here](https://docs.google.com/presentation/d/1CXDg5SdAqqoztQeh1k4IQ-GJeayA-BS54hAZje7T5CA/edit#slide=id.g24f3c5c0928_0_72), and has a filter but it was never committed.  That's what I'm doing here.  

In addition, we are adding 4 and 8 gev detectors with no ECal or HCal (just tracker and scintillator).  Removing the calorimeters greatly speeds up simulation.

bloodyyugo:
This resolves issue 1423...as title says, add a filter on the tagger hits and adds detectors with no calorimeters.


Biasing/include/Biasing/TaggerHitFilter.h
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @param[in] step Geant4 step
+   */
+  void stepping(const G4Step* step) final override;
+
+  /**
+   * Action called once tracking of all particles has concluded. This is being
+   * used to clear the hit count set in preparation for the next event.
+   */
+  void EndOfEventAction(const G4Event* event) final override;
```
tvami, line 50:
```suggestion
  void endOfEventAction(const G4Event* event) final override;
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+  //  track->GetParentID()<<std::endl;
+  //}
+  //  if (copy_number>20 && track->GetParentID() != 0)
+  //  return;
+  //  std::cout<<"Inserting hit "<<copy_number<<"  Parent ID  "<<
+  //  track->GetParentID()<<std::endl;
+  layer_count_.insert(copy_number);
+}
+
+void TaggerHitFilter::EndOfEventAction(const G4Event* event) {
```
tvami, line 57:
```suggestion
void TaggerHitFilter::endOfEventAction(const G4Event* event) {
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+#include "G4RunManager.hh"
+#include "G4Step.hh"
+
+namespace biasing {
+
+TaggerHitFilter::TaggerHitFilter(const std::string& name,
+                                 framework::config::Parameters& parameters)
+    : simcore::UserAction(name, parameters) {
+  layers_hit_ = parameters.getParameter<int>("layers_hit", 8);
+  std::cout << "TaggerHitFilter:: layers_hit_ = " << layers_hit_ << std::endl;
```
tvami:
```suggestion
  ldmx_log(debug) << " layers_hit_ = " << layers_hit_;
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+
+  // Get the PDG ID of the track and make sure it's an electron. If
+  // another particle type is found, thrown an exception because the generator
+  // being used is likely wrong.
+  //  std::cout<<"Track pdgID =
+  //  "<<track->GetParticleDefinition()->GetPDGEncoding()<<std::endl; June 21,
+  //  2024 MG:  similar here, though it's unlikely to be aything else if (auto
+  //  pdgID{track->GetParticleDefinition()->GetPDGEncoding()}; pdgID != 11){
+  //  return;
+  //}
```
tvami:
```suggestion
```
For clarity if you think these comments are not helpful, it's better to remove them

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+  // Require that track is charged
+  if (auto pdgCh{track->GetParticleDefinition()->GetPDGCharge()};
+      abs(pdgCh) != 1) {
```
tvami: I wonder if we ever get alpha particles, should we maybe have?
```suggestion
      abs(pdgCh) == 0) {
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+    return;
+  }
+
+  // Only electrons in the Tagger region are of interest.
+  auto volume{track->GetVolume()};
+  //  std::cout<<"Volume Name =
+  //  "<<volume->GetLogicalVolume()->GetRegion()->GetName()<<std::endl;
```
tvami:
```suggestion
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+  if (auto
+  nregion{track->GetNextVolume()->GetLogicalVolume()->GetRegion()->GetName()};
+      (nregion.compareTo("tagger") != 0) ||
+      (track->GetKineticEnergy() == 0)) {
+      checkAbortEvent(track);
+      return;
+  }
+  */
+  // MG:  remove the kinetic energy requirement
+  //      this basically just checks if we are exiting the tagger
```
tvami, line 67:
```suggestion
  //      Checks if we are exiting the tagger
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+          track->GetNextVolume()->GetLogicalVolume()->GetRegion()->GetName()};
+      (nregion.compareTo("tagger") != 0)) {
+    checkAbortEvent(track);
+    return;
+  }
+
+  // A particle will only leave hits in the active silicon so other volumes can
+  // be skipped for now.
+  //  std::cout<<"check active Si volume
+  //  "<<track->GetVolume()->GetName()<<std::endl;
```
tvami:
```suggestion
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+  // if layers 1 or 2 (copy_number = 10 or 20) count
+  // from any source
+  //  if (copy_number<21){
+  //  std::cout<<"Found inner layer hit "<<copy_number<<"  Parent ID  "<<
+  //  track->GetParentID()<<std::endl;
+  //}
+  //  if (copy_number>20 && track->GetParentID() != 0)
+  //  return;
+  //  std::cout<<"Inserting hit "<<copy_number<<"  Parent ID  "<<
+  //  track->GetParentID()<<std::endl;
```
tvami:
```suggestion
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+       << std::endl;
+    */
+    return;
+  }
+
+  /*
+  std::cout << "[ ";
+  for (auto &l : layer_count_) std::cout << l << ", ";
+  std::cout << " ]" << std::endl;
+  */
```
tvami:
```suggestion
```

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+    //  if (layer_count_.size() < layers_hit_) {
+    if (track != nullptr) track->SetTrackStatus(fKillTrackAndSecondaries);
+    G4RunManager::GetRunManager()->AbortEvent();
+    /*
+          std::cout << "Aborting" << std::endl;
+          std::cout<<"layer_count = "<<layer_count_.size()<<std::endl;
+          std::cout<<"layer_count_.count(10) "<<layer_count_.count(10) <<
+       std::endl; std::cout<<"layer_count_.count(20) "<<layer_count_.count(20)
+       << std::endl;
+    */
```
tvami:
```suggestion
```

Detectors/data/ldmx-det-v14-no-cals/constants.gdml
tvami, line 1: Hmm what's the point of adding this geometry? My understanding is that at this point we'll not have a 4 GeV beam at the time when we have the full tracker apparatus ready

Detectors/data/ldmx-det-v14-8gev-no-cals/detector.gdml
```diff
@@ -0,0 +1,229 @@
...
+        <rotationref ref="identity"/> 
+      </physvol>
+      <auxiliary auxtype="DetElem" auxvalue="Top"/>
+    </volume>
+  </structure>
+    
+  <userinfo>
+    <!-- detector version -->
+    <auxiliary auxtype="DetectorVersion" auxvalue="14">
+      <auxiliary auxtype="DetectorName" auxvalue="ldmx-det-v14-8gev"/>
```
tvami:
```suggestion
      <auxiliary auxtype="DetectorName" auxvalue="ldmx-det-v14-8gev-no-cals"/>
```

Detectors/data/ldmx-det-v14-8gev-no-cals/detector.gdml
```diff
@@ -0,0 +1,229 @@
...
+      <auxiliary auxtype="Description" 
+                 auxvalue="The Late 2022 Design Report Geometry."/>
```
tvami:
```suggestion
                 auxvalue="The Late 2022 Design Report Geometry w/o calorimeters."/>
```

Detectors/data/ldmx-det-v14-8gev-no-cals/detector.gdml
```diff
@@ -0,0 +1,229 @@
...
+      <physvol copynumber="6"> 
+        <file name="ecal.gdml"/> 
+        <positionref ref="em_calorimeter_pos"/> 
+        <rotationref ref="identity"/> 
+	    </physvol>
+      <physvol copynumber="7"> 
+        <file name="hcal.gdml"/> 
+        <positionref ref="hadron_calorimeter_pos"/> 
+        <rotationref ref="identity"/> 
+      </physvol> -->
```
tvami:
```suggestion
```
I looked at the diff wrt to the base and seeing only the commenting makes it less obs, while the full removal makes it clear

Detectors/data/ldmx-det-v14-no-cals/constants.gdml
tvami: One more thing is that these files are copied at compile, so I'd prefer to have the number of geometries kept at minimal, if possible

tvami (changes requested):
- I have a bunch of cleaning up comments
- The `auxiliary auxtype="DetectorName"` is the biggest "real" suggestion
- I'd really prefer to not have the 4 GeV geometry if possible

Biasing/src/Biasing/TaggerHitFilter.cxx
```diff
@@ -0,0 +1,137 @@
...
+}
+
+void TaggerHitFilter::EndOfEventAction(const G4Event* event) {
+  checkAbortEvent(nullptr);
+  layer_count_.clear();
+}
+
+void TaggerHitFilter::checkAbortEvent(G4Track* track) {
+  if ((layer_count_.size() < layers_hit_) ||
+      ((layer_count_.count(10) == 0) && (layer_count_.count(20) == 0))) {
```
EBerzin: I just have a general question based on this selection:

I think these hit constraints are tighter than what is currently in place for truth tracking and track reconstruction, where we just require a minimum number of hits (unless this is implicitly accounted for in the target extrapolation step in the CKFProcessor). If so, do we want to add something to track reconstruction to also check for hit placement close to the target? Or maybe just for truth tracks?

Biasing/include/Biasing/TaggerHitFilter.h
tomeichlersmith: I know this is against our styling but it is the correct capitalization for the override.

Biasing/src/Biasing/TaggerHitFilter.cxx
tomeichlersmith: I know this is against our styling but it is the correct capitalization for the override.

Biasing/src/Biasing/TaggerHitFilter.cxx
tomeichlersmith: I second this suggestion, we will certainly get wacky particles although we may want the requirement to be "_simple_ and charged"?

Biasing/src/Biasing/TaggerHitFilter.cxx
bloodyyugo: Good question...in Omar's presentation, he suggested this requirement and it seems sensible.  I think, currently, the strategies may require these layers be hit anyway,  [https://github.com/user-attachments/files/16818852/acts-v19-v36-tagger-comparisons-4gev.pdf](see this plot included for v36 upgrade),  though we will want to extend the strategies used,.   The extrapolation doesn't require any sort of layers-hit-pattern as far as I know (and if it does, we should get rid of it).  

That said, I personally haven't studied the 2-sensors-closest to target requirement in this context...I don't know how much it a) cuts down on the background we might see, and b) how much is speeds up generation.  With the way the code is right now, I don't think we loose any efficiency because I'm pretty sure they are required anyway.  Or at least they are always included in tracks we see.

Biasing/include/Biasing/TaggerHitFilter.h
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+ private:
+  void checkAbortEvent(G4Track* track);
+
+  /// Set used to keep track which layers were hit by a particle.
+  std::set<int> layer_count_;
+  /// Total number of hits required to persist an event.
+  int layers_hit_{8};
+
```
tvami, line 66:
```suggestion
  /// Enable logging
  enableLogging("TaggerHitFilter")

```
This is needed for the `ldmx_log(debug)` to work

Detectors/data/ldmx-det-v14-no-cals/constants.gdml
bloodyyugo: I don't really care if we add this or not...totally up to you.

Biasing/src/Biasing/TaggerHitFilter.cxx
tvami: Maybe have these numbers as named `constexpr` so they dont look more like magic number too?

Detectors/data/ldmx-det-v14-no-cals/constants.gdml
tvami: OK then let's just have the 8 GeV only, and remove these! Thanks!

Biasing/include/Biasing/TaggerHitFilter.h
```diff
@@ -0,0 +1,67 @@
...
+//~~ C++ StdLib ~~//
+#include <set>
+#include <string>
+
+//~~ SimCore ~~//
+#include "SimCore/UserAction.h"
+
+//~~ Framework ~~//
+#include "Framework/Configure/Parameters.h"
+
```
tvami, line 13:
```suggestion
#include "Framework/EventProcessor.h"

```

tvami (approved):
I squashed the many commits into one, also removed the 4GeV geometry, I'm happy with the current state now!

tomeichlersmith (approved):

EBerzin (approved):

PR #1425, opened 2024-08-31
Deprecate ECal cellxy.txt and use EcalGeometry instead

### LINKED ISSUE #1413: Deprecate ECal cellxy.txt and use EcalGeometry instead

Chatted with Tom about the ECAL cell sizes, and realized that
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Ecal/data/cellxy.txt
is actually not up to date with the v14 geometry.

For example in that file the leftmost and rightmost cells are at 242.4 mm , I assume that’s the center of the cell. So then adding the radios of 5 it will still not be 3*R+1.5mm = 256.5mm as we'd expect.
Furthermore, the horizontal shifting is not included :( 

Move completely to `EcalGeometry`.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1413

The code certainly got much simpler like this, no weird stepping in the `mapsx`/`mapsy` anymore.

## Check List

As a test I fixed the X location to -9999, or to -270 --> indeed it shows that it's not fiducial.
Otherwise ran the new kaon config, the fiducial case looked ok too.
I expect the bigger effects to show up in signal, especially with the 1 GeV case.


tvami:
@tomeichlersmith I think we should cut gold and then have this PR in, like that it nicely separates from other developments (like the lin-reg changes)

DetDescr/src/DetDescr/EcalGeometry.cxx
```diff
@@ -629,4 +629,54 @@ bool EcalGeometry::isInside(double normX, double normY) const {
...
+  // deduce cell ID
+  int cell_id = cell_id_in_module_.FindBin(p, q) - 1;
+
+  if (cell_id < 0) {
+    return false;
+  } else {
+    return true;
+  }
+}
+
```
tomeichlersmith: I am interested in reducing duplication so that this rather complicated logic is eaiser to maintain. My proposal is to add a `fallible` flag to the `getID` cascade of functions. If `fallible` is `false`, then the current behavior of throwing exceptions is done. This is helpful for SimCore where we want any `getID` call not being found to be treated as a run-ending bug in the geometry. If `fallible` is `true`, then an ID is returned that needs to be checked for validity. We could use the null ID `ldmx::EcalID(0)` (checked with `id.null()`) as the "invalid" ID or [std::optional](https://en.cppreference.com/w/cpp/utility/optional).

tomeichlersmith (commented):
This is an excellent update and one that a student I'm working with also has a use for!

My only comment is about reducing duplication.

tomeichlersmith (commented):
Awesome, thank you!

The last thing is just to update the documentation of `getID` to include the `fallible` parameter. And maybe an example about how to use it, e.g.

(doxycomments can contain codeblocks just like markdown).

```cpp
auto id = geometry.getID(x, y, ilayer, true);
if (id.null()) {
  // position (x,y) is not contained within a cell in layer ilayer
}
```

tomeichlersmith (approved):

danyi211 (approved):

PR #1441, opened 2024-09-06
Add validation installation to the just commands and rename the dir

### LINKED ISSUE #1439: Add validation installation to the just commands & rename the dir

This describtion
https://github.com/LDMX-Software/ldmx-sw/tree/d7392d661674d7ca0f2b35ef4d4778899eec0d95/Validation#inside-ldmx
doesn't work anymore with `ldmx` being an alias for `just`

Also there is this text in the README
> This module cannot be run from ldmx-sw directory. import Validation prefers the local directory instead of the installed path so it tries to load from the ldmx-sw/Validation directory.
> We could fix this by renaming the package inside Validation or renaming the directory.

- Add validation installation to the just commands.
- Rename the directory to Validate

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1439

## Check List
```
just install-validation
```
installs Validation module
and then
```
cd Validation/src
denv python3 -m Validate -h
```
works, and then 
```
denv python3 -m Validate compareDir/ --systems ecal.shower_feats
```
makes nice plots.


Validation/README.md
```diff
@@ -44,39 +32,35 @@ Using Validation without a virtual environment and outside of the container is n
...
-We could fix this by renaming the package inside Validation or renaming the directory.
 
 ### CLI
 The Validation module is constructed to do some common tasks quickly on the command line.
 Printing out its help message shows how to run it and gives you the details on what
 parameters to provide.
 ```
-python3 -m Validation -h
+cd Validation/src
+denv python3 -m Validate -h
```
tomeichlersmith: This brings up a large usage question. The whole point of installing the python package is to allow users to run `denv python3 -m Validate` from _anywhere_. In this example, you are particularly running from within the source code for the python module which isn't bad, it just makes the install step not necessary.

There are two ways forward:
1. Remove the install step: we don't need a `install-validation` recipe and instead we have a `validation-plot` recipe which does the necessary directory movement and passes the arguments to the module.
2. Keep the install step and make sure it can be run from anywhere. The current recipe does install the `Validate` python and I can get the help message from other places besides within the source directory.

I like the install step since then folks can run `denv python3 -m Validate` from where-ever their data happens to reside, but I also don't use the Validation module much and so if you prefer running the Validation source directly (maybe because you are editing it frequently?), then we could move in that direction. I just think we should only have one way of running the code to avoid confusion.

justfile
```diff
@@ -174,3 +174,15 @@ compile ncpu=num_cpus() *CONFIG='': (configure CONFIG) (build ncpu)
...
+
+# install the validation module
+# `python3 -m pip install Validation/` is the standard `pip` install method.
+# We add `--upgrade` to tell `pip` it should overwrite the package if it already has been
+# installed before which is helpful in the case where someone is updating the code and running
+# the new code within the container. The `--target install/python/` arguments tell `pip`
+# where to install the package. This directory is where we currently store our python modules
+# and is where the container expects them to be. The `--no-cache` argument tells `pip` to
+# not use a cache for downloading any dependencies from the internet which is necessary since
+# `pip` will not be able to write to the cache location within the container.
```
tomeichlersmith: The last comment line before a recipe is the one printed with the help message, so I am just reordering this so that the help message is "install the validation python module".

```suggestion
# `python3 -m pip install Validation/` is the standard `pip` install method.
# We add `--upgrade` to tell `pip` it should overwrite the package if it already has been
# installed before which is helpful in the case where someone is updating the code and running
# the new code within the container. The `--target install/python/` arguments tell `pip`
# where to install the package. This directory is where we currently store our python modules
# and is where the container expects them to be. The `--no-cache` argument tells `pip` to
# not use a cache for downloading any dependencies from the internet which is necessary since
# `pip` will not be able to write to the cache location within the container.
# install the python Validation plotting module
```

tomeichlersmith (commented):
My only other broad comment is to update the Validation/README.md to avoid references to `ldmx`. I forgot to update it when merging `denv`+`just` and so now is the time. I can do that if you don't want to as well.

Validation/README.md
tvami: Yeah I'm good with the solution that allows us to run this from anywhere, I'll modify the README accordingly

tvami: Done now!

tomeichlersmith (approved):

danyi211 (approved):

PR #1454, opened 2024-09-13
Update tracking to use Acts v36.0

bloodyyugo:
Pull request [addresses issue 1356](https://github.com/LDMX-Software/ldmx-sw/issues/1365), upgrading tracking to use ACTs version 36.0.  
 
The development branch is actsv36.0 and compiles, builds and runs. Comparisons between tracking using the previous ACTs version (19.X ... something) are shown on the issue page linked above.  

There is one outstanding problem in that we see fewer hits-on-track for tagger tracks with this update.  We don't see this in the recoil for some reason.  We are investigating but this issue can be closed without it.

PR #1458, opened 2024-09-16
logging fix and the long-awaited setting log-level by processor

### LINKED ISSUE #922: Allow setting of log level for a processor
Currently, the log level can only be set globally.  This is an issue when trying to debug a processor because messages from all other processors are also printed.  It would be useful to set the log level such that it only applies to a given processor.

### LINKED ISSUE #1301: Add event number to logger as attribute
I think this is helpful and would be helpful for anyone using the logger to debug their development.

Adding attributes to Boost Log is already well documented, so I hope that this isn't too difficult.

### LINKED ISSUE #1455: Multiple printouts when using logFrequency
**Describe the bug**

With the settings of
```
p.logFrequency = 100
p.termLogLevel = 1
```
I'd expect a single msg every 100 event to show up, instead, every 100 event the following
```
[ Process ] 1 : Processing 100 Run 2 Event 100  (2024-09-13 17:11:04.004336000-0700)
[ Process ] 1 : Processing 100 Run 2 Event 100  (2024-09-13 17:11:04.005039001-0700)
```
this comes up >5000 times!

**To Reproduce**
I was running this
```
p=ldmxcfg.Process("v14_nonfid")

import LDMX.Ecal.ecal_hardcoded_conditions as ecal_conditions
import LDMX.Hcal.hcal_hardcoded_conditions as hcal_conditions
from LDMX.Biasing import ecal

mysim = ecal.nonfiducial_photo_nuclear('ldmx-det-v14-8gev', generators.single_8gev_e_upstream_tagger())
mysim.description = "ECal Non-Fiducial Test Simulation"

#from LDMX.Biasing import util
#mysim.actions.append( util.StepPrinter(1) )

import LDMX.Ecal.vetos as ecal_vetos

p.outputFiles = [f'events_nonfiducial_test_production.root']
p.histogramFile = f'hist_nonfiducial_test_production.root'

p.maxTriesPerEvent = 10000
p.maxEvents = 100
p.totalEvents = 1000
p.run = 2
p.logFrequency = 100
p.termLogLevel = 1

p.sequence=[ mysim,
        ecal_digi.EcalDigiProducer(),
        ecal_digi.EcalRecProducer(),
        ecal_vetos.EcalVetoProcessor()
        ]

from LDMX.DQM import dqm
p.sequence.extend(dqm.ecal_dqm)
```

**Desired behavior**

Just have that print out once after ever N event

**Environment:**
```
denv_workspace="/home/vamitamas/NonFiducialSimu/v4.1.2"
denv_name="v4.1.2"
denv_image="ldmx/pro:v4.1.2"
denv_mounts="/home/vamitamas/NonFiducialSimu/v4.1.2 "
denv_shell="/bin/bash -i"
denv_network="true"
apptainer version 1.2.5
```

tomeichlersmith:
These developments do the following
- interpret the message's logging level into a human readable string for easier parsing
- add event number to logging messages via a custom Formatter which resolves #1301 
- allow logging levels to differ between different channels by implementing a custom Filter which resolves #922 
- update the configuration so that the logging has its own configuration class, keeping its parameters organized and separate from the Process (the current syntax is supported by overriding `__getattr__` so this doesn't require folks to update their configs).
- only printout an event status timestamp on the first try of every event which resolves #1455  

to lower the level for everyone
```python
p.logger.termLevel = 0
# OR, with legacy syntax
p.termLogLevel = 0
```
to debug a specific processor
```python
p.logger.debug(my_processor)
# OR
p.logger.debug("MyProcessorName")
```
to silence a specific processor
```python
p.logger.silence(my_processor)
# OR
p.logger.silence("MyProcessorName")
```
or something in between
```python
p.logger.custom(my_processor, level = 1)
```

## Check List

These developments were tested with a simple standalong processor and config that I've copied here for completeness.

```python
p = ldmxcfg.Process('log')
p.sequence = [ldmxcfg.Producer.from_file('EgLog.cxx')]
p.outputFiles = [ '/dev/null' ]
p.maxEvents = 5
p.logFrequency = 1
p.termLogLevel = 0
p.logger.debug(p.sequence[0])
```

```cpp

class EgLog : public framework::Producer {
 public:
  EgLog(const std::string& name, framework::Process& p)
    : framework::Producer(name, p) {}
  virtual ~EgLog() = default;
  void produce(framework::Event& event) override {
    ldmx_log(info) << "Info message";
    ldmx_log(error) << "Error message";
  }
};

DECLARE_PRODUCER(EgLog);
```

### DIFF
--- a/Framework/src/Framework/Logger.cxx
+++ b/Framework/src/Framework/Logger.cxx
@@ -154,21 +154,19 @@ void Formatter::operator()(const log::record_view& view,
   os << "[ " << log::extract<std::string>("Channel", view) << " ] "
      << event_number_ << " ";
   /**
-   * We _copy_ the value out of the log into our own type
+   * We de-reference the value out of the log into our own type
    * so that we can compare and convert it into a string.
-   * I expect this copying to be okay since its just a
-   * enum (int equivalent), but its good to be clear
    */
-  level msg_level = *log::extract<level>("Severity", view);
+  const level& msg_level = *log::extract<level>("Severity", view);
   switch (msg_level) {
     case level::debug:
       os << "debug";
       break;
     case level::info:
-      os << "info ";
+      os << " info";
       break;
     case level::warn:
-      os << "warn ";
+      os << " warn";
       break;
     case level::error:
       os << "error";
@@ -177,6 +175,12 @@ void Formatter::operator()(const log::record_view& view,
       os << "fatal";
       break;
     default:
+      // this should never happen since the loggers created
+      // with makeLogger use the level enum and thus check
+      // at compile time that the given level is an good option.
+      // That being said, we leave this in case someone creates
+      // their own logger circumventing makeLogger for whatever
+      // reason and changes the enum defining the severity.
       os << "?????";
       break;
   }
--- a/Framework/src/Framework/Logger.cxx
+++ b/Framework/src/Framework/Logger.cxx
@@ -13,7 +13,16 @@ namespace framework {
 
 namespace logging {
 
-level convertLevel(int &iLvl) {
+/**
+ * Convert an integer to the severity level enum
+ *
+ * Any integer below zero will be set to 0 (debug),
+ * and any integer above four will be set to 4 (fatal).
+ *
+ * @param[in] iLvl integer level to be converted
+ * @return converted enum level
+ */
+level convertLevel(int iLvl) {
   if (iLvl < 0)
     iLvl = 0;
   else if (iLvl > 4)
@@ -26,11 +35,14 @@ logger makeLogger(const std::string &name) {
   return boost::move(lg);
 }
 
-void open(const level termLevel, const level fileLevel,
-          const std::string &fileName) {
+void open(const framework::config::Parameters& p) {
   // some helpful types
   typedef sinks::text_ostream_backend ourSinkBack_t;
   typedef sinks::synchronous_sink<ourSinkBack_t> ourSinkFront_t;
+
+  level termLevel{convertLevel(p.getParameter<int>("termLevel", 4))};
+  level fileLevel{convertLevel(p.getParameter<int>("fileLevel", 0))};
+  std::string filePath{p.getParameter<std::string>("filePath", "")};
 
   // allow our logs to access common attributes, the ones availabe are
   //  "LineID"    : counter increments for each record being made (terminal or
@@ -43,11 +55,11 @@ void open(const level termLevel, const level fileLevel,
   boost::shared_ptr<log::core> core = log::core::get();
 
   // file sink is optional
-  //  don't even make it if no fileName is provided
-  if (not fileName.empty()) {
+  //  don't even make it if no filePath is provided
+  if (not filePath.empty()) {
     boost::shared_ptr<ourSinkBack_t> fileBack =
         boost::make_shared<ourSinkBack_t>();
-    fileBack->add_stream(boost::make_shared<std::ofstream>(fileName));
+    fileBack->add_stream(boost::make_shared<std::ofstream>(filePath));
 
     boost::shared_ptr<ourSinkFront_t> fileSink =
         boost::make_shared<ourSinkFront_t>(fileBack);
--- a/Framework/src/Framework/Logger.cxx
+++ b/Framework/src/Framework/Logger.cxx
@@ -55,21 +55,9 @@ void open(const level termLevel, const level fileLevel,
     // this is where the logging level is set
     fileSink->set_filter(log::expressions::attr<level>("Severity") >=
                          fileLevel);
-
-    // TODO change format to something helpful
-    // Currently:
-    //  [ Channel ] int severity : message
-    fileSink->set_formatter([](const log::record_view &view,
-                               log::formatting_ostream &os) {
-      os
-          //                            <<
-          //                            log::extract<boost::date_time::int_adapter>(
-          //                            "TimeStamp" , view )
-          << " [ " << log::extract<std::string>("Channel", view) << " ] "
-          << /*humanReadableLevel.at*/ (log::extract<level>("Severity", view))
-          << " : " << view[log::expressions::smessage];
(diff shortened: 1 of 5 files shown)


tomeichlersmith:
first round of PR Validations: https://github.com/LDMX-Software/ldmx-sw/actions/runs/10890162256/job/30218290716

Framework/src/Framework/Logger.cxx
```diff
@@ -116,6 +142,47 @@ void close() {
...
+  level msg_level = *log::extract<level>("Severity", view);
+  switch (msg_level) {
+    case level::debug:
+      os << "debug";
+      break;
+    case level::info:
+      os << "info ";
+      break;
+    case level::warn:
+      os << "warn ";
```
tvami: Extremely minor suggestion, but speaking of parsing, it could be important
```suggestion
      os << "info";
      break;
    case level::warn:
      os << "warn";
```

tvami (changes requested):
Awesome work resolving several issues! I have a small comment below, otherwise great!

Framework/src/Framework/Logger.cxx
tomeichlersmith: So I left the spaces in on purpose so that the message label is constant width. Do we want to keep that constant width? (And maybe apply a width to the event number and processor names) or not worry about constant width

tvami: I see, well with the `debug` the printout is `debug:` while with the info that becomes `info :` which looked a bit weird. And indeed with the processor name being not the same length I didnt even think of the reason for having a space is the same length. I dont think it's realistic to have a fixed length for the processors, sometimes we have very long names, so I'd just give up on this.

tvami: I guess it's true that it does make it nicer inside a given processor
```
[ ElectronCounter ] 9838 debug: Found 1 electrons (tracks) using input collection TriggerPadTracksY_
[ trigger ] 9838 debug: Got trigger energy cut 1500 for 1 electrons counted in the event.
[ trigger ] 9838 debug: Got trigger energy sum 966.672; and decision is pass = true
[ ecalVeto ] 9839 debug: Straight tracks found (before merge): 0
[ ecalVeto ] 9839 debug: Beginning track merging using 0 tracks
[ ecalVeto ] 9839 debug: Straight tracks found (after merge): 0
[ ecalVeto ] 9839 debug: Finding linreg tracks from 4 hits
[ ecalVeto ] 9839 info :   MIP tracking completed; found 0 straight tracks and 0 lin-reg tracks
[ ecalVeto ] 9839 debug:   The pred > bdtCutVal = true
[ ElectronCounter ] 9839 debug: Found 1 electrons (tracks) using input collection TriggerPadTracksY_
```

tvami: So what if we put the space in front, like ` info` and ` warn`? That will satisfy both the fixed length, and will not introduce a space between the text and the colon

bryngemark: sounds like an easy way out

Framework/src/Framework/Process.cxx
```diff
@@ -484,8 +481,10 @@ void Process::newRun(ldmx::RunHeader &header) {
   if (performance_) performance_->stop(performance::Callback::onNewRun, 0);
 }
 
-bool Process::process(int n, Event &event) const {
-  if ((logFrequency_ != -1) && ((n + 1) % logFrequency_ == 0)) {
+bool Process::process(int n, int n_try, Event &event) const {
+  if ((logFrequency_ != -1) && ((n + 1) % logFrequency_ == 0) && (n_try < 2)) {
+    // only printout event counter if we've enabled log frequency, the event
+    // matches the frequency and we are on the first try
```
bryngemark, line 487: this change reminds me, i've grown used to having to set the termLogLevel to 1 or less to get event number printouts even with logFrequency set to something reasonable (like 10 printouts per job). does that change in the current implementation?

bryngemark (approved):
thanks for finding the time to do this, @tomeichlersmith. long-awaited indeed :) 

when this is in, am i right that we should go through all processors (maybe not that many! but most of mine) which have a separate verbosity functionality and move them to only using the logger's severity?

Framework/src/Framework/Process.cxx
tomeichlersmith: This does not; however, I could implement something to avoid that.

From the logging's perspective, both processors and the `Process` object (and anything that logs) are just "channels" that are sending logging messages. I could lower the Process channel's log level if a logFrequency_ > 0 is provided so that the Process's messages are displayed.

tomeichlersmith:
> when this is in, am i right that we should go through all processors (maybe not that many! but most of mine) which have a separate verbosity functionality and move them to only using the logger's severity?

That is correct @bryngemark ! Instead of `verbose` flags I would imagine that these messages would be put into `debug` and then people could opt-in to view the debug messages for certain processors with
```python
p.logger.debug(processor)
```

I'm open to hearing other ideas on the user interface for this. While I like having the Process and Logger configuration separate on the C++ side, the Python side is very flexible and (as can be seen with `__setattr__`), we can get pretty :magic_wand: magical :magic_wand: with it.

tvami (approved):
Perfect, thanks!

tomeichlersmith:
So the `ecal_pn` sample is failing to run

```
    File "config.py", line 179, in <module>
      recoil_dqm.measurement_collection=digi_recoil.out_collection
  NameError: name 'digi_recoil' is not defined. Did you mean: 'digiRecoil'?
  Configuration Error [ConfigureError] : Problem running python script.
    at /home/runner/work/ldmx-sw/ldmx-sw/Framework/src/Framework/ConfigurePython.cxx:319 in ConfigurePython
```

which was introduced in #1454 

I'm guessing this is what is being referred to by
> we updated the tracking configs

in #1459

tvami:
> I'm guessing this is what is being referred to by

I confirm that, I already have a working version in the branch, but we should have https://github.com/LDMX-Software/ldmx-sw/pull/1460 merged and then I can do the other items in that issue and push it

Issue #1459, opened 2024-09-17
Updates to CI: use just, update tracking, enforce clang-tidy


tvami:
**Is your feature request related to a problem? Please describe.**
 - the CI compilion still uses the old approach
 - we updated the tracking configs
 - after https://github.com/LDMX-Software/ldmx-sw/issues/1176 is resolved, we should enforce those settings for the future code

**Describe the solution you'd like**
 - move CI compile to `just`
 - propagate new tracking config to the CI config too
 - enforce clang-tidy at CI level

**Describe alternatives you've considered**

We could discuss the last point at a software meeting, alternative seems more complicated when we do compile log diffs, in the end the developer should fix those problems

PR #1460, opened 2024-09-17
Part 3 of static analysis: clang-tidy

### LINKED ISSUE #1176: Static analysis meta-issue
Similar to #1166 I'm creating this issue to track warnings from compilers or static analysis.  Generally these aren't urgent to patch (I'll note if/when I think they need more immediate attention) but for future validation/PR workflows reducing the amount of these will allow us to use warnings that currently generate too many false-ish positives (true positives but probably not all that impactful) to be useful. 

To do this I'm currently building a custom version of the container environment with some additional tools (clang, clang-tidy so-far). If this turns out to be something we want to use in the future we could consider adding to the regular container. I'm also adding some additional support for adding additional warnings in the cmake module while working on https://github.com/LDMX-Software/cmake/issues/17. 

Whatever we decide for an approach for #1168 probably should guide if/how we would apply this.

Potential bugs/worth looking into

- [x] `-Wreturn-type` Control reaching the end of non-void functions. 
    - [x] GetGammaProcessManager (SimCore)
    - [x] getVisAttributes (SimCore)
    - [x] rotateGlobalToLocalBarPosition/getScintillatorOrientation (DetDescr)
- [x] `-Wmaybe-uninitialized` Using or recording uninitialized variables can be a real issue. 

Medium priority (fix when convenient)
- [x] Missing virtual for destructors. These are all over the place, mostly in processors, and probably not an issue most of the time. However, if you have a destructor that actually needs to do something then this could be a problem. This seems straight-forward to patch whenever you are touching a corresponding part of the code 
- [x] Shadowed virtual functions (e.g. `virtual void onFileOpen() {}` instead of `virtual void onFileOpen(EventFile &eventFile) {}`
- [x] `-Wunused` Unused variables. Often just left over from development but can also be a mistake. 
- [x] `-Wreorder`: C++ member variables are always initialized in the order they are declared, even if the constructor initializer list has a different order. Usually harmless but if there are any dependencies on the order of parameters then it can be a big problem 

Low priority (might not be worth fixing)
- [x] Signed/unsigned int conversions. These are also all over the place but I don't think these need to be patched in the same way. It comes up a lot when dealing with DetectorIDs (where most things are unsigned internally) 
- [x] Double promotion. This I suspect is even less important than the signed/unsigned int conversions, we can probably just turn this off (note: this is not narrowing from double to float, which could be an issue) 
- [ ]  

Related issues/PRs
- [x] https://github.com/LDMX-Software/ldmx-sw/issues/1208 
- [x] https://github.com/LDMX-Software/Framework/issues/70 
- [x] https://github.com/LDMX-Software/ldmx-sw/issues/1199, https://github.com/LDMX-Software/ldmx-sw/pull/1207
- [x] https://github.com/LDMX-Software/ldmx-sw/issues/1172 

PRs with warning cleanup as part of the PR 
- [x] https://github.com/LDMX-Software/Hcal/pull/70 
- [x] https://github.com/LDMX-Software/TrigScint/pull/54 
- [x] https://github.com/LDMX-Software/SimCore/pull/95

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1176

## Check List

I killed all the warning. 
```
rm -rf build
just configure -DADDITIONAL_WARNINGS=ON  -DENABLE_CLANG_TIDY=ON  -DENABLE_LTO=ON  -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang
just build
// And switch back to a build with GCC
rm -rf build 
just configure -DADDITIONAL_WARNINGS=ON  -DENABLE_CLANG_TIDY=ON  -DENABLE_LTO=ON 
just build
```
is now nice and clean.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,7 +20,7 @@ project(LDMX_SW VERSION 3.1.1
                 LANGUAGES CXX    
 )
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 
 # Load additional macros used by this project. 
 list(APPEND CMAKE_MODULE_PATH ${LDMX_SW_SOURCE_DIR}/cmake/)
@@ -35,6 +35,9 @@ include(BuildMacros RESULT_VARIABLE build_macros_found)
 include_directories(SYSTEM /usr/local/include/root)
 include_directories(SYSTEM Tracking/acts/Core/include)
 include_directories(SYSTEM Trigger/HLS_arbitrary_Precision_Types/include)
+
+# FunctionalCoreTest.cxx doesnt comply with clang-tidy but that's ok
+set_source_files_properties(Framework/test/FunctionalCoreTest.cxx PROPERTIES COMPILE_OPTIONS "-Wno-null-dereference")
 
 # If an install location hasn't been set via CMAKE_INSTALL_PREFIX, set it to 
 # a reasonable default ($PWD/install).  
--- a/Framework/src/Framework/Event.cxx
+++ b/Framework/src/Framework/Event.cxx
@@ -122,7 +122,10 @@ void Event::setInputTree(TTree* tree) {
   // find the names of all the existing branches
   TObjArray* branches = inputTree_->GetListOfBranches();
   if (!branches) {
-    EXCEPTION_RAISE("RunHeaderError", "Branch doesnt exists");
+    EXCEPTION_RAISE(
+        "BadInputFile",
+        "Input tree doesn't have a list of branches but still made it to this "
+        "point. This is bad and has never been seen before!");
     return;
   }
   for (int i = 0; i < branches->GetEntriesFast(); i++) {
--- a/Framework/src/Framework/Event.cxx
+++ b/Framework/src/Framework/Event.cxx
@@ -121,8 +121,15 @@ void Event::setInputTree(TTree* tree) {
 
   // find the names of all the existing branches
   TObjArray* branches = inputTree_->GetListOfBranches();
+  if (!branches) {
+    EXCEPTION_RAISE("RunHeaderError", "Branch doesnt exists");
+    return;
+  }
   for (int i = 0; i < branches->GetEntriesFast(); i++) {
-    std::string brname = branches->At(i)->GetName();
+    std::string brname;
+    if (branches->At(i)) {
+      brname = branches->At(i)->GetName();
+    }
     if (brname != ldmx::EventHeader::BRANCH) {
       size_t j = brname.find("_");
       auto br = dynamic_cast<TBranchElement*>(branches->At(i));
--- a/justfile
+++ b/justfile
@@ -58,14 +58,16 @@ install-denv:
     curl -s https://raw.githubusercontent.com/tomeichlersmith/denv/main/install | sh
 
 # configure how ldmx-sw will be built
+# added ADDITIONAL_WARNINGS and CLANG_TIDY to help improve code quality
 configure *CONFIG:
     denv cmake -B build -S .  -DADDITIONAL_WARNINGS=ON  -DENABLE_CLANG_TIDY=ON  {{ CONFIG }}
 
-configure-force-error *CONFIG:
-    just  configure -DWARNINGS_AS_ERRORS=ON {{ CONFIG }}
+# This is the same as just configure but reports all (non-3rd-party) warnings as errors
+configure-force-error *CONFIG: (configure "-DWARNINGS_AS_ERRORS=ON" CONFIG)
 
-configure-clang-lto:
-    denv cmake -B build -S . -DADDITIONAL_WARNINGS=ON  -DENABLE_CLANG_TIDY=ON  -DENABLE_LTO=ON  -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang 
+# This is an alternative compiler (clang) together with enabling the LTO sanitizer
+# This is very helpful but for now it optimizes away our module linking, use for testing only
+configure-clang-lto: (configure "-DENABLE_LTO=ON -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang") 
 
 # compile and install ldmx-sw
 build ncpu=num_cpus():
--- a/justfile
+++ b/justfile
@@ -59,7 +59,13 @@ install-denv:
 
 # configure how ldmx-sw will be built
 configure *CONFIG:
-    denv cmake -B build -S . {{ CONFIG }}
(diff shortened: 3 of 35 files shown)


danyi211 (approved):

tvami:
Tests ran fine https://github.com/LDMX-Software/ldmx-sw/actions/runs/10894469361
(except for the ecal pn, but that's broken bc of the tracking change, I'll fix it with https://github.com/LDMX-Software/ldmx-sw/issues/1459 tomorrow)

CMakeLists.txt
```diff
@@ -36,6 +36,9 @@ include_directories(SYSTEM /usr/local/include/root)
 include_directories(SYSTEM Tracking/acts/Core/include)
 include_directories(SYSTEM Trigger/HLS_arbitrary_Precision_Types/include)
 
+# FunctionalCoreTest.cxx doesnt comply with clang-tidy but that's ok
+set_source_files_properties(Framework/test/FunctionalCoreTest.cxx PROPERTIES COMPILE_OPTIONS "-Wno-null-dereference")
+
```
tomeichlersmith, line 41: Is it possible to move this into `Framework/CMakeLists.txt`?

Framework/src/Framework/Event.cxx
```diff
@@ -121,8 +121,15 @@ void Event::setInputTree(TTree* tree) {
 
   // find the names of all the existing branches
   TObjArray* branches = inputTree_->GetListOfBranches();
+  if (!branches) {
+    EXCEPTION_RAISE("RunHeaderError", "Branch doesnt exists");
```
tomeichlersmith:
```suggestion
    EXCEPTION_RAISE("BadInputFile", "Input tree doesn't have a list of branches but still made it to this point. This is bad and has never been seen before!");
```

justfile
```diff
@@ -59,7 +59,13 @@ install-denv:
 
 # configure how ldmx-sw will be built
 configure *CONFIG:
-    denv cmake -B build -S . {{ CONFIG }}
+    denv cmake -B build -S .  -DADDITIONAL_WARNINGS=ON  -DENABLE_CLANG_TIDY=ON  {{ CONFIG }}
+
+configure-force-error *CONFIG:
+    just  configure -DWARNINGS_AS_ERRORS=ON {{ CONFIG }}
```
tomeichlersmith:
```suggestion
configure-force-error *CONFIG: (configure -DWARNINGS_AS_ERRORS=ON CONFIG)
```

justfile
```diff
@@ -59,7 +59,13 @@ install-denv:
...
+
+configure-clang-lto:
+    denv cmake -B build -S . -DADDITIONAL_WARNINGS=ON  -DENABLE_CLANG_TIDY=ON  -DENABLE_LTO=ON  -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang
```
tomeichlersmith: Is falling back to `denv` necessary here?

```suggestion
configure-clang-lto: (configure -DENABLE_LTO=ON -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang) 
```

justfile
```diff
@@ -59,7 +59,13 @@ install-denv:
 
 # configure how ldmx-sw will be built
 configure *CONFIG:
-    denv cmake -B build -S . {{ CONFIG }}
+    denv cmake -B build -S .  -DADDITIONAL_WARNINGS=ON  -DENABLE_CLANG_TIDY=ON  {{ CONFIG }}
```
tomeichlersmith, line 63: If we really want these CMake options to be ON by default, maybe we should put them into the CMakeLists.txt like that?

e.g. I think the following would still allow us to disable on the command line but would change the default to `ON` if not specified.

```cmake
if (NOT ADDITIONAL_WARNINGS)
  set(ADDITIONAL_WARNINGS ON)
endif()
```

tomeichlersmith (commented):
C++ changes look good, my comments are mainly focused on passing arguments to recipes within the justfile simply to avoid the recursive call to the program itself and maybe moving default build settings into the CMakeLists.txt.

CMakeLists.txt
tvami: I tried this, it does not work! :(
EDIT: oh maybe it's that I copied the whole thing and not think about the relative path

tvami: Nope, it doesnt work with the below either
```
set_source_files_properties(test/FunctionalCoreTest.cxx PROPERTIES COMPILE_OPTIONS "-Wno-null-dereference")
```
I'm going back to the original

Framework/src/Framework/Event.cxx
tvami: done in the next commit

justfile
tvami: Implemented it as `configure-force-error *CONFIG: (configure "-DWARNINGS_AS_ERRORS=ON" CONFIG)`

tomeichlersmith: We found out that `just` treats `-` as tokens within its own syntax and so we need to wrap the `-DWARNINGS_AS_ERRORS=ON` with `"` to inform `just` that it should be treated as an argument to the recipe and not apart of its own syntax.

justfile
tvami: Done as 
```
configure-clang-lto: (configure "-DENABLE_LTO=ON -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang")
```

justfile
tvami: So it's yes and no. It does seem to work, but gives this other warning
```
CMake Warning (dev) at cmake/CompilerWarnings.cmake:13 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'ADDITIONAL_WARNINGS'.
Call Stack (most recent call first):
  cmake/BuildMacros.cmake:143 (enable_compiler_warnings)
  Recon/CMakeLists.txt:37 (setup_library)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at cmake/StaticAnalysis.cmake:5 (option):
  Policy CMP0077 is not set: option() honors normal variables.  Run "cmake
  --help-policy CMP0077" for policy details.  Use the cmake_policy command to
  set the policy and suppress this warning.

  For compatibility with older versions of CMake, option is clearing the
  normal variable 'ENABLE_CLANG_TIDY'.
Call Stack (most recent call first):
  cmake/BuildMacros.cmake:145 (enable_clang_tidy)
  Recon/CMakeLists.txt:37 (setup_library)
This warning is for project developers.  Use -Wno-dev to suppress it.
```
which makes it pretty ugly... so I'll just keep this as it was. Honestly I think it's more transparent too to have it explicitly like this in the justfile

tomeichlersmith (approved):

PR #1461, opened 2024-09-17
Implementation of TS clustering and tracking as envisioned in LUT firmware.

rodwyer100:
This push request takes the validated HLS firmware code and reincorporates it into the ldmx-sw structure. I have included in TrigScintFirmwareProducer a producer which takers clusterproducer_sw and trackproducer_sw (faithful reproductions of the code used to implement tracking in HLS), stages a collection of digis for insertion, and then obtains a collection of tracks for use thereafter. The producer as of yet does not save the collected digis into the track object, nor does it produces clusters, but it does faithfully reproduce the tracks of the existing horizontal geometry. There are two caveats with this:

1. It completely reproduces ldmx reconstruction results with no inefficiency in HLS because HLS does a better job at staging the rather poorly behaves ap_int classes it uses. I had some initial issues getting this code to be 100% efficient (its 98% efficient now) because ap_int behaves irregularly in how its implemented here.
2. This Firmware design was used to implement the original horiztonal 3 pad geometry (ldmx-det-v14) and does not have a LUT that corresponds to vertical bars. If this is desired, one could envision changing trackproducer_hw and testing it, but it will require further work.

I validated that the producer made the same number of tracks using print out statements like this:
10 10 10
18 18 18
29 29 30
 [ ElectronCounter ] 0 : Found 3 electrons (tracks) using input collection TriggerPadTracks_sim
I have not validated this code as extensively as the firmware version, so it could presumably make tracks from different clusters but my intuition about how the code works tells me that is very unlikely.

Here is a picture of the track number distribution for both:
Track no for original reconstruction:
image
Track no for firmware reconstruction:
image
If any more plots are necessary please lmk


tvami:
> I am adding some includes in /includes/ which are necessary for the primitive non floating point objects that HLS uses to create firmware from C++ files. 

How are these files different from what we have in `Trigger/HLS_arbitrary_Precision_Types/include`? If they are not you should not add them again, but rather reuse them

TrigScint/exampleConfigs/firmwareEx.py
```diff
@@ -0,0 +1,175 @@
...
+tsDigisDown.input_pass_name = "sim"
+
+tsClustersTag  =TrigScintClusterProducer.pad2()
+tsClustersUp  =TrigScintClusterProducer.pad1()
+tsClustersDown  =TrigScintClusterProducer.pad3()
+
+if "v12" in version :
+     tsClustersTag.pad_time = -2.
+     tsClustersUp.pad_time = 0.
+     tsClustersDown.pad_time = 0.
```
tvami:
```suggestion
```

TrigScint/exampleConfigs/firmwareEx.py
```diff
@@ -0,0 +1,175 @@
...
+
+from LDMX.TrigScint.trigScint import TrigScintDigiProducer
+from LDMX.TrigScint.trigScint import TrigScintClusterProducer
+from LDMX.TrigScint.trigScint import trigScintTrack
+from LDMX.TrigScint.trigScint import TrigScintFirmwareTracker
+
+if "v12" in version :
+     tsSimColls=[ "TriggerPadTagSimHits", "TriggerPadUpSimHits", "TriggerPadDnSimHits" ]
+else :
+     tsSimColls=[ "TriggerPad2SimHits", "TriggerPad3SimHits", "TriggerPad1SimHits" ]
```
tvami: Is any of the v12 stuff still relevant?
```suggestion
tsSimColls=[ "TriggerPad2SimHits", "TriggerPad3SimHits", "TriggerPad1SimHits" ]
```

TrigScint/include/TrigScint/TrigScintFirmwareTracker.h
```diff
@@ -0,0 +1,124 @@
...
+
+  ldmx::TrigScintTrack makeTrack(Track outTrk);
+
+  /**
+   * add a hit at index idx to a cluster
+   */
+
+  void onProcessStart() override;
+
+  void onProcessEnd() override;
```
tvami:
```suggestion
```

TrigScint/src/TrigScint/TrigScintFirmwareTracker.cxx
```diff
@@ -0,0 +1,290 @@
...
+  std::cout<<"I GOT HERE 2"<<std::endl;
+  // TODO remove this once verified that the noise overlap bug is gone
+
+  int occupied[50];
+  for(int i = 0; i<50;i++){
+    occupied[i]=-1;
+  }
+  int count=0;
+  for (const auto &digi : digis1_) {
+    std::cout<<"I AM LOOPING IN DIGIS1 "<<digi.getPE()<<" "<<digi.getTime()<<std::endl;
```
tvami: Please move all the `cout`s to `ldmx_log` (probably debug level)
There are other `cout`s elsewhere so my comment here is rather general

TrigScint/src/TrigScint/TrigScintFirmwareTracker.cxx
```diff
@@ -0,0 +1,290 @@
...
+
+  return;
+}
+
+void TrigScintFirmwareTracker::onProcessEnd() {
+  ldmx_log(debug) << "Process ends!";
+
+  return;
+}
+
```
tvami:
```suggestion
```

TrigScint/src/TrigScint/TrigScintFirmwareTracker.cxx
```diff
@@ -0,0 +1,290 @@
...
+  //std::cout<<"I GOT HERE 1"<<std::endl;
+  const auto digis1_{
+     event.getCollection<ldmx::TrigScintHit>(digis1_collection_, passName_)};
+  const auto digis3_{
+     event.getCollection<ldmx::TrigScintHit>(digis2_collection_, passName_)};
+  const auto digis2_{
+     event.getCollection<ldmx::TrigScintHit>(digis3_collection_, passName_)};
+
+  std::cout<<"I GOT HERE 2"<<std::endl;
+  // TODO remove this once verified that the noise overlap bug is gone
```
tvami: Is this bug gone?

TrigScint/src/TrigScint/TrigScintFirmwareTracker.cxx
```diff
@@ -0,0 +1,290 @@
...
+	HPad1[count].Amp=(ap_int<12>)(digi.getPE());
+	HPad1[count].Time=(ap_int<12>)(digi.getTime());
+	occupied[digi.getBarID()]=count;
+	count++;
+      }
+    }
+  }
+
+  std::cout<<"I GOT HERE 3"<<std::endl;
+  for(int i = 0; i<50;i++){
```
tvami: please make the 50 a const expression, or a configurable if it is expected to change.

TrigScint/src/TrigScint/TrigScintFirmwareTracker.cxx
```diff
@@ -0,0 +1,290 @@
...
+	topSeed=Point3[i].Seed.bID;
+      }
+    }
+  }
+
+   if (verbose_) {
+    ldmx_log(debug) << "Got digi collection " << digis1_collection_ << "_"
+                    << passName_ << " with " << digis1_.size() << " entries ";
+  }
+  std::cout<<"I GOT HERE 5"<<std::endl;
```
tvami: There is a few of these, please delete them
```suggestion
```

TrigScint/src/TrigScint/clusterproducer_sw.cxx
```diff
@@ -0,0 +1,77 @@
...
+#include "TrigScint/clusterproducer.h"
+
+Cluster* clusterproducer_sw(Hit inHit[NHITS]){
+
+	ap_int<12> SEEDTHR = 30;
+	ap_int<12> CLUSTHR = 30;
+
+	ap_int<12> mapL1[NCHAN];
+
+	Cluster* outClus = new Cluster[NCLUS];
```
tvami: can this become a smart pointer instead? I dont see any `delete`'s anywhere, so I'm worried for a memory leak

TrigScint/src/TrigScint/trackproducer_hw.cxx
```diff
@@ -0,0 +1,98 @@
...
+
+
+void trackproducer_hw(Cluster Pad1[NTRK],Cluster Pad2[NCLUS],Cluster Pad3[NCLUS],Track outTrk[NTRK],ap_int<12> lookup[NCENT][COMBO][2]){
+	#pragma HLS ARRAY_PARTITION variable=Pad1 dim=0 complete
+	#pragma HLS ARRAY_PARTITION variable=Pad2 dim=0 complete
+	#pragma HLS ARRAY_PARTITION variable=Pad3 dim=0 complete
+	#pragma HLS ARRAY_PARTITION variable=outTrk dim=0 complete
+	#pragma HLS ARRAY_PARTITION variable=lookup dim=0 complete
+	#pragma HLS PIPELINE II=10
+	std::cout<<"HELLO 2 HERE 1"<<std::endl;
```
tvami: This file needs to be cleaned up too with all the HELLOs
```suggestion
```

scripts/ldmx-env.sh
```diff
@@ -198,7 +198,6 @@ if hash docker &> /dev/null; then
       -e LDMX_BASE \
       -e DISPLAY=${LDMX_CONTAINER_DISPLAY}:0 \
       $_envs \
-      -v /tmp/.X11-unix:/tmp/.X11-unix \
```
tvami: why was this changed needed?

tvami (changes requested):
I have a few smaller comments, please generally clean up the files, and you'll also have to run `just format-cpp` to have the clang formatting. I can also run the other sanitizers tomorrow for more comments, or maybe Tom will catch those before I run it tomorrow.

scripts/ldmx-env.sh
rodwyer100: This change is necessary for anyone running on s3df. I had to use it but I can remove it.

TrigScint/exampleConfigs/firmwareEx.py
rodwyer100: No. I adopted an old working script here. Any naming convention works for me.

TrigScint/src/TrigScint/TrigScintFirmwareTracker.cxx
rodwyer100: Done.

TrigScint/src/TrigScint/trackproducer_hw.cxx
rodwyer100: done.

TrigScint/src/TrigScint/TrigScintFirmwareTracker.cxx
rodwyer100: done

rodwyer100:
We should talk about how ap_int is implemented in the ldmx-sw framework. I know TS in the Trigger repository doesn't use it, and I suspect it had something to do with the <*> in the fixed point library requiring ap_int to be found as a standard directory (and therefore not being found). While there are a couple repositories in the Ecal Trigger repositories that use it I am not certain how Christian managed to work with the relative paths indicated by the use of quotation marks and still have it built with <*>. That doesn't seem to work for me, though Im happy to take suggestions. I have implemented a fix that should work for both of us, and that is that in my local copy I have all of the includes of the HLS objects switched (wherever relevant) from <> to "". The issue is that since these are contained in a submodule we don't have access to, I can push that version which would work for us both. I know Tom worked on dereferening TrigScint and I am wondering if that is called for here, or if there is something better one should do. Otherwise I am not certain how to proceed on that front.

TrigScint/src/TrigScint/clusterproducer_sw.cxx
rodwyer100: I'll try to do that :). Gonna hit the sack for tonight though!

TrigScint/python/trigScint.py
```diff
@@ -285,6 +285,42 @@ def __init__(self,name) :
...
+        self.input_pass_name="" #take any pass
+        self.output_collection="TriggerPadTracks"
+        self.verbosity = 0
+
+class TrigScintFirmwareTracker(ldmxcfg.Producer) :
+    """Configuration for the track producer from the Firmware Tracker"""
+    def __init__(self,name) :
+        super().__init__(name,'trigscint::TrigScintFirmwareTracker','TrigScint')
+        self.clustering_threshold=40.0
+        self.digis1_collection='trigScintDigisPad1'
```
bryngemark, line 315: should these be the QIEDigis?

bryngemark (commented):
I have a general structure/philosophy comment/question. First, are these meant to be used in the HLS implementation or in sw clustering/tracking for comparison to HLS? If the former, I wonder if we should put them in some separate directory (be it within `TrigScint` or `Trigger`, I don't know). If the latter, I'd like the producer naming to be more explicit. You do use one explicit name "`TrigScintFirmwareTracker`" and that model is great. Overall it would be nice if you follow e.g. the capitalization conventions we're used to.

I would also appreciate if you commented the code more, so the next person can follow what's going on.

scripts/ldmx-env.sh
tvami: Ah but there is a trick you can do on s3df, which you'll only have to do once per iana node you are on (so maybe twice ever? I usually end up on  006): 
```
touch /tmp/.X11-unix
```
and that's it

tvami:
> The issue is that since these are contained in a submodule we don't have access to,

I'm confused by this statement. Doesn't this mean you just forgot the "recursive" when cloning in `git clone --recursive git@github.com:LDMX-Software/ldmx-sw.git` ? It should really be available like that

scripts/ldmx-env.sh
tomeichlersmith: (This is one of a few small usability issues that are resolved with `denv` as well so please look at moving your workflow to using that program instead of `scripts/ldmx-env.sh`. https://ldmx-software.github.io/developing/quick-switch.html)

tomeichlersmith:
Christian got the relative path includes working by telling CMake where to find the necessary headers.

https://github.com/LDMX-Software/ldmx-sw/blob/77fffe17e87e4237d9aaa106d9c343fc7f805947/Trigger/Algo/CMakeLists.txt#L39

This type of thing would be necessary in TrigScint as well if the `ap_*` headers are of use. Without a line like this in the CMake, there would be no way for the compiler to figure out where the headers are and thus would emit a `unable to find ap_XXX.h` style compiler error.

tomeichlersmith:
I did a quick test and this branch still compiles after the following changes.
```
git rm -r TrigScint/include/TrigScint/etc
```
```diff
diff --git a/TrigScint/CMakeLists.txt b/TrigScint/CMakeLists.txt
index 2d1a8473..05382878 100644
--- a/TrigScint/CMakeLists.txt
+++ b/TrigScint/CMakeLists.txt
@@ -44,6 +44,7 @@ setup_library(module TrigScint
               dependencies Framework::Framework Recon::Event DetDescr::DetDescr
                            Tools::Tools SimCore::Event
 )
+target_include_directories(TrigScint PUBLIC ../Trigger/HLS_arbitrary_Precision_Types/include)
 
 setup_python(package_name LDMX/TrigScint)
 
diff --git a/TrigScint/include/TrigScint/objdef.h b/TrigScint/include/TrigScint/objdef.h
index 230cf112..b9706db9 100755
--- a/TrigScint/include/TrigScint/objdef.h
+++ b/TrigScint/include/TrigScint/objdef.h
@@ -1,7 +1,7 @@
 #ifndef OBJDEF_H
 #define OBJDEF_H
 
-#include "../../../Trigger/HLS_arbitrary_Precision_Types/include/ap_int.h"
+#include "ap_int.h"
 #define NTIMES 6
 #define NHITS 25
 #define NCLUS 25
```

TrigScint/python/trigScint.py
rodwyer100: I will comment the code more thoroughly for sure :)! I am including this code for two reasons. The first is that Elizabeth needs the code for efficiency studies to reflect firmware stuff. The other reason is along the lines of having firmware directly reflect software, and since the firmware uses pattern matching so will the software. This processor does both clustering and tracking in one step, using clusterproducer_sw and trackerproducer_hw to do those steps respectively. These files are the raw ones that are the validated pieces of HLS code, the processor itself just stages the digis. If I wanted to include the hitmaker I could as well, in which case I guess the input would be adc and tdc stuff from QIEDecoder, but I haven't done that just yet.

rodwyer100:
I have added every change (smart pointers, general adoption of existing capitalization structure, etc) with the exception of clang formatting. I am trying to get my clang formatter working, but s3df doesnt allow for sudo so I have to git clone it and make it and Im on hour 6 or 7 of that. So hopefully it will all be done tomorrow. Please lmk if the chosen way of aligning with the capitalization structure is sufficient!

tvami:
@rodwyer100 after we merge https://github.com/LDMX-Software/ldmx-sw/pull/1464 the clang format will automatically be applied if it's too much work for you. (Tho I use s3df too and `just format-cpp` works really nicely w/o any sudo needs)

scripts/ldmx-env.sh
tvami: @rodwyer100 please revert the changes in this file

rodwyer100:
The files have been clang formatted. Lmk if you think it needs anything else :)!

tvami:
> The files have been clang formatted. Lmk if you think it needs anything else :)!

The only thing missing from my review is https://github.com/LDMX-Software/ldmx-sw/pull/1461#discussion_r1764339136

rodwyer100:
Whoopsie just saw your comment. Fixed that change now. Thanks so much!

tvami (approved):
It's good from my side now, one more thing to change is the PR description, now that's not in line with the changes since you didnt move the headers in there, please modify the text accordingly

TrigScint/include/TrigScint/Firmware/trackproducer.h
```diff
@@ -0,0 +1,15 @@
...
+#include "objdef.h"
+
+void copyCluster1(Cluster One, Cluster Two);
+void copyCluster2(Cluster One, Cluster Two);
+void trackproducer_ref(Cluster Pad1[NTRK], Cluster Pad2[NCLUS],
+                       Cluster Pad3[NCLUS], Track outTrk[NTRK],
+                       ap_int<12> lookup[NCENT][COMBO][2]);
+void trackproducer_hw(Cluster Pad1[NTRK], Cluster Pad2[NCLUS],
+                      Cluster Pad3[NCLUS], Track outTrk[NTRK],
+                      ap_int<12> lookup[NCENT][COMBO][2]);
```
tomeichlersmith, line 13: It feels weird that Pad1 would have length `NTRK` while the other `Cluster` arrays would have length `NCLUS`. Is this on purpose?

TrigScint/src/TrigScint/Firmware/clusterproducer_sw.cxx
```diff
@@ -0,0 +1,77 @@
+#include <stdio.h>
+
+#include <iostream>
+
+#include "TrigScint/Firmware/clusterproducer.h"
+#include "TrigScint/Firmware/objdef.h"
+
+std::unique_ptr<Cluster[]> clusterproducer_sw(Hit inHit[NHITS]) {
```
tomeichlersmith: What you want to use is `std::array<Cluster, NCLUS>` since you know the length of the array at compile time. This allows you to pass around C-style, fixed-length arrays in C++ without needing to have bare pointers or any `new`.

tomeichlersmith (changes requested):
I have one technical question and a technical change request.

I haven't looked through to understand the algorithm, but that's what future testing and bug reports are for I suppose.

TrigScint/include/TrigScint/Firmware/trackproducer.h
rodwyer100: Yes. The NTRK should be able to be smaller ultimately because we dont have the bandwidth in firmware to read out every track if there are more than 5 or so in a 25 ns clock cycle. The way the firmware handles this is that it handles the first cluster a little differently throughout, allowing it to read in a smaller number of clusters there. The number of tracks is less than or equal to the number of pad1 clusters and therefore the track max is inbuilt by a pad1 cluster max.

rodwyer100: I think its set to 10, which is way more clusters than we expect in the ldmx environment on a single layer but it is still handleable by our firmware. This processor will try to model the firmware precisely.

tomeichlersmith: That makes sense, can you add a comment explaining this so future developers don't think its a typo and replace it?

tvami:
(not directly connected to this PR, but I'm happy to see the bot is doing the clang update as expected, Rory dont be surprised for this extra commit in your branch)
Screenshot 2024-09-18 at 11 12 07

TrigScint/src/TrigScint/Firmware/clusterproducer_sw.cxx
tomeichlersmith: This is what I'm imagining: https://godbolt.org/z/zbYzxKPzc

You'll have to manually enforce the shortening of `NCLUS` to `NTRK` but I think that is beneficial since it points out to the user a design of the firmware that we are intentionally emulating in the software.

tomeichlersmith (approved):
Thank you :) looks good

PR #1462, opened 2024-09-17
Update tracking CI, add new CI action for LTO and modify other actions to use just

### LINKED ISSUE #1459: Updates to CI: use just, update tracking, enforce clang-tidy

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1459

## Check List

Well the test of the pudding is to see what github does with these actions:
 * it does work according to the expectation


tomeichlersmith (commented):
This is great :) definitely a step in the right direction.

I don't really have any notes except a question. Do you think it would be helpful to summarize the default ci building in a just recipe to avoid copying it around?

```
[private]
ci-build: install-denv init (configure-force-error) build
```
(The parentheses help inform `just` that we aren't giving any more arguments to the configure step which would normally swallow all following arguments as build configuration.)

Then the CI calls would be
```yaml
run: just ci-build
```

might be helpful, might not be.

tvami:
Hmm, I dont know, I see your point that this would certainly be much more compact, but I kinda like that it's spelled out, just for making it more obvious. And also it's more segmented right now in case some of the prev steps failed. So I'm kinda leaning forward keeping it as is.

tvami:
Compilation part looks good! Running the pn validation here
https://github.com/LDMX-Software/ldmx-sw/actions/runs/10911000232

tvami:
> Compilation part looks good! Running the pn validation here https://github.com/LDMX-Software/ldmx-sw/actions/runs/10911000232

Yaay, ecalPN is back to work!

danyi211 (approved):

.github/workflows/pr_validation.yml
```diff
@@ -23,10 +23,18 @@ jobs:
...
+ 
     - name: Compile and Install ldmx-sw
-      uses: ./.github/actions/setup
+      run: |
+        just install-denv init configure-force-error
+        just build
+      shell: bash
 
-    - name: Git Gold Histograms from trunk
+    - name: Get Gold Histograms from trunk
```
tomeichlersmith: This mis-spelling was on purpose ;) but I recognize that might be confusing for some

tomeichlersmith (approved):

.github/workflows/pr_validation.yml
tvami: I can put it back, I just did that change so there is a commit to trigger tests

PR #1463, opened 2024-09-17
dereference boost.log's attribute pointer only if defined

tomeichlersmith:
this avoids clang-tidy complaining about de-referencing nullptrs since
there is a possibility the logs will contain messages that don't have a
Channel and/or Severity attribute

## Check List

### DIFF
--- a/Framework/src/Framework/Logger.cxx
+++ b/Framework/src/Framework/Logger.cxx
@@ -35,6 +35,16 @@ logger makeLogger(const std::string& name) {
   return boost::move(lg);
 }
 
+template <typename T>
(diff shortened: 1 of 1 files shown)


tvami:
Let me link this to being the follow-up to https://github.com/LDMX-Software/ldmx-sw/pull/1458

tomeichlersmith:
Running a more strict build locally to align with #1462 

```
just clean
just configure-force-error
just build $(nproc) test
```

... taking a long time since my tiny computer is tiny ...

tvami (approved):
Thanks for the quick fix for this! I checked that it works with the compiler and also tested the functionality. All good!

PR #1466, opened 2024-09-18
Add condition for not being on trunk when pushing in format CI

tvami:

Format check fails on trunk
https://github.com/LDMX-Software/ldmx-sw/actions/runs/10923712245/job/30321004467
because there is nothing to commit (which I tested already, and works) BUT now we are also not on a PR branch but on `trunk` (this I couldn't test until the prev PR was merged).

## Check List

Post-merge comment: indeed push was skipped on trunk https://github.com/LDMX-Software/ldmx-sw/actions/runs/10924826800/job/30324837501

### DIFF
--- a/.github/workflows/format-check.yml
+++ b/.github/workflows/format-check.yml
@@ -37,7 +37,7 @@ jobs:
       id: commit
 
     - name: Push changes
-      if: steps.commit.outcome == 'success'
+      if: github.ref != 'refs/heads/trunk' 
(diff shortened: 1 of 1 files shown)


tomeichlersmith (approved):
Going to use my [unlimited power](https://www.youtube.com/watch?v=SWwFogRQVnk) :tm: to merge this since we can only really test this on trunk.

PR #1469, opened 2024-09-20
Change default trigger thresholds

### LINKED ISSUE #1468: Change trigger threshold defaults

https://github.com/LDMX-Software/ldmx-sw/blob/c1164a4452c210840cec065115b2b3bcecbf62ed/Recon/python/simpleTrigger.py#L43
```
 self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
```
leads to 3000, 10000, 17000, 24200, while
@EBerzin 's studies lead to different numbers
https://indico.fnal.gov/event/64932/#25-multi-electron-trigger-thre
(although I think Elizabeth was looking at the 34 layers of ECAL, which is not the same as the one we have )
 
Agree on the final numbers and have them in ldmx-sw.
I should add that I strongly believe we should keep the 1e threshold at 3000, so we dont need to redo all the 1e studies (trig skim, BDT, etc), even if it's not fully optimal.

EBerzin:

This resolves #1468. I've updated the 8 GeV default thresholds to reflect the optimized values calculated for a mu=1 Poisson beam profile, and 20 ECal layers. The 1e threshold was kept at 3 GeV, which also happens to be consistent with the optimized value (3.01 GeV).

Thresholds for 4 GeV were also updated to reflect thresholds calculated by Megan, as reported here: https://indico.fnal.gov/event/51366/contributions/225984/attachments/148282/190432/MeganLoh_multielectronTriggering.pdf

The 1e- threshold at 4 GeV was kept at 1.5 GeV.

## Check List

### DIFF
--- a/Recon/python/simpleTrigger.py
+++ b/Recon/python/simpleTrigger.py
@@ -41,7 +41,7 @@ class TriggerProcessor(ldmxcfg.Producer) :
           
         self.beamEnergy = beamEnergy
         if (self.beamEnergy == 4000.): 
-            self.thresholds = [ 3000., 10000., 17000., 24200. ]
+            self.thresholds = [ 1500., 5000., 8500., 12100. ]
         else:
             self.thresholds = [ 3000., 10790., 18540., 26250. ]
         self.mode = 0
--- a/Recon/python/simpleTrigger.py
+++ b/Recon/python/simpleTrigger.py
@@ -41,7 +41,7 @@ class TriggerProcessor(ldmxcfg.Producer) :
           
         self.beamEnergy = beamEnergy
         if (self.beamEnergy == 4000.): 
-            self.thresholds = [ 3000, 10000, 17000, 24200 ]
+            self.thresholds = [ 3000., 10000., 17000., 24200. ]
         else:
             self.thresholds = [ 3000., 10790., 18540., 26250. ]
         self.mode = 0
--- a/Recon/python/simpleTrigger.py
+++ b/Recon/python/simpleTrigger.py
@@ -41,7 +41,7 @@ class TriggerProcessor(ldmxcfg.Producer) :
           
         self.beamEnergy = beamEnergy
         if (self.beamEnergy == 4000.): 
-            self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
+            self.thresholds = [ 3000, 10000, 17000, 24200 ]
         else:
             self.thresholds = [ 3000., 10790., 18540., 26250. ]
         self.mode = 0
--- a/Recon/python/simpleTrigger.py
+++ b/Recon/python/simpleTrigger.py
@@ -40,7 +40,10 @@ class TriggerProcessor(ldmxcfg.Producer) :
         super().__init__(name,'recon::TriggerProcessor','Recon')
           
         self.beamEnergy = beamEnergy
-        self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
+        if (self.beamEnergy == 4000.): 
+            self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
+        else:
+            self.thresholds = [ 3000., 10790., 18540., 26250. ]
         self.mode = 0
         self.start_layer = 0
         self.end_layer = 20
--- a/Recon/python/simpleTrigger.py
+++ b/Recon/python/simpleTrigger.py
@@ -41,7 +41,7 @@ class TriggerProcessor(ldmxcfg.Producer) :
           
         self.beamEnergy = beamEnergy
         if (self.beamEnergy == 4000.): 
-            self.thresholds = [ 1500., 5000., 8500., 12100. ]
+            self.thresholds = [ 1500., 5000., 8200., 11800. ]
         else:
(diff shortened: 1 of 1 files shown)


Recon/python/simpleTrigger.py
```diff
@@ -40,7 +40,10 @@ def __init__(self, name, beamEnergy) :
         super().__init__(name,'recon::TriggerProcessor','Recon')
           
         self.beamEnergy = beamEnergy
-        self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
+        if (self.beamEnergy == 4000.): 
+            self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
```
tvami: At this point, I'd also not have this as a function of energy but have the direct numbers. The reason for having the energy there was because we wanted to hand both energy case in one line, but now that we are moving away from it...
```suggestion
            self.thresholds = [ 3000, 10000, 17000, 24200 ]
```

tvami (changes requested):
Thanks, I have one small comment:

Recon/python/simpleTrigger.py
bryngemark: these defaults were never exact at any energy so if they're a function of energy or not doesn't matter too much. if we want the default numbers to be hardwired to what elizabeth has found then that's fine; if we wanted to have the 4 GeV defaults they should also be hardwired to what megan found a few years ago. 

one thing that could be interesting is to add different defaults for 20 or 34 layers.

Recon/python/simpleTrigger.py
```diff
@@ -40,7 +40,10 @@ def __init__(self, name, beamEnergy) :
         super().__init__(name,'recon::TriggerProcessor','Recon')
           
         self.beamEnergy = beamEnergy
-        self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
+        if (self.beamEnergy == 4000.): 
+            self.thresholds = [ 3000, 10000, 17000, 24200 ]
```
tvami:
```suggestion
            self.thresholds = [ 3000., 10000., 17000., 24200. ]
```
sorry I think these have to be floats

tvami (approved):

Recon/python/simpleTrigger.py
```diff
@@ -40,7 +40,10 @@ def __init__(self, name, beamEnergy) :
         super().__init__(name,'recon::TriggerProcessor','Recon')
           
         self.beamEnergy = beamEnergy
-        self.thresholds = [ self.beamEnergy/4000.*1500.0, self.beamEnergy/4000.*1000. + self.beamEnergy, self.beamEnergy/4000.*500. + 2*self.beamEnergy, self.beamEnergy/4000.*100. + 3*self.beamEnergy ]
+        if (self.beamEnergy == 4000.): 
+            self.thresholds = [ 3000., 10000., 17000., 24200. ]
```
bryngemark: if these are for 4 GeV, they should be 1500, ~5000, ~8500...

bryngemark (changes requested):

Recon/python/simpleTrigger.py
tvami:
```suggestion
            self.thresholds = [ 1500., 5000., 8500., 12100. ]
```
ups, indeed!

Recon/python/simpleTrigger.py
EBerzin: If we were to have different thresholds for 20 vs. 34 layers, would we just use one set of thresholds for <= 20 layers and another for > 20 layers? Also, is this the most recent version of Megan's thresholds? [https://indico.fnal.gov/event/51366/contributions/225984/attachments/148282/190432/MeganLoh_multielectronTriggering.pdf](url)

tvami: > one thing that could be interesting is to add different defaults for 20 or 34 layers.

This, we should do in another PR. 

As for the numbers for 4 GeV, I have the same request to keep 1e as it is for consistency. Based on that talk, 2e should be at 5, 3e at 8.2, 4e at 11.8, right? @EBerzin will you push that change?

tvami:
Thanks @EBerzin !

tvami:
> Thanks @EBerzin !

Actually one more thing: can you please edit the PR description too to reflect you also changed the 4 GeV settings (with a link to the talk from Megan)

tvami:
I'd like to trigger the full test, for that I'll move this to "draft" and back to "ready"

tvami:
Screenshot 2024-09-26 at 14 23 56

OK the only plot failing is this. I think it's expected since that's the only PU workflow we have, and the change on the plot is exactly what we change. It's actually even a bit surprising that it didnt have more visible effects downstream.

tomeichlersmith (approved):

bryngemark:
there are a few places where the trigger decision differs but I guess either a) there are no trigger pass/fail plots in the DQM b) the KS test still judges the distributions to be "from the same underlying distribution"

examples from the log diff:

```
  < [ trigger ] 68 debug: Got trigger energy cut 10000 for 2 electrons counted in the event.
  < [ trigger ] 68 debug: Got trigger energy sum 10147.9; and decision is pass = false
  ---
  > [ trigger ] 68 debug: Got trigger energy cut 10790 for 2 electrons counted in the event.
  > [ trigger ] 68 debug: Got trigger energy sum 10147.9; and decision is pass = true
[...]
  < [ trigger ] 334 debug: Got trigger energy cut 10000 for 2 electrons counted in the event.
  < [ trigger ] 334 debug: Got trigger energy sum 10210.8; and decision is pass = false
  ---
  > [ trigger ] 334 debug: Got trigger energy cut 10790 for 2 electrons counted in the event.
  > [ trigger ] 334 debug: Got trigger energy sum 10210.8; and decision is pass = true
```

PR #1473, opened 2024-09-20
Committing TS Firmware Hit Reconstruction Stagger for the Purpose of Triggering Studies

rodwyer100:
In my last pull request, I added the validated processor which stagged analysis digi objects to be inserted into our most current version of cluster reconstruction. In this pull request, I am adding the validate processor which inserts the most currect  validated version of hit reconstruction into the analysis chain so that Elizabeth may play with it. It is a little different from the S30XL implementation primarily in that its input FIFO pragmas mean that it would need to run around 5 clock cycles as written (though LK said we would likely integrate over 3) and would need 5 copies to run continuously; the S30XL runs continuously and if we were to do that here the pragma structure would need to be altered a little bit.

This is all unnecessary detail, just note that this is the first validated piece of hit making firmware designed for the full LDMX environment that emulates LK's original manner of reconstruction. This is done in TrigScintFirmwareHitProducer, which takes in QIEDigis and outputs TrigScintHit objects. The processor will ultimately be adopted to emulate the NumericalHitProcessor of Andrew and Niramay as that seems the easiest to reproduce in something that can run in firmware. I am including it here because it is the "official" processor I will use to emulate in firmware. If it needs some refashioning or being made pretty, please lmk. I can also include it in a latter pull request alongside the firmware pileUp processor if necessary.

## Check List

Here are proofs that the firmware emulates the software hits (from QIE adcs and tdcs) faithfully. Note that our ap_int resolution here means that if you fiddle with the gains (i.e. make them bigger) you will note get the right hit PE amplitudes:
SAMPLE OUTPUT:
Analysis barID: 2, PE Number: 59.4437
Analysis barID: 3, PE Number: 97.9569
Analysis barID: 22, PE Number: 79.7444
Analysis barID: 23, PE Number: 89.955
Analysis barID: 35, PE Number: 158.878
Analysis barID: 36, PE Number: 61.3231
DID I GET HERE 3
DID I GET HERE 4
DID I GET HERE 5
Firmware barID: 2, PE Number: 59
Firmware barID: 3, PE Number: 98
Firmware barID: 22, PE Number: 79
Firmware barID: 23, PE Number: 90
Firmware barID: 35, PE Number: 159
Firmware barID: 36, PE Number: 61
Firmware barID: 49, PE Number: 8
DID I GET HERE 6
Note that the reason there is an extra hit printed in the Firmware in bar 49 in this case is because I didn't print analysis hits with amplitude below 30. I can fix that, but I have seen enough of the output to know that the firmware is replicating the software appropriattely.

### DIFF
--- a/TrigScint/include/TrigScint/Firmware/hitproducer.h
+++ b/TrigScint/include/TrigScint/Firmware/hitproducer.h
@@ -1,0 +1,13 @@
+#ifndef HITPRODUCER_H
+#define HITPRODUCER_H
+
+#include "objdef.h"
+
+void copyHit1(Hit One, Hit Two);
+void copyHit2(Hit One, Hit Two);
+void hitproducer_ref(ap_uint<14> FIFO[NHITS][5], Hit outHit[NHITS],
+                     ap_uint<8> Peds[NHITS]);
+void hitproducer_hw(ap_uint<14> FIFO[NHITS][5], Hit outHit[NHITS],
+                    ap_uint<8> Peds[NHITS]);
+
+#endif
--- a/TrigScint/include/TrigScint/Firmware/hitproducer.h
+++ b/TrigScint/include/TrigScint/Firmware/hitproducer.h
@@ -3,10 +3,12 @@
 
 #include "objdef.h"
 
+#ifdef TS_NOT_EMULATION
 void copyHit1(Hit One, Hit Two);
 void copyHit2(Hit One, Hit Two);
 void hitproducer_ref(ap_uint<14> FIFO[NHITS][5], Hit outHit[NHITS],
                      ap_uint<8> Peds[NHITS]);
+#endif
 void hitproducer_hw(ap_uint<14> FIFO[NHITS][5], Hit outHit[NHITS],
                     ap_uint<8> Peds[NHITS]);
 
--- a/TrigScint/include/TrigScint/TrigScintFirmwareHitProducer.h
+++ b/TrigScint/include/TrigScint/TrigScintFirmwareHitProducer.h
@@ -52,9 +52,6 @@ class TrigScintFirmwareHitProducer : public framework::Producer {
    */
 
  private:
-  /// Class to set the verbosity level.
-  // TODO: Make use of the global verbose parameter.
-  bool verbose_{false};
 
   /// Name of the input collection containing the sim hits
   std::string inputCollection_;
--- a/TrigScint/include/TrigScint/TrigScintFirmwareHitProducer.h
+++ b/TrigScint/include/TrigScint/TrigScintFirmwareHitProducer.h
@@ -1,0 +1,92 @@
+/**
+ * @file TrigScintFirmwareHitProducer.h
+ * @brief Staging of Real Hits
+ * @author Lene Kristian Bryngemark, Stanford University
+ */
+
+#ifndef TRIGSCINT_TRIGSCINTFIRMWAREHITPRODUCER_H
+#define TRIGSCINT_TRIGSCINTFIRMWAREHITPRODUCER_H
+
+/*~~~~~~~~~~*/
+/*   ROOT   */
+/*~~~~~~~~~~*/
+#include "TRandom3.h"
+
+// LDMX
+#include "DetDescr/TrigScintID.h"
+#include "Recon/Event/EventConstants.h"
+#include "Tools/NoiseGenerator.h"
+#include "TrigScint/Event/TrigScintHit.h"
+#include "TrigScint/Event/TrigScintQIEDigis.h"
+
+/*~~~~~~~~~~~~~~~*/
+/*   Framework   */
+/*~~~~~~~~~~~~~~~*/
+#include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
+
+/*~~~~~~~~~~~*/
+/* TrigScint */
+/*~~~~~~~~~~~*/
+#include "TrigScint/Firmware/objdef.h"
+#include "TrigScint/SimQIE.h"
+
+namespace trigscint {
+
+/**
+ * @class TrigScintFirmwareHitProducer
+ * @brief
+ */
+class TrigScintFirmwareHitProducer : public framework::Producer {
+ public:
+  TrigScintFirmwareHitProducer(const std::string& name,
+                               framework::Process& process)
+      : Producer(name, process) {}
+
+  void configure(framework::config::Parameters& ps) override;
+
+  void produce(framework::Event& event) override;
+
+  /**
+   * add a hit at index idx to a cluster
+   */
+
+ private:
+  /// Class to set the verbosity level.
+  // TODO: Make use of the global verbose parameter.
+  bool verbose_{false};
+
+  /// Name of the input collection containing the sim hits
+  std::string inputCollection_;
+
+  /// Name of the pass that the input collection is on (empty string means take
+  /// any pass)
+  std::string inputPassName_;
+
+  /// Name of the output collection that will be used to stored the
+  /// digitized trigger scintillator hits
+  std::string outputCollection_;
+
+  /// SiPM gain
+  double gain_{1e6};
+
+  /// QIE pedestal
+  double pedestal_{6.0};
+
+  /// Total MeV per MIP
+  double mevPerMip_{1.40};
+
+  /// Total number of photoelectrons per MIP
+  double pePerMip_{13.5};
+
+  /// Total number of photoelectrons per MIP
+  int sample_of_interest_{2};
+
+  std::string testCollection_;
+
+  bool doTest_{true};
+};
+
+}  // namespace trigscint
+
+#endif /* TRIGSCINT_TRIGSCINTFIRMWAREHITPRODUCER_H */
--- a/TrigScint/include/TrigScint/TrigScintFirmwareHitProducer.h
+++ b/TrigScint/include/TrigScint/TrigScintFirmwareHitProducer.h
@@ -52,7 +52,6 @@ class TrigScintFirmwareHitProducer : public framework::Producer {
    */
 
  private:
-
   /// Name of the input collection containing the sim hits
   std::string inputCollection_;
 
--- a/TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
+++ b/TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
@@ -1,0 +1,167 @@
+#include <stdio.h>
+
+#include <iostream>
+
+#include "TrigScint/Firmware/hitproducer.h"
+#include "TrigScint/Firmware/objdef.h"
+
+void hitproducer_hw(ap_uint<14> FIFO[NHITS][5], Hit outHit[NHITS],
+                    ap_uint<8> Peds[NHITS]) {
+#pragma HLS ARRAY_PARTITION variable = FIFO complete
+#pragma HLS ARRAY_PARTITION variable = amplitude complete
+#pragma HLS ARRAY_PARTITION variable = Peds complete
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[0]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[1]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[2]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[3]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[4]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[5]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[6]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[7]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[8]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[9]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[10]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[11]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[12]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[13]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[14]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[15]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[16]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[17]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[18]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[19]
+
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[20]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[21]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[22]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[23]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[24]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[25]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[26]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[27]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[28]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[29]
+
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[30]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[31]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[32]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[33]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[34]
+#pragma HLS INTERFACE ap_fifo depth = 16 port = FIFO[35]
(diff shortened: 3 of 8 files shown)


tvami:
@rodwyer100 let me rebase this for you, you have all these commits from the past, I'll push in a second

rodwyer100:
I think the NumericalRecHitProducer may be introducing errors that prevent the build test from working. I am trying to remove it for another commit; its the other stuff I really only need commited on the short term.

tvami:
> let me rebase this for you

so I tried this, but it leads to way to many conflicts, so I dont wanna touch too much of the code. This is what I did
```
git fetch
git switch trunk
git pull
git switch -
just format-cpp
git rebase trunk
```
this will then give you options to pick if you want the new or old

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
```diff
@@ -0,0 +1,144 @@
...
+    outHit[i].bID = -1;
+    outHit[i].mID = 0;
+    outHit[i].Time = 0;
+    outHit[i].Amp = 0;
+    ap_uint<14> word1 = FIFO[i][0];
+    ap_uint<14> word2 = FIFO[i][1];
+    ap_uint<14> word3 = FIFO[i][2];
+    ap_uint<14> word4 = FIFO[i][3];
+    ap_uint<14> word5 = FIFO[i][4];
+    ap_uint<14> word6 = FIFO[i][5];
```
tvami:
```
ap_uint<14> FIFO[NHITS][5]
```
i.e. the highest index you can have is `FIFO[i][4];` not the `FIFO[i][5];`

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
```diff
@@ -0,0 +1,97 @@
...
+
+  return;
+}
+
+void TrigScintFirmwareHitProducer::produce(framework::Event &event) {
+  if (doTest_) {
+    const auto rechits{event.getCollection<ldmx::TrigScintHit>(testCollection_,
+                                                               inputPassName_)};
+    for (const auto &hit : rechits) {
+      std::cout << "Analysis barID: " << hit.getBarID()
```
tvami: Move this to the logger system

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
```diff
@@ -0,0 +1,97 @@
...
+  ap_uint<8> Peds[NCHAN];
+  for (int i = 0; i < NCHAN; i++) {
+    Peds[i] = 0;
+    FIFO[i][0] = (Peds[i] << 6) + 63;
+    FIFO[i][1] = (Peds[i] << 6) + 63;
+    FIFO[i][2] = (Peds[i] << 6) + 63;
+    FIFO[i][3] = (Peds[i] << 6) + 63;
+    FIFO[i][4] = (Peds[i] << 6) + 63;
+  }
+  std::cout << "DID I GET HERE 3" << std::endl;
```
tvami:
```suggestion
```

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
```diff
@@ -0,0 +1,97 @@
...
+  for (const auto &digi : digis) {
+    std::vector<int> adcs = digi.getADC();
+    std::vector<int> tdcs = digi.getTDC();
+    for (int i = 0; i < 5; i++) {
+      FIFO[(int)(digi.getChanID())][i] =
+          (ap_uint<14>)((adcs[i] << 6) + (tdcs[i]));
+    }
+  }
+  std::cout << "DID I GET HERE 4" << std::endl;
```
tvami:
```suggestion
```

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
```diff
@@ -0,0 +1,97 @@
...
+      }
+      ldmx::TrigScintHit hit;
+      hit.setModuleID(outHit[i].mID);
+      hit.setBarID(outHit[i].bID);
+      hit.setTime(outHit[i].Time);
+      hit.setPE(outHit[i].Amp);
+      trigScintHits.push_back(hit);
+    }
+  }
+  std::cout << "DID I GET HERE 6" << std::endl;
```
tvami:
```suggestion
```

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+  std::cout << "DID I GET HERE 4" << std::endl;
+  hitproducer_hw(FIFO, outHit, Peds);
+  std::cout << "DID I GET HERE 5" << std::endl;
+  std::vector<ldmx::TrigScintHit> trigScintHits;
+  for (int i = 0; i < NHITS; i++) {
+    if (outHit[i].Amp >= 3) {
+      if (doTest_) {
+        std::cout << "Firmware barID: " << outHit[i].bID
```
tvami: move it to log debug

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
```diff
@@ -0,0 +1,97 @@
...
+    std::vector<int> adcs = digi.getADC();
+    std::vector<int> tdcs = digi.getTDC();
+    for (int i = 0; i < 5; i++) {
+      FIFO[(int)(digi.getChanID())][i] =
+          (ap_uint<14>)((adcs[i] << 6) + (tdcs[i]));
+    }
+  }
+  std::cout << "DID I GET HERE 4" << std::endl;
+  hitproducer_hw(FIFO, outHit, Peds);
+  std::cout << "DID I GET HERE 5" << std::endl;
```
tvami:
```suggestion
```

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
```diff
@@ -0,0 +1,144 @@
...
+              sense_[4 * rr + ss] / 2 - 1;
+
+    rr = (word6 >> 6) / 64;
+    v1 = (word6 >> 6) % 64;
+    ss = 1 * (v1 > nbins_[1]) + 1 * (v1 > nbins_[2]) + 1 * (v1 > nbins_[3]);
+    charge6 = edges_[4 * rr + ss] + (v1 - nbins_[ss]) * sense_[4 * rr + ss] +
+              sense_[4 * rr + ss] / 2 - 1;
+
+    outHit[i].bID = i;
+    outHit[i].Time = (word1 & 63);
```
tvami, line 155: is this 63 on purpose? everything above was with 64. Maybe add a comment in the code where these numbers are coming from

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
```diff
@@ -0,0 +1,144 @@
...
+    outHit[i].Amp =
+        ((charge1 + charge2 + charge3 + charge4 + charge5 + charge6 - 36) *
+         .00625);
```
tvami: can we make the `0.00625` a name const expr? also why are we subtracting 36? (prob another point where adding a comment would be helpful to the reader of the code)

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
```diff
@@ -0,0 +1,97 @@
...
+    FIFO[i][1] = (Peds[i] << 6) + 63;
+    FIFO[i][2] = (Peds[i] << 6) + 63;
+    FIFO[i][3] = (Peds[i] << 6) + 63;
+    FIFO[i][4] = (Peds[i] << 6) + 63;
+  }
+  std::cout << "DID I GET HERE 3" << std::endl;
+  for (const auto &digi : digis) {
+    std::vector<int> adcs = digi.getADC();
+    std::vector<int> tdcs = digi.getTDC();
+    for (int i = 0; i < 5; i++) {
```
tvami: is the 5 the anyhow connected to the `NTIMES` variable?

tvami (changes requested):
I have some cosmetics comments, but also some real issue with the line `ap_uint<14> word6 = FIFO[i][5];`. Otherwise cleanup, addition of comments + moving magic numbers to const expressions

tvami:
>  but I have seen enough of the output to know that the firmware is replicating the software appropriattely.

I think we should introduce a DQM analyzer that plots these variables, and then we can compare things like what you mention above more easily (I guess this can be done in another PR)

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
rodwyer100: Done

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
rodwyer100: Done

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
rodwyer100: Done

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
rodwyer100: Done

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
rodwyer100: I put a comment; it is needed to extract the tdc alone.

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
rodwyer100: Yes. The number 5 used to be wholely determined by the clock speed of the FPGA and our bunch crossing rate. Namely if we are able to have a ~5 ns clock cycle with ~25-26 many clocks between bunches, than we have a latency of 5. Originally this latency was what was going to restrict how many clockcycles we integrated a pulse over, but now we just need it as long as a pulse as we'll run a couple of these in parrallel. This number, however, will not change and is somewhat nonconfigurable; the firmware is built around it. I would need to delete lines to make it 3 or add some to make it 6.

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
rodwyer100: Yes good catch. This was a big problem; I think I bungled the number just before submitting the pull request because right after I was having build issues. Thank you!

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
rodwyer100: The 36 is an artefact of how the QIE relates incident charge to ADCs; it is necessary to have as it is the only way zero charge corresponds to zero adc. The .00625 is a conglomerate effect of gains and other things that will vary depending on the amounts of shunts activated in your QIE card. It is a default value, I will include a shunt value that can change.

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
rodwyer100: done I think? You mean use verbose, correct?

rodwyer100:
Okay so I am trying to pass the build test with warnings interpreted as errors. It seems that there are many warnings unrelated to what I am doing in code thats old. Lmk if you want me to try to fix it, but seeing as its in the HLS primitives used in Trigger I doubt Id be able to.

tomeichlersmith:
@rodwyer100 See the changes here: https://github.com/LDMX-Software/ldmx-sw/pull/1470

We can choose to ignore the warnings from HLS but there are several other parts of TS code that can be patched by us.

You can also quicken development by testing the warnings-as-errors compilation locally with

```
just configure-force-error
```

before building to update your configuration to treat warnings as errors.

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
tvami: no no I meant 
```
ldmx_log(debug)
```
and then you dont need the `<< std::endl` either

rodwyer100: Gotcha! Implemented it :)

rodwyer100:
Alright everythings been rebased, checked that its building, and fixed w.r.t the changes yall made on the last push

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
```diff
@@ -0,0 +1,167 @@
+#include <stdio.h>
+
+#include <iostream>
+
+#include "TrigScint/Firmware/hitproducer.h"
+#include "TrigScint/Firmware/objdef.h"
+
+void hitproducer_hw(ap_uint<14> FIFO[NHITS][5], Hit outHit[NHITS],
+                    ap_uint<8> Peds[NHITS]) {
+#pragma HLS ARRAY_PARTITION variable = FIFO complete
```
tomeichlersmith, line 11: These `pragma`s are likely an issue. I encountered a warning that they were undefined when compiling and so I wrapped the ones currently on `trunk` in a `#ifdef` to avoid them during compilation.

https://github.com/LDMX-Software/ldmx-sw/blob/ebbf27abb5b46398e8539c7ab02cfd9e77adcc69/TrigScint/src/TrigScint/Firmware/trackproducer_hw.cxx#L10-L22

Is this the proper way to do that? These `pragma` makes me think these lines of code are actually compiled by some other tool into some other product. Is that accurate? In which case, how can we make sure they are compiling as intended in this other environment?

TrigScint/include/TrigScint/Firmware/hitproducer.h
```diff
@@ -0,0 +1,13 @@
+#ifndef HITPRODUCER_H
+#define HITPRODUCER_H
+
+#include "objdef.h"
+
+void copyHit1(Hit One, Hit Two);
+void copyHit2(Hit One, Hit Two);
+void hitproducer_ref(ap_uint<14> FIFO[NHITS][5], Hit outHit[NHITS],
+                     ap_uint<8> Peds[NHITS]);
```
tomeichlersmith, line 10: Similar comment to the `pragma` comment. These are not defined in the source file and so they will lead to an error. Should we wrap them in some compiler definition `#ifdef` to avoid them when we are emulating (here) and keep them for when this file is not being used for emulation (somewhere else?).

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
```diff
@@ -0,0 +1,96 @@
...
+    ldmx_log(info) << "\nPedestal: " << pedestal_ << "\nGain: " << gain_
+                   << "\nMEV per MIP: " << mevPerMip_
+                   << "\nPE per MIP: " << pePerMip_
+                   << "\ninput collection:     " << inputCollection_
+                   << "\ntest collection:	" << testCollection_
+                   << "\nAre we testing:        " << doTest_
+                   << "\nInput pass name:     " << inputPassName_
+                   << "\nOutput collection:    " << outputCollection_
+                   << "\nVerbosity: " << verbose_;
+  }
```
tomeichlersmith: I know that this pattern exists on `trunk` currently but we do have better control over the logging now and so Please convert these messages to use `ldmx_log(debug)` instead of `ldmx_log(info)` and remove the `verbose_` check.

(This comment goes for other locations of `verbose_` as well).

TrigScint/include/TrigScint/TrigScintFirmwareHitProducer.h
```diff
@@ -0,0 +1,92 @@
...
+  void produce(framework::Event& event) override;
+
+  /**
+   * add a hit at index idx to a cluster
+   */
+
+ private:
+  /// Class to set the verbosity level.
+  // TODO: Make use of the global verbose parameter.
+  bool verbose_{false};
```
tomeichlersmith: Like other comment said - goal is to remove this as a member variable and put any high-verbosity messages into the `ldmx_log(debug)` channel.

tomeichlersmith (commented):
Looking good! I have some questions about how these source files within `Firmware` are being used outside of ldmx-sw and whether particular declarations and `pragma` are necessary to keep around.

Besides that, just some cleanup with the logging.

TrigScint/include/TrigScint/Firmware/hitproducer.h
rodwyer100: These are old pieces of code that used to mean something. The hitproducer_ref might be something HLS requires defined however, so while the copyHit1 may not need to be there in some capacity the _ref will be. So here I think the ifdef solution is reasonable.  I will investigate whether in vitis HLS there is a special bool somewhere that they use to enable the reading of pragmas; doubt it but I will check. Ruckus uses a cmake file unlike vitis HLS's standalone distribution, so if nothing else I can put a copy of whatever convention we adopt here in there.

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
rodwyer100: That is true. They are being compiled in Vitis HLS' propriety software, so not sure how to ensure they can read it in while we ignore it. In the next comment I mention that although I am not certain how this would be done in vitis hls standalone, we could probably use the same files with this ifdef scheme if we compile firmware using Larry's cmake build oriented tools

TrigScint/include/TrigScint/TrigScintFirmwareHitProducer.h
rodwyer100: done :)!

TrigScint/src/TrigScint/TrigScintFirmwareHitProducer.cxx
rodwyer100: Done!

TrigScint/include/TrigScint/Firmware/hitproducer.h
rodwyer100: Done!

TrigScint/src/TrigScint/Firmware/hitproducer_hw.cxx
rodwyer100: done

tvami (approved):
Looks good! My only request is a follow-up PR that does a DQM analyzer on these developments

tomeichlersmith (approved):

PR #1474, opened 2024-09-24
Add ASAN/UBSAN recipe to just, make a CI that runs it + adding GDB

### LINKED ISSUE #1043: Improved debugging QOL
Debugging using std::cout and its friends works but is rather limited compared to using more standard debugging tools/features. In particular, I find myself missing two features when working on LDMX-sw (especially when something crashes!), a dedicated debugger and the various sanitizers (in particular address and undefined behavior, see https://github.com/google/sanitizers). 

I also believe that supporting these things can lead to a much less unpleasant experience for people that aren't familiar with the framework when things go wrong. Rather than having to know how all the pieces fit together so you know where to place your std::cout statements, running with sanitizers enabled can often directly pinpoint where the problem is located. For example, I had a lovely time trying to figure out the cause of intermittent crashes caused by a hard coded string literal in the prototype geometry code (I get what I deserve for writing that in the first place!). After working on it for ~2 hours, I went ahead and made a bare metal setup with sanitizers enabled. UBsan caught the issue directly and provided local information about the problem.

I would like to see two updates to the container environment with corresponding support in the build setup (i.e. the Cmake module). These are technically distinct so if people are fine with GDB but don't want sanitizer support in the container then that's possible. 

1. Adding support for using a basic debugger. This would require
    - Adding GDB to the container
2. Adding support for building LDMX-sw with sanitizers enabled
    - Adding the runtime libraries required by the sanitizers (note that the compile time instrumentation features come with the compiler already) 
    -  Add corresponding options in the CMake setup to let people be able to use sanitizers without having to fiddle around with compilation flags directly 

Keeping on having a bare metal version of LDMX-sw lying around which isn't ideal since 

1. The benefits of the container workflow is lost 
2. I need to maintain two separate setups on my machine 
3. I can't teach people how to do this easily but I can tell someone to enable some CMake flags 

**Additional context**
I've already tried this out and both the required changes for the container and the CMake modules are straight forward. If we want to do this, I can create issues in the corresponding repositories and describe the (brief) technical details there

There was some brief discussion regarding this in the #software-dev channel on the LDMX slack, December 15th

God fortsättning as we say in Sweden :)

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1043 (the last bit of it)

## Check List

I did
```
just configure-asan-ubsan
just fire .github/validation_samples/kaon_enhanced/config.py 
```
give a lot of indirect memory leaks, which is fine, none of them are big. This is kinda expected since we fixed the ASAN / UBSAN problems in previous PRs. 

Added it to the CI with the option w/o the leak sanitizer, so currently there is no output from ASAN/UBSAN (given that we cleaned all the issues up already).

Then
```
just configure-gdb
denv gdb build/run_test 
```
this runs too, although I'm not sure how to actually make use of it for a real config file with `fire`.
I introduced
```
just debug
```
as well.

### DIFF
--- a/.github/workflows/asan_ubsan_build.yml
+++ b/.github/workflows/asan_ubsan_build.yml
@@ -34,7 +34,7 @@ jobs:
     - name: Run ecal_pn config
       continue-on-error: true
       run: |
-        just setenv LDMX_NUM_EVENTS=1500
+        just setenv LDMX_NUM_EVENTS=15
         just setenv LDMX_RUN_NUMBER=1
         just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py 2> output.log
       shell: bash
--- a/.github/workflows/asan_ubsan_build.yml
+++ b/.github/workflows/asan_ubsan_build.yml
@@ -35,7 +35,7 @@ jobs:
       #continue-on-error: true
       run: |
         just setenv LDMX_NUM_EVENTS=15
-        just setenv LDMX_RUN_NUMBER=1
+        just setenv LDMX_RUN_NUMBER=1500
         #just setenv ASAN_OPTIONS=log_path=output.log
         just setenv ASAN_OPTIONS=detect_leaks=0
         just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py
--- a/.github/workflows/asan_ubsan_build.yml
+++ b/.github/workflows/asan_ubsan_build.yml
@@ -9,7 +9,6 @@ on:
 
 env:
   LDMX_DOCKER_TAG: ldmx/dev:latest
-  LDMX_NUM_EVENTS: 7500
 
 jobs:
   build-with-asan-ubsan:
@@ -34,9 +33,9 @@ jobs:
 
     - name: Run ecal_pn config
       run: |
-        just setenv LDMX_NUM_EVENTS=7500
+        just setenv LDMX_NUM_EVENTS=1500
         just setenv LDMX_RUN_NUMBER=1
-        export LDMX_NUM_EVENTS=7500
+        export LDMX_NUM_EVENTS=1500
         export LDMX_RUN_NUMBER=1
         just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py
       shell: bash
--- a/.github/workflows/asan_ubsan_build.yml
+++ b/.github/workflows/asan_ubsan_build.yml
@@ -32,16 +32,18 @@ jobs:
       shell: bash
 
     - name: Run ecal_pn config
-      continue-on-error: true
+      #continue-on-error: true
       run: |
         just setenv LDMX_NUM_EVENTS=15
         just setenv LDMX_RUN_NUMBER=1
-        just setenv ASAN_OPTIONS=log_path=output.log
+        #just setenv ASAN_OPTIONS=log_path=output.log
+        just setenv ASAN_OPTIONS=detect_leaks=0
         just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py
       shell: bash
 
-    - name: Upload log artifact
-      uses: actions/upload-artifact@v4
-      with:
-        name: ASAN_UBSAN_output
-        path: output.log*
+      # For the future, we could try to save the output of ASAN/UBSAN
+      #    - name: Upload log artifact
+      #      uses: actions/upload-artifact@v4
+      #      with:
+      #        name: ASAN_UBSAN_output
+      #        path: output.log*
--- a/.github/workflows/asan_ubsan_build.yml
+++ b/.github/workflows/asan_ubsan_build.yml
@@ -32,6 +32,11 @@ jobs:
       run: |
         just install-denv init configure-asan-ubsan 
         just build
-        just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py 
+        just setenv LDMX_NUM_EVENTS=7500
+        just setenv LDMX_RUN_NUMBER=1
       shell: bash
+
+    - name: Run ecal_pn config
+      run: |
+        just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py
 
--- a/.github/workflows/asan_ubsan_build.yml
+++ b/.github/workflows/asan_ubsan_build.yml
@@ -35,7 +35,7 @@ jobs:
       #continue-on-error: true
       run: |
         just setenv LDMX_NUM_EVENTS=15
-        just setenv LDMX_RUN_NUMBER=1500
+        just setenv LDMX_RUN_NUMBER=7500
         #just setenv ASAN_OPTIONS=log_path=output.log
         just setenv ASAN_OPTIONS=detect_leaks=0
         just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py
--- a/.github/workflows/asan_ubsan_build.yml
+++ b/.github/workflows/asan_ubsan_build.yml
@@ -36,6 +36,8 @@ jobs:
       run: |
         just setenv LDMX_NUM_EVENTS=7500
         just setenv LDMX_RUN_NUMBER=1
+        export LDMX_NUM_EVENTS=7500
+        export LDMX_RUN_NUMBER=1
         just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py
       shell: bash
 
--- a/.github/workflows/asan_ubsan_build.yml
+++ b/.github/workflows/asan_ubsan_build.yml
@@ -36,13 +36,12 @@ jobs:
       run: |
         just setenv LDMX_NUM_EVENTS=15
         just setenv LDMX_RUN_NUMBER=1
-        just setenv LSAN_OPTIONS=verbosity=1
-        just setenv LSAN_OPTIONS=log_threads=1
-        just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py 2> output.log
+        just setenv ASAN_OPTIONS=log_path=output.log
+        just fire  ${GITHUB_WORKSPACE}/.github/validation_samples/ecal_pn/config.py
       shell: bash
 
     - name: Upload log artifact
       uses: actions/upload-artifact@v4
       with:
         name: ASAN_UBSAN_output
-        path: output.log
+        path: output.log*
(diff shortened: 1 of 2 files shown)


.github/workflows/asan_ubsan_build.yml
```diff
@@ -0,0 +1,42 @@
...
+ 
+    - name: Compile and Install ldmx-sw
+      run: |
+        just install-denv init configure-asan-ubsan 
+        just build
+      shell: bash
+
+    - name: Run ecal_pn config
+      run: |
+        just setenv LDMX_NUM_EVENTS=1500
```
EinarElen: I would consider setting the ASAN_OPTIONS flag to not show leaks

justfile
```diff
@@ -82,6 +91,11 @@ test *ARGS:
 fire config_py *ARGS:
     denv fire {{ config_py }} {{ ARGS }}
 
+# run gdb on a config file
+[no-cd]
+gdb-fire  config_py *ARGS:
+    denv gdb fire {{ config_py }} {{ ARGS }}
```
EinarElen, line 100: I would name it debug rather than gdb-fire

EinarElen:
Thinking a bit more about this, do we need a separate sanitizer build? Why not just run all of the CI with a sanitized build?

tvami:
> Thinking a bit more about this, do we need a separate sanitizer build? Why not just run all of the CI with a sanitized build?

It takes forever... we usual test 10k events, I tried with 7.5k for the ecal pn, and was killed at 6 hours. I'm now converging to the point of only testing 1500 events which will take about an hour. (But for some reason I cant keep the ASAN output in the CI)

.github/workflows/asan_ubsan_build.yml
```diff
@@ -0,0 +1,47 @@
...
+        just setenv LDMX_RUN_NUMBER=1
+        export LDMX_NUM_EVENTS=15
+        export LDMX_RUN_NUMBER=1
```
tomeichlersmith:
```suggestion
```

These lines don't get passed into the containerized environment since `just init` tells `denv` to not share host environment variables by default. And when setting an environment variable, `denv` overrides the host's value (you can see that fire ran for 1500 events rather than just 15).

.github/workflows/asan_ubsan_build.yml
```diff
@@ -0,0 +1,47 @@
...
+      with:
+          just-version: 1.26.0
+ 
+    - name: Compile and Install ldmx-sw
+      run: |
+        just install-denv init configure-asan-ubsan 
+        just build
+      shell: bash
+
+    - name: Run ecal_pn config
```
tomeichlersmith, line 34:
```suggestion
    - name: Run ecal_pn config
      continue-on-error: true
```

We want the job to continue even if this step fails (which it will since ASAN will exit with non-zero if any issues are encountered). https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

justfile
```diff
@@ -62,13 +62,22 @@ install-denv:
...
+
 # This is the same as just configure but reports all (non-3rd-party) warnings as errors
 configure-force-error *CONFIG: (configure "-DWARNINGS_AS_ERRORS=ON" CONFIG)
 
 # This is an alternative compiler (clang) together with enabling the LTO sanitizer
 # This is very helpful but for now it optimizes away our module linking, use for testing only
 configure-clang-lto: (configure "-DENABLE_LTO=ON -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang") 
 
+configure-gdb: (configure-quick "-DCMAKE_BUILD_TYPE=Debug")
+
```
tomeichlersmith: I'm kind of a DRY fiend and so I'm imagining a "parent" and private recipe that the "child" and public recipes use. This would mean we'd have one place that defines how `cmake` is called within the denv and the rest are just passing different configure options.

Not sure if this is worth it though since the code that is being repeated is pretty simple and short.

```suggestion
# base configure command defining how cmake is called, private so only experts call it
[private]
configure-base *CONFIG:
    denv cmake -B build -S . {{ CONFIG }}

# default configure of build when developing
configure *CONFIG: (configure-base "-DADDITIONAL_WARNINGS=ON -DENABLE_CLANG_TIDY=ON" CONFIG)

# configure minimal option for faster compilation
configure-quick: (configure-base)

# configure with Address Sanitizer (ASAN) and  UndefinedBehaviorSanitizer (UBSan)
configure-asan-ubsan: (configure-base "-DENABLE_SANITIZER_UNDEFINED_BEHAVIOR=ON -DENABLE_SANITIZER_ADDRESS=ON")

# This is the same as just configure but reports all (non-3rd-party) warnings as errors
configure-force-error: (configure "-DWARNINGS_AS_ERRORS=ON")

# Use alternative compiler and enable LTO (test compiling only, won't run properly)
configure-clang-lto: (configure "-DENABLE_LTO=ON -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER=clang") 

# Keep debug symbols so running with gdb provides more helpful detail
configure-gdb: (configure-base "-DCMAKE_BUILD_TYPE=Debug")
```

tomeichlersmith (commented):
One DRY comment that is optional and some guidance on writing the workflow.

.github/workflows/asan_ubsan_build.yml
tvami: Perfect thank you! I really was looking for this!

EinarElen:
> > Thinking a bit more about this, do we need a separate sanitizer build? Why not just run all of the CI with a sanitized build?
> 
> 
> 
> It takes forever... we usual test 10k events, I tried with 7.5k for the ecal pn, and was killed at 6 hours. I'm now converging to the point of only testing 1500 events which will take about an hour. (But for some reason I cant keep the ASAN output in the CI)

This sounds really really strange. The sanitizers should have an extremely small impact on performance

.github/workflows/asan_ubsan_build.yml
EinarElen: You can tell asan to continue on error if needed

tvami:
> > > Thinking a bit more about this, do we need a separate sanitizer build? Why not just run all of the CI with a sanitized build?
> > 
> > 
> > It takes forever... we usual test 10k events, I tried with 7.5k for the ecal pn, and was killed at 6 hours. I'm now converging to the point of only testing 1500 events which will take about an hour. (But for some reason I cant keep the ASAN output in the CI)
> 
> This sounds really really strange. The sanitizers should have an extremely small impact on performance

Hmm, I dont know, here is the action that was killed after 6 hours https://github.com/LDMX-Software/ldmx-sw/actions/runs/11007030465 if it helps

tvami:
Ok the output is being saved now
https://github.com/LDMX-Software/ldmx-sw/actions/runs/11038784203/artifacts/1978699120
but it still doesnt have what I want... I'll play with it more later

tvami:
(sorry I'm going back and forth between draft and ready -- that's my only way to trigger tests until it's merged [after that we'll be able to do this from the Actions with the dispatch options])

tvami:
Hmm, I dont understand, in local the ASAN output is in stderr. But in the action I only get

https://github.com/LDMX-Software/ldmx-sw/actions/runs/11044190876
```
denv fire /home/runner/work/ldmx-sw/ldmx-sw/.github/validation_samples/ecal_pn/config.py 
==5==LeakSanitizer has encountered a fatal error.
==5==HINT: For debugging, try setting environment variable LSAN_OPTIONS=verbosity=1:log_threads=1
==5==HINT: LeakSanitizer does not work under ptrace (strace, gdb, etc)
error: Recipe `fire` failed on line 95 with exit code 1
```
even tho ` LSAN_OPTIONS=verbosity=1:log_threads=1` are enabled... (which btw I dont need when I run locally)

EinarElen:
That seems strange... But the simulation finishes and the leak sanitizer issue is only at the end? I'm still a bit skeptical to the value of running leaksanitizer in CI for us so I might just turn it off isntead

tvami:
> But the simulation finishes and the leak sanitizer issue is only at the end? 

I'm running ASAN + UBSAN, and they somehow invoke LSAN too. And yes, they only show up in the end, bc of the setting to dont stop on issues (which is a default set by you @EinarElen I believe, but correct me if I'm wrong [given that we fixed all the issues, it's hard to tell]).

I can be convinced either way about the leak part. But the issue about github dealing with stderr from ASAN/UBSAN is not going to be connected to the leak part. Maybe @tomeichlersmith has an idea about the environment settings on the machines used  by the actions.

tomeichlersmith:
The runners do not swallow anything from stdout or stderr, but they _do not_ have a TTY and so programs that are "smart" and disable certain behaviors in a non-interactive environment (mainly determined by detecting if a TTY is present) will disable those behaviors.

You can mimic running without a TTY to see if that is the issue.

https://superuser.com/a/1430883

EinarElen:
LSAN is a part of ASAN so it gets enabled by default. You turn if off with ASAN_OPTIONS=detect_leaks=0

tvami:
> You can mimic running without a TTY to see if that is the issue.

so none of this managed to reproduce what I see in the actions, but I think I found a workaround with specifying the ASAN output, I'll test that now

tvami:
OK I think I'm ready now, this is the module operandi I propose: we dont save the log for now (it didnt work even with specifying the output for ASAN), we dont run the leak part --> this means currently there is no issues. Then we run this every time a new PR comes in, and if any of those introduce a problem, this test will fail. It will not show what the issue is (given the output issue), but it will show that there is an issue, so we can check out the branch locally and run this and we'll see the problem.

tvami:
Screenshot 2024-09-26 at 13 40 41

ok so w/o the the leak part, it doesnt actually take extremely long: 1500 events took 30 min, so the 10k would take 3.3 hours. That's still more than the usual 1.5-2h time. I could increase this to 7500. I think the ecal_pn at this point runs all kinda processors so we are safe to just run that with ASAN/UBSAN.

tvami:
OK, tests are fine, @tomeichlersmith @EinarElen I wont push anymore, please approve if you agree with the changes

tomeichlersmith (approved):

EinarElen (approved):

.github/workflows/asan_ubsan_build.yml
tvami: Yeah I have that now `just setenv ASAN_OPTIONS=detect_leaks=0`

justfile
tvami: ok sure

justfile
```diff
@@ -82,6 +94,11 @@ test *ARGS:
 fire config_py *ARGS:
     denv fire {{ config_py }} {{ ARGS }}
 
+# run gdb on a config file
+[no-cd]
+gdb-fire  config_py *ARGS:
```
tvami:
```suggestion
debug config_py *ARGS:
```

PR #1475, opened 2024-09-25
Hcal tot digitization

vrm7fj:
I am updating _ldmx-sw_, with implemented Hcal TOT digitization.

## Check List


tomeichlersmith (changes requested):
Thank you for getting this started!

Generally, I would like to see some more detailed explanation on what you've done and then some plots showing that your developments are doing what is intended. Additionally, you have edited HgcrcoEmulator which is also used within the Ecal digitiziation pipeline. Have you confirmed that these edits do not change that pipeline at all?

(Side note: remove the `.DS_Store` files from the branch (using `git rm`). These are not related to the software and just tells me that you are on MacOS!)

vrm7fj:
**New Class 'PulseRecord':**

This class is intended solely for plotting purposes and should not interfere with the existing workflow. The PulseRecord(double volts, int adc, int tot) function temporarily stores the digitized pulses when the HgcrocEmulator is called by either the Hcal or Ecal pipelines and is cleared each time the emulator is invoked. If the pulse is digitized using ADC, the TOT value is set to zero, and vice versa.

To my understanding, the HgcrocEmulator cannot be called by the Ecal and Hcal pipelines at the exact same time. Therefore, when the getter for PulseRecord is called within the HcalDigiProducer, it will only retrieve the digitized information corresponding to the pulses sent for digitization at that moment. If this understanding is incorrect, the class may only be suitable for studying the back Hcal.

**HgcrocEmulator**

HgcrocEmulator.h has been amended with a mutable vector ‘pulseRecord_’ and a getter to access ‘pulseRecord_’ to store and access digitizations.

Originally, the TOT digitization for an incoming pulse was set by comparing the voltage (vpeak) of the hit, calculated using the pulse function at the time of the hit that comes with the pulse. However, in ADC digitization, the voltage (bxvolts) is calculated using iADC and the clock cycle. There is a slight difference between the two voltage values. See the following debug print.

Screenshot 2024-09-26 at 2 15 34 PM

If we used the original setting:

if (vpeak > totThreshold) {
    startTOT = true;
}

ADC and TOT digitizations look like this. Note that the ADC digitization does not reach up to 1024 counts here.

ADCTOT2

Instead, if we modify the code as follows, the complete ADC range is utilized in the digitization:

         if (bxvolts > totThreshold){
        /startTOT = true;
        }
        
ADCTOT

In this branch, the ‘bxvolts > totThreshold’ condition is currently used.

**HcalDigiProducer** 

Once each section in HcalDigiProducer sends pulses to HgcrocEmulator, they are accessed through the getter and temporarily stored in the vector ‘DigiToPlot’, which is cleared for each ‘hitsByID’ in the loop. At the end of each section, the recorded digitized info is written into a text file following the format [Voltage ADC TOT]. The plots I showed above were generated using that text file.

**hcal_hardcoded_conditions.py  and proposed TOT reconstruction**

Please see the following presentation to understand how changes were made to hcal_hardcoded_conditions and how TOT reconstruction can be implemented.

[LDMX_HCAL.pdf](https://github.com/user-attachments/files/17331780/LDMX_HCAL.pdf)

**Discussion**

Before the implementation of TOT digitization, the histogram of Hcal reconstructed energy was piling up towards the last bins. However, after the implementation, we now observe a smoothed decay in the reconstructed energy. Please note that the appropriate parameters for reconstruction have not yet been configured.

Image 10-10-24 at 1 25 PM

Hcal/src/Hcal/HcalDigiProducer.cxx
```diff
@@ -246,7 +257,44 @@ void HcalDigiProducer::produce(framework::Event& event) {
...
+        }
+      }
+
+      // Print Voltages, adc/tot values to the same txt file
+      std::ofstream dataFile("Voltage_ADC_TOT.txt", std::ios::app);
+      for (const auto& entry : DigiToPlot) {
+        dataFile << std::get<0>(entry) << " " << std::get<1>(entry) << " "
+                 << std::get<2>(entry) << "\n";
+      }
+      dataFile.close();
```
tomeichlersmith: This chunk of code (and the others like it) is incredibly helpful for you the developer when tuning the digitization parameters; however, most people running this producer will not be tuning the parameters and creating an extra txt file would at best be confusing and at worst cause extra issues.

My request here is to put these code blocks behind a configuration parameter such that the user can choose whether to save the digitization information. Maybe a `digi_emulation_file` parameter that is the file to save the information to? (e.g. it could be `"Voltage_ADC_TOT.txt"`) and when it is empty, nothing is saved.

Hcal/src/Hcal/HcalDigiProducer.cxx
```diff
@@ -246,7 +257,44 @@ void HcalDigiProducer::produce(framework::Event& event) {
...
+      }
+
+      if (hgcroc_->digitize(posendID.raw(), pulses_negend, digiToAddNegend) &&
+          hgcroc_->digitize(posendID.raw(), pulses_posend, digiToAddPosend)) {
+        const auto& pulseRecord = hgcroc_->getPulseRecord();
+
+        for (const auto& record : pulseRecord) {
+          DigiToPlot.emplace_back(record.getVolts(), record.getADC(),
+                                  record.getTOT());
+        }
```
tomeichlersmith, line 290: One of the code blocks that could be skipped if the user doesn't want to save the digi emulation detailed information.

Hcal/src/Hcal/HcalDigiProducer.cxx
```diff
@@ -262,18 +310,54 @@ void HcalDigiProducer::produce(framework::Event& event) {
...
+
+        // Print Voltages, adc/tot values to the same txt file
+        std::ofstream dataFile("Voltage_ADC_TOT.txt", std::ios::app);
+        for (const auto& entry : DigiToPlot) {
+          dataFile << std::get<0>(entry) << " " << std::get<1>(entry) << " "
+                   << std::get<2>(entry) << "\n";
+        }
+        dataFile.close();
+
```
tomeichlersmith, line 362: One of the code blocks that could be skipped if the user doesn't want to save the digi emulation detailed information.

Hcal/src/Hcal/HcalDigiProducer.cxx
```diff
@@ -262,18 +310,54 @@ void HcalDigiProducer::produce(framework::Event& event) {
...
+                                  record.getTOT());
+        }
+
+        // Print Voltages, adc/tot values to the same txt file
+        std::ofstream dataFile("Voltage_ADC_TOT.txt", std::ios::app);
+        for (const auto& entry : DigiToPlot) {
+          dataFile << std::get<0>(entry) << " " << std::get<1>(entry) << " "
+                   << std::get<2>(entry) << "\n";
+        }
+        dataFile.close();
```
tomeichlersmith, line 388: One of the code blocks that could be skipped if the user doesn't want to save the digi emulation detailed information.

Tools/src/Tools/HgcrocEmulator.cxx
```diff
@@ -101,8 +102,20 @@ bool HgcrocEmulator::digitize(
...
+      // << std::endl; std::cout << " " << std::endl;
+
+      // if (vpeak > totThreshold){
+      // startTOT = true;
+      // if (toverTOT < hit.second)
+      // toverTOT = hit.second;  // use the latest time in the window
+      //}
 
-      if (vpeak > totThreshold) {
+      if (bxvolts > totThreshold) {
```
tomeichlersmith, line 117: This is the part that I am worried about effecting the ECal digi emulation as well. This is a substantial change and could affect things there and, while the change is well motivated, I would like to understand what the results are.

Tools/src/Tools/HgcrocEmulator.cxx
```diff
@@ -176,14 +191,19 @@ bool HgcrocEmulator::digitize(
...
       // TODO: properly handle saturation and recovery, eventually.
       // Now just kill everything...
       while (digiToAdd.size() < nADCs_) {
         digiToAdd.emplace_back(true, false,  // flags to mark type of sample
                                0x3FF, 0x3FF, 0);
       }
 
-      return true;  // always readout
+      for (const auto &digi : digiToAdd) return true;  // always readout
+
```
tomeichlersmith: You don't need the for loop. It will exit on the first entry in the for loop anyways.

```suggestion
      return true;  // always readout
```

tomeichlersmith (changes requested):
Thank you for the extra detail. I agree that the new `PulseRecord` class would not be affecting the emulation. The HgcrocEmulator not only is called at separate times but the Ecal and Hcal subsystems have their own _copies_ of the class so they will not interfere. My concern about how this affects the Ecal emulation is from the change in the value used to determine if a specific pulse is above the tot threshold or not. I've highlighted the specific code below. This change is well motivated, I just want to understand the downstream effects.

Besides more specific comments below, my other requests are to remove the `.DS_Store` files that are not interesting for us and write some more documentation into the `PulseRecord` header file. You've already written a nice summary in your comment above so that is a good place to start.

Hcal/src/Hcal/HcalDigiProducer.cxx
```diff
@@ -501,7 +502,7 @@ void HcalDigiProducer::produce(framework::Event& event) {
 =======
     }
 >>>>>>> trunk
```
tomeichlersmith: :warning: conflict leftovers :warning:

tvami:
@vrm7fj @cmantill what's the status of this? (There are new conflicts)

tvami:
@cmantill should we close this PR for the time being, until you have somebody to take over? Or maybe @erik-wallin could help in with this development?

erik-wallin:
#1740 mostly achieves the same goals as this PR, apart from updating the default HGCROC emulator conditions. I'm currently producing MC samples for the testbeam analysis and working on tuning the conditions to match the real data, and they don't quite match the values here. I think it makes sense to use numbers that reflect what we saw at the TB, rather than to pick numbers from the data sheet.

The larger project here is actually using TOT in the reconstruction. In the TB MC samples the transition between the ADC region and TOT region looks quite good: 
image
So I think it will be quite easy to update HCalRecHitProducer to make use of TOT. I probably won't do that in this particular PR though

cmantill:
I think that most the work to use TOT in the reconstruction was done in the changes to the hcal reco conditions in this PR. So we can close this PR but we should open an issue to address this. I hope Erik can do this (I can help guide!) given that #1740 was merged already.

tvami:
> So we can close this PR but we should open an issue to address this. 

@cmantill @erik-wallin could one of you make that issue?

tvami:
OK I made the issue: https://github.com/LDMX-Software/ldmx-sw/issues/1787

Please feel free to edit it!

PR #1488, opened 2024-10-16
GENIE simulation with HepMC3 format

wesketchum:

This PR enables GENIE simulation that include the following:

- A HepMC3GenEvent object to add to the LDMX event bus, inheriting from the HepMC3::GenEvent.
- A EventWeights object to add to the LDMX event bus, for use in event reweighting studies. 
- A GenieGenerator generator that interfaces with the GENIE neutrino generator to run lepton scattering models. Currently the generator is limited to starting interactions at the target. It uses a HepMC3GenEvent as an interface to the event bus.
- A GenieReweightProducer that can run GENIE's Reweight functionality, to reweight events based on turns of systematic knobs. It reads in HepMC3GenEvent recrods, recreates the GENIE internal event record based on that, and then runs reweighting.
- A GenieTruthDQM module that extracts truth GENIE interaction information from the HepMC3GenEvent metadata, and stores it in a flat tree.

## Check List
(note, this requires use of the updated docker image described by this PR: https://github.com/LDMX-Software/docker/pull/105)

An example ldmxsw config is here: https://github.com/wesketchum/LDMX_eN_GENIE/blob/main/ldmxsw_configs/genie_sim.py. From the output of that, one can look at the `genie_truth` tree from the DQM. Attached is a plot of the initial and final lepton energy for a 4 GeV run. 
lepton_energy_example

Additionally, to see the event reweighting working, here is a screenshot of some screen debug outputs that show weights applied per event.
Screenshot 2024-10-16 at 14 26 29


tomeichlersmith (changes requested):
As a first pass, I just have some big comments
- Please review the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html), we solidified the format of variable names and such which you should apply here.
- Is `ldmx::HepMC3GenEvent` necessary? If you just need helper functions, I would prefer you to write those helper function somewhere else rather than derive a new event data type.
- Instead of using `if (verbosity ...)` checks, please use `ldmx_log`. I know we haven't transitioned other parts of SimCore to the logging system, but I would really like to avoid adding new code using the old method. It looks like your verbosity of `1` corresponds to the `debug` level.
- Does this compile with ldmx/dev:4.2.2? I just want to update the compatibility table if it doesn't.

wesketchum:
Working on this...

On `ldmx::HepMC3GenEvent`, I added i because I thought it was necessary for the `Print()` and `Clear()` functions were needed for things put onto the event bus. If that's not the case/there's a way around that, then I'd be very happy to get rid of it, and then yes, can have a helper class that is separate for anything that may be needed.

tomeichlersmith:
You're correct, the `Print` and `Clear` functions are still necessary so I withdraw that comment.

wesketchum:
OK, so:

- Reviewed code style and tried to comply. format-cpp run.
- Ran asan/ubsan and LTO builds. Minor changes after that.
- Merged with trunk (resolving conflicts on `DQM/python/dqm.py`)

Think we are ready for a re-review.

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
```diff
@@ -0,0 +1,203 @@
...
+GenieReweightProducer::GenieReweightProducer(const std::string& name,
+                                             framework::Process& process)
+    : Producer(name, process) {
+  hepMC3Converter_ = new genie::HepMC3Converter;
+  genie_rw_ = new genie::rew::GReWeight;
+}
+
+GenieReweightProducer::~GenieReweightProducer() {
+  delete hepMC3Converter_;
+  delete genie_rw_;
```
tomeichlersmith: Have you tried just using plain objects instead of pointers? Or if you need pointers, maybe `std::unique_ptr` so that the `delete` happens no-matter what?

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
```diff
@@ -0,0 +1,203 @@
...
+}
+
+void GenieReweightProducer::configure(framework::config::Parameters& ps) {
+  hepmc3CollName_ = ps.getParameter<std::string>("hepmc3CollName");
+  hepmc3PassName_ = ps.getParameter<std::string>("hepmc3PassName");
+  eventWeightsCollName_ = ps.getParameter<std::string>("eventWeightsCollName");
+  verbosity_ = ps.getParameter<int>("verbosity");
+  seed_ = ps.getParameter<int>("seed");
+  n_weights_ = (size_t)(ps.getParameter<int>("n_weights"));
```
tomeichlersmith: I have a vendetta against C-style casts[^1]

[^1]: reason: https://www.youtube.com/watch?v=SmlLdd1Q2V8

```suggestion
  n_weights_ = static_cast<size_t>(ps.getParameter<int>("n_weights"));
```

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
```diff
@@ -0,0 +1,203 @@
...
+  genie_rw_ = new genie::rew::GReWeight;
+}
+
+GenieReweightProducer::~GenieReweightProducer() {
+  delete hepMC3Converter_;
+  delete genie_rw_;
+}
+
+void GenieReweightProducer::configure(framework::config::Parameters& ps) {
+  hepmc3CollName_ = ps.getParameter<std::string>("hepmc3CollName");
```
tomeichlersmith: Fun fact, in a Framework update a while ago, I changed `Parameters` to have a shorter accessor function.

Not really necessary, just using this as a soap-box to try to advertise this update a bit more.

```suggestion
  hepmc3CollName_ = ps.get<std::string>("hepmc3CollName");
```

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
```diff
@@ -0,0 +1,203 @@
...
+  genie_rw_ = new genie::rew::GReWeight;
+
+  genie::RunOpt::Instance()->SetTuneName(tune_);
+  if (!genie::RunOpt::Instance()->Tune()) {
+    EXCEPTION_RAISE("ConfigurationException", "No TuneId in RunOption.");
+  }
+  genie::RunOpt::Instance()->BuildTune();
+
+  genie_rw_->AdoptWghtCalc("hadro_fzone", new genie::rew::GReWeightFZone);
+  genie_rw_->AdoptWghtCalc("hadro_intranuke", new genie::rew::GReWeightINuke);
```
tomeichlersmith, line 87: Does `genie_rw_` own these objects you created with `new`? (will it `delete` them?)

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
```diff
@@ -0,0 +1,203 @@
...
+  for (auto const& vt_str : var_types_strings) {
+    auto vtype = ldmx::EventWeights::string_to_variation_type(vt_str);
+    for (size_t i_w = 0; i_w < n_weights_; ++i_w)
+      variation_map_[vtype].push_back(normal_distribution(generator));
+  }
+}
+
+void GenieReweightProducer::reinitializeGenieReweight() {
+  delete genie_rw_;
+  genie_rw_ = new genie::rew::GReWeight;
```
tomeichlersmith: If this object doesn't have a reset-in-place option making it possible to use a non-pointer type, you can still use `std::unique_ptr` with the `reset` function.
```suggestion
// if genie_rw_ is a std::unique_ptr<genie::rew::GReWeight>
genie_rw_.reset(new genie::rew::GReWeight);
```

cmake/BuildMacros.cmake
```diff
@@ -67,6 +67,147 @@ macro(setup_geant4_target)
...
+#message(STATUS "Adding library Genie::${gname}")
+
+add_library(Genie::${gname} SHARED IMPORTED GLOBAL) # or STATIC instead of SHARED
+set_target_properties(Genie::${gname} PROPERTIES
+  IMPORTED_LOCATION "/usr/local/lib/lib${gname}.so"
+  INTERFACE_INCLUDE_DIRECTORIES "/usr/local/include/GENIE"
+)
+endfunction()
+
+macro(setup_genie_target)
```
tomeichlersmith, line 80: I would like to update this CMake to be
- formatted so its easier to see function definition boundaries
- target-based so we can just use `target_link_libraries` instead of having to `include_directories` and `${GENIE_LIBS}` in the `SimCore/CMakeLists.txt`
I can figure this out if you are unfamiliar with CMake and don't have the time to learn.

SimCore/src/SimCore/Simulator.cxx
```diff
@@ -157,6 +158,15 @@ void Simulator::produce(framework::Event& event) {
...
   event_header.setStringParameter("eventSeed", stream.str());
 
+  auto event_info = static_cast<UserEventInformation*>(
+      runManager_->GetCurrentEvent()->GetUserInformation());
+
+  auto hepmc3_events = event_info->getHepMC3GenEvents();
+  for (auto& hepmc3ev : hepmc3_events) {
+    hepmc3ev.event_number = event.getEventHeader().getEventNumber();
+  }
+  event.add("SimHepMC3Events", hepmc3_events);
```
tomeichlersmith: I'm curious on if we could avoid including this collection when not running the GENIE generator. Right now, this looks like we will be carrying around an empty branch with all of our other run configurations and I think that is a waste of space and potentially confusing. Could this `event.add` call be skipped if `hepmc3_events` is empty?

tomeichlersmith (commented):
More comments, apologies that this might take several rounds given how big this PR is. If you are out of time and need to just get this merged, let me know and I can prioritize that.

SimCore/include/SimCore/Generators/GenieGenerator.h
```diff
@@ -0,0 +1,117 @@
...
+  std::string spline_file_;
+
+  std::string message_threshold_file_;
+
+  std::vector<double> ev_weighting_integral_;
+  size_t n_events_generated_;
+  std::vector<size_t> n_events_by_target_;
+  std::vector<double> xsec_by_target_;
+
+  double xsec_total_;
```
tvami: I think these could be `float`-s, or do you see the need for `double` precision?

SimCore/include/SimCore/Event/EventWeights.h
```diff
@@ -0,0 +1,151 @@
+//
+// Created by Wesley Ketchum on 4/27/24.
+//
+
+#ifndef SIM_CORE_EventWeights_H
+#define SIM_CORE_EventWeights_H
```
tvami:
```suggestion
#ifndef SIMCORE_EventWeights_H
#define SIMCORE_EventWeights_H
```
this is how you did it in other cases too

SimCore/include/SimCore/Reweight/GenieReweightProducer.h
```diff
@@ -0,0 +1,110 @@
...
+
+namespace simcore {
+
+class GenieReweightProducer : public framework::Producer {
+ public:
+  // constructor
+  GenieReweightProducer(const std::string& name, framework::Process& process);
+
+  // default destructor
+  virtual ~GenieReweightProducer();
```
tvami:
```suggestion
  virtual ~GenieReweightProducer() = default;
```

SimCore/python/generators.py
```diff
@@ -193,6 +193,78 @@ def __init__(self,name,initCommands) :
...
+                         abundances = [ 1.0 ],
+                         time = 0.0,
+                         position = [0.,0.,0.],
+                         direction = [0.,0.,1.],
+                         tune='G18_02a_00_000',
+                         verbosity=True)
+    """
+
+    def __init__(self,name,
+                     energy=4.0,
```
tvami: Should this be 8?
```suggestion
                     energy=8.0,
```

SimCore/python/generators.py
```diff
@@ -193,6 +193,78 @@ def __init__(self,name,initCommands) :
...
+                     targets = [],
+                     target_thickness = 0.3504,
```
tvami, line 240: what unit is this in? (my comment is more like please add what unit it is in the Attributes above)
Is this Ti specific?

SimCore/python/genie_reweight.py
```diff
@@ -0,0 +1,13 @@
+from LDMX.Framework import ldmxcfg
+
+class GenieReweightProducer(ldmxcfg.Producer) :
+
+    def __init__(self,name='genieEventWeights'):
+        super().__init__(name,"simcore::GenieReweightProducer","SimCore::Reweight")
+
+        self.verbosity = 0
```
tvami: we should not have this, but rather use the ldmx logging system

SimCore/src/SimCore/Generators/GenieGenerator.cxx
```diff
@@ -0,0 +1,396 @@
...
+  ldmx_log(debug) << "---------- " << "Generated Event "
+                  << n_events_generated_ + 1 << " ----------";
+
+  for (int i_p = 0; i_p < n_entries; ++i_p) {
+    genie::GHepParticle* p = (genie::GHepParticle*)(*genie_event)[i_p];
+
+    // make sure it's a final state particle
+    if (p->Status() != 1) continue;
+
+    ldmx_log(debug) << "\tAdding particle " << p->Pdg() << " with status "
```
tvami, line 355: I think this might be under `trace`
https://ldmx-software.github.io/developing/Logging.html

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
```diff
@@ -0,0 +1,203 @@
...
+        par.first.compare(par.first.size() - genie_tune_par.size(),
+                          genie_tune_par.size(), genie_tune_par) == 0) {
+      new_tune = par.second;
+      break;
+    }
+
+  if (tune_ != new_tune) {
+    if (verbosity_ >= 0)
+      std::cout << "Found new tune " << new_tune << " (used to be " << tune_
+                << ")" << std::endl;
```
tvami: Please move these into the logging system

tvami (commented):
I had a few comments about doubles, logging system, and some cosmetics

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
wesketchum: Was following examples from within GENIE, which (criminally) don't use smart ptrs ... so yes, good idea. Implemented (and I hope correctly).

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
wesketchum: OK, done. I'll take the opportunity to advocate for an unsigned int on getParameter if it's not there? (I assume I tried this and it did not work, and assume you'd tell me to do that rather than cast if it did.)

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
wesketchum: Updated!

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
wesketchum: Yes (or at least it claims to own), and yes, it does delete them.

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
wesketchum: Just calling a make_unique here now, so should be better.

cmake/BuildMacros.cmake
wesketchum: I would appreciate your help here. I _should_ take the time to learn, but the chances of that happening quickly are not too high.

SimCore/src/SimCore/Simulator.cxx
wesketchum: Yes, that's a very valid point. I don't think it should be a problem to skip the add if `hepmc3_events.size()==0`. Changed to that.

SimCore/include/SimCore/Generators/GenieGenerator.h
wesketchum: Likely right. Updated.

SimCore/include/SimCore/Event/EventWeights.h
wesketchum: Thanks. Fixed.

SimCore/include/SimCore/Reweight/GenieReweightProducer.h
wesketchum: Right -- the comment was wrong before actually, but now using unique ptrs, I can use the default constructor again. Done.

SimCore/python/generators.py
wesketchum: yes ...

SimCore/python/generators.py
wesketchum: mm (comment added).

SimCore/python/genie_reweight.py
wesketchum: OK, so ... fair general point, but this is actually controlled a call to a `Print()` call on event objects. How would that be done via the logging system?

SimCore/src/SimCore/Generators/GenieGenerator.cxx
wesketchum: It's not so many particles (O(10) per event), as it's only the 'final state' particles passed from GENIE to G4, and, in 'debugging', seeing the particles produced and passed to G4 is very useful, so I prefer to keep as is.

SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx
wesketchum: This one moved, yes.

SimCore/include/SimCore/Generators/GenieGenerator.h
wesketchum: Well, actually, as get<float> doesn't appear to work, and there's not an easy(that I know of) casting of vector<double> to vector<float>, I've left the vector<double> ones that are read in by config. The size here is not a problem, so I hope convenience can win out.

wesketchum:
OK, huge thanks to @tomeichlersmith and @tvami for the comments. I've implemented what I can and left comments.

Two things with respect to logging:

(1) What should be done for a conditional print that is calling an event object's `Print()` call, like so:
```
if (verbosity_ >= 1) ev_weights.Print();
```
Should event objects `Print()` calls also use the logging system? (Seems ... no.) I think maintaining a verbosity flag here may be warranted.

(2) I had some messages being printed in the destructor of GenieGenerator, in order to print out some 'end of run' information, summarizing total events generated, xsecs, etc. When these were couts, no problem, but making them into `ldmx_log(info)` messages gives me an error at the end of processing:
```
terminate called after throwing an instance of 'boost::wrapexcept<boost::exception_detail::error_info_injector<boost::log::v2_mt_posix::system_error> >'
  what():  Failed to set TLS value: Invalid argument [system:22]
 ```
pointing back to the GenieGenerator destructor.

Arguably the destructor is a bad place for this, but I did not know where else. Is there a way to print such 'summary' stats in the Simulator? If not, I may look into adding it ... and may in the meantime request an exemption to use a cout for now.

File: SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx (hunk 2 above)
Comment (tomeichlersmith) [reply SimCore/src/SimCore/Reweight/GenieReweightProducer.cxx]: We are converting from Python types and so long-story-short we will not support an unsigned int from getParameter. The supported types are
- `list` (1D and 2D) to `std::vector` of type in the first element
- `object` and `dict` to `Parameters` holding their members
- `str` to `std::string`
- `int` to `int`
- `float` to `double`
- `bool` to `bool`

https://github.com/LDMX-Software/ldmx-sw/blob/e3c9d41b105ff88451ebcbad77dbce7ba466830e/Framework/src/Framework/Configure/Python.cxx#L146C1-L152C50

File: SimCore/include/SimCore/Generators/GenieGenerator.h (hunk 8 above)
Comment (tomeichlersmith) [reply SimCore/include/SimCore/Generators/GenieGenerator.h]: Parameters extracted from Python `float` are the size of C++ `double` so we need to initially extract them to `double`. I don't want to support quietly extracting a Python `float` to a C++ `float` for the same reason I don't want to support a Python `int` to a C++ `unsigned int`. I support @wesketchum in keeping the **parameters** `double` as long as the objects that end up in the output event file are `float` in order to save space.

**Comment by tomeichlersmith** [comment]
1. Please refactor the `Print`. You should use a `operator<<(std::ostream&, ldmx::EventWeight&)` so that you can send an event weight to the logging stream and then `Print` just becomes `{ std::cout << *this; }`. See `Recon/Event/HgcrocTrigDigi` for an example. Other event classes don't do this and we just manually construct a logging message from their accessors (which may be what we have to do if this streaming `operator<<` thing doesn't work).
2. You are correct this is due to the logging system. The Boost exceptions are from Boost logging already being "closed" before you try to send messages to it. We do not have a callback in PrimaryGenerators for end-of-run summary statistics. Personally, I'm okay leaving it with `cout` **without the verbosity flag**. This will be a helpful reminder to add a callback to PrimaryGenerators if we want to insert this into the logging system in order to quiet it.

File: SimCore/src/SimCore/Generators/GenieGenerator.cxx (hunk 14 above)
Comment (tvami) [reply SimCore/src/SimCore/Generators/GenieGenerator.cxx:355]: Ok sounds good!

File: SimCore/python/genie_reweight.py (hunk 13 above)
Comment (tvami) [reply SimCore/python/genie_reweight.py]: I agree with Tom's suggestion below 
https://github.com/LDMX-Software/ldmx-sw/pull/1488#issuecomment-2922509201
to have the Print as cout, but to not have the verbosity flag.

File: SimCore/include/SimCore/Generators/GenieGenerator.h (hunk 8 above)
Comment (tvami) [reply SimCore/include/SimCore/Generators/GenieGenerator.h]: OK indeed what matters is the format of what's in the output event file

Comment (EinarElen) [reply SimCore/include/SimCore/Generators/GenieGenerator.h]: Regarding this, I would even argue that we should prioritize using double precision for anything that doesn't end up in the event records. Floating point issues are so much easier to run into with single precision and are a nightmare to deal with

Comment (tvami) [reply SimCore/include/SimCore/Generators/GenieGenerator.h]: Maybe we should make this a dedicated issue @EinarElen because I'm the side of having `float`-s where `float`-s are enough, and having `double`-s only where they are needed.

File: cmake/BuildMacros.cmake (hunk 6 above)
Comment (tomeichlersmith) [reply cmake/BuildMacros.cmake:80]: I wasn't able to figure out the target-based approach so I just formatted it.

File: SimCore/include/SimCore/Event/HepMC3GenEvent.h [hunk 16]
```diff
@@ -0,0 +1,34 @@
...
+
+#include "HepMC3/Data/GenEventData.h"
+#include "HepMC3/GenEvent.h"
+#include "TObject.h"
+
+namespace ldmx {
+
+class HepMC3GenEvent : public HepMC3::GenEventData {
+ public:
+  virtual ~HepMC3GenEvent() {}
```
tomeichlersmith:
```suggestion
  virtual ~HepMC3GenEvent() = default;
```

SimCore/include/SimCore/Reweight/GenieReweightProducer.h
```diff
@@ -0,0 +1,114 @@
...
+  // on new run
+  virtual void onNewRun(const ldmx::RunHeader& runHeader);
+
+  // produce on the event
+  virtual void produce(framework::Event& event);
+
+ private:
+  // verbosity
+  int verbosity_;
+
```
tomeichlersmith:
```suggestion
```

SimCore/python/generators.py
```diff
@@ -193,6 +193,76 @@ def __init__(self,name,initCommands) :
...
+
+    Parameters
+    ----------
+    name : str
+        name of new primary generator
+
+    Attributes
+    ----------
+    verbosity : int, optional
+        Verbosity flag for this generator
```
tomeichlersmith:
```suggestion
```

SimCore/include/SimCore/Generators/GenieGenerator.h
```diff
@@ -0,0 +1,117 @@
...
+ */
+class GenieGenerator : public simcore::PrimaryGenerator {
+ public:
+  /**
+   * Constructor.
+   *
+   * @param parameters Parameters used to configure GENIE generator.
+   *
+   * Parameters:
+   *  verbosity : > 1 means print configuration
```
tomeichlersmith:
```suggestion
```

tomeichlersmith (changes requested):
Just some leftover cleanup, but its looking good :)

SimCore/src/SimCore/Generators/GenieGenerator.cxx
```diff
@@ -0,0 +1,395 @@
...
+  if (!validateConfig())
+    EXCEPTION_RAISE("ConfigurationException", "Configuration not valid.");
+
+  n_events_generated_ = 0;
+  initializeGENIE();
+  calculateTotalXS();
+}
+
+GenieGenerator::~GenieGenerator() {
+  std::cout << "--- GENIE Generation Summary BEGIN ---" << std::endl;
```
tvami, line 250: I'm confused here: I thought these could become part of the logging system, is that not right?

DQM/src/DQM/GenieTruthDQM.cxx
```diff
@@ -0,0 +1,203 @@
...
+  if (interaction_type_ptr)
+    ntuple_.setVar<int>("interaction_type", interaction_type_ptr->value());
+  auto scattering_type_ptr = hepmc3_ev.attribute<HepMC3::IntAttribute>(
+      "GENIE.Interaction.ScatteringType");
+  if (scattering_type_ptr)
+    ntuple_.setVar<int>("scattering_type", scattering_type_ptr->value());
+  auto rescatter_code_ptr =
+      hepmc3_ev.attribute<HepMC3::IntAttribute>("GENIE.RescatterCode");
+  if (rescatter_code_ptr)
+    ntuple_.setVar<int>("rescatter_code", scattering_type_ptr->value());
```
tvami: this is a copy paste error
```suggestion
    ntuple_.setVar<int>("rescatter_code", rescatter_code_ptr->value());
```

DQM/src/DQM/GenieTruthDQM.cxx
```diff
@@ -0,0 +1,203 @@
...
+    ntuple_.setVar<double>("hnuc_e", hitnuc_4vec[3]);
+  }
+  // hit quark info
+  // note: it's only there for some interaction types!
+  auto hqrkpdg_ptr = hepmc3_ev.attribute<HepMC3::IntAttribute>(
+      "GENIE.Interaction.HitQuarkPDG");
+  if (hqrkpdg_ptr) ntuple_.setVar<int>("hqrk_pdg", hqrkpdg_ptr->value());
+  auto hqrksea_ptr = hepmc3_ev.attribute<HepMC3::IntAttribute>(
+      "GENIE.Interaction.HitSeaQuark");
+  if (hqrksea_ptr) ntuple_.setVar<int>("hqrk_sea", hqrkpdg_ptr->value());
```
tvami: here too
```suggestion
  if (hqrksea_ptr) ntuple_.setVar<int>("hqrk_sea", hqrksea_ptr->value());
```

SimCore/src/SimCore/Event/EventWeights.cxx
```diff
@@ -0,0 +1,37 @@
...
+
+std::ostream &operator<<(std::ostream &o, const EventWeights &ew) {
+  o << "Num weights: " << ew.getNumWeights() << std::endl;
+  for (size_t i_w = 0; i_w < ew.getNumWeights(); ++i_w) {
+    o << "\t" << i_w << ": weight=" << ew.getNthWeight(i_w);
+    for (auto const &v : ew.getVariationsNthWeight(i_w)) {
+      o << "\n\t var_type=" << v.first << ", var_value=" << v.second;
+    }
+    o << std::endl;
+  }
```
tvami, line 32: does this not need to return o?
```
return o;
```

tvami (changes requested):
I also found some important typos

SimCore/src/SimCore/Generators/GenieGenerator.cxx
tomeichlersmith: The logging system is already closed by the time the generator desctructor is called.

https://github.com/LDMX-Software/ldmx-sw/pull/1488#issuecomment-2921291585

I said it was okay to leave these to `cout` until we add a callback to the PrimaryGenerator class for end-of-run activities.

tvami: ok thanks!

SimCore/include/SimCore/Event/HepMC3GenEvent.h
wesketchum: done

SimCore/include/SimCore/Reweight/GenieReweightProducer.h
wesketchum: done

SimCore/python/generators.py
wesketchum: done.

SimCore/include/SimCore/Generators/GenieGenerator.h
wesketchum: done

DQM/src/DQM/GenieTruthDQM.cxx
wesketchum: good catch! thank you!

DQM/src/DQM/GenieTruthDQM.cxx
wesketchum: another good catch. thank you!

SimCore/src/SimCore/Event/EventWeights.cxx
wesketchum: yes! done.

tvami (approved):
thanks a lot Wes! all good from my side!

tomeichlersmith (approved):
Looks good to me too :) thank you

tvami:
ups ASAN failed
https://github.com/LDMX-Software/ldmx-sw/actions/runs/15781041353/job/44486508221?pr=1488

tvami:
Do I understand the msg that the issue is that `UserEventInformation` doesnt have a `.cxx` file?

tomeichlersmith:
The `UserEventInformation` does have a implementation file. That error is saying that the `simcore::UserEventInformation` (i.e. the constructor) is not defined for that library. For us, this means the library where `UserEventInformation` is defined (SimCore) has not been loaded when the library where it is being used (SimCore_Generators) is being loaded.
I'm not sure how to solve this problem or why its showing up now, I guess this is the first generator using `UserEventInformation` directly.

wesketchum:
OK, I've managed to get this to run at least one of the failing ASAN-compiled validation checks. I imagine the rest will follow ...

From my understanding, I think there are two options here:

- We make UserEventInformation and other similar ones (UserTrackInformation, UserRegionInformation, and UserPrimaryParticleInformation) header only. 
- We move them to some compiled library (like G4User), and then adjust links and dependencies throughout the code.

I've done the second option in local code, and that appears to be working (well, outside of many warnings on memory leaks, which I think are unrelated). I've not tried the first option...I suspect it would work, and it may be 'cleaner' in the amount of code that needs to be changed. But I don't have strong preferences there.

Assuming those are our options, I'd recommend we decide what we would like to do, make a separate PR that does that reorganization, test that it works for everything in trunk as-is, and then I can pull those changes into my branch for testing with the GenieGenerator.

tomeichlersmith:
I agree the second option is best - started a branch: https://github.com/LDMX-Software/ldmx-sw/tree/re-ord-simcore-libs

wesketchum:
OK, with #1736 now merged into this branch, the build is working and I can run a simulation change. I've pushed up the minor updates needed here.

tomeichlersmith (approved):
Just re-approving because I am ready to merge now.

PR #1489, opened 2024-10-16
Introduce `ldmx-reduced-v2`

### LINKED ISSUE #1483: Reduced v2 Geometry Implementation
Given that we will potentially have a test run in 2025 summer, we should create the expected reduced detector within ldmx-sw to simulate with.

This is an update to the first version of the reduced geometry where now, the absorber widths have been changed. For the six layers, the absorber widths are:

v1: [0, 0, 1, 2, 2, 3]
v2: [2, 4, 4, 4, 4, 4]

We considered moving the layers further back in the ECal, but realized we'd lose sensitivity to wide-angle particles which is important for WAB studies. We are still looking into optimizing absorber widths.

SanjitMasanam:

This resolved #1483 

## Check List

Simulating 3 GeV electrons showering in the ECal, we see a significant improvement in the shower containment of the ECal between v1 and v2.

image

### DIFF
--- a/DetDescr/python/EcalGeometry.py
+++ b/DetDescr/python/EcalGeometry.py
@@ -124,9 +124,8 @@ class EcalGeometry() :
         eg.layer_shift_x = 2*eg.moduleMinR / eg.nCellRHeight
         return eg
         
-        
     def reduced() :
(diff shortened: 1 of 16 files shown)


tvami (approved):
Looks good thanks Sanjit

tomeichlersmith (approved):

PR #1492, opened 2024-11-05
use denv v1.1.0's quiet no-overwrite init flags

tomeichlersmith:
This simplifies `just init` for newer `denv` users and avoids a confusing error message that is ignored because we immediately follow up a missing workspace with creating the workspace.

### DIFF
--- a/justfile
+++ b/justfile
@@ -102,18 +102,28 @@ debug config_py *ARGS:
 # initialize a containerized development environment
 init:
     #!/usr/bin/env sh
-    # while setting the denv_workspace is helpful for other
-    # commands that can assume the denv is already initialized,
-    # we need to unset this environment variable to make sure
-    # the test is done appropriately.
-    # just makes sure this recipe runs from the directory of
-    # the justfile so we know we are in the correct location.
-    unset denv_workspace
-    if denv check --workspace --quiet; then
-      echo "\033[32mWorkspace already initialized.\033[0m"
-      denv config print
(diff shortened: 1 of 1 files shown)


tvami:
I tried to test this, but I get 
```
/run/user/3852/just/just-JaiBXy/init: line 107: [: : integer expression expected
```
on POD when I do `just init` in this branch with `apptainer`

tvami:
Doing it on SLAC, where I had older `denv` I get
```
[tamasvami@sdfiana006 ldmx-sw]$ just init
/run/user/46072/just/just-GnFnRN/init: line 107: [: : integer expression expected
This directory does not exist. Would you like to create it for the denv? [Y/n] Y
mkdir: unrecognized option '--no-mkdir'
Try 'mkdir --help' for more information.
error: Recipe `init` failed with exit code 1
```
so maybe we should add another msg to make sure the user uses new enough `denv`

tomeichlersmith:
Apologies! I should have opened this as draft. I did not do any testing myself on anywhere outside my laptop. I'll mark as ready-for-review when I feel like it is ready.

tomeichlersmith:
@tvami @EinarElen are either of you available to double check this? I feel like its ready, but I don't want to merge it without double checks.

tvami:
I can have a look at this today!

tvami (approved):
I tested this, I have `denv v1.1.1` and it works nicely out of the box

PR #1495, opened 2024-11-15
Use recoil tracking's extrapolated position in EcalVeto

### LINKED ISSUE #1494: Use tracking for projections in EcalVeto

Right now the EcalVeto uses the scoring plane to determine the recoil electron position, and thus the photon projection.

To be more realistic we should use the extrapolation from the recoil tracking to do this.

For real data-taking this has to be done as there is no SP info in data.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1494

Introduced `recoil_from_tracking_` boolean, which if switched on will propagate the recoil track to the ECAL face and use that to make the projections.
For now I kept the default boolean False, and switched it on for the ecalPN sample, where we run the tracking anyway. 

## Check List

Unfortunately the ecalPN sample we have did not save the tracking simhits so I cannot rerun tracking on with.
I ran on the kaon samples tho which did keep the hits. I'll show the results at the SWAN meeting on Monday.

Much of this work was done by @JYoo001 who also made plots quantifying the distance between the projected and real positions in the ECAL (this we showed at a tracking meeting)

### DIFF
--- a/Ecal/src/Ecal/EcalVetoProcessor.cxx
+++ b/Ecal/src/Ecal/EcalVetoProcessor.cxx
@@ -245,18 +245,18 @@ void EcalVetoProcessor::produce(framework::Event &event) {
   }
   // Get recoilPos using recoil tracking
   if (recoil_from_tracking_) {
-    std::vector<float> recoilTrackStates;
+    std::vector<float> recoil_track_states;
     if (verbose_) {
       ldmx_log(debug) << "   Propagate recoil tracks to ECAL face";
     }
     // Get the recoil track collection
     auto recoil_tracks{event.getCollection<ldmx::Track>(track_collection_)};
 
-    ldmx::TrackStateType tsType = ldmx::TrackStateType::AtECAL;
-    recoilTrackStates = trackProp(recoil_tracks, tsType, "ecal");
+    ldmx::TrackStateType ts_type = ldmx::TrackStateType::AtECAL;
+    recoil_track_states = trackProp(recoil_tracks, ts_type, "ecal");
     // Redefining recoilPos now to come from the track state
     // track_state_loc0 is recoilPos[0] and track_state_loc1 is recoilPos[1]
-    recoilPos = recoilTrackStates;
+    recoilPos = recoil_track_states;
   }
 
   if (verbose_) {
@@ -1283,10 +1283,10 @@ std::vector<float> EcalVetoProcessor::trackProp(const ldmx::Tracks &tracks,
                                                 ldmx::TrackStateType ts_type,
                                                 const std::string &ts_title) {
   // Vector to hold the new track state variables
-  std::vector<float> newTrackStates;
+  std::vector<float> new_track_states;
 
   // Return if no tracks
-  if (tracks.empty()) return newTrackStates;
+  if (tracks.empty()) return new_track_states;
 
   // Otherwise loop on the tracks
   for (auto &track : tracks) {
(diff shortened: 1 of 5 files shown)


tvami:
OK the non-ecal-pn tests failed just because I added an extra printout for the BDT score, but all the K-S passes. 
Otherwise, for the ecal-pn, there is quite some change: 
1) fidiciality changes as expected
 * recoil y a little: 
Screenshot 2024-11-15 at 21 16 29

 * recoil x quite a lot:
Screenshot 2024-11-15 at 21 17 42

Honestly the y distribution with the recoil tracking seems more natural. Sure we expect this to be asymmetric. But I dont think we expect it to be completely on the left side, so having some tracking ending in the right looks OK to me. One thing that's interesting is that the tracking propagation should include both magnetic field and material. I dont actually know how the SP y is made.

2) This will change the projections
 * photon separation
Screenshot 2024-11-15 at 21 19 29

 * ep angle
Screenshot 2024-11-15 at 21 20 21

looks less peaky in the high end, which to me a good sign

 * photon containment energy
Screenshot 2024-11-15 at 21 20 50

 * ele cont energy
Screenshot 2024-11-15 at 21 21 24

3) As a consequence the MIP tracking changes too 
 * lin-reg tracks
 
Screenshot 2024-11-15 at 21 22 36

I'm kinda surprised this failed the K-S, usually the changes in the tails dont bother the K-S, but ok, from first principles this is expected

 * str8 tracks
Screenshot 2024-11-15 at 21 22 15

So here it's a bit weird, I don't know what to make of it. It's not bad for a cutflow tho.

tvami:
@therwig I thought I'd bring your attention to the validation plots I posted here too. Quite a lot more tracks in the ecal PN as well

tvami:
I was talking with @danyi211  about this, she pointed out that if the recoil track is on the left, that means that the photon projection is on the right. But if you mismeasure by the tracking that the recoil track is on the right, then the photon projection will be on left -- but that's where the real electron shower is!! so you'll find MIP tracks bc there are MIP tracks from the electron.

therwig:
Thanks for sharing these - interesting. The straight tracks are quite striking :)
Thinking about your explanation, can we correlate this pathological reconstruction with the electron containment energy?  I wonder if some requirement that E/p~1 (something loose, like within ~2 sigma?) or some level of track-to-cluster matching for the electron would clean up this case.  We could also look at track quality.  
OK, probably you already had some similar ideas about how to fix this, but its gratifying to see a nice explanation here!

Ecal/src/Ecal/EcalVetoProcessor.cxx
```diff
@@ -1261,6 +1279,44 @@ float EcalVetoProcessor::distPtToLine(TVector3 h1, TVector3 p1, TVector3 p2) {
...
+  // Return if no tracks
+  if (tracks.empty()) return newTrackStates;
+
+  // Otherwise loop on the tracks
+  for (auto &track : tracks) {
+    // Get track state for ts_type
+    auto trk_ts = track.getTrackState(ts_type);
+    // Continue if there's no value
+    if (!trk_ts.has_value()) continue;
+    ldmx::Track::TrackState &TargetState = trk_ts.value();
```
tvami: This should actually be 
```suggestion
    ldmx::Track::TrackState &EcalTrackState = trk_ts.value();
```

tomeichlersmith: Also, please have the variable be `snake_case` instead of `CamelCase` so that it is easy to visually distinguish from classes (`CamelCase`) and functions (`someWeirdJavaThing`).

Ecal/src/Ecal/EcalVetoProcessor.cxx
```diff
@@ -1261,6 +1279,44 @@ float EcalVetoProcessor::distPtToLine(TVector3 h1, TVector3 p1, TVector3 p2) {
...
+    if (TargetState.params.size() < 5) continue;
+
+    float track_state_loc0 = static_cast<float>(TargetState.params[0]);
+    float track_state_loc1 = static_cast<float>(TargetState.params[1]);
+
+    // Store the new track state variables
+    newTrackStates.push_back(track_state_loc0);
+    newTrackStates.push_back(track_state_loc1);
+    // z-position at the ECAL scoring plane
+    newTrackStates.push_back(239.999);
```
tomeichlersmith: I fear that this hard-coded number will lead to an obscure bug in the future. Can we link this value to the value that gets passed to `getTrackState`?

Also not sure if we need this many decimal points. Why not `240.`?

tvami: I share your fear, but the track state does not save the Z position, it just does it at the ECAL SP's position. I wish the parameters in the GMDL could be invoked here. I think it's not 240, so it doesnt overlap with the real ECAL in the GDML. I'm open to ideas for how to not have this hard-coded.

tomeichlersmith: Let's leave it hardcoded for now. I think resolving this would require some more serious development of the Tracking infrastructure that no one has the time to do right now. I forgot that the TrackStates are pre-computed at Tracking-time and I feel like the z-locations of the track states that are pre-computed should be hardcoded along with their names.

tvami: >  hardcoded along with their names.

well it adds to the confusing that in tracking X axis = beam axis (i.e. our Z). But yeah this should be indeed done in tracking. I'll "resolve" this comment now then

tvami:
We figured out that the pathological events can be killed by requiring 1 recoil track. I have had this comment in the code saying that we need to match the extrapolated track to the CLUE cluster, that will lead to the same results, but that needs the CLUE PR to be merged.
Screenshot 2025-01-10 at 13 30 09

So I propose we merge this in this state and will have the promised improvements as CLUE merges. Please also note the use of the tracking in Ecal is OFF by default. I also plan to include ACTS tracking in the signal + kaon validation configs so we can enable this by default in the future. But my point is that merging this will cause no harm.

danyi211 (approved):

tomeichlersmith (approved):

Issue #1491, opened 2024-11-04
Adding a reduced geometry tracking producer


fdelzanno:
**Is your feature request related to a problem? Please describe.**
Adding a producer for running tracking on samples with the reduced geometry within ldmx-sw/Tracking

**Describe the solution you'd like**
Within ldmx-sw/Tracking, add a file(s) that: 1) transforms digi hits on recoil double-layers into 3D positions in space with sensor uncertainty, 2) perform linear track fitting on combinations of Recoil Tracker hits and a reference measurement in the first layer of the Ecal, 3) do tracking by selecting the fitted tracks by minimizing the track chi-2 value.

**Describe alternatives you've considered**
We considered modifying the existing processor in ldmx-sw/Tracking to handle the reduced geometry, but initial results were poor. 

**Additional context**

fdelzanno:
The tracking logic is as follows:

**LinearSeedFinder.cxx** - this is the "seeder", which:
- Converts hits on the recoil layers (from RecoilDigiSim) into "sensor" hits. There are two sensors, made from an energy-weighted averaging of the positions on Layers 1 and 2, and those on Layers 3 and 4. Uncertainty from the Recoil geometry is assigned to the position in x and y. 
- Constructs every combination of 2 sensor hits + 1 EcalRecHit on the first layer of the Ecal and uses those points to fit a straight line in 3D. Using the track parameters of the seed, we extrapolate the track to the first layer of the Ecal, and keep only seeds within a threshold distance of the RecHit used in fitting.

**LinearTrackFinder.cxx** - this is the track processor, which:
- Takes the seed collection from the LinearSeedFinder and processes them by the EcalRecHit associated with the seed.
- By iterating through all the seeds corresponding to a RecHit, we select the seed with the lowest chi2 value as the track for that RecHit. 
- We remove any remaining seed that uses the two sensors and/or EcalRecHit used by the "best track". If there are any seeds left, we select the next EcalRecHit, find the best track, and remove seeds with shared hits -- iterating until we have fit all RecHits or processed all seeds. 

**LinearTruthTracking.cxx** - this is the "truth tracking", which:
- Finds 4 points in the RecoilDigiSim Hits with trackID == 1 (i.e. the recoil electron hits) and fits a straight line to those 4 points.
- The output track is the true Recoil electron for that event.

[...]

Issue #1486, opened 2024-10-07
Extend ECAL DQM histogram bins


tvami:
**Is your feature request related to a problem? Please describe.**

I think the ECAL DQM histo bins were selected with 4 GeV was in mind. 
Now if we shoot an 8 GeV beam into the ECAL the plots looks like this
<img width="1200" alt="Screenshot 2024-10-07 at 15 48 52" src="https://github.com/user-attachments/assets/c9f1223c-79f0-4024-a19b-312d584f650b">

**Describe the solution you'd like**
Change the 8000 to 10000 here
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/DQM/python/dqm.py#L270
Possibly other places should be updated too. I can do a more diligent review when I get to this

tvami:
+ add BDT pass / fail histogram
+ rename lin-reg so it's not called tracks, extend the bins there too

PR #1497, opened 2024-11-20
Optimize ECAL lin-reg MIP tracking

### LINKED ISSUE #1420: Optimize ECAL lin-reg

This is a follow-up to https://github.com/LDMX-Software/ldmx-sw/pull/1407
There are several things to make the lin-reg better. List below.

1) Implement that we really are selecting hits from the closest hits
https://github.com/LDMX-Software/ldmx-sw/blob/e6c6935a7b0426d5302b270e2f10bc08737f33f9/Ecal/src/Ecal/EcalVetoProcessor.cxx#L996
and then
https://github.com/LDMX-Software/ldmx-sw/blob/e6c6935a7b0426d5302b270e2f10bc08737f33f9/Ecal/src/Ecal/EcalVetoProcessor.cxx#L1028
it should prob take the hits from `trackingHitList[hitsInRegion[jHit]]` and also `trackingHitList[hitsInRegion[kHit]]` and propagate this below in the code too.
2) That opens up immediately that 50 in `int hitsInRegion[50];` might be not realistic
3) And then the distance in https://github.com/LDMX-Software/ldmx-sw/blob/e6c6935a7b0426d5302b270e2f10bc08737f33f9/Ecal/src/Ecal/EcalVetoProcessor.cxx#L1018 was not optimized at all.

Besides the physics changes discussed above we could also consider to move away from the ROOT approach and use Eigen instead.

Keep it as it is, currently it's not really used, but I think it has potential.

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1420

## Check List

This is a work together with @kawaterrysit
We optimized kaons vs signal, I'll attach the plot when we fully converge, but the 23 mm seems already pretty good. Edit: the best option actually seems to be 35 mm.

With the original setting we really were just looking in the front ECAL since the 9 mm setting would not reach the next layer in the back ECAL. So we expect much more tracks to be built like this.

### DIFF
--- a/Ecal/include/Ecal/EcalVetoProcessor.h
+++ b/Ecal/include/Ecal/EcalVetoProcessor.h
@@ -148,6 +148,7 @@ class EcalVetoProcessor : public framework::Producer {
   double bdtCutVal_{0};
 
   double beamEnergyMeV_{0};
+  double linreg_radius_{0};
 
   bool verbose_{false};
 
--- a/Ecal/python/vetos.py
+++ b/Ecal/python/vetos.py
@@ -21,7 +21,7 @@ class EcalVetoProcessor(ldmxcfg.Producer) :
         self.bdt_file = makeBDTPath( "segmip" )
         self.roc_file = makeRoCPath( 'RoC_v14_8gev' )
         self.beam_energy = 8000.0  # in MeV
-        self.linreg_radius = 23.0 # in mm
+        self.linreg_radius = 35.0 # in mm
         self.disc_cut = 0.99741
         self.collection_name = "EcalVeto"
         self.rec_coll_name = 'EcalRecHits'
--- a/Ecal/python/vetos.py
+++ b/Ecal/python/vetos.py
@@ -21,6 +21,7 @@ class EcalVetoProcessor(ldmxcfg.Producer) :
         self.bdt_file = makeBDTPath( "segmip" )
         self.roc_file = makeRoCPath( 'RoC_v14_8gev' )
         self.beam_energy = 8000.0  # in MeV
+        self.linreg_radius = 23.0 # in mm
         self.disc_cut = 0.99741
         self.collection_name = "EcalVeto"
         self.rec_coll_name = 'EcalRecHits'
--- a/Ecal/src/Ecal/EcalVetoProcessor.cxx
+++ b/Ecal/src/Ecal/EcalVetoProcessor.cxx
@@ -127,6 +127,7 @@ void EcalVetoProcessor::configure(framework::config::Parameters &parameters) {
   ecalLayerTime_.resize(nEcalLayers_, 0);
 
   beamEnergyMeV_ = parameters.getParameter<double>("beam_energy");
+  linreg_radius_ = parameters.getParameter<double>("linreg_radius");
 
   // Set the collection name as defined in the configuration
   collectionName_ = parameters.getParameter<std::string>("collection_name");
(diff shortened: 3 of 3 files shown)


tvami:
> With the original setting we really were just looking in the front ECAL since the 9 mm setting would not reach the next layer in the back ECAL. So we expect much more tracks to be built like this.

And indeed we see much more tracks.
For signal
Screenshot 2024-11-19 at 20 14 08

For the kaons
Screenshot 2024-11-19 at 20 14 31

For ecal PN
Screenshot 2024-11-19 at 20 14 44


The plot that shows why 23 is pretty good already will come tomorrow

tvami:
Here is the presentation that shows the effect of the different values
[Optimizing Linear Regression MIP Tracking.pdf](https://github.com/user-attachments/files/17852007/Optimizing.Linear.Regression.MIP.Tracking.pdf)

tvami:
signal
Screenshot 2024-11-22 at 12 27 35

kaons
Screenshot 2024-11-22 at 12 23 39

ecalPN
Screenshot 2024-11-22 at 12 24 04

tomeichlersmith (commented):
This feels like a lot of tracks, but I also suppose we have low requirements and an entire shower to analyze.

One feature I would like to be considered is to "extend" tracks. How many of these tracks are just several 3-hit segments of something we would consider a single track? This may be too big of an ask and is venturing close to doing full-scale tracking which we could probably just use ACTS for at some point.

tvami:
Yeah calling these "tracks" is actually an overstatement. All of these are 3-hit segments, so if you have a long track that makes 15 hits, that's going to lead to 5 "track" on this plot. And that's one reason for the inflation. The other is that in the past with the old radius we technically didnt use the back ECAL at all, so now adding all those makes more "tracks". 

I think you have a good point, but I dont see how to do it with the current code. I kinda agree that we should just move away from the lin-reg approach for the further future. We actually played with ACTS w/ B=0 field for the reduced tracker that didnt really work out well, but the algo mentioned in https://github.com/LDMX-Software/ldmx-sw/issues/1491 could be explored for ECAL usage too

tomeichlersmith (commented):
Yea, the ECal situation is weird. We don't need a good track fit, we just want to be able to algorithmically recognize tracks out of a sea of hits.

I think this is good as is, I'm glad to hear there are two reasons for such a large increase in the number of "tracks". Since both signal and background overflow our histograms, maybe we need to extend the axis in DQM as well? Might need to be reserved for future, I have no idea how that will interact with our CI testing.

I also am _very_ invested in using good vocabulary, so maybe the histograms and some other important locations could use a different word that track? tracklets? track segments?

tvami:
I propose to extend and rename in the PR that addresses https://github.com/LDMX-Software/ldmx-sw/issues/1486 

I do actually plan to tackle that soon too, although next week is tricky with Thanksgiving. But it should prob be done before the next ldmx-sw is cut.

Here is a normalized plot with an overflow bin btw:
Screenshot 2024-11-22 at 16 46 58

(this is after trigger and fiducial, seems to change the ecalPN quite a lot)

tomeichlersmith (approved):

danyi211 (approved):

tvami:
Ok I'm gonna merge this now, and it will make the golds obsolete, but I think we should not yet cut a new release until some of the other PRs with the DQM histo changes are in. Unless it's useful for other purposes

tvami:
> I propose to extend and rename in 

This is now done in https://github.com/LDMX-Software/ldmx-sw/pull/1508

PR #1498, opened 2024-11-25
Modified the LYSO target design

### LINKED ISSUE #1370: LYSO active target for ldmx-lyso-r2-v14-8gev

zwl0331:

This resolves #1370 and implements the new design for the LYSO target. It consiste of two layers. The first layer has 32 bar, while the second has 33 bars. 

## Check List


bryngemark:
Hi Sam,
if your PR is ready for review (necessary before it gets merged), you need to pick at least two reviewers who will take a look and approve when satisfied. @tvami, @tomeichlersmith or myself would all be qualified. 

To help us understand how you know that your changes are good, please upload some plots in this discussion thread, explain what you ran to get them, and what they show. (This is the purpose of the second box, that you already checked when you made the PR. I unchecked it until you have shared some more material.)

zwl0331:
I implemented the following geometry:
Picture1
In Geant4 this looks like this:
- Side view:
Screenshot 2024-11-25 at 5 26 02 PM

- Front view:
Screenshot 2024-11-25 at 5 26 36 PM

tomeichlersmith (commented):
Thank you for the images! Those look great :)

I just want to confirm that this runs and the geometry doesn't have any sneaky issues (e.g. overlaps that are not visible because they are small but could cause issues during the simulation). Can you share a configuration you've used to confirm that this geometry successfully runs a simulation?

We can then run the same configuration with

```python
# sim is the simulator configuration object already created
sim.validate_detector = True
sim.verbosity = 5
```
so that ldmx-sw takes extra time to make sure the geometry is valid.

zwl0331:
I have talked to Dr. James Oyang, @joyang8caltech, and he agreed to run some simulations using this geometry. Dr. Oyang, I forked a version of ldmx-sw into my github account and made the geometry modifications. It is at https://github.com/zwl0331/ldmx-sw You can clone it to run simulations with it.

Detectors/data/ldmx-lyso-r1-v14-8gev/constants.gdml
```diff
@@ -11,13 +11,48 @@
...
+  <!-- target array2-->
+<constant name="number_of_target_bars2" value="33" />
+<constant name="target_array2_dx" value="target_bar_dx" />
+<constant name="target_array2_dy" value="number_of_target_bars2 * target_bar_dy + (number_of_target_bars2 - 1) * target_bar_gap" />
+<constant name="target_array2_thickness" value="target_bar_thickness" />
+
+  <!-- target box-->
+<constant name="target_dim_x" value="target_bar_dx" />
+<constant name="target_dim_y" value="target_array2_dy" />
+<constant name="target_thickness" value="2*target_bar_thickness + target_bar_gap" />
```
tvami: Can you add a comment what this adds up to? Is it still 1.1 mm? If not what's the radiation length of the new thickness?

Detectors/data/ldmx-lyso-r1-v14-8gev/constants.gdml
```diff
@@ -11,13 +11,48 @@
...
+
+<!-- Target positions -->
+
+  <!-- target array1-->
+<constant name="target_array1_first_bar_y" value="-(number_of_target_bars1 - 1) * (target_bar_dy + target_bar_gap) / 2" />
```
tvami: I tried to identify what's array1 and array2 in the picture you attached, but I dont see it, can you please clarify?

Detectors/data/ldmx-lyso-r1-v14-8gev/target.gdml
```diff
@@ -83,7 +113,7 @@
       <materialref ref="Vacuum"/>
       <solidref ref="target_area_box" />
 
-      <loop for="x" to="number_of_bars" step="1">
+      <loop for="x" from="1" to="number_of_bars" step="1">
```
tvami: this is not bad, but since you had 
```
<variable name="x" value="1" />
```
this wont change anything. Anyway it's fine to change it

Detectors/data/ldmx-lyso-r1-v14-8gev/target.gdml
```diff
@@ -100,16 +130,12 @@
         </physvol>
       </loop>
 
-      <physvol copynumber="3">
-        <volumeref ref="target"/>
+      <physvol name="lvTarget_phys">
```
tvami: here too, I think when we add the sensitive volumes later on, that will look at the copy number

Detectors/data/ldmx-lyso-r1-v14-8gev/target.gdml
```diff
@@ -100,16 +130,12 @@
...
+        <volumeref ref="lvTarget"/>
         <positionref ref="center"/>
         <rotationref ref="identity"/>
       </physvol>
 
-      <auxiliary auxtype="Region" auxvalue="target" />
-      <!-- <auxiliary auxtype="VisAttributes" auxvalue="InvisibleShowDau"/> -->
-      <auxiliary auxtype="DetElem" auxvalue="target"/>
-
```
tvami: again this aux info can be helpful, I'd suggest to keep them.

tvami (changes requested):
Sorry it took me some time to look at this (DR writing...), I made some suggestions / questions in below.

Generarly I'd really like to see some tests that use this geometry, furthermore I'd suggest to either copy or rename the whole directory to r2, i.e. `ldmx-lyso-r2-v14-8gev` the `r` was for the revision, so it was exactly meant for this purposes. If you chose to rename, please add the info in the README that v4.1.4 is the last one to have `ldmx-lyso-r1-v14-8gev`

Detectors/data/ldmx-lyso-r1-v14-8gev/target.gdml
bryngemark: i think making this explicit is a good idea! navigating gdml files for variable definitions is usually not very fun.

bryngemark:
> I have talked to Dr. James Oyang, @joyang8caltech, and he agreed to run some simulations using this geometry. Dr. Oyang, I forked a version of ldmx-sw into my github account and made the geometry modifications. It is at https://github.com/zwl0331/ldmx-sw You can clone it to run simulations with it.

How is this going? Please see the specific geometry validation instructions posted here by @tomeichlersmith.

I also wonder, @tomeichlersmith and @tvami, if we should have a validation workflow for geometry changes? (This is not the last time we'll make an update.) We do expect a whole lot of changes to distributions, so it's not clear what tests are meaningful, apart from that things run, and the overlap check tom mentioned.

bryngemark:
By the way @zwl0331, @joyang8caltech, the plastic geometry doesn't look right. We expect to have 24 bars in each plastics array, so they are equal height but staggered in _y_.

joyang8caltech:
We need to have thirty two 3mm height and 40mm long LYSO bars so that the active LYSO target will have the same coverage as tungsten target.

On Dec 16, 2024, at 6:17 AM, bryngemark ***@***.***> wrote:



By the way @zwl0331<https://github.com/zwl0331>, @joyang8caltech<https://github.com/joyang8caltech>, the plastic geometry doesn't look right. We expect to have 24 bars in each plastics array, so they are equal height but staggered in y.

tvami:
> if we should have a validation workflow for geometry changes?

I agree we need this, but I think we should have this after the sensitive volumes are defined, then we can do a PN sample with LYSO in the CI

The validate_detector doesnt work, see https://github.com/LDMX-Software/ldmx-sw/issues/1533

So we need to do this semi-blind...

@zwl0331 can you please update the code with the comments from me and Tom?

Detectors/data/ldmx-lyso-r1-v14-8gev/constants.gdml
zwl0331: This adds up to 1.34 mm. I will add a comment in the file as well. @joyang8caltech may know the radiation length.

Detectors/data/ldmx-lyso-r1-v14-8gev/constants.gdml
zwl0331: I will add a comment in the code here, but layer 1 is the layer with 1 less LYSO bar than layer 2, so it's the one closer to the TS.

Detectors/data/ldmx-lyso-r1-v14-8gev/target.gdml
zwl0331: Okay. I have added the copynumber back.

Detectors/data/ldmx-lyso-r1-v14-8gev/constants.gdml
zwl0331: According to [this paper](https://cds.cern.ch/record/2748152/files/1912.00397.pdf) from CERN, the radiation length of LYSO is about 11.2 mm. The 2 laryers of target bar have total thickness 1.2 mm (not including the gap in between), which is about ~10% of the radiation length.

tvami: Great, please add these as a comment in the GDML

zwl0331:
I have successfully built ldmx-sw with the new geometry ran a simple simulation with it. Here is the simulation configuration:
```
p = ldmxcfg.Process( "sim" )
import sys
p.maxEvents = 10 #Note: use small number as you really are just checking the geom
if len(sys.argv) > 1 :

# we want to see every event
p.logFrequency = 1 if p.maxEvents <= 10 else 100
p.termLogLevel = 1

# Set a run number
p.run = 9001

p.outputFiles = [ "elecGun_8GeV.root" ]

import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry
mySim.setDetector( 'ldmx-det-v14-8gev' , True )
#sim.beamSpotSmear = [20., 80., 0.]

# Get a pre-written generator

myGun = gen.gun('myGun')
myGun.particle = "e-" #'neutron','mu-','pi-','e-','gamma'
myGun.position = [ 0., 0., 0. ]
myGun.direction = [ 0., 0., 1.]
myGun.energy = 8.0 #GeV
mySim.generators.append( myGun )

# sim is the simulator configuration object already created--> From Tom
sim.validate_detector = True
sim.verbosity = 5

p.sequence.append( mySim )

# #Note: reco stuff --> use what you need, probably you dont care too much about this stuff
import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry
import LDMX.Ecal.vetos as ecal_vetos
import LDMX.Hcal.digi as hcal_digi

from LDMX.Hcal import hcal_trig_digi
from LDMX.Ecal import ecal_trig_digi
from LDMX.Hcal import hcal
p.sequence.extend([
        ecal_digi.EcalDigiProducer(),
        ecal_digi.EcalRecProducer(),
        hcal_digi.HcalDigiProducer(),
        hcal_digi.HcalRecProducer()
        ])
```
The simulation ran successfully, giving the following output:
```
zwl0331@Samoffice:~/Documents/Research/LDMX/software/ldmx-sw$ sudo env "PATH=$PATH" denv fire elecGun.py
[sudo] password for zwl0331:
[ Process ] 1  info: Processing 1 Run 9001 Event 1  (2025-02-03 20:28:04.616878000+0000)
[ Process ] 2  info: Processing 2 Run 9001 Event 2  (2025-02-03 20:28:05.031306001+0000)
[ Process ] 3  info: Processing 3 Run 9001 Event 3  (2025-02-03 20:28:05.328579000+0000)
[ Process ] 4  info: Processing 4 Run 9001 Event 4  (2025-02-03 20:28:05.613524001+0000)
[ Process ] 5  info: Processing 5 Run 9001 Event 5  (2025-02-03 20:28:05.908598000+0000)
[ Process ] 6  info: Processing 6 Run 9001 Event 6  (2025-02-03 20:28:06.211264000+0000)
[ Process ] 7  info: Processing 7 Run 9001 Event 7  (2025-02-03 20:28:06.502809000+0000)
[ Process ] 8  info: Processing 8 Run 9001 Event 8  (2025-02-03 20:28:06.796944000+0000)
[ Process ] 9  info: Processing 9 Run 9001 Event 9  (2025-02-03 20:28:07.090554000+0000)
[ Process ] 10  info: Processing 10 Run 9001 Event 10  (2025-02-03 20:28:07.383509000+0000)
[ Process ] 10  info: RunHeader { run: 9001, numTries: 10, detectorName: ldmx-det-v14-8gev, description:
  intParameters:
    Included Scoring Planes = 1
    RandomNumberMasterSeed[sim] = 9001
    Use Random Seed from Event Header = 0
  floatParameters:
    Gen0 Direction X = 0
    Gen0 Direction Y = 0
    Gen0 Direction Z = 1
    Gen0 Energy [GeV] = 8
    Gen0 Time [ns] = 0
    Gen0 X [mm] = 0
    Gen0 Y [mm] = 0
    Gen0 Z [mm] = 0
  stringParameters:
    Geant4 revision =
    Gen0 Class = simcore::generators::ParticleGun
    Gen0 Particle = e-
    ldmx-sw revision = 2d449dc907f3b131b1c3c48c575695d0eae795f0
    ldmx-sw version = 4.2.8
}
```

zwl0331:
> By the way @zwl0331, @joyang8caltech, the plastic geometry doesn't look right. We expect to have 24 bars in each plastics array, so they are equal height but staggered in _y_.

Sorry about this. I put up the wrong figures. The correct ones are these. As you can see, they are 24 bars on both TS layers and they are staggered in y.
Screenshot 2025-02-03 at 12 37 17 PM
Screenshot 2025-02-03 at 12 37 35 PM

tomeichlersmith:
There are two points of confusion with your configuration.

```python
sim.validate_detector = True
sim.verbosity = 5
```
As written, this won't work. `sim` is the `simulator` module and not the object your created `mySim`. It is just being silently ignored right now, but don't worry about that. This check won't work anyways because Geant4 can't seem to download the schema necessary to do the check itself. #1533 

```python
mySim.setDetector( 'ldmx-det-v14-8gev' , True )
```
This is saying to run the simulation with the `ldmx-det-v14-8gev` detector. You are adding the `ldmx-lyso-r2-v14-8gev` detector, so the simulation you have run is not using the detector you have been editing.

tomeichlersmith:
Unrelated, you can avoid having to use `sudo` and type your password in if you give your user access to manage the docker group on your computer.

https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

bryngemark:
thanks @zwl0331 for sharing your config file! as you can see this was a great way to catch some issues with your validation procedure. 
in addition to what @tomeichlersmith has spotted, i also notice that you are manually setting up an electron gun to start at the origin (`myGun.position = [ 0., 0., 0. ]`) , which ought to be in the middle of the LYSO target. I suggest using the predefined generator `single_8gev_e_upstream_tagger()` (double check the exact name) for this instead.

zwl0331:
> thanks @zwl0331 for sharing your config file! as you can see this was a great way to catch some issues with your validation procedure. in addition to what @tomeichlersmith has spotted, i also notice that you are manually setting up an electron gun to start at the origin (`myGun.position = [ 0., 0., 0. ]`) , which ought to be in the middle of the LYSO target. I suggest using the predefined generator `single_8gev_e_upstream_tagger()` (double check the exact name) for this instead.

Thanks for the comments! This is the new simulation configuration I ran:
``` python
p = ldmxcfg.Process( "sim" )
import sys
p.maxEvents = 10 #Note: use small number as you really are just checking the geom
if len(sys.argv) > 1 :

# we want to see every event
p.logFrequency = 1 if p.maxEvents <= 10 else 100
p.termLogLevel = 1

# Set a run number
p.run = 9001

p.outputFiles = [ "elecGun_8GeV.root" ]

import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry
mySim.setDetector( 'ldmx-lyso-r2-v14-8gev' , True )
#sim.beamSpotSmear = [20., 80., 0.]

# Get a pre-written generator

myGun = gen.single_8gev_e_upstream_tagger()
mySim.generators.append( myGun )

# sim is the simulator configuration object already created--> From Tom
sim.validate_detector = True
sim.verbosity = 5

p.sequence.append( mySim )

# #Note: reco stuff --> use what you need, probably you dont care too much about this stuff
import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry
import LDMX.Ecal.vetos as ecal_vetos
import LDMX.Hcal.digi as hcal_digi

from LDMX.Hcal import hcal_trig_digi
from LDMX.Ecal import ecal_trig_digi
from LDMX.Hcal import hcal
p.sequence.extend([
        ecal_digi.EcalDigiProducer(),
        ecal_digi.EcalRecProducer(),
        hcal_digi.HcalDigiProducer(),
        hcal_digi.HcalRecProducer()
        ])
```
And it ran successfully:
```
zwl0331@Samoffice:~/Documents/Research/LDMX/software/ldmx-sw$ denv fire elecGun.py
[ Process ] 1  info: Processing 1 Run 9001 Event 1  (2025-02-06 00:42:07.534521000+0000)
[ Process ] 2  info: Processing 2 Run 9001 Event 2  (2025-02-06 00:42:07.919035000+0000)
[ Process ] 3  info: Processing 3 Run 9001 Event 3  (2025-02-06 00:42:08.215431001+0000)
[ Process ] 4  info: Processing 4 Run 9001 Event 4  (2025-02-06 00:42:08.511326000+0000)
[ Process ] 5  info: Processing 5 Run 9001 Event 5  (2025-02-06 00:42:08.815088000+0000)
[ Process ] 6  info: Processing 6 Run 9001 Event 6  (2025-02-06 00:42:09.110338000+0000)
[ Process ] 7  info: Processing 7 Run 9001 Event 7  (2025-02-06 00:42:09.396317000+0000)
[ Process ] 8  info: Processing 8 Run 9001 Event 8  (2025-02-06 00:42:09.690388000+0000)
[ Process ] 9  info: Processing 9 Run 9001 Event 9  (2025-02-06 00:42:09.979721000+0000)
[ Process ] 10  info: Processing 10 Run 9001 Event 10  (2025-02-06 00:42:10.281946000+0000)
[ Process ] 10  info: RunHeader { run: 9001, numTries: 10, detectorName: ldmx-lyso-r2-v14, description:
  intParameters:
    Included Scoring Planes = 1
    RandomNumberMasterSeed[sim] = 9001
    Use Random Seed from Event Header = 0
  floatParameters:
    Gen0 Direction X = 0.0485658
    Gen0 Direction Y = 0
    Gen0 Direction Z = 0.99882
    Gen0 Energy [GeV] = 8
    Gen0 Time [ns] = 0
    Gen0 X [mm] = -21.7459
    Gen0 Y [mm] = 0
    Gen0 Z [mm] = -883
  stringParameters:
    Geant4 revision =
    Gen0 Class = simcore::generators::ParticleGun
    Gen0 Particle = e-
    ldmx-sw revision = 8c1b6fcdf39f0055ee138d14d26efe25e65d80f9
    ldmx-sw version = v4.2.9
}
[ Simulator ] : Started 10 events to produce 10 ev
```

tvami:
I'd like to point out that 
1) This doesnt validate the geometry (we cant until https://github.com/LDMX-Software/ldmx-sw/issues/1533 is resolved)
2) This does check that passive element part, but then how different are the results from running it thru a W target?
3) For the r1 version of this we used signal to compare, maybe that's the best way to do this now too
4) Or we should implement the sensitive volume and then have a pn interaction on it (and this is what I think we should have in the CI when we have it)

bryngemark:
thanks @tvami, those are some good points. @zwl0331 how is this coming along?

zwl0331:
> thanks @tvami, those are some good points. @zwl0331 how is this coming along?

I cannot validate the geometry until #1533 is resolved.
As for the rest, @joyang8caltech has agreed to take on the tasks, because I am mainly working on a different experiment.

tvami:
This PR is from Sam's trunk to ldmx-sw's trunk, which is not the recommended way of doing it: insted it should be a branch in ldmx-sw to trunk in ldmx-sw. So now I modified the target of this PR to go into `LDMX-Software:iss1370-lyso-new-geom`. I'll merge this and then immediately open a PR, which should be totally equivalent to this, but it will allow us to rebase and do some other changes collaboratively

PR #1501, opened 2024-12-23
add function to shoot electrons with settable energy  from upstream 

### LINKED ISSUE #1500: Add capability to shoot beam electrons starting upstream of magnet, with settable energies
For the off-energy beam-electrons study we need to start the electrons before they feel any field from the analyzing magnet. 

We need to add a single-particle gun configuration that starts upstream of magnet with correct position & direction for nominal (8GeV , arriving at center of target) electrons that also has settable electron energy.  

This can all be done in generators.py and we already know the position and direction at ~8m upstream of target (plenty of distance for zero field effect).  It's just a few lines of python given the existing functions.

bloodyyugo:
Resolves iss1500

### DIFF
--- a/SimCore/python/generators.py
+++ b/SimCore/python/generators.py
@@ -311,7 +311,7 @@ def single_8gev_e_upstream_tagger():
     )
 
 
-def single_8gev_beam_pipe(ene = 8.0):
+def single_e_beam_pipe(ene = 8.0):
     """Configure a particle gun to fire an electron of settable energy
     upstream of the tagger tracker.  
 
--- a/SimCore/python/generators.py
+++ b/SimCore/python/generators.py
@@ -311,6 +311,28 @@ def single_8gev_e_upstream_tagger():
     )
 
 
+def single_8gev_e_upstream_magnet_setEnergy(ene=8.0): 
+    """Configure a particle gun to fire an electron of settable energy
+    upstream of the tagger tracker.  
+
+    The starting position here is well upstream of the analyzing magnet
+    the position/angle of the gun is such that 8 gev electrons arrive 
+    at the target z=0 at xy=(0,0).  This generator is used to study 
+    off-energy beam electrons.  
+
+    Note that if an energy != 8gev, the trajectory will be different. 
+
+    Returns
+    -------
+    Instance of a particle gun configured to fire a single 8 GeV electron 
+    upstream of the entire detector apparatus.
+    """
+    return _single_e_upstream_tagger(
+        [ -299.2386690686212, 0.0, -6000.0 ],
+        [ 434.59663056485   , 0.0, 7988.698356992288],
+        ene
+    )
+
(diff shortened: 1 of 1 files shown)


SimCore/python/generators.py
```diff
@@ -311,6 +311,28 @@ def single_8gev_e_upstream_tagger():
     )
 
 
+def single_8gev_e_upstream_magnet_setEnergy(ene=8.0):
```
tomeichlersmith: I like this idea and I'm just trying to think of a better name for this function.

What could we call upstream of the tagger? The magnet and the tagger overlap in z so upstream of the magnet is kinda the same place as upstream of the tagger. Maybe `beam_pipe` instead of `upstream_magnet` to try to say we are modeling the electron arriving infront of everything?

I would also remove the `_setEnergy` suffix, I think the argument makes the point well enough.

```suggestion
def single_8gev_beam_pipe(ene = 8.0):
```

SimCore/python/generators.py
```diff
@@ -311,6 +311,28 @@ def single_8gev_e_upstream_tagger():
...
+    """Configure a particle gun to fire an electron of settable energy
+    upstream of the tagger tracker.  
+
+    The starting position here is well upstream of the analyzing magnet
+    the position/angle of the gun is such that 8 gev electrons arrive 
+    at the target z=0 at xy=(0,0).  This generator is used to study 
+    off-energy beam electrons.  
+
+    Note that if an energy != 8gev, the trajectory will be different. 
+
```
tomeichlersmith, line 326:
```suggestion
    And many electrons with energies sufficiently lower than 8GeV will just curve
    into the side of the magnet and not reach the target.

```

SimCore/python/generators.py
tvami: I dont think it should include the `_8gev_` either, since if `ene = 4.0` then this is not an 8 GeV beam. The `_e_` does make sense tho, right? i.e. it should start with `single_e`

SimCore/python/generators.py
```diff
@@ -311,6 +311,30 @@ def single_8gev_e_upstream_tagger():
     )
 
 
+def single_8gev_beam_pipe(ene = 8.0):
```
tvami:
```suggestion
def single_e_beam_pipe(ene = 8.0):
```

EBerzin (approved):

tvami (approved):

tomeichlersmith:
I do not know why the C++ compilation is failing on this branch when the only difference is in the Python configuration. Since this just adds a new Python configuration function, I'm going to use my :sparkles: great admin powers :sparkles: to go ahead and merge this.

tvami:
> I do not know why the C++ compilation is failing on this branch

I think that's because this branch was branched off from a version of ldmx-sw that had issues. If one were to rebase the branch with `main`, it would work

PR #1504, opened 2025-01-05
Missing argument in just debug

erik-wallin:
"just debug" passes cmd line arguments, such as the config file name, into gdb instead of into fire. This is shown by the following two examples:
```
>>>just debug config.py
[...]
Reading symbols from fire...
"/home/ewallin/Programming/justbug/ldmx-sw/build/config.py" is not a core dump: file format not recognized
```

```
>>>just debug config.py --nevents 5
denv gdb fire config.py --nevents 5
gdb: unrecognized option '--nevents'
Use `gdb --help' for a complete list of options.
error: Recipe `debug` failed on line 100 with exit code 1
```

## Check List
- [X ] I successfully compiled _ldmx-sw_ with my developments

 Now
```
>>>just debug config.py
>>>just debug config.py --nevents 5
```
both make it to the gdb prompt without complaint, and can run fire with the intended passed arguments :)

### DIFF
--- a/justfile
+++ b/justfile
@@ -97,7 +97,7 @@ fire config_py *ARGS:
 # run gdb on a config file
 [no-cd]
 debug config_py *ARGS:
-    denv gdb fire {{ config_py }} {{ ARGS }}
+    denv gdb --args fire {{ config_py }} {{ ARGS }}
 
 # initialize a containerized development environment
 init:


tomeichlersmith (approved):
Glad to see folks are using this workflow :)

tvami (approved):
Thanks for fixing this!

PR #1506, opened 2025-01-06
Optionally save HCalDigis in every readout channel

erik-wallin:
This PR adds a new argument to the HCalDigiProducer, digitizeAllChannels. If true, digis will be saved from every channel, not only in bars with large simhits in them. This mostly just adds a digi with some noise around the pedestal to a lot of channels. If false, as it is by default, then nothing changes from how it worked previously. 

With this change, we can more easily compare MC and data from the test beam. To e.g. find and fit pedestals in the simulated data, so we have the same calibration and analysis chain in both data and MC.  

In a simulation with the HCal prototype geometry, running only the digi producer and the rec hit producer, it becomes 3x slower and produces 13% larger file sizes when enabling this flag.  

Some details: 

- If the HgcrocEmulator produces a digi that doesn't pass above the readout threshold (i.e. when the HgcrocEmulator returns _false_), it is replaced by random noise instead (ignoring whatever small simhit there is in the bar).

- There is a new and optional noise simulation in HcalDigiProducer, that produces noise digis in every bar with no simhits. The ADC in the sample of interest is not treated specially, and may be below the readout threshold. 

- In the rare case where one end of the bar passes over the readout threshold, while the other end does not, the normal digi producer would not produce digis for either end. With this flag enabled, the bar that has some (probably very low) activity gets its digi saved properly, while the other end gets random noise.

## Check List
- [x ] I successfully compiled _ldmx-sw_ with my developments
- [ x] I ran my developments and the following shows that they are successful:

Simulating 1000 events with the prototype geometry yields exactly 384 channels * 1000 events = 384000 saved digis. So, it doesn't seem to be missing any channels or accidentally produce duplicate digis in the same one! 

Here's a plot of the reconstructed pulse amplitudes from MIPs. With digitizeAllChannels = false in red, and = true in blue: 
[NoiseSimulationComparison.pdf](https://github.com/user-attachments/files/18324773/NoiseSimulationComparison.pdf)
Here we see a lot of hits near the pedestal, which is exactly what I wanted :)

### DIFF
--- a/Hcal/include/Hcal/HcalDigiProducer.h
+++ b/Hcal/include/Hcal/HcalDigiProducer.h
@@ -86,7 +86,7 @@ class HcalDigiProducer : public framework::Producer {
 
   /// If false, save digis from all channels, even pure noise in empty bars
   /// Helpful when comparing with test beam data
-  bool zeroSuppression_;
+  bool zeroSuppression_{true};
 
   /// Hgcroc Emulator to digitize analog voltage signals
   std::unique_ptr<ldmx::HgcrocEmulator> hgcroc_;
--- a/Hcal/include/Hcal/HcalDigiProducer.h
+++ b/Hcal/include/Hcal/HcalDigiProducer.h
@@ -84,6 +84,10 @@ class HcalDigiProducer : public framework::Producer {
   /// development
   bool noise_{true};
 
+  /// Save digis from all channels, even pure noise in empty bars
+  /// Helpful when comparing with test beam data
+  bool digitizeAllChannels_;
+
   /// Hgcroc Emulator to digitize analog voltage signals
   std::unique_ptr<ldmx::HgcrocEmulator> hgcroc_;
 
--- a/Hcal/include/Hcal/HcalDigiProducer.h
+++ b/Hcal/include/Hcal/HcalDigiProducer.h
@@ -84,9 +84,9 @@ class HcalDigiProducer : public framework::Producer {
   /// development
   bool noise_{true};
 
-  /// Save digis from all channels, even pure noise in empty bars
+  /// If false, save digis from all channels, even pure noise in empty bars
   /// Helpful when comparing with test beam data
-  bool digitizeAllChannels_;
+  bool zeroSuppression_;
 
   /// Hgcroc Emulator to digitize analog voltage signals
   std::unique_ptr<ldmx::HgcrocEmulator> hgcroc_;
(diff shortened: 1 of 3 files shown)


tomeichlersmith (approved):
This looks good and is similar to what I implemented for the ECal. The only difference is that I use the phrase "zero suppression".

https://github.com/LDMX-Software/ldmx-sw/blob/d5a93d462449a0aa0a9bf58f962f0e3421a1e7d0/Ecal/src/Ecal/EcalDigiProducer.cxx#L43

The HGCROC chip is intended to have this "zero suppression" feature where channels below the readout threshold are not readout (oversimplifying). This means the default procedure (what you've called `not digitizeAllChannels_` with the "fast noise sim") is the designed behavior and the other behavior where every channel has a digi no matter what is what would happen if we configure the HGCROC to disable the zero suppression feature. It might be helpful to alter the vocabulary in the code to reference the "zero suppression", but I'll leave it up to you.

erik-wallin:
Thanks Tom, the term 'zero suppression' slipped my mind, that is exactly what it should be called. Will correct that. 

I skimmed through the HGCROCv3 data sheet, and I can't really find anything related to turning on or off zero suppression, or how that threshold would be configured. For the trigger data path there is a threshold ADC_TH, which can be used to force the ADC to 0 if it is too close to the pedestal, but for the DAQ data path I can't see anything. I may be mistaken! But, we could always manually apply some zero suppression further downstream I suppose? (Also, this isn't the right place for this discussion)

tomeichlersmith:
I may be mistaken as well. I thought the zero suppression was done on the HGCROC but it could be in a downstream chip. I'm fairly confident it is done on the frontend because of how many ECal cells there are.

cmantill:
Hi @tomeichlersmith @Skelpdar , zero-suppresion is done in the ECON-D chip.

Hcal/include/Hcal/HcalDigiProducer.h
```diff
@@ -84,6 +84,10 @@ class HcalDigiProducer : public framework::Producer {
   /// development
   bool noise_{true};
 
+  /// If false, save digis from all channels, even pure noise in empty bars
+  /// Helpful when comparing with test beam data
+  bool zeroSuppression_;
```
tvami: I think we should have a default value here
```suggestion
  bool zeroSuppression_{true};
```

tvami (changes requested):
I have a small technical comment below. Otherwise I'm a bit surprised the KS tests for HCAL didnt show significant changes, maybe we should rerun the tests again?

tvami:
> maybe we should rerun the tests again?

Running it here: https://github.com/LDMX-Software/ldmx-sw/actions/runs/12678697860

tvami (approved):
None of the distributions in the HCAL DQM are affected.

PR #1508, opened 2025-01-10
Improvements to DQM axis in ECal and PN, adding ECal BDT pass histo

### LINKED ISSUE #1486: Extend ECAL DQM histogram bins

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1486

## Check List

Changes are pretty trivial, but now we dont have a bunch of overflow bins populated. I also did similar changes to the PN part of the DQM not just Ecal as described in the issue.

### DIFF
--- a/DQM/python/dqm.py
+++ b/DQM/python/dqm.py
@@ -247,7 +247,7 @@ class EcalDigiVerify(ldmxcfg.Analyzer) :
                 "Number of noisy RecHits" , 100 , -0.5 , 99.5 )
 
         self.build1DHistogram( "total_rec_energy"      ,
-                "Total Reconstructed Energy in ECal [MeV]" , 800 , 0. , 8000. )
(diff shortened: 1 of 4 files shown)


danyi211 (approved):

tvami:
The tests ofc failed, but the plots failing are the ones that I touched, so it makes sense

tomeichlersmith (approved):

PR #1514, opened 2025-01-16
Introduce v15 geometry with Air

### LINKED ISSUE #1513: Stop having our detectors in vacuum
We use vacuum in our gdmls. ESA will contain, among other pieces of junk and museum artifacts, air. 

That we specify air as the material containing everything else. 

Leaving it as is. That is fine if we can convince ourselves it is. Which requires switching to air to compare. 

**Additional context**
Partly discussed already in issue #1435

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1513

## Check List

See the comment in here https://github.com/LDMX-Software/ldmx-sw/pull/1514#issuecomment-2597072483


tvami:
"Error: 507 plots failed the KS test against gold." haha

The signal is not affected tho, I guess that's because of how we deal with the LHE, so the electron doesnt see the air? 

Otherwise changes are everywhere and are small:
https://github.com/LDMX-Software/ldmx-sw/actions/runs/12816250967/artifacts/2443537846

@bryngemark @tomeichlersmith what do you think? We prob should merge this in v14 but if you think this makes sense I can edit the PR to a new geom with v15

tvami:
Talked to @bryngemark this morning, we agreed to have a v15 geometry started and then have a higher stat sample made

tvami:
Let me put some of the validation plots here just so they stay here
with 500 MeV cut
Screenshot 2025-01-17 at 09 59 53

with 2000 MeV 
Screenshot 2025-01-17 at 10 00 34

photon vertex in Z
Screenshot 2025-01-17 at 10 02 52

tvami:
> The signal is not affected tho, I guess that's because of how we deal with the LHE, so the electron doesnt see the air?

I realized this has nothing to do with this, and was due to https://github.com/LDMX-Software/ldmx-sw/issues/1520

Given that we dont test this geometry, I'd say we can merge this and then with some bigger timeline I'll have a student scrutinize the details with a high stat sample

tomeichlersmith (approved):

bryngemark:
Sounds like a good plan to do a more detailed comparison. I assume there will be several changes before we reach a final `v15`. I propose we keep updating it step by step after introducing air. Our validation suite won't catch any changes to a different detector version though so there will be some manual work going forward.

bryngemark (approved):

tvami:
Yes I agree! Maybe we can have another issue in which we collect all the developments planned for v15 as subissues. For example the new tracker geometry that allows the space for LYSO, and the 16 double layers of ECAL.

PR #1515, opened 2025-01-16
Update hcal prototype in reduced LDMX

### LINKED ISSUE #1512: Add the HCAL demonstrator into the `ldmx-reduced-v2`

Currently `ldmx-reduced-v2` has the full HCAL.

We already have a demonstrator geometry, I'd like to have that added to the  `ldmx-reduced-v2`, I'm sure somebody with more HCAL knowledge could this quickly.

I think we have to do this for the reduced-LDMX tests coming up

cmantill:

This resolves #1512 

## Check List
- [x] I ran my developments and the following shows that they are successful:
I used `geant4` GDML reader (https://gitlab.cern.ch/geant4/geant4/tree/f18089fbc5147e33302b542e1135ebda79fcf18b/examples/extended/persistency/gdml/G01) in the updated `ldmx-reduced-v2` directory. 
It compiled without problems (i.e. all variables/constants/volumes) are defined.

- [ ] I visualized the changes
I do not have a nice `*.mac` file so that I can visualize the volumes. If anyone has that and could share it, that would be awesome.

### DIFF
--- a/Detectors/data/ldmx-reduced-v2/constants.gdml
+++ b/Detectors/data/ldmx-reduced-v2/constants.gdml
@@ -146,106 +146,6 @@
 -->
 <constant name="trigger_pad2_actual_z"
     value="ts_center_z + (trig_scint_area_envelope_z/2 - trigger_pad_thickness/2)" />
-
-
-<!--
-   HCal
---> 
-
-<!-- Absorber -->
-<variable name="absorber_width" value="665"/>
-<variable name="absorber_height" value="700"/>
-<variable name="absorber_thickness" value="20."/>
-<variable name="absorber_hole_radius" value="6.5"/>
-<variable name="absorber_hole_distance_from_edge" value="16."/>
-
-<!-- Mounting plates  -->
-<variable name="bar_mounting_plate_thickness" value="3."/>
-
-<!-- Scintillators  -->
-<variable name="scint_thickness" value="20."/>
-<variable name="scint_bar_width" value="50."/>
-<variable name="scint_bar_length" value="2000."/>
-<variable name="scint_bar_cover_thickness" value="0.5"/>
-<variable name="num_bars_front" value="8"/>
-<variable name="num_bars_back" value="12"/>
-
-<!-- Layers -->
-<variable name="layer_thickness" value="45."/>
-<variable name="air_thickness"
-          value="layer_thickness - (absorber_thickness + bar_mounting_plate_thickness + scint_thickness + scint_bar_cover_thickness)"/>
-<variable name="num_layers_front_vertical" value="5"/>
-<variable name="num_layers_front_horizontal" value="4"/>
-<variable name="num_layers_front"
-          value="num_layers_front_vertical + num_layers_front_horizontal"/>
-<variable name="num_layers_back_vertical" value="5"/>
-<variable name="num_layers_back_horizontal" value="5"/>
-<variable name="num_layers_back"
-          value="num_layers_back_vertical + num_layers_back_horizontal"/>
-<variable name="num_layers" value="num_layers_front + num_layers_back"/>
-<variable name="num_absorbers" value="num_layers + 1"/>
-<variable name="back_start" value="num_layers_front * layer_thickness"/>
-
-<!-- Distances between layers -->
-<constant name="distance_to_subsequent_absorber_layer " value="layer_thickness"/>
-<constant name="distance_to_subsequent_scint_layer" value="2*layer_thickness"/>
-
-<!-- Prototype dimensions-->
-<variable name="bar_x_displacement" value="0"/>
-<variable name="hcal_dx" value="3000. + 2*bar_x_displacement"/>
-<variable name="hcal_dy" value="3000."/>
-<variable name="hcal_dz" value="num_layers * layer_thickness"/>
-
-<!-- Positional constants -->
-<variable name="support_box_shift" value="12.7*1.5" />
-<constant name="hadron_calorimeter_pos_z" value="ecal_front_z + hcal_dz/2." />
-<position name="hadron_calorimeter_pos" x="0.0" y="support_box_shift" z="hadron_calorimeter_pos_z"/>
-<variable name="first_layer_zpos" value="-hcal_dz/2"/>
-<variable name="first_layer_absorber_zpos"
-          value="first_layer_zpos + absorber_thickness/2"/>
-<variable name="first_layer_scint_cover_zpos"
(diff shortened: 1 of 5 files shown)


tvami:
so when we try to run this inside ldmx-sw we get
```
Evaluator : unknown variable
Aborted (core dumped)
```
I looked for hours now which of these could be unknown, but I didnt find it :/

cmantill:
What kind of test is running that gives: "Evaluator : unknown variable"?

tvami:
```
wget https://raw.githubusercontent.com/LDMX-Software/ldmx-sw/e375990bf13c7073634bf2eab3bff7e84a9d991b/.github/validation_samples/reduced_ldmx/config.py
just setenv LDMX_NUM_EVENTS=1
just setenv LDMX_RUN_NUMBER=1
just fire config.py
```

tvami:
@cmantill can you confirm that you can reproduce the error?

tvami (commented):
OK, so I found two things:
 * the config fails bc of the scoring planes, probably there is some overlap there --> if I switch if off, it works, see https://github.com/LDMX-Software/ldmx-sw/pull/1515/commits/280d31528cf81af5672c4fb18b1936837da145ed
 * To check the overlap I tried to validated the detector, well, see https://github.com/LDMX-Software/ldmx-sw/issues/1533

I think after the CI finishes I'm ready to sign this

cmantill:
So you are OK with not having SPs for now?

tvami:
> So you are OK with not having SPs for now?

Yes, unless you see where the problem is already, then let's fix it; otherwise having the SP is less crucial (right @SanjitMasanam , do you use the SP for anything at this point?)

SanjitMasanam:
Correct, I don't use the SP for my studies so excluding them will be fine!

tvami:
OK it technically works, but we lost a lot of ECAL which makes no sense since only the HCAL was modified, I'm worried it might be because of the overlap
Screenshot 2025-01-24 at 13 38 51

CI results are in https://github.com/LDMX-Software/ldmx-sw/actions/runs/12956457526?pr=1515

tvami:
With the new gold
https://github.com/LDMX-Software/ldmx-sw/actions/runs/12961371139?pr=1515
Red looks like the old blue
Screenshot 2025-01-25 at 09 59 16

I dont really understand why these fluctuate back and forth, but the blue in this PR looks reasonable... We need to look into this in details, but I'm ok to merge this as is for now

tvami (approved):

SanjitMasanam (approved):

SanjitMasanam:
So trying out these changes locally (both on my system and on UCSB's HPC cluster POD), I'm still getting the error Tamas brought up before
```
Evaluator : unknown variable
Aborted
```
I'm using the most updated version of this branch:
```
$ git log --oneline
a9d89a4 Add hcal veto
f9d130a  Dont include any SP for the reduced geom
f84d666 Merge branch 'trunk' into iss1512
...
```
When I `git checkout trunk`, my WAB configs run fine (before even when going back to trunk this issue would persist but this seems not to be the case anymore). Does anyone have some more insight into fixing this issue so I can test if these changes affect my WAB study?

tvami:
Looks good
https://github.com/LDMX-Software/ldmx-sw/actions/runs/12997477816?pr=1515

tvami (changes requested):
Actually (after Sanjit pointed out), the logs do say a bad config
```
[ ElectronCounter ] 762  info: Found 1 electrons (tracks) using input collection TriggerPadTracksY_
[ trigger ] 762  info: Got trigger energy cut 1500 for 1 electrons counted in the event.
[ trigger ] 762  info: Got trigger energy sum 5.31643; and decision is pass = true
[ TrigScintDigiPad1 ] 762 debug: Looping over hits in trigScintDigisPad1
[ TrigScintDigiPad2 ] 762 debug: Looping over hits in trigScintDigisPad2
[ TrigScintDigiPad3 ] 762 debug: Looping over hits in trigScintDigisPad3
[ fire ] 763 fatal: [BadConf] : Relative cell coordinates (-85.10, -30.82) mm derived from world coordinates (-30.82, 85.10) mm with layer = 0 and module = 0 are outside module hexagon
  at /home/runner/work/ldmx-sw/ldmx-sw/DetDescr/src/DetDescr/EcalGeometry.cxx:174 in getID
```
we really need to fix that the CI fails if `fire` returns `fatal`, @tomeichlersmith

tvami:
OK so we established that w/o this change the ecal part looks fine and run, with this change the ecal geometry is messed up... I do not understand how that's possible when we only add HCAL here...

Detectors/data/ldmx-reduced-v2/constants.gdml
```diff
@@ -150,85 +150,99 @@
...
+<constant name="distance_to_subsequent_scint_layer" value="2*layer_thickness"/>
+
+<!-- Prototype dimensions-->
+<variable name="bar_x_displacement" value="0"/>
+<variable name="hcal_dx" value="3000. + 2*bar_x_displacement"/>
+<variable name="hcal_dy" value="3000."/>
+<variable name="hcal_dz" value="num_layers * layer_thickness"/>
+
+<!-- Positional constants -->
+<position name="hadron_calorimeter_pos" x="0.0" y="0.0" z="0.0"/>
```
tvami: I think this is the issue!!
```suggestion
<variable name="support_box_shift" value="12.7*1.5" />
<constant name="hadron_calorimeter_pos_z" value="ecal_front_z + hcal_dz/2." />
<position name="hadron_calorimeter_pos" x="0.0" y="support_box_shift" z="hadron_calorimeter_pos_z"/>
```

tvami:
I think we are getting there... now upto event 900 it works and then:
```
[ fire ] 923 fatal: [BadConf] : Relative cell coordinates (90.18, 69.35) mm derived from world coordinates (74.17, -90.18) mm with layer = 1 and module = 0 are outside module hexagon
    at /home/runner/work/ldmx-sw/ldmx-sw/DetDescr/src/DetDescr/EcalGeometry.cxx:174 in getID
  [ fire ] 923 debug:     0 /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libDetDescr.so(+0x30b5c) [0x7faf98e12b5c]
      1 simcore::EcalSD::ProcessHits(G4Step*, G4TouchableHistory*) + 469 
      2 G4SteppingManager::Stepping() + 492 
      3 G4TrackingManager::ProcessOneTrack(G4Track*) + 284 
```
kinda wild this didnt come up in the first 900 events

tvami:
I propose we close this PR too for now. As I understand we cant have this version of the HCAL at the test beam given the absorber constraint

PR #1518, opened 2025-01-18
Add ACTS tracking in all CIs

### LINKED ISSUE #1487: Update PU CI validation scripts

This 
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/.github/validation_samples/it_pileup/gen_main.py#L12
and 
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/.github/validation_samples/it_pileup/gen_pileup.py#L11
are still at 4 GeV.

This line
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/.github/validation_samples/it_pileup/config.py#L100
would be better with `dqm.ecal_dqm`

Generally now that we've been running 2e samples this should be updated.

Update the configs

### LINKED ISSUE #1517: Add tracking to all CIs

For now only the ecalPN has the tracking running, so the projections in the ECAL only use that. 

Let's do it for all of them, and then the default for the projections can rely on the tracking by default

We have to do this in data.

tvami:

First step to resolve https://github.com/LDMX-Software/ldmx-sw/issues/1517

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1487

## Check List

### DIFF
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -24,11 +24,6 @@ p.run = int(os.environ['LDMX_RUN_NUMBER'])
 p.histogramFile = f'hist.root'
 p.outputFiles = [f'events.root']
 
-# The Tracking modules produce a lot of helpful messages
-# but (at the debug level) is too much for commiting the gold log
-# into the git working tree on GitHub
-p.termLogLevel = 1
-
 # Load the tracking module
 from LDMX.Tracking import tracking
 from LDMX.Tracking import geo
@@ -199,6 +194,22 @@ recoil_dqm.buildHistograms()
 ecalVeto = ecal_vetos.EcalVetoProcessor()
 ecalVeto.recoil_from_tracking = True
 
+# The Tracking modules produce a lot of helpful messages
+# but (at the debug level) is too much for commiting the gold log
+# into the git working tree on GitHub
+p.termLogLevel = 0
+p.logger.custom(truth_tracking, level = 2)
+p.logger.custom(digi_tagger, level = 2)
+p.logger.custom(digi_recoil, level = 2)
+p.logger.custom(seeder_tagger, level = 2)
+p.logger.custom(seeder_recoil, level = 2)
+p.logger.custom(tracking_tagger, level = 0)
+p.logger.custom(tracking_recoil, level = 0)
+p.logger.custom(seed_tagger_dqm, level = 2)
+p.logger.custom(seed_recoil_dqm, level = 2)
+p.logger.custom(tagger_dqm, level = 2)
+p.logger.custom(recoil_dqm, level = 2)
+
 p.sequence.extend([
         digi_tagger,
         digi_recoil,
@@ -223,5 +234,7 @@ p.sequence.extend([
         tagger_dqm,
         seed_recoil_dqm,
         recoil_dqm,
-        ] + dqm.all_dqm)
+        ])
+
+p.sequence.extend(dqm.all_dqm)
 
--- a/.github/validation_samples/inclusive/config.py
+++ b/.github/validation_samples/inclusive/config.py
@@ -49,7 +49,7 @@ truth_tracking.p_cutEcal         = 0. # In MeV
 # The Tracking modules produce a lot of helpful messages
 # but (at the debug level) is too much for commiting the gold log
 # into the git working tree on GitHub
-p.termLogLevel = 0
+p.logger.termLevel = 0
 p.logger.custom(truth_tracking, level = 2)
 
 # These smearing quantities are default. We expect around 6um hit resolution in bending plane
@@ -199,7 +199,7 @@ ecalVeto.recoil_from_tracking = False
 # The Tracking modules produce a lot of helpful messages
 # but (at the debug level) is too much for commiting the gold log
 # into the git working tree on GitHub
-p.termLogLevel = 0
+p.logger.termLevel = 0
 p.logger.custom(truth_tracking, level = 2)
 p.logger.custom(digi_tagger, level = 2)
 p.logger.custom(digi_recoil, level = 2)
--- a/.github/validation_samples/inclusive/config.py
+++ b/.github/validation_samples/inclusive/config.py
@@ -23,7 +23,113 @@ p.maxEvents = int(os.environ['LDMX_NUM_EVENTS'])
 
 p.histogramFile = 'hist.root'
 p.outputFiles = ['events.root']
+
+# Load the tracking module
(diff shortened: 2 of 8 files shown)


.github/validation_samples/ecal_pn/config.py
```diff
@@ -199,6 +194,22 @@
 ecalVeto = ecal_vetos.EcalVetoProcessor()
 ecalVeto.recoil_from_tracking = True
 
+# The Tracking modules produce a lot of helpful messages
+# but (at the debug level) is too much for commiting the gold log
+# into the git working tree on GitHub
+p.termLogLevel = 0
```
tomeichlersmith, line 200: Do you want to use the newer syntax? (I also want to double check that #1458 works.)
```suggestion
p.logger.termLevel = 0
```

.github/validation_samples/inclusive/config.py
```diff
@@ -23,7 +23,113 @@
...
+truth_tracking.z_min             = 0.
+truth_tracking.track_id          = -1
+truth_tracking.p_cut             = 0.05 # In MeV
+truth_tracking.pz_cut            = 0.03
+truth_tracking.p_cutEcal         = 0. # In MeV
+
+# The Tracking modules produce a lot of helpful messages
+# but (at the debug level) is too much for commiting the gold log
+# into the git working tree on GitHub
 p.termLogLevel = 0
```
tomeichlersmith:
```suggestion
p.logger.termLevel = 0
```

.github/validation_samples/inclusive/config.py
```diff
@@ -45,17 +151,78 @@
...
 count.input_pass_name = ''
 
+# Load ecal veto and use tracking in it
+ecalVeto = ecal_vetos.EcalVetoProcessor()
+ecalVeto.recoil_from_tracking = False
+
+# The Tracking modules produce a lot of helpful messages
+# but (at the debug level) is too much for commiting the gold log
+# into the git working tree on GitHub
+p.termLogLevel = 0
```
tomeichlersmith:
```suggestion
p.logger.termLevel = 0
```

.github/validation_samples/it_pileup/config.py
```diff
@@ -87,7 +222,30 @@
 count = ElectronCounter(1,'ElectronCounter')
 count.input_pass_name = ''
 
+# The Tracking modules produce a lot of helpful messages
+# but (at the debug level) is too much for commiting the gold log
+# into the git working tree on GitHub
+p.termLogLevel = 0
```
tomeichlersmith:
```suggestion
p.logger.termLevel = 0
```

.github/validation_samples/kaon_enhanced/config.py
```diff
@@ -133,9 +234,65 @@
...
+recoil_dqm.trackStates = ["ecal","target"]
+recoil_dqm.title = ""
+recoil_dqm.measurement_collection=digi_recoil.out_collection
+recoil_dqm.truth_hit_collection="RecoilSimHits"
+recoil_dqm.buildHistograms()
+
+# The Tracking modules produce a lot of helpful messages
+# but (at the debug level) is too much for commiting the gold log
+# into the git working tree on GitHub
+p.termLogLevel = 0
```
tomeichlersmith:
```suggestion
p.logger.termLevel = 0
```

.github/validation_samples/signal/config.py
```diff
@@ -56,11 +156,71 @@
...
+recoil_dqm.buildHistograms()
+
+# Load ecal veto and use tracking in it
+ecalVeto = ecal_vetos.EcalVetoProcessor()
+ecalVeto.recoil_from_tracking = False
+
+# The Tracking modules produce a lot of helpful messages
+# but (at the debug level) is too much for commiting the gold log
+# into the git working tree on GitHub
+p.termLogLevel = 0
```
tomeichlersmith:
```suggestion
p.logger.termLevel = 0
```

tomeichlersmith (changes requested):
My only request is focused on using the newer syntax for configuring the logger. While both work, since these configs are used as reference by many folks, I'd like them to contain the newer syntax as example.

fdelzanno (approved):
I assume the values of the cuts on the SeedFinding (i.e. pmin, pmax, d0min, d0max, etc.) are up to date from the tracking people (I haven't followed their developments in the last weeks)? If so, the implementation looks good to me.

bryngemark:
> I assume the values of the cuts on the SeedFinding (i.e. pmin, pmax, d0min, d0max, etc.) are up to date from the tracking people

I agree with this comment, but even more so, I think whatever has been settled on in recent tracking optimizations should go in as defaults for all these processors, and not be specified in detail here. As @tomeichlersmith pointed out, the validation configs are very useful examples and having people set all kinds of variables themselves is not sustainable in the long term.

tvami:
@bloodyyugo @EBerzin can you please confirm?

bloodyyugo:
> I agree with this comment, but even more so, I think whatever has been settled on in recent tracking optimizations should go in as defaults for all these processors, and not be specified in detail here. As @tomeichlersmith pointed out, the validation configs are very useful examples and having people set all kinds of variables themselves is not sustainable in the long term.

Yeah, these values are good.  Setting them as default values is a good idea, but we use the same class for both tagger and recoil tracking and they potentially have different values (though right now they are so loose they might as well be the same).  That said, we could easily add a switch to tell the code if it's doing tagger or recoil and set the parameters appropriately.  

I plan on changing how these cuts are implemented in the nearish future, so I can add this then.

EBerzin:
The values look good for the tagger. But these config files use `RecoilTruthSeeds ` for recoil tracking, which are much cleaner than reconstructed seeds. In this case the parameters that are set for `seeder_recoil ` I think actually don't matter. But if we want to use reconstructed seeds in these configs, the optimized parameters I found are a bit different, and the full tracking chain also requires an ambiguity resolution step. I was planning to open a PR with those ambiguity resolution changes later today, which will also include an example config with the optimized seeding parameters.

tvami:
OK I suggest we do this in two steps, like that the CI tests will actually show the improvements with the full suite of tracking in Elisabeth's PR

bryngemark:
Thanks @bloodyyugo and @EBerzin for checking the numbers. Great to hear that the tracking parameters are going to be updated.

How about we do it this way. @tvami removes the specific parameters being set from these configs, and when you update the defaults to the better ones, the CI test will show the difference it makes?

bryngemark:
Haha @tvami is clearly onboard at least :)

tomeichlersmith (approved):

PR #1523, opened 2025-01-19
Follow the electron to HCAL and potentially dont use its hits for max PE

### LINKED ISSUE #1519: Follow the recoil e- to the HCAL and make sure the highest PE hits are not caused by it

In the non-fid studies we see that we loose a lot of signal due to the HCAL veto's current behaviour. 

Follow the recoil e- to the HCAL and make sure the highest PE hits are not caused by it

**Additional context**

See non-fiducial studies in the DR

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1519

## Check List


tvami:
I switched OFF the new development, i.e. right now this PR adds the HCAL veto as it used to be for the CIs, together with a dedicated DQM. I'd like to have this merged before the next patch release, so that the next PR when I switch this ON can show the changes in those DQM histograms.

tvami:
Oh I also renamed to `snake_case` so that leads to a lot of code changes that do nothing.
I should point out tho that I added if the HCAL veto passes or not into the logs, so together with the new plots, we do expect the validation to show differences

tomeichlersmith (approved):

danyi211 (approved):

PR #1525, opened 2025-01-20
update default LHE ID for A' to be 1023

### LINKED ISSUE #1521: Change ldmx-sw to be able to read MG5 inputs

We moved the LHE production to MG5 in https://github.com/LDMX-Software/dark-brem-lib-gen/pull/19

Now next step is to change ldmx-sw to be able to read those LHE files

Keep using MG4 files

**Additional context**

Follow-up to https://github.com/LDMX-Software/ldmx-sw/issues/1520

tomeichlersmith:
This update then makes the default configuration of the dark brem model (from our point of view) align with the libraries that are produced by v5.0 of dark-brem-lib-gen.

Users can still use earlier versions of dark-brem-lib-gen, but then they will need to change their configuration script to look for the old LHE ID 622 rather than the new one.
```python
mySim.dark_brem.model.aprime_lhe_id = 622
```

This change *does not* alter the PDG ID of the dark photon within our simulation. It will still have the PDG ID of 622 according to Geant4 and the output SimParticle.

This resolves #1521 

## Check List

When opening this PR as a draft, the signal validation sample action will fail to run because it will be looking for ID 1023 in v4 dark brem libraries where the ID is 622. I'll then edit the CI to use v5 dark brem libraries and then they will succeed when I mark this PR as ready for review.

### DIFF
--- a/.github/validation_samples/signal/init.sh
+++ b/.github/validation_samples/signal/init.sh
@@ -11,7 +11,7 @@ start_group Produce Dark Brem Library
 wget https://raw.githubusercontent.com/LDMX-Software/dark-brem-lib-gen/main/env.sh
 source env.sh
 # commented out lines are dbgen's defaults for reference
-dbgen use v4.6
+dbgen use v5.0
 #dbgen cache ${HOME} <- only matters for apptainer/singularity
 #dbgen work /tmp
 #dbgen dest $PWD
--- a/SimCore/src/SimCore/APrimePhysics.cxx
+++ b/SimCore/src/SimCore/APrimePhysics.cxx
@@ -108,7 +108,7 @@ void APrimePhysics::ConstructProcess() {
               model.getParameter<double>("epsilon"), scaling_method_it->second,
               g4db::G4DarkBreMModel::XsecMethod::Auto,
               model.getParameter<double>("max_R_for_full", 50.0),
-              model.getParameter<int>("aprime_lhe_id", 622),
+              model.getParameter<int>("aprime_lhe_id", 1023),
               true,  // always load the library
               model.getParameter<bool>("scale_APrime", false),
               model.getParameter<double>("dist_decay_min", 0.0),


tomeichlersmith:
Annoyingly (and why we didn't notice this before), the signal fails to generate any events but then that validation job completes "successfully". I can confirm that the exception that was observed in #1521 is observed here.
```
   [ ParallelWorld ] 0 debug: Adding : sp_hcal_left_right_PV to parallel world.
  Unrecognized Exception: BadConf : Unable to find any library entries at 'electron_tungsten_MaxE_8.0_MinE_4.0_RelEStep_0.1_UndecayedAP_mA_0.01_run_1'
    The library is either a single CSV file or a directory of LHE files.
    Any individual file can be compressed with `gzip`.
    This means the valid extensions are '.lhe', '.lhe.gz', '.csv', and '.csv.gz'
```

tvami:
I'd say we should wait a little bit with merging this into the release, and have a patch before with v4.6, and then maybe in the next patch we should put this in

tomeichlersmith:
The PR Validation run from after I marked this as Ready for Review did have the signal sample fully generated. The diff is very large because the current gold has no events in it due to this bug, so do not go to that action's page unless you can wait for the large page to load.

Dropping this back to draft so we can re-run validation with an updated patch and compare MG5 to MG4 more directly.

tvami:
@tomeichlersmith did you rebase this?

tomeichlersmith:
I have not. I don't know why the rebase was necessary for you, we specifically copy the gold from trunk so it should only matter _when_ the workflow is triggered.

https://github.com/LDMX-Software/ldmx-sw/blob/4b012fbf7f77894db2963c100201ee4fc8883652/.github/workflows/pr_validation.yml#L37-L39

tvami:
Hmmm interesting, I'm quite confused about the other PR then, but ok sounds good!

tvami:
CI plots look fine, it's interesting that we give more energy to the dark photon

Screenshot 2025-01-24 at 10 33 28

Consistently less energy for the recoil ele
Screenshot 2025-01-24 at 10 34 16

OTOH the KS test did not flag the pT of the dark photon, and the first bin is hidden in this plot so I cant say much 
Screenshot 2025-01-24 at 10 34 54

But, we do see the first bin for the recoil electron, that matches very nicely (if you zoom in)
Screenshot 2025-01-24 at 10 36 36

tvami (approved):

PR #1529, opened 2025-01-23
Move some 'debug' to 'info' and change CI log level to 'info'

tvami:

This is a further attempt to consolidate the logs in the CI. 
We also agreed that 
`info`: anything event level, e.g. number of tracks in the event, number of HCAL PE, max PE hits information --> i.e. anything that's adding 1 line of output per event into the logs
`debug`: lower than that, i.e. hit level, e.g. hit locations, each tracks momentum, covariance matrices, etc 

## Check List

The logs are significantly reduced

### DIFF
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -197,7 +197,7 @@ ecalVeto.recoil_from_tracking = True
 # The Tracking modules produce a lot of helpful messages
 # but (at the debug level) is too much for commiting the gold log
 # into the git working tree on GitHub
-p.logger.termLevel = 0
+p.logger.termLevel = 1
 p.logger.custom(truth_tracking, level = 2)
 p.logger.custom(digi_tagger, level = 2)
 p.logger.custom(digi_recoil, level = 2)
--- a/.github/validation_samples/inclusive/config.py
+++ b/.github/validation_samples/inclusive/config.py
@@ -193,7 +193,7 @@ ecalVeto.recoil_from_tracking = False
 # The Tracking modules produce a lot of helpful messages
 # but (at the debug level) is too much for commiting the gold log
 # into the git working tree on GitHub
-p.logger.termLevel = 0
+p.logger.termLevel = 1 
 p.logger.custom(truth_tracking, level = 2)
 p.logger.custom(digi_tagger, level = 2)
 p.logger.custom(digi_recoil, level = 2)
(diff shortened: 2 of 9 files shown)


tomeichlersmith:
This looks good to me, I am now just curious on the log size before we merge. For context, GitHub prevents files larger than 100MB from being pushed via `git push`.

I also put this advice on the logging levels onto the docs page about logging: https://ldmx-software.github.io/developing/Logging.html

tvami:
Here is the connected CI results
https://github.com/LDMX-Software/ldmx-sw/actions/runs/12933909558

The ecal PN has the biggest log, with the size of 7.5 MB

tomeichlersmith:
Just for future reference, I did a survey of the logs for all samples (except `it_pileup` because I am impatient).

Sample | Old | New | Diff
---|---|---|---
ecal_pn | 950.51KB | 7.18MB | 6.26MB
hcal | 3.37KB | 3.37KB | 0.00B
inclusive | 18.26MB | 5.84MB | -12.42MB
kaon_enhanced | 16.08MB | 3.55MB | -12.53MB
reduced_ldmx | 7.40MB | 10.50MB | 3.11MB
signal | 7.96KB | 5.90MB | 5.90MB

We save about 25MB in the `inclusive` and `kaon_enhanced` samples but we grow the other logs from the raising of some ecal veto messages to `info`. Still an overall decrease that I am in favor of.

<details>

  <summary>File That Makes this Table after Download Artifacts</summary>

Just dumping this here in case I want to include it in future CI refactoring.

```bash
#!/bin/bash

set -e

sizefmt() {
        numfmt --to=iec --suffix=B --format "%.2f" ${1}
}

logdiff () 
{ 
    w=$(mktemp -d)
    cd ${w} 
    unzip ~/Downloads/${1}-pr-validation.zip &> /dev/null
    gold=$(tar -tvf ${1}_recon_validation_plots.tar.gz gold.log | awk '{print $3}');
    outp=$(tar -tvf ${1}_recon_validation_plots.tar.gz output.log | awk '{print $3}');
    diff=$((outp - gold))
    if (( diff < 0 )); then
            diff_str="-$(sizefmt $((-1*diff)))"
    else
            diff_str="$(sizefmt ${diff})"
    fi
    printf -- "${1} | %s | %s | %s\n" $(sizefmt ${gold}) $(sizefmt ${outp}) ${diff_str}
    cd - &> /dev/null
    rm -r ${w}
}

printf -- "Sample | Old | New | Diff\n"
printf -- "---|---|---|---\n"
for sample in ecal_pn hcal inclusive kaon_enhanced reduced_ldmx signal; do
    logdiff ${sample}
done
```

</details>

PR #1539, opened 2025-01-27
Persist ldmx-sw version and sha in run header

### LINKED ISSUE #1538: ldmx-sw version of previously generated files
I want to ensure samples that I am generating are using the same *ldmx-sw* version, but there is no way for me to check what version I used in my previously generated files, as far as I know.

It would be good if there was a simple command to do this, but at the very least, we need the *ldmx-sw* version encoded into the generated files in some readable manner.

Writing down (externally) the version number for every generated file is a possibility, but tedious, and information can easily slip through the cracks as files are transferred, re-run, etc. 

**Additional context**
@tvami and @tomeichlersmith had a conversation on Slack about this issue, perhaps they can summarize what was discussed, or continue the conversation here.

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1538

## Check List

I ran
```
just setenv LDMX_NUM_EVENTS=1
just setenv LDMX_RUN_NUMBER=1
just fire  .github/validation_samples/inclusive/config.py
```
and then
```
just fire my_ana.py events.root 
```
and got
```
[tamasvami@sdfiana007 ldmx-sw]$ just fire my_ana.py events.root 
denv fire my_ana.py events.root
['events.root']
ldmx-sw version = v4.2.5
ldmx-sw sha = 0987a39511a7a1a70ec2d11c53521468ba805182
```

Where `my_ana.py` and `MyAna.cxx` are in the details below.

<details>

```
import os
import sys
fileIn = sys.argv[1:]
fileName =  " ".join(sys.argv[1:])

p = ldmxcfg.Process('myAna')

from LDMX.Hcal import hcal
from LDMX.Ecal import vetos

p.maxEvents = -1
p.run = 2

print(fileIn)
p.inputFiles  = fileIn
p.histogramFile = "myHistoOnEventFile.root"


myAna = ldmxcfg.Analyzer.from_file('MyAna.cxx')


p.sequence = [myAna]
```
and `MyAna.cxx` is
```

#include <math.h>

class MyAna : public framework::Analyzer {
public:
  MyAna(const std::string& name, framework::Process& p)
  : framework::Analyzer(name, p) {}
  ~MyAna() = default;
  void onNewRun(const ldmx::RunHeader& rh);
  void analyze(const framework::Event& event) final;
};


void MyAna::onNewRun(const ldmx::RunHeader& runHeader) {
   std::cout << "ldmx-sw version = " << runHeader.getLdmxswVersion() << std::endl;
   std::cout << "ldmx-sw sha = " << runHeader.getSoftwareTag() << std::endl;
}

void MyAna::analyze(const framework::Event& event) {}

DECLARE_ANALYZER(MyAna);
```

</details>

### DIFF
--- a/Framework/CMakeLists.txt
+++ b/Framework/CMakeLists.txt
@@ -41,12 +41,33 @@ find_package(ROOT 6.16 CONFIG REQUIRED)
 # to generate the dictionary.
 include("${ROOT_DIR}/RootMacros.cmake")
 
-# Execute the command to extract the SHA1 hash of the current git tag. The
-# variable GIT_SHA1 will be set to contain the hash.
-execute_process(COMMAND git rev-parse HEAD OUTPUT_VARIABLE GIT_SHA1)
+# Execute the command to extract the SHA1 hash of the current git tag.
+# 'git' is removed from within the container to discourage opening a shell
+# in the container, so we need to go to some lengths in order to avoid using 'git'
+# The variable GIT_SHA1 will be set to contain the hash.
+set(git_dir "${PROJECT_SOURCE_DIR}/../.git")
+if (NOT IS_DIRECTORY "${git_dir}")
+  # we are a submodule of another project
+  file(READ "${git_dir}" git_dir)
+  string(REGEX REPLACE "^gitdir: " "" git_dir ${git_dir})
+  string(REGEX REPLACE "\n$" "" git_dir ${git_dir})
+endif()
+file(READ "${git_dir}/HEAD" current_ref)
+string(REGEX REPLACE "\n$" "" current_ref ${current_ref})
+if (current_ref MATCHES "^ref:.*$")
+  # on a branch
+  string(REGEX REPLACE "^ref: " "" current_ref "${current_ref}")
+  file(READ "${git_dir}/${current_ref}" GIT_SHA1)
+  string(REGEX REPLACE "\n$" "" GIT_SHA1 ${GIT_SHA1})
+else()
+  # in detached head state (probably on a tag)
+  set(GIT_SHA1 ${current_ref})
+endif()
+message(STATUS "Deduced git SHA: ${GIT_SHA1}")
 
-# Remove the newline character
-string(REGEX REPLACE "\n$" "" GIT_SHA1 "${GIT_SHA1}")
+# Include ldmx-sw version in the Run Header
+file(READ ${PROJECT_SOURCE_DIR}/../.github/actions/validate/gold_label LDMXSW_VERSION)
+string(STRIP "${LDMXSW_VERSION}" LDMXSW_VERSION)
 
 # Copies the file 'Version.h.in', substitutes the value of GIT_SHA1 and writes
 # it out to Version.h.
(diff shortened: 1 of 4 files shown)


bryngemark:
One question, what happens to this information when I run over an input file with a different version and produce a new output file?

tvami:
> One question, what happens to this information when I run over an input file with a different version and produce a new output file?

I just tested this by running myAna again while I modified the gold_label. It kept the "old" version.

-- > I think this is the desired behavior anyway, what we really care about is the simulation step, the other producers can run in the top of that. Otherwise just by running skimming in a new version should not overwrite this, so this behavior makes sense to me

tomeichlersmith:
Yea, whenever running in recon mode, the RunHeader is _copied_ from the input file to the output file. We could have subsequent processes add a new run parameter (as Tamas suggested on slack with `setStringParameter`), but then we have to figure out how to handle the case when we run recon mode on a file that was already recon'ed.

I think just leaving the base version is the best for now. We could develop a situation where we associate the pass name with a ldmx-sw version at some point in the future.

tomeichlersmith (approved):

jpasc27:
So the ldmx-sw version gets printed during the output of newly generated files - great! How would you check the version for files that have already been run (after this update is pushed of course)? Or is that a separate problem?

tomeichlersmith:
That is a separate problem. This helps avoid that problem in the future.

I can help you a bit more on slack if need be, but I'd ask you to more clearly define what information you want. Do you need to know if a specific simulation bug was present? Do you just want to document for replicability?

danyi211 (approved):

jpasc27:
Was just wanting to document for replicability! We can chat on Slack Tom.

jpasc27 (approved):

tvami:
Ehhh...

https://github.com/LDMX-Software/ldmx-sw/actions/runs/13034690037/job/36362081247

```
#8 0.823 CMake Error at Framework/CMakeLists.txt:69 (file):
#8 0.823   file failed to open for reading (No such file or directory):
#8 0.823 
#8 0.823     /code/Framework/../.github/actions/validate/gold_label
```

tomeichlersmith:
I'm guessing its because we need to copy the source code into the image https://github.com/LDMX-Software/ldmx-sw/blob/b3195994ada82b76b2e540bcca00de8140ed26d5/Dockerfile#L9
and this copy skips hidden directories.

It might be easier to just move gold_label somewhere else, maybe even have a version file at repository root that can be read by everyone.

tvami:
> One question, what happens to this information when I run over an input file with a different version and produce a new output file?

@bryngemark I was thinking about this: we could easily adopt a convention in central production to have the `pass` name be the software version. `sim` and `test` and etc are not very useful names anyway. Not sure if it makes sense to automate it, maybe it would break some of the validation code, tho from a quick look nothing seems extract anything with `test` in it, @tomeichlersmith ?

tvami:
Made an issue for further discussion: https://github.com/LDMX-Software/ldmx-sw/issues/1559

PR #1541, opened 2025-01-27
Fail validation CI if errors occur in validate action

### LINKED ISSUE #1540: CI does not fail if there is a fatal error in fire
**Describe the bug**

Even tho ldmx gives a fatal error the CI
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/.github/workflows/pr_validation.yml
still goes on and actually returns success instead of failing.

**To Reproduce**
See https://github.com/LDMX-Software/ldmx-sw/pull/1515#pullrequestreview-2576711776

**Desired behavior**

Fail if there is a failure...

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1540

## Check List

I have to test this on the fly.


tvami:
I think this worked, now `reduced_ldmx` is failing (as it should)

tomeichlersmith (approved):
<a href="https://imgflip.com/i/9i8gac"></a><div><a href="https://imgflip.com/memegenerator">from Imgflip Meme Generator</a></div>

SanjitMasanam (approved):

PR #1545, opened 2025-01-29
Add hcal veto to reduced ldmx CI

tvami:

After https://github.com/LDMX-Software/ldmx-sw/pull/1541 the CI fails (as it should)
but I didnt realize that will mean we cant build patch releases until one of the CIs is failing, see
https://github.com/LDMX-Software/ldmx-sw/actions/runs/13021076970

So this PR is going to fix that, then we can cut a patch release...

## Check List


tvami:
OK this all looks good, the CIs fail for known reasons of log diffs + new DQM plots. The reduced ldmx CI does complete correctly

PR #1546, opened 2025-01-29
Move ldmx-sw versioning to dedicated file

tvami:

This is a follow up to https://github.com/LDMX-Software/ldmx-sw/pull/1539#issuecomment-2622007262

## Check List

Locally this works, for github we are testing on the fly...


.github/actions/common.sh
```diff
@@ -68,7 +68,9 @@ warn() {
 start_group Deduce Common Environment Variables
 export LDMX_GOLD_LABEL_FILE=${GITHUB_WORKSPACE}/.github/actions/validate/gold_label
 export LDMX_BASE=$(cd ${GITHUB_WORKSPACE}/../ && pwd)
+export LDMX_VERSION_FILE=${LDMX_BASE}/VERSION
```
tomeichlersmith:
```suggestion
export LDMX_VERSION_FILE=${LDMX_BASE}/ldmx-sw/VERSION
```

`LDMX_BASE` is set to ldmx-sw's parent directory so we need to re-enter ldmx-sw to find VERSION.

tomeichlersmith (changes requested):
Just some other updates.

- [x] We could move the reading of the VERSION file earlier in the CMake so that the CMake can be informed of the VERSION as well.
Right now, CMake thinks we are pretty old

https://github.com/LDMX-Software/ldmx-sw/blob/b3195994ada82b76b2e540bcca00de8140ed26d5/CMakeLists.txt#L18

Not sure if we can set the version with a variable value though.

- [x] Update the function that prints the gold label to just dump this file.

https://github.com/LDMX-Software/ldmx-sw/blob/b3195994ada82b76b2e540bcca00de8140ed26d5/.github/actions/common.sh#L25

- [x] update to copy the VERSION along with the golds from trunk

https://github.com/LDMX-Software/ldmx-sw/blob/b3195994ada82b76b2e540bcca00de8140ed26d5/.github/workflows/pr_validation.yml#L38

.github/actions/common.sh
```diff
@@ -23,6 +23,7 @@
 # Print the gold label
 ldmx_gold_label() {
   cat ${LDMX_GOLD_LABEL_FILE}
```
tomeichlersmith:
```suggestion
```
don't need gold label file anymore

.github/actions/common.sh
```diff
@@ -68,7 +69,9 @@ warn() {
 start_group Deduce Common Environment Variables
 export LDMX_GOLD_LABEL_FILE=${GITHUB_WORKSPACE}/.github/actions/validate/gold_label
```
tomeichlersmith:
```suggestion
```

.github/actions/common.sh
```diff
@@ -68,7 +69,9 @@ warn() {
...
 export LDMX_BASE=$(cd ${GITHUB_WORKSPACE}/../ && pwd)
+export LDMX_VERSION_FILE=${LDMX_BASE}/ldmx-sw/VERSION
 echo "LDMX_GOLD_LABEL_FILE=${LDMX_GOLD_LABEL_FILE}"
```
tomeichlersmith:
```suggestion
```

.github/workflows/generate_pr_gold.yml
```diff
@@ -102,6 +102,7 @@ jobs:
         start_group Update Gold Label
         echo "Old Label: $(ldmx_gold_label)"
         echo "${GITHUB_REF##refs/tags/}" > ${LDMX_GOLD_LABEL_FILE}
```
tomeichlersmith:
```suggestion
```

.github/workflows/pr_validation.yml
```diff
@@ -38,6 +38,10 @@ jobs:
       run: git checkout origin/trunk -- .github/validation_samples/*/gold.root .github/actions/validate/gold_label
       shell: bash
 
+    - name: Git VERSION from trunk
+      run: git checkout origin/trunk -- VERSION
+      shell: bash
+
```
tomeichlersmith:
```suggestion
      run: git checkout origin/trunk -- .github/validation_samples/*/gold.root VERSION
      shell: bash
```

CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ if ("${srcdir}" STREQUAL "${bindir}")
 endif()
 
 # Set the project name
-project(LDMX_SW VERSION 3.1.1
+project(LDMX_SW VERSION ${LDMXSW_VERSION}
```
tomeichlersmith:
```suggestion
file(READ ${PROJECT_SOURCE_DIR}/VERSION LDMXSW_VERSION)
string(STRIP "${LDMXSW_VERSION}" LDMXSW_VERSION)
project(LDMX_SW VERSION ${LDMXSW_VERSION} 
```

Framework/CMakeLists.txt
```diff
@@ -66,7 +66,7 @@ endif()
 message(STATUS "Deduced git SHA: ${GIT_SHA1}")
 
 # Include ldmx-sw version in the Run Header
-file(READ ${PROJECT_SOURCE_DIR}/../.github/actions/validate/gold_label LDMXSW_VERSION)
+file(READ ${PROJECT_SOURCE_DIR}/../VERSION LDMXSW_VERSION)
 string(STRIP "${LDMXSW_VERSION}" LDMXSW_VERSION)
```
tomeichlersmith:
```suggestion
```
don't need this here anymore since it is defined in the central CMakeLists.txt and the value is inherited

tomeichlersmith (changes requested):
I'm just trying to remove the gold_label so that it doesn't confuse anyone in the future. I think these are the necessary edits and then we can just `git rm .github/actions/validate/gold_label`.

CMakeLists.txt
```diff
@@ -15,7 +15,9 @@ if ("${srcdir}" STREQUAL "${bindir}")
 endif()
 
 # Set the project name
-project(LDMX_SW VERSION 3.1.1
+file(READ ${PROJECT_SOURCE_DIR}/VERSION LDMXSW_VERSION)
```
tvami: This didnt work

```
CMake Error at CMakeLists.txt:18 (file):
  file failed to open for reading (No such file or directory):

    /VERSION
```

tvami: If I do 
```
-file(READ ${PROJECT_SOURCE_DIR}/VERSION LDMXSW_VERSION)
+file(READ ${CMAKE_CURRENT_SOURCE_DIR}/VERSION LDMXSW_VERSION)
```
it does work here, but then
```
CMake Error at CMakeLists.txt:20 (project):
  VERSION "v4.2.6" format invalid.
```

tvami: I guess I have to remove the `v`

tvami: OK this works now.

VERSION
```diff
@@ -0,0 +1 @@
+v4.2.6
```
tomeichlersmith:
```suggestion
v4.2.7
```

Framework/CMakeLists.txt
```diff
@@ -66,8 +66,6 @@ endif()
 message(STATUS "Deduced git SHA: ${GIT_SHA1}")
 
 # Include ldmx-sw version in the Run Header
```
tvami:
```suggestion
```

tomeichlersmith (approved):
One last more cosmetic thing: we could update the patch number now to keep it consistent or just do another patch release to help test it.

tvami:
> just do another patch release to help test it.

I'll add the tracking to ecal veto next, we can cut a patch release after that which can also act as a test for this

danyi211 (approved):

PR #1563, opened 2025-02-02
Inverse skimming in vetos

### LINKED ISSUE #1561: Add the ability to the vetos to flip the skimming logic

Many times I'd like to understand events that *dont* pass a given veto. For example what signal events do we kill with a given veto. 

I'd like to add a inverse skimming variable that now "keeps" events that fail the veto.

One could do this at analysis level, but it's just easier to deal with smaller files after skims

**Additional context**

Studying non-fiducial events where the HCAL veto fails and thus removes quite some signal

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1561

## Check List

I ran a skim and it contains events that failed the HCAL veto

### DIFF
--- a/Ecal/include/Ecal/EcalVetoProcessor.h
+++ b/Ecal/include/Ecal/EcalVetoProcessor.h
@@ -175,6 +175,7 @@ class EcalVetoProcessor : public framework::Producer {
   std::string rec_coll_name_;
   bool recoil_from_tracking_;
   std::string track_collection_;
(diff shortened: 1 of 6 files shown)


tvami:
Test show no real changes (just whatever we already had in `trunk` with the versioning), as they should.

tomeichlersmith (approved):

danyi211 (approved):

PR #1566, opened 2025-02-03
Ecal biasing factor developments

### LINKED ISSUE #1535: "Biasing factor is too large" warning

After running https://github.com/LDMX-Software/ldmx-sw/issues/1533 successfully I see
```
[ PhotoNuclearXsecBiasingOperator ]: [ WARNING ]: Biasing factor is too large.
```

This means

https://github.com/LDMX-Software/ldmx-sw/blob/9458b6baf1a4d0beb04dba82073037b1e40cacc6/SimCore/src/SimCore/BiasOperators/PhotoNuclear.cxx#L63C25-L63C40

```
emXsecBiased == pnXsecUnbiased_
```

which kinda makes me wonder if we are not actually biasing?

Well this is more of a question on what all this means

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1535

## Check List

See https://github.com/LDMX-Software/ldmx-sw/issues/1535#issuecomment-2625960556


tvami:
I'm gonna use this PR to test https://github.com/LDMX-Software/ldmx-sw/pull/1569

tvami:
I think this works now!

tvami:
I'll test if it also works in a draft mode

tvami:
```
  Switched to a new branch 'iss1535-biasing-factor-optimization'
  branch 'iss1535-biasing-factor-optimization' set up to track 'origin/iss1535-biasing-factor-optimization'.
```

Yaay it does! Sorry everybody if you got more than than a dozen msg because of this!

tvami:
I'm opening this for review, as we agreed today that this will not affect anything in the DR and is safe to merge

tomeichlersmith (commented):
Idk... do we want to wait and see if we are _really_ over-biasing a high energy child-of-primary photon? I'm thinking that maybe there is a bug in our checks on which particles get biased...

tvami:
Here are the ecal shower plots:
[compare_230_550_bias_factor_ecal_shower.pdf](https://github.com/user-attachments/files/18774947/compare_230_550_bias_factor_ecal_shower.pdf)

Looks pretty good

tvami:
I meant this to be a sanity check... 
No warnings in 10k events!
But the ECAL and PN (and HCAL & TS even) distributions change a lot
[ecal_pn_recon_validation_plots.pdf](https://github.com/user-attachments/files/18815687/ecal_pn_recon_validation_plots.pdf)

bryngemark:
@tvami LOL yes that _z_ distribution is telling. All interactions in the first possible instance!

tvami:
Exactly :D 
Would be good to have an earlier measure that helps to find the optimal number before having to look at the Z distribution

Biasing/python/ecal.py
```diff
@@ -52,7 +52,7 @@ def photo_nuclear( detector, generator ) :
 
     # Set run parameters
     if generator.energy == 8.0:
-          xsec_bias = 550.
+          xsec_bias = 19500000000.
```
tvami:
```suggestion
          xsec_bias = 1000.
```

tvami:
this is the Z vertex with a factor of a 1000:
Screenshot 2025-02-17 at 14 13 22

tvami:
Here is the rest of the KS-failed distributions:
[change_bias_factor_1000-ecal_pn_recon_validation_plots.pdf](https://github.com/user-attachments/files/18833502/change_bias_factor_1000-ecal_pn_recon_validation_plots.pdf)

bryngemark:
With a factor 1000 it looks like we're getting about 10% more events in the first two bins than we used to (btw is the reference with 8 GeV/550, or 4 GeV/450?). I think this could still mean that we're getting a large enough xsc to trigger a warning but it is suppressed. 

Did you get around to running with `old_ecal`? That should factor the low-_Z_ materials out of the problem. If we can run without these materials, maybe finding the maximum warning-free biasing factor would be a good procedure to optimize without having to look at _z_.

tvami:
>  (btw is the reference with 8 GeV/550, or 4 GeV/450?).

with ref to 8 GeV/550

> Did you get around to running with old_ecal?

Not yet, but sounds good

bryngemark:
I suggest we discuss this at the sw dev meeting today and type up what we learn from that and the email thread with natalia in this issue afterwards.

tvami:
Let's test with the new gold

tvami:
Screenshot 2025-02-20 at 17 52 09

Screenshot 2025-02-20 at 17 52 33

tvami:
So a little bit more interactions in the non-W part of the ecal, but it seems the with a factor of a 1000 it's not too different.

Reminder about the Z dist (now with the clean gold, and normalized plots with error bars):
Screenshot 2025-02-20 at 17 56 43

tvami:
The rest of the plots that fail the KS (reminder the red is with 550 and the blue is with 1000)
[bias-opt-ecal_pn_recon_validation_plots.pdf](https://github.com/user-attachments/files/18898182/bias-opt-ecal_pn_recon_validation_plots.pdf)

Biasing/python/ecal.py
```diff
@@ -52,7 +52,7 @@ def photo_nuclear( detector, generator ) :
 
     # Set run parameters
     if generator.energy == 8.0:
-          xsec_bias = 550.
+          xsec_bias = 1000.
```
tvami: Let me see what this will look like
```suggestion
          xsec_bias = 5000.
```

tvami:
Results in https://github.com/LDMX-Software/ldmx-sw/actions/runs/13469761175

The z with using biasing factor of 5000  

Screenshot 2025-02-25 at 14 38 09

Material
Screenshot 2025-02-25 at 14 38 38

Volume
Screenshot 2025-02-25 at 14 38 57

So more is happening now in the non-W based elements, I guess this was expected

Biasing/python/ecal.py
```diff
@@ -52,7 +52,7 @@ def photo_nuclear( detector, generator ) :
 
     # Set run parameters
     if generator.energy == 8.0:
-          xsec_bias = 550.
+          xsec_bias = 5000.
```
tvami:
```suggestion
          xsec_bias = 1.
```

tvami:
I'll check now how things look like for biasing factor of 1

tvami:
> I'll check now how things look like for biasing factor of 1

Ohh so with biasing factor of 1 the ecalPN takes longer than 6 hours to run, so github kills it...

bryngemark:
@tvami did this move along lately? were you able to run some tests outside of the CI action?

tvami:
> @tvami did this move along lately? were you able to run some tests outside of the CI action?

Not really, after I'm done with the DR stuff I'll come back to this

tvami:
[PN_pn_gamma_int_z.pdf](https://github.com/user-attachments/files/20742813/PN_pn_gamma_int_z.pdf)
[PN_pn_interaction_material.pdf](https://github.com/user-attachments/files/20742815/PN_pn_interaction_material.pdf)
[PN_pn_vertex_volume.pdf](https://github.com/user-attachments/files/20742816/PN_pn_vertex_volume.pdf)

I conclude that we are good with the 550 as is.

Btw, the 1100 still doesn't trigger the warning, but it clearly biases the z distribution

tomeichlersmith (commented):
Why the `old_ecal.py` module? I understand keeping it around for testing, but I am kinda opposed to keeping around code that we have marked as "old". (That's what `git` is for, folks can poke around an old release to see what the configuration parameters were if they need the details.)

tvami:
@tomeichlersmith I agree with that sentiment, but "old_ecal" is in the software: https://github.com/LDMX-Software/ldmx-sw/blob/8f4cc8a0416855eb7647dd5ad79b615649a1972c/SimCore/src/SimCore/DetectorConstruction.cxx#L67-L68 etc it just didnt have the python config to use it, so I thought the consistent behavior is to add it. But I'm fine to remove it if you feel strongly about this.

tomeichlersmith:
I would be happy to remove that `old_ecal` now. It was in reference to which volumes were being biased and I made the regretful decision to do the physics testing on trunk instead of on a separate branch (like you've done here). I don't think anyone should use `old_ecal` and thus, removing it is preferable.

tomeichlersmith (approved):

SimCore/src/SimCore/BiasOperators/PhotoNuclear.cxx
```diff
@@ -68,9 +68,12 @@ G4VBiasingOperation* PhotoNuclear::ProposeOccurenceBiasingOperation(
 
     double emXsecBiased = std::max(
         emXsecUnbiased + pnXsecUnbiased_ - pnXsecBiased_, pnXsecUnbiased_);
-    if (emXsecBiased == pnXsecUnbiased_) {
-      ldmx_log(debug) << "EM XS = PN unbiased XS! The biasing factor ("
-                      << factor_ << ") is too large";
+    auto material = track->GetMaterial();
+    if ((emXsecBiased == pnXsecUnbiased_) && material->GetName() == "G4_W") {
```
bryngemark: Is there a way to do the material check with a pointer instead? I know @EinarElen has found a large number of time consuming string comparisons where a pointer would work just fine and importantly, much faster. (I don't know how often we will in practice fulfill the first condition of this AND – hopefully we find the sweet spot before launching big productions – so it might not be a big deal, but could serve as an example of how to access this type of G4 info.)

bryngemark (commented):
Looks good to me, I'll approve once my curiosity question is answered 🙂

SimCore/src/SimCore/BiasOperators/PhotoNuclear.cxx
tvami: @bryngemark I did this in https://github.com/LDMX-Software/ldmx-sw/pull/1566/commits/a394dd7185d933c4e9a095953309e6e2db9c6fde

SimCore/src/SimCore/BiasOperators/PhotoNuclear.cxx
```diff
@@ -68,9 +68,14 @@ G4VBiasingOperation* PhotoNuclear::ProposeOccurenceBiasingOperation(
 
     double emXsecBiased = std::max(
         emXsecUnbiased + pnXsecUnbiased_ - pnXsecBiased_, pnXsecUnbiased_);
-    if (emXsecBiased == pnXsecUnbiased_) {
-      ldmx_log(debug) << "EM XS = PN unbiased XS! The biasing factor ("
-                      << factor_ << ") is too large";
+    auto material_tungsten = simcore::g4user::ptrretrieval::getMaterial("G4_W");
```
tomeichlersmith: Should be `static` so that we only retrieve the pointer by name once per run instead of on every call. (Retrieving the pointer by name also requires string comparisons so we only want to do that once and the save the resulting pointer which Geant4 doesn't change over the course of running).

```suggestion
    static auto material_tungsten = simcore::g4user::ptrretrieval::getMaterial("G4_W");
```

tomeichlersmith (approved):

tvami:
@bryngemark any other comments from your side?

bryngemark (approved):

PR #1575, opened 2025-02-06
Move VERSION def before build in PR Gold CI

### LINKED ISSUE #1573: Gold log files are always one version behind

When a new gold is added
https://github.com/LDMX-Software/ldmx-sw/pull/1567/files#diff-7b60b8e351cbb80c47459ffe2c79f1a26404871f49294780fe47ad0e58c09350

We update VERSION to be the version that is the patch we are updating to. But the logs still contain the version before the update. With this the test are perpetually failing on the log diff.

Image

Have these the same, tho I don't see what we did wrong in the first place.

Just not be annoyed that the checks are never perfect

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1573 (hopefully)

## Check List

I cant test this... we'd need to cut a patch release to test it

### DIFF
--- a/.github/workflows/generate_pr_gold.yml
+++ b/.github/workflows/generate_pr_gold.yml
@@ -25,6 +25,15 @@ jobs:
       with:
           just-version: 1.26.0
 
+    - name: Take care of versioning
+      run: | 
+        source ${GITHUB_WORKSPACE}/.github/actions/common.sh
+        start_group Update Gold Label
+        echo "Old Label: $(ldmx_gold_label)"
+        echo "${GITHUB_REF##refs/tags/}" > ${LDMX_VERSION_FILE}
+        echo "New Label: $(ldmx_gold_label)"
+        end_group
+
     - name: Compile and Install ldmx-sw
       run: | 
         just install-denv init configure-force-error
@@ -98,13 +107,6 @@ jobs:
     - name: Update the Gold
       id: update-gold
       run: |
-        source ${GITHUB_WORKSPACE}/.github/actions/common.sh
-        start_group Update Gold Label
-        echo "Old Label: $(ldmx_gold_label)"
-        echo "${GITHUB_REF##refs/tags/}" > ${LDMX_VERSION_FILE}
-        echo "New Label: $(ldmx_gold_label)"
-        git add ${LDMX_GOLD_LABEL_FILE}
-        end_group
         start_group Copy over new Gold Histograms
         artifacts_d=${{steps.download.outputs.download-path}}
         echo ${artifacts_d}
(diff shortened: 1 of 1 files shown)


.github/workflows/generate_pr_gold.yml
```diff
@@ -25,6 +25,15 @@ jobs:
...
           just-version: 1.26.0
 
+    - name: Take care of versioning
+      run: | 
+        source ${GITHUB_WORKSPACE}/.github/actions/common.sh
+        start_group Update Gold Label
+        echo "Old Label: $(ldmx_gold_label)"
+        echo "${GITHUB_REF##refs/tags/}" > ${LDMX_VERSION_FILE}
+        echo "New Label: $(ldmx_gold_label)"
+        end_group
```
tomeichlersmith:
```suggestion
```

.github/workflows/generate_pr_gold.yml
```diff
@@ -98,13 +107,6 @@ jobs:
     - name: Update the Gold
       id: update-gold
       run: |
-        source ${GITHUB_WORKSPACE}/.github/actions/common.sh
-        start_group Update Gold Label
-        echo "Old Label: $(ldmx_gold_label)"
-        echo "${GITHUB_REF##refs/tags/}" > ${LDMX_VERSION_FILE}
-        echo "New Label: $(ldmx_gold_label)"
-        git add ${LDMX_GOLD_LABEL_FILE}
-        end_group
```
tomeichlersmith: We still need to update the label here because this is in a new job (`pr-new-gold` instead of `generation` or `compile-ldmx-sw`) with a new clone of ldmx-sw and our previous edit of VERSION has not been applied (the `generation` job gets the updated VERSION from the tar-ball of the ldmx-sw build it downloads). I would just keep these lines as is maybe with a comment line about why we need to do both.

.github/workflows/generate_pr_gold.yml
```diff
@@ -25,6 +25,15 @@ jobs:
       with:
           just-version: 1.26.0
 
+    - name: Take care of versioning
+      run: | 
+        source ${GITHUB_WORKSPACE}/.github/actions/common.sh
+        start_group Update Gold Label
```
tomeichlersmith: The `start_group` and `end_group` are helpful for separating messages within a single job step. Since this job step is only changing the version, I don't think nesting the messages within a group is necessary.

```suggestion
```

tomeichlersmith (changes requested):
Some cosmetic and one functional thing.

LiamBrennan-UCSB (approved):

tomeichlersmith (approved):

PR #1577, opened 2025-02-06
Persist the HCAL bar's orientation

### LINKED ISSUE #1576: Persist if HCAL bar is in X or Y

In theory we know that some bars in HCAL we can have a good X coordinate, and in some others we can have a good Y coordinate.

There is a `getIsX` function that indicates that people wanted to keep this:
https://github.com/LDMX-Software/ldmx-sw/blob/ea2a21c94057cf6eba09ce2b75f555976f6097df/Hcal/include/Hcal/Event/HcalHit.h#L197

But there is no setIsX, and consequently this is not set in 
https://github.com/LDMX-Software/ldmx-sw/blob/ea2a21c94057cf6eba09ce2b75f555976f6097df/Hcal/src/Hcal/HcalRecProducer.cxx

Add a setter + keep the info.

**Additional context**

Improving the HCAL veto for non-fiducial events

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1576

## Check List

### DIFF
--- a/Hcal/include/Hcal/Event/HcalHit.h
+++ b/Hcal/include/Hcal/Event/HcalHit.h
@@ -119,9 +119,21 @@ class HcalHit : public ldmx::CalorimeterHit {
 
   /**
    * Get if bar is oriented in X
-   * @return isX
+   * @return orientationX
    */
-  int getIsX() const { return isX_; }
+  int getOrientationX_() const { return orientationX_; }
+
+  /**
+   * Get if bar is oriented in Y
+   * @return orientationY
+   */
+  int getOrientationY_() const { return orientationY_; }
+
+  /**
+   * Get if bar is oriented in Z
+   * @return orientationZ
+   */
+  int getOrientationZ_() const { return orientationZ_; }
 
   /**
    * Get the time difference between the two ends
@@ -208,17 +220,25 @@ class HcalHit : public ldmx::CalorimeterHit {
   /**
    * Set original position
    */
-  void setPositionUnchanged(double position, int isX) {
+  void setPositionUnchanged(double position, bool orientation) {
     position_ = position;
-    isX_ = isX;
+    orientation_ = orientation;
   }
 
   /**
    * Set if the bar is orientied in X
-   * 0 means oriented in Y
-   * 1 means oriented in X
    */
-  void setIsX(int isX) { isX_ = isX; }
+  void setOrientationX(bool orientationX) { orientationX_ = orientationX; }
+
+  /**
+   * Set if the bar is orientied in Y
+   */
+  void setOrientationY(bool orientationY) { orientationY_ = orientationY; }
+
+  /**
+   * Set if the bar is orientied in Z
+   */
+  void setOrientationZ(bool orientationZ) { orientationZ_ = orientationZ; }
 
  private:
   /** The number of PE estimated for this hit. */
@@ -239,7 +259,10 @@ class HcalHit : public ldmx::CalorimeterHit {
 
   double timeDiff_{-9999.};
   double position_{-9999.};
-  int isX_{-9999};
+  int orientation_{-9999};
+  bool orientationX_{false};
+  bool orientationY_{false};
+  bool orientationZ_{false};
 
   double toaPos_{-9999.};
   double toaNeg_{-9999.};
--- a/Hcal/include/Hcal/Event/HcalHit.h
+++ b/Hcal/include/Hcal/Event/HcalHit.h
@@ -112,6 +112,26 @@ class HcalHit : public ldmx::CalorimeterHit {
   int getAmplitudeNeg() const { return amplitudeNeg_; }
 
   /**
+   * Get the position of the hit
+   * @return position
+   */
+  double getPosition() const { return position_; }
+
+  /**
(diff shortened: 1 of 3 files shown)


Hcal/include/Hcal/Event/HcalHit.h
```diff
@@ -193,9 +213,12 @@ class HcalHit : public ldmx::CalorimeterHit {
     isX_ = isX;
   }
 
-  double getPosition() const { return position_; }
-  int getIsX() const { return isX_; }
-  double getTimeDiff() const { return timeDiff_; }
+  /**
```
bechenard, line 230: It would be better to use a bool if it has only two values. Otherwise we should extend it to X,Y and Z.

Hcal/include/Hcal/Event/HcalHit.h
```diff
@@ -111,6 +111,26 @@ class HcalHit : public ldmx::CalorimeterHit {
...
+   * Get if bar is oriented in X
+   * @return isX
+   */
+  int getIsX() const { return isX_; }
+
+  /**
+   * Get the time difference between the two ends
+   * @return timeDiff
+   */
+  double getTimeDiff() const { return timeDiff_; }
```
cmantill, line 144: is timeDiff_ ever set?

tvami: Not that I know of; but this has been here earlier, I just moved it up to the getters, would you rather have it deleted?

tvami: It's set in a different producer: https://github.com/LDMX-Software/ldmx-sw/blob/ea2a21c94057cf6eba09ce2b75f555976f6097df/Hcal/src/Hcal/HcalDoubleEndRecProducer.cxx#L153

So I wont remove it

Hcal/include/Hcal/Event/HcalHit.h
tvami: OK, I did it for all directions

Hcal/include/Hcal/Event/HcalHit.h
```diff
@@ -216,7 +259,10 @@ class HcalHit : public ldmx::CalorimeterHit {
 
   double timeDiff_{-9999.};
   double position_{-9999.};
-  double isX_{-9999.};
+  int orientation_{-9999};
+  bool orientationX_{false};
+  bool orientationY_{false};
+  bool orientationZ_{false};
```
EinarElen: Do we have to store these three? Is it not sufficient to store the enum and have the member functions return comparisons with that instead?

Hcal/include/Hcal/Event/HcalHit.h
```diff
@@ -111,6 +111,38 @@ class HcalHit : public ldmx::CalorimeterHit {
...
+   * Get the position of the hit
+   * @return position
+   */
+  double getPosition() const { return position_; }
+
+  /**
+   * Get if bar is oriented in X
+   * @return orientationX
+   */
+  int getOrientationX_() const { return orientationX_; }
```
EinarElen: On a similar note, it is probably nice if these are named in a more descriptive manner. E.g. something like "is oriented in x" etc so the meaning is obvious to people that are new to the code

Hcal/include/Hcal/Event/HcalHit.h
```diff
@@ -111,6 +111,38 @@ class HcalHit : public ldmx::CalorimeterHit {
...
+   * Get if bar is oriented in Z
+   * @return orientationZ
+   */
+  int getOrientationZ_() const { return orientationZ_; }
+
+  /**
+   * Get the time difference between the two ends
+   * @return timeDiff
+   */
+  double getTimeDiff() const { return timeDiff_; }
```
EinarElen, line 144: Might be worth mentioning here that this only applies for double ended readout

Hcal/src/Hcal/HcalDoubleEndRecProducer.cxx
```diff
@@ -105,21 +105,21 @@ void HcalDoubleEndRecProducer::produce(framework::Event& event) {
...
     if (orientation ==
         ldmx::HcalGeometry::ScintillatorOrientation::horizontal) {
       position_unchanged = position.X();
-      isX = 1;
+      bar_orientation = 1;
       position.SetX(position_bar);
     } else {
       position_unchanged = position.Y();
-      isX = 0;
+      bar_orientation = 0;
```
EinarElen: Again i would just persist the enumeration directly. It should be written to disk as an integer anyways

tvami:
Just for the record: I cannot store the enum directly as it requires the HCal geom class, that's under DetDesc so we'd need to compile in that directory. Instead I made an integer version of the orientation, and that's what we are keeping now.

bechenard (approved):
Looks good

Hcal/include/Hcal/Event/HcalHit.h
cmantill: I was going to suggest to add something similar in the HcalRecProducer while you are at it. i.e. to only add it if the hit is in the backHcal. Either way works. I'll approve now

cmantill (approved):

PR #1578, opened 2025-02-06
Ability to run on an event range when reading files

### LINKED ISSUE #1570: Ability to run on an event range

I'd like to have the ability to run on a run range with there is an input file.
This I'd need now to debug events, I dont need to resimulate, just run the tracking on the n-th to m-th event.

I also think that this is a feature we'll need when we have real data, you often want to debug some weird event at event number 124241

I think here
https://github.com/LDMX-Software/ldmx-sw/blob/ea2a21c94057cf6eba09ce2b75f555976f6097df/Framework/src/Framework/Process.cxx#L333
this needs to become a for loop and we add the lower and upper ends in there.

Even to resimulation should work on simulation, that's not a solution to real data.

**Additional context**

For example in https://github.com/LDMX-Software/ldmx-sw/issues/1562 happens at 122nd event

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1570

## Check List

I had a file which had
```
...
[ Recoil_TrackFinder ] 51  info: found tracks size = 1
[ Recoil_TrackFinder ] 52  info: found tracks size = 1
[ Recoil_TrackFinder ] 53  info: found tracks size = 1
[ Recoil_TrackFinder ] 54  info: found tracks size = 0
[ Recoil_TrackFinder ] 55  info: found tracks size = 0
[ Recoil_TrackFinder ] 56  info: found tracks size = 1
[ Recoil_TrackFinder ] 57  info: found tracks size = 1
...
```

now if I run
```
p.minEvents = 54
p.maxEvents = 55 
```
I only get 
[ Recoil_TrackFinder ] 54  info: found tracks size = 0
[ Recoil_TrackFinder ] 55  info: found tracks size = 0
```
Given that it was surrounded with events that had 1 track, I'm pretty certain it's the correct event.

### DIFF
--- a/Framework/src/Framework/Process.cxx
+++ b/Framework/src/Framework/Process.cxx
@@ -29,6 +29,7 @@ Process::Process(const framework::config::Parameters &configuration)
 
   maxTries_ = configuration.getParameter<int>("maxTriesPerEvent", 1);
   eventLimit_ = configuration.getParameter<int>("maxEvents", -1);
+  minEvents_ = configuration.getParameter<int>("minEvents", -1);
   totalEvents_ = configuration.getParameter<int>("totalEvents", -1);
   logFrequency_ = configuration.getParameter<int>("logFrequency", -1);
   compressionSetting_ =
@@ -327,6 +328,11 @@ void Process::run() {
         // empty output file list, use inputFile as master file
         inFile.setupEvent(&theEvent);
         masterFile = &inFile;
+      }
+
+      // In case we'd like to skip up to the event of minEvents_
+      while(n_events_processed < (minEvents_-1) && masterFile->nextEvent()) {
+        n_events_processed++;
       }
 
       bool event_completed = true;
--- a/Framework/src/Framework/Process.cxx
+++ b/Framework/src/Framework/Process.cxx
@@ -331,7 +331,8 @@ void Process::run() {
       }
 
       // In case we'd like to skip up to the event of minEvents_
-      while(n_events_processed < (minEvents_-1) && masterFile->nextEvent()) {
+      while (n_events_processed < (minEvents_ - 1) &&
+             masterFile->nextEvent(false)) {
         n_events_processed++;
       }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: Framework/src/Framework/Process.cxx [hunk 1]
```diff
@@ -329,6 +330,11 @@ void Process::run() {
         masterFile = &inFile;
       }
 
+      // In case we'd like to skip up to the event of minEvents_
+      while (n_events_processed < (minEvents_ - 1) && masterFile->nextEvent()) {
```
tomeichlersmith: Since we are skipping these events, we want to tell `nextEvent` to *not* store the event in the output file (if there is an output file).

```suggestion
      while (n_events_processed < (minEvents_ - 1) && masterFile->nextEvent(false)) {
```

tomeichlersmith (commented):
I am happy with this implementation, there are two comments I would add into the documentation of `minEvents`.

1. Since we actually scan the events with `masterFile->nextEvent()`, this process is relatively slow. If a user wants to use this for anything outside of debugging, they should probably skim to a new file and then run again rather than use this.
2. As implemented, this skips the first `minEvents_ - 1` events _of *each* input file_. I don't think this is an implementation bug, it just needs to be emphasized since this feature is really focused on use only with a single input file.

tvami:
I'm going to run validation again, not related to this PR, just related to the testing... A new gold is merged so this should give completely clean results. /run-validation

tvami:
As a comparison , here is the job that's triggered by the "ready for review":
https://github.com/LDMX-Software/ldmx-sw/actions/runs/13193852544/job/36831522017?pr=1578

tvami:
@tomeichlersmith @bryngemark as you can see the "ready for review" type of testing shows no changes, as expected. Please approve when you get the chance

tomeichlersmith (approved):

danyi211 (approved):

tvami:
This will be also handy for https://github.com/LDMX-Software/ldmx-sw/issues/1586

PR #1588, opened 2025-02-10
Decrease the number of events in PU CI

### LINKED ISSUE #1581: Run overlay PU CI on half the events

Right now the overlay PU CI runs about twice as long as anything else, which makes sense since it generates `LDMX_NUM_EVENTS` + `LDMX_NUM_EVENTS**0.95` events, and then overlays `LDMX_NUM_EVENTS`...

Image

I'd like to have `LDMX_NUM_EVENTS*0.5` events generated, it should be sufficient for the testing purposes and will make the CI finish in the same timescale as everything else

Be patient

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1581
and with that will make all the CIs finish in a similar time scale.

## Check List

### DIFF
--- a/.github/validation_samples/it_pileup/config.py
+++ b/.github/validation_samples/it_pileup/config.py
@@ -14,7 +14,7 @@ p=ldmxcfg.Process(thisPassName)
 
 det = 'ldmx-det-v14-8gev'
 p.run = int(os.environ['LDMX_RUN_NUMBER'])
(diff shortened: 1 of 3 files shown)


tomeichlersmith (approved):

bryngemark (approved):
Looks fine as long as the number of events being halved is reasonable.

Issue #1388, opened 2024-08-13
Use central Random Number Seed Service rather than config


tomeichlersmith:
https://github.com/LDMX-Software/ldmx-sw/blob/4e76f4267d6986cbdef9ecdcf76efe7d5330b2a9/TrigScint/src/TrigScint/TrigScintDigiProducer.cxx#L29-L30

:cry: 

#1385 reminded me of this since we need to set the `randomSeed` for the TS digi processors within the CI.

https://github.com/LDMX-Software/ldmx-sw/blob/4e76f4267d6986cbdef9ecdcf76efe7d5330b2a9/Ecal/src/Ecal/EcalDigiProducer.cxx#L73-L78

The seeding in ECal is done in `produce` because it was written before a patch to Framework allowing use of conditions within `onNewRun`. I would suggest putting the same code within `onNewRun` to avoid confusion of future developers and to avoid unnecessary checks. Moving the seeding to `onNewRun` cleans up the code and makes it obvious that the RNG _does not_ need to be seeded on every event. This also gives us the opportunity to switch to using STL like Tracking instead of `TRandom`.

## Solution
One can seed their noise generator within `onNewRun` which is the first callback that happens after the conditions are configured. This gives you access to the RNSS and allows you to have your seeds be produced via the central service which will change the seeds whenever all of the other seeds should change.

**Examples**
https://github.com/LDMX-Software/ldmx-sw/blob/4e76f4267d6986cbdef9ecdcf76efe7d5330b2a9/Tracking/src/Tracking/Reco/DigitizationProcessor.cxx#L71-L75

where `generator_` is a `std::default_random_engine` which can be passed to an instance of a distribution (for example `std::normal_distribution<float>`) to sample it.

tvami:
This is the same development as I did in https://github.com/LDMX-Software/ldmx-sw/pull/1359/commits/bda6d5c7145d052d7bac36cfc0eb3f8ea2640192 right?
If yes, seems easy, I could put this on my todo list.

tomeichlersmith:
Kinda. If you choose to use STL[^1], then you'll need to change `TRandom` to an equivalent STL distribution ([`std::uniform_int_distribution`](https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution) I believe) **or** you could follow whats in ECal for the ROOT way which just waits to create `TRandom` until access to the RNSS is available.

[...]

PR #1591, opened 2025-02-11
bring back old Hcal digi/reco chain for testing/debug

### LINKED ISSUE #1351: Bring back the old simple reconstruction chain
When developing, debugging, or trying to understand reconstructed quantities people often have to work with simulation level hits which are quite different from their reconstructed counterparts. This means that people have to re-create an entire (often worse) reco-chain in their analysis.

Having the old reconstruction tool available as an alternative would solve this problem. It should of course be clearly marked as simplified & not the default use.

tomeichlersmith:
This is a transfer of https://github.com/LDMX-Software/Hcal/pull/75 into this repository.

This resolves #1351 

## Check List
- [x] I ran my developments and the following shows that they are successful. See https://github.com/LDMX-Software/Hcal/pull/75

### DIFF
--- a/Hcal/include/Hcal/HcalSimpleDigiAndRecProducer.h
+++ b/Hcal/include/Hcal/HcalSimpleDigiAndRecProducer.h
@@ -1,13 +1,13 @@
 #ifndef HCALSIMPLEDIGIANDRECPRODUCER_H
 #define HCALSIMPLEDIGIANDRECPRODUCER_H
+#include <random>
+
 #include "DetDescr/DetectorID.h"
 #include "DetDescr/HcalGeometry.h"
 #include "DetDescr/HcalID.h"
 #include "Framework/Configure/Parameters.h"
-#include "Framework/EventDef.h"
 #include "Framework/EventProcessor.h"
 #include "Framework/RandomNumberSeedService.h"
-#include "TRandom3.h"
 #include "Tools/NoiseGenerator.h"
 
 namespace hcal {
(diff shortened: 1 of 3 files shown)


Hcal/src/Hcal/HcalSimpleDigiAndRecProducer.cxx
```diff
@@ -0,0 +1,149 @@
...
+    recHit.setID(hitID.raw());
+    recHit.setXPos(xpos);
+    recHit.setNoise(false);
+    recHit.setYPos(ypos);
+    recHit.setZPos(zpos);
+    recHit.setTime(time);
+    recHit.setSection(hitID.section());
+    recHit.setStrip(hitID.strip());
+    recHit.setLayer(hitID.layer());
+    recHit.setEnergy(edep);
```
tvami, line 137: Can we add the `setOrientation` function like I did in https://github.com/LDMX-Software/ldmx-sw/pull/1577/files please?

Hcal/include/Hcal/HcalSimpleDigiAndRecProducer.h
```diff
@@ -0,0 +1,40 @@
...
+
+namespace hcal {
+class HcalSimpleDigiAndRecProducer : public framework::Producer {
+ public:
+  HcalSimpleDigiAndRecProducer(const std::string& name,
+                               framework::Process& process)
+      : framework::Producer{name, process} {}
+  ~HcalSimpleDigiAndRecProducer() override = default;
+  void configure(framework::config::Parameters& ps) override;
+  void SetupRandomNumberGeneration();
```
tvami: shouldnt this be lower case, since it's a function, not a class?

Hcal/src/Hcal/HcalSimpleDigiAndRecProducer.cxx
```diff
@@ -0,0 +1,149 @@
...
+    framework::config::Parameters& ps) {
+  input_coll_name = ps.getParameter<std::string>("input_coll_name");
+  input_pass_name = ps.getParameter<std::string>("input_pass_name");
+  output_coll_name = ps.getParameter<std::string>("output_coll_name");
+  mev_per_mip = ps.getParameter<double>("mev_per_mip");
+  pe_per_mip = ps.getParameter<double>("pe_per_mip");
+  attenuation_length = ps.getParameter<double>("attenuation_length");
+  readout_threshold = ps.getParameter<int>("readout_threshold");
+  mean_noise = ps.getParameter<double>("mean_noise");
+  position_resolution = ps.getParameter<double>("position_resolution");
```
tvami: also if we follow our (not-yet-written-down) coding rules, these should end with a trailing underscore, showing that these are global variables

Hcal/include/Hcal/HcalSimpleDigiAndRecProducer.h
tomeichlersmith: yep! and its content can be moved to `onNewRun` after a Framework patch awhile ago

Hcal/src/Hcal/HcalSimpleDigiAndRecProducer.cxx
tomeichlersmith: will do :+1:

tomeichlersmith:
Alright, I resolved your comments @tvami and switched from TRandom3 to the STL distributions/generator so that we have a concrete example of what I outlined in https://github.com/LDMX-Software/ldmx-sw/issues/1388#issuecomment-2286459236

tvami:
> switched from TRandom3 to the STL distributions/generator so that we have a concrete example of what I outlined in https://github.com/LDMX-Software/ldmx-sw/issues/1388#issuecomment-2286459236

This is great! I think we can resolve that next after we cut a gold with some of the recent changes

tvami (approved):

PR #1551, opened 2025-01-30
Factor out tracking, update all CIs to full sequence

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1549

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments
- [x] I ran my developments and the following shows that they are successful.

I tested with the ecal PN sample.

I did two things in this PR
 1) update the CIs to use the latest tracking, for this I made a new file so I dont have to copy a lot of things, I also updated reco.py @bloodyyugo I took the numbers from https://github.com/LDMX-Software/ldmx-sw/pull/1526 (and rounded them).

 2) Ecal Veto now relies on the tracking for the photon projection, when it can (i.e. if a recoil track exists). With this we also introduced another ep angle as measured at the target, or as it used to be measured at the ECAL face. Much of this work was done by @JYoo001

PR #1596, opened 2025-02-12
Factor out tracking, move to digi based seeds in the CI tests

### LINKED ISSUE #1595: Factor out tracking, move to digi based seeds in the CI tests

https://github.com/LDMX-Software/ldmx-sw/pull/1551 does too many things at the same time, so I'd like to break it up to sub-PRs

In this first PR to do the factorization, and to move to the digi based seeds, but not include the optimized parameters.

tvami:

https://github.com/LDMX-Software/ldmx-sw/pull/1551 does too many things at the same time, so I'd like to break it up to sub-PRs --> in this one I factor out tracking, and move to digi based seeds in the CI tests --> we expect the tests to show degraded (more realistic) performance. In a next PR I'll change to the optimized parameters.

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1595

## Check List


tvami:
Here are the plots that fail the CI
[signal_recon_validation_plots.pdf](https://github.com/user-attachments/files/18774754/signal_recon_validation_plots.pdf)
[it_pileup_recon_validation_plots.pdf](https://github.com/user-attachments/files/18774756/it_pileup_recon_validation_plots.pdf)
[inclusive_recon_validation_plots.pdf](https://github.com/user-attachments/files/18774758/inclusive_recon_validation_plots.pdf)
[ecal_pn_recon_validation_plots.pdf](https://github.com/user-attachments/files/18774761/ecal_pn_recon_validation_plots.pdf)

I'll concentrate on the inclusive for now:
  * we seem to be having less track in general, which makes sense since now we use the digi seeds, not the truth seeds
  * the quality of these tracks is worse than before, starting with the fact that we get more tracks per event: fig 4
  * fig 8 - 14: in the `trunk` with truth seeds dont seem to have duplicates, now we do, which makes sense, that's why we have the ambiguity resolution (which is not in these plots)
  * fig 27 - 32 is similar story
  * the rest just shows degraded performance as well

@bloodyyugo @EBerzin I'd like to have this merged, so that we can update the gold files, and then I'll include the optimized parameters from @EBerzin in a next PR (which by that time can include the amb resolution)

EBerzin (approved):
The changes in tracking performance look consistent with what I saw when initially switching to reconstructed seeds from truth seeds.

bloodyyugo (approved):

PR #1600, opened 2025-02-13
Add TOT return condition in HGCROC emulation

### LINKED ISSUE #1586: Negative/null energy hit in ecal PN
**Describe the bug**

https://github.com/LDMX-Software/ldmx-sw/actions/runs/13248358110/job/36981301291?pr=1585
```
> [ ecalVeto ] 7403 fatal: ECal hit has negative or zero energy, this should never happen, check the threshold settings in HgcrocEmulator
```

**To Reproduce**
Run the ecalPN CI with seed = 1 and it will show up at 7403th event

**Desired behavior**
A clear and concise description of what you want to happen (i.e. instead of the bug).

**Environment:**

```
 ldmx-sw revision = 53c73d52a6ffee68f125cb31b378a66335a5b5b7
```

**Additional context**

Came up in https://github.com/LDMX-Software/ldmx-sw/pull/1585

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1586

## Check List

I made a printout in the issue


tvami:
Let's test this after the gold is merged

tvami:
yay(ish)!
All plots pass the KS test, and the hit with fatal error is gone!
Now the GSF is really noisy but that has nothing to do with this PR
(Edit: the GSF is showing up in the logs only bc the time is included in the printout that clearly is a different number than the gold log)

LiamBrennan-UCSB (approved):

danyi211 (approved):

Tools/src/Tools/HgcrocEmulator.cxx
```diff
@@ -163,30 +173,40 @@ bool HgcrocEmulator::digitize(
...
-      );
+      if (verbose)
+        ldmx_log(debug) << "    Adding TOT hit with toa = " << toa
+                        << ", tdc_counts = " << tdc_counts
+                        << " adc_t at prev iADC = "
+                        << digiToAdd.at(iADC - 1).adc_t();
+      // ADC at t-1
+      auto adc_at_tminus1 =
+          (iADC > 0) ? digiToAdd.at(iADC - 1).adc_t() : pedestal;
+      auto i_tot_sample = digiToAdd.size() - 1;
```
tomeichlersmith:
```suggestion
      auto i_tot_sample = digiToAdd.size();
```
we haven't `emplace_back` the TOT sample yet so the index of the TOT sample is the current size of the vector. (If this line is moved to after the `emplace_back` directly below on line 185, then you would keep the `-1` in).

Tools/src/Tools/HgcrocEmulator.cxx
```diff
@@ -46,8 +46,10 @@ bool HgcrocEmulator::digitize(
     const int &channelID,
     std::vector<std::pair<double, double>> &arriving_pulses,
     std::vector<ldmx::HgcrocDigiCollection::Sample> &digiToAdd) const {
+  // The emulation is run on each sim hit in each crossing, this is very
+  // verbose, so ldmx_log(debug) is not enough
+  bool verbose{false};
```
tomeichlersmith, line 51: Even if we have `p.logger.silence('HgcrocEmulator')`? I'm not opposed to keeping the verbose flag, but I do think we have the tools necessary to avoid it (unless there is a use-case I'm not thinking of).

tomeichlersmith (changes requested):

Tools/src/Tools/HgcrocEmulator.cxx
tvami: I considered this... but that means we need to put this to every cfg file, right? That might be annoying... I feel like we'd need another level below debug, `debug` shows things that are hit level, but this is a loop on the rechits, then simhits, and samples, so it's one level down from `debug` in my mind.

tvami: And call it `ldmx_log(spam)` lol ?

tvami: If we dont feel too funny, then `ldmx_log(diag)` ?

tomeichlersmith: Yea, in the past, I've just commented out these messages and treated it like "if you are editing this file, then these messages might be helpful to uncomment while developing"

tvami: Made an issue: https://github.com/LDMX-Software/ldmx-sw/issues/1611
I'll keep this as it is for now

Ecal/test/EcalDigiPipelineTest.cxx
```diff
@@ -248,40 +248,41 @@ class EcalCheckEnergyReconstruction : public framework::Analyzer {
...
-    auto trigDigi = trigDigis.at(0);
-    float tp_energy = 8 * trigDigi.linearPrimitive() * 320. / 1024 * MeV_per_fC;
-
-    CHECK_THAT(tp_energy, isCloseEnough(truth_energy, MAX_ENERGY_ERROR_TP,
-                                        MAX_ENERGY_PERCENT_ERROR_TP));
-    ntuple_.setVar<float>("TrigPrimEnergy", tp_energy);
-    ntuple_.setVar<int>("TrigPrimDigiEncoded", trigDigi.getPrimitive());
-    ntuple_.setVar<int>("TrigPrimDigiLinear", trigDigi.linearPrimitive());
-
+    if (daqDigis.getNumDigis() == 1) {
```
tomeichlersmith, line 251: Why did you put this in? The number of digis (i.e. number of emulated hits) should be 1 by construction. Anything different should fail the test. If the test is falling apart after this, then you could do `REQUIRE(daqDigis.getNumDigis() == 1)` so that the test is stopped if it fails.

tomeichlersmith (commented):
Sorry, missed something on the first pass

Ecal/test/EcalDigiPipelineTest.cxx
tvami: if the charge is high enough it will go to TOT mode so the read out digi might not be 1 after this PR. So the tests failed... I'm not sure I understand your proposal: if the test fails then we cant even trigger the PR validation CI

tvami: Here is the action when it failed: https://github.com/LDMX-Software/ldmx-sw/actions/runs/13301313095/job/37143110007

tvami: > might not be 1 after this PR

and I mean that's the whole point of the PR that there are hits we should (and wont) read out

tomeichlersmith: Ok, to be clear about vocab I use in this code. Each "hit" aka "digi" has 10 "samples", nothing about the specifics of the measurement within the chip changes how many samples are read out. The `getNumDigis()` function returns the number of digis that have been readout (i.e. this number is the number of hits that would be reconsturcted). If a simulated hit would not pass readout (TOT too late or ADC too low), then I could see `getNumDigis() == 0`.

tvami: >  If a simulated hit would not pass readout (TOT too late or ADC too low), then I could see getNumDigis() == 0.

Exactly, and then this test in its original form will fail; but I dont want it to fail thus the condition I added here

tomeichlersmith (approved):

Issue #1481, opened 2024-09-30
Add pass name to tracking code (CKFProcessor, Vertexing, etc)


tvami:
**Is your feature request related to a problem? Please describe.**

Filippo was running ACTSv35 on an existing eN sample where the tracking branches were already created and ran into the issue of
```
[ fire ] 1 fatal: [ProductAmbiguous] : Multiple products found for name 'RecoilTruthSeeds' without specified pass name :
{ name = RecoilTruthSeeds, pass = genie_reco, type = vector<ldmx::Track>}
{ name = RecoilTruthSeeds, pass = genie_reco_v36, type = vector<ldmx::Track>}
  at /Users/fdelzanno/Desktop/Incandela/Coding/ldmx-sw/Framework/include/Framework/Event.h:300 in getObject
```

**Describe the solution you'd like**

this
https://github.com/LDMX-Software/ldmx-sw/blob/34d43a553320a549b81a9dd30f31ab7c826a0d93/Tracking/src/Tracking/Reco/CKFProcessor.cxx#L239
```
 event.getCollection<ldmx::Track>(seed_coll_name_);
```
should be
```
 event.getCollection<ldmx::Track>(seed_coll_name_, input_pass_name_);
```
probably true for the vertexer and other processors in the tracking.

**Describe alternatives you've considered**

Dropping the old collection and running on the new one in two steps. But this wont allow comparisons in one config for sure

bryngemark:
good catch, configurable input collection and pass names, and output collection names, should be standard for all processors in my opinion

tvami:
Yes, I agree!

tvami:
Resolved in https://github.com/LDMX-Software/ldmx-sw/pull/1607

PR #1603, opened 2025-02-13
Resolves Issue 1481 - adds the functionality to use input names in tracking

LiamBrennan-UCSB:
This resolves Issue #1481 where Track reconstruction lacked configurable input collection and pass names. The feature has been added.

## Check List


Tracking/include/Tracking/Reco/CKFProcessor.h
```diff
@@ -205,6 +205,8 @@ class CKFProcessor final : public TrackingGeometryUser {
   // The seed track collection
   std::string seed_coll_name_{"seedTracks"};
 
+  std::string input_pass_name_{"input_pass_name"};
```
tvami:
```suggestion
  std::string input_pass_name_{""};
```

Tracking/src/Tracking/Reco/CKFProcessor.cxx
```diff
@@ -729,6 +729,9 @@ void CKFProcessor::configure(framework::config::Parameters& parameters) {
   seed_coll_name_ =
       parameters.getParameter<std::string>("seed_coll_name", "seedTracks");
 
+  input_pass_name_ =
+      parameters.getParameter<std::string>("input_pass_name", "input_pass_name");
```
tvami:
```suggestion
      parameters.getParameter<std::string>("input_pass_name", "");
```

tvami (changes requested):
The pass name should be the empty string by default. Also please add the same to the Vertex producer too

PR #1605, opened 2025-02-13
Improve ECAL digi random number generation

tvami:

This is part 1) of https://github.com/LDMX-Software/ldmx-sw/issues/1388
I'll make the same for TS after we agreed that this is the way to go.

## Check List

ECAL PN config runs


tvami:
No KS failure triggered, the logs show changes in the energy for the trigger to fluctuate, but it's not visible in the plots
Screenshot 2025-02-14 at 07 20 51

LiamBrennan-UCSB (approved):

tomeichlersmith (approved):

PR #1614, opened 2025-02-16
Improve TrigScint digi random number generation

tvami:

This is part 2 of https://github.com/LDMX-Software/ldmx-sw/issues/1388

## Check List

Inclusive sample runs fine

### DIFF
--- a/.github/validation_samples/it_pileup/config.py
+++ b/.github/validation_samples/it_pileup/config.py
@@ -61,9 +61,17 @@ ts_digis = [
         TrigScintDigiProducer.pad2(),
         TrigScintDigiProducer.pad3(),
         ]
-for d in ts_digis :
-    d.randomSeed = 1
-    d.input_collection += overlayStr
+for digi in ts_digis :
+    digi.input_collection += overlayStr
+
+ts_clusters = [
+        TrigScintClusterProducer.pad1(),
+        TrigScintClusterProducer.pad2(),
+        TrigScintClusterProducer.pad3(),
+        ] 
+for clu in ts_clusters :
+    clu.input_collection += overlayStr
+
 
 # Load the ECAL modules                           
 ecalDigi   =eDigi.EcalDigiProducer('ecalDigis')
@@ -107,9 +115,7 @@ p.sequence.extend([
     ecalDigi, ecalReco, ecalVeto,
     hcalDigi, hcalReco, hcal_veto,
     *ts_digis,
-    TrigScintClusterProducer.pad1(),
-    TrigScintClusterProducer.pad2(),
-    TrigScintClusterProducer.pad3(),
+    *ts_clusters,
     trigScintTrack,
     count, TriggerProcessor('trigger', 8000.),
     dqm.PhotoNuclearDQM(verbose=False),
--- a/.github/validation_samples/it_pileup/config.py
+++ b/.github/validation_samples/it_pileup/config.py
@@ -7,8 +7,8 @@ from LDMX.Framework import ldmxcfg
 
 # first, we define the process, which must have a name which identifies this
 # processing pass ("pass name").
-simPassName="test"
-pileupFilePassName="test"
+simPassName="ecal_pn"
+pileupFilePassName="pileup"
 thisPassName="overlay"
 p=ldmxcfg.Process(thisPassName)
 
@@ -69,9 +69,6 @@ ts_clusters = [
         TrigScintClusterProducer.pad2(),
         TrigScintClusterProducer.pad3(),
         ] 
-for clu in ts_clusters :
-    clu.input_collection += overlayStr
-
 
 # Load the ECAL modules                           
 ecalDigi   =eDigi.EcalDigiProducer('ecalDigis')
--- a/TrigScint/src/TrigScint/TrigScintDigiProducer.cxx
+++ b/TrigScint/src/TrigScint/TrigScintDigiProducer.cxx
@@ -1,10 +1,5 @@
 #include "TrigScint/TrigScintDigiProducer.h"
 
-#include <iostream>
-
-#include "Framework/Exception/Exception.h"
-#include "Framework/RandomNumberSeedService.h"
-
 namespace trigscint {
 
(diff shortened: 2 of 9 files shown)


tvami:
The failing plots are consistent with the noise change. Looking at the channel ID (Fig 5) it seems a little bit even more uniform actually.
[inclusive_recon_validation_plots.pdf](https://github.com/user-attachments/files/18813621/inclusive_recon_validation_plots.pdf)

tomeichlersmith (approved):

.github/validation_samples/it_pileup/config.py
```diff
@@ -61,9 +61,17 @@
...
+for digi in ts_digis :
+    digi.input_collection += overlayStr
+
+ts_clusters = [
+        TrigScintClusterProducer.pad1(),
+        TrigScintClusterProducer.pad2(),
+        TrigScintClusterProducer.pad3(),
+        ] 
+for clu in ts_clusters :
+    clu.input_collection += overlayStr
```
bryngemark: are you sure this is working? is `overlayStr` appended to the output collection of the digi step? otherwise it shouldn't be able to find it...

TrigScint/src/TrigScint/TrigScintDigiProducer.cxx
```diff
@@ -92,14 +84,15 @@ void TrigScintDigiProducer::produce(framework::Event &event) {
...
           beamFrac[id] = contrib.edep;
-        else
+        } else {
           beamFrac[id] += contrib.edep;
+        }
       }
     }
 
-    // for now, we take am energy weighted average of the hit in each stip to
+    // for now, we take am energy weighted average of the hit in each strip to
```
bryngemark: then you might as well correct the typo am --> an 😄

TrigScint/src/TrigScint/TrigScintDigiProducer.cxx
```diff
@@ -137,9 +130,11 @@ void TrigScintDigiProducer::produce(framework::Event &event) {
     Ypos[id] = Ypos[id] / Edep[id];
     Zpos[id] = Zpos[id] / Edep[id];
     double meanPE = depEnergy / mevPerMip_ * pePerMip_;
-    cellPEs[id] = random_->Poisson(meanPE + meanNoise_);
+    std::poisson_distribution<int> poisson_dist(meanPE + meanNoise_);
+    cellPEs[id] = poisson_dist(rng_);
 
     // If a cell has a PE count above threshold, persit the hit.
+    // TAV: we should move the threshold to be an input variable
```
bryngemark: I think this is 1 PE as it is the minimum thinkable quantum so I'm not so sure it needs to be configurable

bryngemark (commented):
Looks reasonable to me, thanks for picking this up after so long. Just a few detailed comments to answer before merge, the validation plots look alright.

TrigScint/src/TrigScint/TrigScintDigiProducer.cxx
tvami: haha I missed that indeed

TrigScint/src/TrigScint/TrigScintDigiProducer.cxx
tvami: yes but dont you see a chance for needing to go higher?

.github/validation_samples/it_pileup/config.py
tvami: I'm not sure but 1) this is how it's been in this config, 2) I'd expect that the CI would fail if this doesnt exists, right?

tvami: >  this is how it's been in this config

on this I mean that the original config never appended the overlayStr to the digi output, maybe that's been wrong all along. Still confusing why 2) would work

TrigScint/src/TrigScint/TrigScintDigiProducer.cxx
bryngemark: no. we let the hits be whatever they are. thresholds are introduced (and configurable) in clustering. 

as a bit of context, only the electronics noise (SiPM shot noise) is this low, and, indeed it is this low. this cut only suppresses artifical noise that is below one light quantum in the SiPM and unphysical.

.github/validation_samples/it_pileup/config.py
bryngemark: i think you are right about 2) but i am also confused.

bryngemark: oh but the CI did fail.

```
  [ fire ] 1 fatal: [ProductNotFound] : No product found for name 'trigScintDigisPad1Overlay'
    at /home/runner/work/ldmx-sw/ldmx-sw/Framework/include/Framework/Event.h:292 in getObject
```

bryngemark: there are two ways to get the right collection.
- set a new output name from digi
- leave as is but use the pass name of the overlay process for the clustering input

Looking at the cluster plots you attached to the PR, one could suspect that they didn't contain pileup before? Considering the TS granularity, in 1/24~4% of the 2e cases you should see 2 MIPs per cluster which should cause extra bumps in the cluster PE distribution.

bryngemark (commented):
Once we get the clustering part right this should be ok.

TrigScint/src/TrigScint/TrigScintDigiProducer.cxx
tvami: Thanks this is great, I'll add this comment in the code

.github/validation_samples/it_pileup/config.py
tvami: Ok... that whole config is wrong, almost everything is taking the 1e instead of the 2e. I'll fix it in a next PR, for now I just made this functional

tvami:
It ran correctly now:
https://github.com/LDMX-Software/ldmx-sw/actions/runs/13425821216/job/37509148060

bryngemark (approved):
looks good now. The clusters seem to have the right number of PEs for 2e

PR #1617, opened 2025-02-18
Error bars for validation histograms

### LINKED ISSUE #1616: Error bars for validation histograms
When we're looking at histograms failing the KS test it is sometimes hard to assess if differences are statistically significant.

It would be great guidance to the human eye to have error bars at least on the test distribution (but also on the reference distribution if possible).

Keep doing the math in my head

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1616

## Check List

Screenshot 2025-02-18 at 12 58 47

### DIFF
--- a/.github/actions/validate/compare.py
+++ b/.github/actions/validate/compare.py
@@ -150,11 +150,10 @@ def compare(gold_f, gold_label, test_f, test_label) :
         The label used in the histogram legends for the test histograms
     """
 
-    gold = HistogramFile(gold_f,gold_label,ROOT.kRed ,True )
+    gold = HistogramFile(gold_f,gold_label,ROOT.kRed ,False )
     test = HistogramFile(test_f,test_label,ROOT.kBlue,False)
 
     c = ROOT.TCanvas()
-    c.SetLogy()
 
     os.makedirs(f'plots/pass',exist_ok=True)
     os.makedirs(f'plots/fail',exist_ok=True)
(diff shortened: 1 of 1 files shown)


tvami:
Plots have error bars now:
https://github.com/LDMX-Software/ldmx-sw/actions/runs/13400553664

And given that all the 1D are normalized now, every KS test failed, as expectedd

bryngemark:
> And given that all the 1D are normalized now, every KS test failed, as expectedd

I think they actually failed on the logs? Both histograms are normalized so that shouldn't make the KS test fail. Looking at the ecal PN ones for instance, there are no failed ones and most seem to be indistinguishable. But in the logs there is an ECal energy difference of a few MeV. 

[EcalMipTrackingFeatures_EcalMipTrackingFeatures_n_straight_tracks.pdf](https://github.com/user-attachments/files/18864899/EcalMipTrackingFeatures_EcalMipTrackingFeatures_n_straight_tracks.pdf) --> some difference 
[EcalDigiVerify_EcalDigiVerify_rec_sim_hit_residual_y.pdf](https://github.com/user-attachments/files/18865085/EcalDigiVerify_EcalDigiVerify_rec_sim_hit_residual_y.pdf) --> complete overlap


This is picky I know: could we keep the gold filled? I think it would make them easier to distinguish (right now, when there is no difference, the reference histogram looks empty since it is drawn first and hidden).

tvami:
> I think they actually failed on the logs?

You are right! But that's not coming from this PR. I guess it's time for new gold. I'll do that after https://github.com/LDMX-Software/ldmx-sw/pull/1614 is merged

> could we keep the gold filled? 

No that didnt work, I did try that first. If I include the filling, it doesnt fill it with the error option on.

But also isnt it just fine to see complete overlap? It's clear sign that nothing changed

bryngemark (approved):
Looks good to me. We agreed in sw dev that people will learn that complete overlap means Nothing Changed.

tomeichlersmith (approved):

PR #1622, opened 2025-02-20
Add pass names and Overlay strings in 2e CI config

### LINKED ISSUE #1620: Fix 2e config CI config to use the overlaid collections

Right now the pass-names are not passed correctly so we are not actually using the overlaid collections

Use the overlaid collections

N/A

**Additional context**

Came up in https://github.com/LDMX-Software/ldmx-sw/pull/1614

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1620

## Check List

### DIFF
--- a/Recon/src/Recon/OverlayProducer.cxx
+++ b/Recon/src/Recon/OverlayProducer.cxx
@@ -65,8 +65,8 @@ void OverlayProducer::onNewRun(const ldmx::RunHeader &) {
         framework::RandomNumberSeedService::CONDITIONS_OBJECT_NAME);
     rndm_ = std::make_unique<TRandom2>(rnss.getSeed("OverlayProducer::rndm"));
   }
-
-  int start_event = rndm_->Uniform(20., 1e4);
+  // TAV: These boundaries should be configurable
+  int start_event = rndm_->Uniform(20., 4570);
   // EventFile::skipToEvent handles actual number of events in file
   int evNb = overlayFile_->skipToEvent(start_event);
   if (evNb < 0) {
@@ -118,6 +118,7 @@ void OverlayProducer::produce(framework::Event &event) {
         event.getCollection<ldmx::SimCalorimeterHit>(collName, simPassName_);
     // but don't copy ecal hits immediately: for them, wait until overlay
     // contribs have been added. then add everything through the hitmap
+    // TAV: the string "Overlay" should be configurable too
     if (!needsContribsAdded) {
       caloCollectionMap[collName + "Overlay"] = simHitsCalo;
     }
--- a/.github/validation_samples/it_pileup/config.py
+++ b/.github/validation_samples/it_pileup/config.py
@@ -119,6 +119,28 @@ hcal_veto.input_hit_pass_name = thisPassName
 # Load the DQM modules
 from LDMX.DQM import dqm
 
+trigScint_sim_dqm = [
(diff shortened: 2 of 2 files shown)


tomeichlersmith (approved):

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -65,8 +65,8 @@ void OverlayProducer::onNewRun(const ldmx::RunHeader &) {
         framework::RandomNumberSeedService::CONDITIONS_OBJECT_NAME);
     rndm_ = std::make_unique<TRandom2>(rnss.getSeed("OverlayProducer::rndm"));
   }
-
-  int start_event = rndm_->Uniform(20., 1e4);
+  // TAV: These boundaries should be configurable
+  int start_event = rndm_->Uniform(20., 4570);
```
bryngemark, line 69: what's this change about?

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -118,6 +118,7 @@ void OverlayProducer::produce(framework::Event &event) {
         event.getCollection<ldmx::SimCalorimeterHit>(collName, simPassName_);
     // but don't copy ecal hits immediately: for them, wait until overlay
     // contribs have been added. then add everything through the hitmap
+    // TAV: the string "Overlay" should be configurable too
```
bryngemark, line 121: Why? A string needs to be added becasue you might (in fact people often do) run sim and overlay in the same pass, so there needs to be a different collection name for the overlay procedure – combining two simhit collections from the same system – to work. Always using "Overlay" makes it very clear what these simhits are. I think making it configurable is asking for trouble.

bryngemark (commented):
thanks for pulling this together. i had a few questions on your changes to the overlay producer itself only.

Recon/src/Recon/OverlayProducer.cxx
tvami: Every (maybe just most?) collection name is configurable in ldmx-sw, and the same argument you used here could be said about those too, like changing the name of the track collection, etc. I dont expect this to be changed, but right now it's "Overlay" inserted into many places, if you change one of them the rest will not follow, which is not a good practice. And at that point, having the consistency with other producers to be able to change the name if needed would be good, or at least "uniform"

Recon/src/Recon/OverlayProducer.cxx
tvami: Yeah so this is not great in itself, but I defer the fix to a next PR.
What this line does is that it picks a random number between 20 and 10k, and then uses that event to overlay with. So if you ever made a PU library that had more than 10k event, none of the events above 10k were ever used...

The change: the 10k was originally in there (I assume) because in the test we go upto 10k times X events. Now in a prev PR I changed that to go to 5k times X events, so the PU CI finishes in the same timescale as the rest. This change follows that change.

Of course the correct way  would be, as I commented, (a) to have this configurable, or (b) somehow learn how many events are in the PU library and use that.

tvami:
@bryngemark I'd like to cut a patch soon, please approve this PR

Recon/src/Recon/OverlayProducer.cxx
tvami: Made an issue here https://github.com/LDMX-Software/ldmx-sw/issues/1638
we can come back to this in a next PR

LiamBrennan-UCSB (approved):

PR #1624, opened 2025-02-20
Additional information for 'dump files' used in firmware testbenchs

therwig:
I am expanding the information saved in the 'dump file' event format, used the pass trigger primitives to the firmware test bench setups.  Instead of just a packed detector ID and its encoded energy, extra information is saved for the module, layer, cell, and linearized energy in order to avoid duplicating the unpacking logic outside of ldmx-sw.

## Check List

I dumped 20 events into this new format and verified that I can read them in the firmware testbench.

### DIFF
--- a/Trigger/Algo/include/Trigger/DiscreteInputs.h
+++ b/Trigger/Algo/include/Trigger/DiscreteInputs.h
@@ -11,22 +11,47 @@ namespace trigger {
 namespace ldmx_int {
 
 struct EcalTP {
+  uint8_t tp;
   uint32_t tid;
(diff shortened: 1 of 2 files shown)


tvami:
Nothing will run/test this in ldmx-sw, right?

therwig:
That's right. I did compare the outputs produced here with the inputs that are read in the firmware testbench, which look good.  We could think about ways to better validate this workflow for the future; may be interesting to discuss with others in tdaq.

tvami (approved):

tomeichlersmith (approved):

PR #1626, opened 2025-02-20
Persist the pass name's version in the event header

### LINKED ISSUE #1559: Versioning the branches

Many times it would be good to know what version of ldmx-sw a given branch was created. For example re-running the ecal veto with a new BDT. Or just generally understand the files from the past. For now one needs to have a good bookkeeping system, and https://github.com/LDMX-Software/ldmx-sw/pull/1539 and follow-ups made sure the SIM step is persisted correctly. 

Add the version number into the `pass` name. It is many times not useful, so I'd suggest that if the 
```
p = ldmxcfg.Process('test')
```
does not have an argument
```
p = ldmxcfg.Process()
```
then the `pass` name becomes the VERSION.
IF there is an argument, then the pass name becomes the `VERSION_argument`

Let the subsequent processes add a new run parameter with `setStringParameter`

**Additional context**

Here https://github.com/LDMX-Software/ldmx-sw/pull/1539#issuecomment-2628514073

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1559

## Check List

The printout 
```
  stringParameters: 
    BiasOperators::PhotoNuclear::Volume = ecal
    Geant4 revision = 
    Gen0 Class = simcore::generators::ParticleGun
    Gen0 Particle = e-
    ldmx-sw revision = 446e5c3345e00ee7aff45e249d6502d2e69b861f
    ldmx-sw version = v4.2.13
    test_version = v4.2.13
```

### DIFF
--- a/Framework/src/Framework/Process.cxx
+++ b/Framework/src/Framework/Process.cxx
@@ -461,7 +461,8 @@ void Process::newRun(ldmx::RunHeader &header) {
   // the run header through 'beforeNewRun' method
 
   // Put the version into the rh string param
-  header.setStringParameter("Pass = " + pass_name_ + " version", LDMXSW_VERSION);
+  header.setStringParameter("Pass = " + pass_name_ + " version",
+                            LDMXSW_VERSION);
   if (performance_) performance_->start(performance::Callback::beforeNewRun, 0);
   std::size_t i_proc{0};
   for (auto module : sequence_) {
--- a/Framework/src/Framework/Process.cxx
+++ b/Framework/src/Framework/Process.cxx
@@ -461,7 +461,7 @@ void Process::newRun(ldmx::RunHeader &header) {
   // the run header through 'beforeNewRun' method
 
   // Put the version into the rh string param
-  header.setStringParameter(pass_name_ + "_version", LDMXSW_VERSION);
+  header.setStringParameter("Pass = " + pass_name_ + " version", LDMXSW_VERSION);
   if (performance_) performance_->start(performance::Callback::beforeNewRun, 0);
   std::size_t i_proc{0};
   for (auto module : sequence_) {
--- a/Framework/src/Framework/Process.cxx
+++ b/Framework/src/Framework/Process.cxx
@@ -24,7 +24,7 @@ Process::Process(const framework::config::Parameters &configuration)
     : conditions_{*this} {
   config_ = configuration;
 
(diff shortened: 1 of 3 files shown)


tvami:
It shows up nicely in the logs too:
Screenshot 2025-02-21 at 16 27 07

tomeichlersmith:
This is great! thank you (do we even want the `ldmx-sw version` string there anymore?)

tvami:
> (do we even want the ldmx-sw version string there anymore?)

That is for what the simulator use, I think it makes sense to keep it (although renaming might be helpful), if the operations are to have several "rereco"-s of the same SIM dataset.

tomeichlersmith (approved):

Framework/src/Framework/Process.cxx
```diff
@@ -459,6 +459,9 @@ TDirectory *Process::openHistoFile() {
 void Process::newRun(ldmx::RunHeader &header) {
   // Producers are allowed to put parameters into
   // the run header through 'beforeNewRun' method
+
+  // Put the version into the rh string param
+  header.setStringParameter(pass_name_ + "_version", LDMXSW_VERSION);
```
tvami:
```suggestion
  header.setStringParameter("Pass = " + pass_name_ + " version", LDMXSW_VERSION);
```

tvami:
I included SIM in the SIM name, and then also made the printout style simulate to the other things in the run header

tomeichlersmith (approved):

Framework/src/Framework/Process.cxx
```diff
@@ -459,6 +459,10 @@ TDirectory *Process::openHistoFile() {
 void Process::newRun(ldmx::RunHeader &header) {
   // Producers are allowed to put parameters into
   // the run header through 'beforeNewRun' method
+
+  // Put the version into the rh string param
+  header.setStringParameter("Pass = " + pass_name_ + " version",
```
tvami:
```suggestion
  header.setStringParameter("Pass = " + pass_name_ + ", version",
```

tvami:
OK this is fully tested now with https://github.com/LDMX-Software/ldmx-sw/pull/1628

I ran a `test2` on the top of the file with `test` and when I read it in with that PR's content, I see both `test` and `test2`

LiamBrennan-UCSB (approved):

PR #1630, opened 2025-02-21
Move RunManager and Biasing to logging system

tvami:

No issue, just cleanup

## Check List

### DIFF
--- a/SimCore/include/SimCore/XsecBiasingOperator.h
+++ b/SimCore/include/SimCore/XsecBiasingOperator.h
@@ -3,6 +3,7 @@
 
 #include "Framework/Configure/Parameters.h"
 #include "Framework/EventProcessor.h"
+#include "Framework/Logger.h"
 #include "Framework/RunHeader.h"
 #include "SimCore/Factory.h"
 
@@ -177,7 +178,7 @@ class XsecBiasingOperator : public G4VBiasingOperator {
     return nullptr;
   }
   /// Enable logging
-  enableLogging("XsecBiasingOperator")
+  framework::logging::logger theLog_;
 
 };  // XsecBiasingOperator
 }  // namespace simcore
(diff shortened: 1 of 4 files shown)


SimCore/include/SimCore/XsecBiasingOperator.h
```diff
@@ -177,7 +177,7 @@ class XsecBiasingOperator : public G4VBiasingOperator {
     return nullptr;
   }
   /// Enable logging
-  enableLogging("EcalProcessFilter")
+  enableLogging("XsecBiasingOperator")
```
tomeichlersmith: Instead of hard-coding the channel name, I would like the channel name to align with the derived biasing operator's name. We can do this by avoiding the `enableLogging` macro.

This is done by event processors among other factory-created objects.

https://github.com/LDMX-Software/ldmx-sw/blob/446e5c3345e00ee7aff45e249d6502d2e69b861f/Framework/include/Framework/EventProcessor.h#L221-L222

https://github.com/LDMX-Software/ldmx-sw/blob/446e5c3345e00ee7aff45e249d6502d2e69b861f/Framework/src/Framework/EventProcessor.cxx#L13

tomeichlersmith (changes requested):
One update to allow for the name of the biasing operator to be passed into the logging channel.

SimCore/include/SimCore/XsecBiasingOperator.h
tvami: So your suggestion is to simply?
```suggestion
```

tomeichlersmith: You would remove `enableLogging` and then add the declaration `logging::logger theLog_;` in the header and definition `theLog_{logging::makeLogger(name)}` in the constructor's definition in the source.

tvami: Done in https://github.com/LDMX-Software/ldmx-sw/pull/1630/commits/772493fcbd35d12c466a8c4d7be7066af0fa4d58

tomeichlersmith (approved):

LiamBrennan-UCSB (approved):

PR #1631, opened 2025-02-21
Processor to flag fiducial recoil electrons

### LINKED ISSUE #1587: Filter for tagging fiducial events
For consistent definitions of signal efficiency across different analyses, we would like to define a clear fiducial volume, and classify events based on this definition.

We've discussed the following truth-level selections on the recoil electron:

1. |p| < 50 MeV (at production)
2. \> 4 recoil tracker layers with RecoilSimHits
3. \> 0  ECalSimHit

It would be nice to have a filter (or processor) to flag events based on whether or not they satisfy these conditions. 

This could also be done at the analysis level, but this leaves room for inconsistency. 

**Additional context**
This is based on discussions at a recent [SWAN meeting](https://indico.fnal.gov/event/68163/contributions/308977/attachments/185117/254756/SWAN02102025.pdf). @tvami @therwig @bryngemark

EBerzin:

This resolves #1587.

I added a processor to check if the recoil electron in an event satisfies certain fiduciality conditions. The output is saved in a new `FiducialFlag` object, which is structured similarly to `TriggerResult`. It includes booleans for whether the recoil electron has the minimum energy at production (by default 50 MeV), the minimum number of hits in the tracker (by default 5 hits), and has a hit in the ECal/HCal. It also includes an integer that combines this information (this may be redundant).

`FiducialFlag` also saves the variables that were used to set the flags, such as the recoil electron energy at production, the number of recoil electron hits in the tracker, and also includes the detector ID of the last hit the recoil electron leaves in the ECal or HCal. 

## Check List

### DIFF
--- a/Recon/include/Recon/Event/FiducialFlag.h
+++ b/Recon/include/Recon/Event/FiducialFlag.h
@@ -32,7 +32,7 @@ class FiducialFlag {
   /**
    * Class destructor.
    */
-  virtual ~FiducialFlag();
+  virtual ~FiducialFlag() = default;
 
   /**
    * Print a description of this object.
--- a/Recon/include/Recon/Event/FiducialFlag.h
+++ b/Recon/include/Recon/Event/FiducialFlag.h
@@ -48,37 +48,37 @@ class FiducialFlag {
    * Return true/false if event is fiducial.
    * @return True if fiducial.
    */
-  bool isFiducial() const { return isFiducial_; }
+  bool isFiducial() const { return is_fiducial_; }
 
   /**
    * Return fiducial flag bit mask.
    * @return Fiducial flag bit mask.
    */
-  int getFiducialFlag() const { return fiducialFlag_; }
+  int getFiducialFlag() const { return fiducial_flag_; }
 
   /**
    * Return true/false if event has ecal hit.
    * @return True if ecal hit.
    */
-  bool hasEcalHit() const { return hasEcalHit_; }
+  bool hasEcalHit() const { return has_ecal_hit_; }
 
   /**
    * Return true/false if event has hcal hit.
    * @return True if hcal hit.
    */
-  bool hasHcalHit() const { return hasHcalHit_; }
+  bool hasHcalHit() const { return has_hcal_hit_; }
 
   /**
    * Return true/false if event has min. number of tracker hits.
    * @return True if >= min tracker hits.
    */
-  bool hasMinTrackerHits() const { return hasMinTrackerHits_; }
+  bool hasMinTrackerHits() const { return has_min_tracker_hits_; }
 
   /**
    * Return true/false if event has min. recoil energy at production.
    * @return True if >= min energy.
    */
-  bool hasMinEnergy() const { return hasMinEnergy_; }
+  bool hasMinEnergy() const { return has_min_energy_; }
 
   /**
    * Return algorithm variable i (see algorithm code for details).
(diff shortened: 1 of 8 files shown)


Recon/include/Recon/Event/FiducialFlag.h
```diff
@@ -0,0 +1,205 @@
...
+ public:
+  /**
+   * Class constructor.
+   */
+  FiducialFlag();
+
+  /**
+   * Class destructor.
+   */
+  virtual ~FiducialFlag();
```
tvami:
```suggestion
  virtual ~FiducialFlag() = default;
```

Recon/include/Recon/Event/FiducialFlag.h
```diff
@@ -0,0 +1,205 @@
...
+  /**
+   * Set an algorithm variable.
+   * @param element The index of the variable.
+   * @param value The variable's new value.
+   */
+  void setAlgoVar(int element, double value);
+
+ private:
+  /* Bit-mask that represents fiduciality conditions. */
+  int fiducialFlag_{0};
```
tvami: please move the variables to `snake_case_`, I know the Trigger object had the `camelCase_`, but we are trying to move everything to the `snake_case_` convention, especially new code [and at one point old code too -- I guess after the DR business is done]

Recon/include/Recon/RecoilFiducialityProcessor.h
```diff
@@ -0,0 +1,80 @@
...
+   * Class constructor.
+   */
+  RecoilFiducialityProcessor(const std::string& name,
+                             framework::Process& process)
+      : framework::Producer(name, process) {}
+
+  /**
+   * Class destructor.
+   */
+  virtual ~RecoilFiducialityProcessor() {}
```
tvami: With the default destructor this is not needed
```suggestion
  /**
   * Class destructor.
   */
  virtual ~RecoilFiducialityProcessor() = default;
```

bloodyyugo (approved):

Recon/src/Recon/RecoilFiducialityProcessor.cxx
```diff
@@ -0,0 +1,132 @@
...
+  flag.setHasHcalHit(hasHcalHit);
+
+  event.add(output_collection_, flag);
+
+  // Tell the skimmer to keep or drop the event based on whether there
+  // were recoil electron was fiducial.
+  if (isFiducialFlag) {
+    setStorageHint(framework::hint_shouldDrop);
+  } else {
+    setStorageHint(framework::hint_shouldKeep);
```
tvami: In the issue we talked about having an inverse rule possibility similar to https://github.com/LDMX-Software/ldmx-sw/commit/9bf6acc1544ed7e7ac2834eb5d0a16cccefa6499

Recon/src/Recon/RecoilFiducialityProcessor.cxx
```diff
@@ -0,0 +1,132 @@
+/**
+ * @file RecoilFiducialityProcessor.cxx
```
tvami: if you touch this file again, this still has `camelCase` in it, altho they are local variables, so less problematic to rename them later

Recon/src/Recon/RecoilFiducialityProcessor.cxx
EBerzin: Right, thanks for the reminder. I also realized the logic in the standard skim was backwards (oops!). Should be corrected now, including the inverse rule option.

tvami (approved):
Looks good now, thanks!

tvami (commented):
hey @EBerzin I just realized that you didnt put this under the `Skims` as I suggested in https://github.com/LDMX-Software/ldmx-sw/issues/1587#issuecomment-2652412803

The reason I suggested this to be under the `Skims` is because this acts like a skimming  package. But in the end it might be better under the SimCore, since every input is at SIM level. I mentioned that in https://github.com/LDMX-Software/ldmx-sw/issues/1587#issuecomment-2652416155 and I hoped somebody would chime in. Maybe let me just tag @tomeichlersmith about this point here

tomeichlersmith:
I think `Skims` or where it is now makes sense.

I don't think it should go into SimCore because it doesn't run during the simulation. We could look at moving a bunch of sim-input analyzers into a new "SimAna" module but I also think it makes sense to keep them separated by subsystem.

tvami:
Thanks @EBerzin ! I think this is great now!

Recon/src/Recon/Event/FiducialFlag.cxx
```diff
@@ -0,0 +1,68 @@
...
+
+  void FiducialFlag::setFiducialFlag(int fiducial_flag, int nvar) {
+    fiducial_flag_ = fiducial_flag;
+
+    if (nvar > variables_.GetSize()) {
+      variables_.Set(nvar);
+    }
+  }
+
+  void FiducialFlag::setIsFiducial(bool is_fiducial) {
```
bryngemark: i'm used to seeing our one-liner setters and getters declared in the header file. should we have some established style praxis here?

Recon/src/Recon/Skims/RecoilFiducialityProcessor.cxx
```diff
@@ -0,0 +1,143 @@
...
+void RecoilFiducialityProcessor::produce(framework::Event &event) {
+  // Get the collection of simulated particles from the event
+  auto particleMap{event.getMap<int, ldmx::SimParticle>("SimParticles")};
+
+  // Search for the recoil electron
+  auto [recoil_track_id, recoil_electron] = Analysis::getRecoil(particleMap);
+
+  // Get the collection of simulated Ecal hits from the event.
+  const std::vector<ldmx::SimCalorimeterHit> ecal_sim_hits =
+      event.getCollection<ldmx::SimCalorimeterHit>(ecal_collection_);
```
bryngemark: please add pass name here and to all other collections + to the configuration and python config (defaulting to an empty string). selecting among several passes is a very useful capability that all processors should have, imo

bryngemark (commented):
Thanks @EBerzin ! Do you have any plots showing the result or other supporting/documenting material for this PR?

bryngemark:
>  I mentioned that in [#1587 (comment)](https://github.com/LDMX-Software/ldmx-sw/issues/1587#issuecomment-2652416155) and I hoped somebody would chime in. 

I did give a thumbs up 😄  but after @therwig 's comment about only using truth info I am also content with leaving it where it is.

EBerzin:
@bryngemark I don't have any additional supporting material at the moment beyond the SWAN talk. To validate I confirmed that the processor gave me the same results as the analysis I had done to get the results shown in the talk. Is there something else that would be good to show here?

Recon/src/Recon/Skims/RecoilFiducialityProcessor.cxx
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  // Checking if the recoil electron energy is > min energy
+  bool has_min_energy = false;
+  if (recoil_electron->getEnergy() >= min_p_mag_) {
+    has_min_energy = true;
+  }
+
+  // Configure outputs
+  bool is_fiducial = has_min_energy & has_min_tracker_hits & has_ecal_hit;
```
tvami, line 103: I wonder if this was supposed to have been
```
bool is_fiducial = has_min_energy && has_min_tracker_hits && has_ecal_hit;
```
since it's not a bitwise operation? @EBerzin

EBerzin: Yes you're right, this should be &&.

tvami: Fix in https://github.com/LDMX-Software/ldmx-sw/pull/1641

PR #1634, opened 2025-02-24
1612 patch version deduction

### LINKED ISSUE #1612: ldmx-sw version is one release behind true value
**Describe the bug**
We are not updating `VERSION` until we make updated gold histograms, but that means the `VERSION` used when building the production image is still behind by one release.

**To Reproduce**
1. `denv init ldmx/pro:v4.2.11`
2. `denv fire ../ldmx-sw/SimCore/test/basic.py`
3. `denv config print`

**Desired behavior**
The `ldmx-sw version` and the `denv_image` should align.
I'm not sure how best to do this, but we should do the following steps in order whenever we want to make a release.
1. Choose a new version tag
2. Commit the new version tag to VERSION on trunk
3. Tag this commit as the version
4. Make GitHub Release with this tag
5. Build Production Image and Generate New Gold

Currently, the Generate New Gold workflow and Build Production Image workflow are both run at the same time off the same commit (which is properly tagged according to `git` but not according to `VERSION`) meaning that the production image reads a `VERSION` that is wrong.

**Additional context**
Reported by @jpasc27 on slack

Image

tomeichlersmith:

This resolves #1612 

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments. This works great in situations where the entire `git` tree is available, but the CI does a sparse/light checkout so `git describe --tags` fails.
- [x] I ran my developments and the following shows that they are successful

from CI https://github.com/LDMX-Software/ldmx-sw/actions/runs/13504011124
```
fatal: No tags can describe 'c4a44984342c2243848539fadbde5f018edf2be9'.
Try --always, or create some tags.
denv cmake -B build -S . -DADDITIONAL_WARNINGS=ON -DENABLE_CLANG_TIDY=ON -DWARNINGS_AS_ERRORS=ON
CMake Warning at CMakeLists.txt:21 (project):
  VERSION keyword not followed by a value or was followed by a value that
  expanded to nothing.
-- The CXX compiler identification is GNU 11.4.0
```
however, the compiling in the PR Validation workflow is functional becaues `fetch-depth: 0` in that clone (i.e. the CI does a full clone).
```
git fetch --tags && git describe --tags | cut -f 1 -d '-' > VERSION
denv cmake -B build -S . -DADDITIONAL_WARNINGS=ON -DENABLE_CLANG_TIDY=ON -DWARNINGS_AS_ERRORS=ON
-- The CXX compiler identification is GNU 11.4.0
```


tvami:
Should we have this in the next patch?

tomeichlersmith:
Yea I like that idea, I'll try to figure it out today

tomeichlersmith:
Using the `ready_for_review` trigger, we can use the workflow defined here and it shows that the PR Validation is working :tada:

image

tvami (approved):

PR #1635, opened 2025-02-24
Merge reduced LDMX tracking producer 

### LINKED ISSUE #1491: Adding a reduced geometry tracking producer

fdelzanno:

This resolves iss1491-a-reduced-tracking by adding 3 files to the tracking branch of ldmx-sw: a track seeder, a track finder, and a truth tracker optimized for the reduced LDMX geometry.

## Check List
See comments below for summary plots and descriptions of files.


fdelzanno:
The tracking logic is as follows:

**LinearSeedFinder.cxx** - this is the "seeder", which:
- Converts hits on the recoil layers (from RecoilDigiSim) into "sensor" hits. There are two sensors, made from an energy-weighted averaging of the positions on Layers 1 and 2, and those on Layers 3 and 4. Uncertainty from the Recoil geometry is assigned to the position in x and y.
- Constructs every combination of 2 sensor hits + 1 EcalRecHit on the first layer of the Ecal and uses those points to fit a straight line in 3D. Using the track parameters of the seed, we extrapolate the track to the first layer of the Ecal, and keep only seeds within a threshold distance of the RecHit used in fitting.

**LinearTrackFinder.cxx** - this is the track processor, which:
- Takes the seed collection from the LinearSeedFinder and processes them by the EcalRecHit associated with the seed.
By iterating through all the seeds corresponding to a RecHit, we select the seed with the lowest chi2 value as the track for that RecHit.
- We remove any remaining seed that uses the two sensors and/or EcalRecHit used by the "best track". If there are any seeds left, we select the next EcalRecHit, find the best track, and remove seeds with shared hits -- iterating until we have fit all RecHits or processed all seeds.

**LinearTruthTracking.cxx** - this is the "truth tracking", which:
- Finds 4 points in the RecoilDigiSim Hits with trackID == 1 (i.e. the recoil electron hits) and fits a straight line to those 4 points.
- The output track is the true Recoil electron for that event.

I added a new type of track object, called "StraightTrack", to contain the track parameters for the reduced geometry tracking. There is also a dqm available that can be added to a simulation ("StraightTracksDQM")

fdelzanno:
I ran a 5,000 event sample using a 4GeV eGun in the reduced-ldmx-v1 geometry to verify that the tracking is working. Here are the statistics:

There are 5000 events in the sample.
We find 5008 tracks / 4703 events with tracks + 297 empty events in which we don't find a track = **5008 tracks / 5000 events**.
We find 4808 truth tracks / 4808 events with truth tracks + 192 empty events in which we don't fit a truth track = **4808 tracks / 5000 events**

Out of the tracks in the sample, **4760 tracks** have trackID = 1 + **248 tracks** have a different trackID number = 5008 tracks

Out of the tracks with trackID = 1 in the sample:
- there are 4567 that have truthProb = 1 (i.e. all 4 hits used in tracking correspond to the recoil electron)
- 129 tracks that have truthProb = 0.75 (i.e. 3 hits / 4 used in tracking correspond to recoil electron)
- 64 tracks that have truthProb = 0.5 (i.e. 2 hits / 4 used in tracking correspond to recoil electron)

We reconstruct 4895 electron tracks + 112 positron tracks + 1 pion track = 5008 tracks/5008.

Here are the summary plots. In blue are all the reconstructed tracks. In green are the truth tracks. In red is the data only from reconstructed tracks with trackID = 1 (i.e. reconstructed recoil electron tracks).

Image

Image

It makes sense that the chi2 is lower for reconstructed tracks, since the tracking looks for tracks with the lowest chi2.

Image

This is the distance between the track extrapolated to the first Ecal layer and the RecHit used to create the track. The threshold value is 15 mm, so there is no reconstructed track with dRecHit > 15 mm. The radius of an Ecal cell is approximately 4 mm, meaning any track with dRecHit < 4 mm is likely to be in the correct cell. 

I also have the pulls of the reconstructed track beam (theta) and azimuthal (phi) angle (these come from the dqm, and are for unique tracks):

Image

Image

For events in which we construct a truth track, I plotted the difference between the location of the trackID = 1 reconstructed track and the truth track at both the Target and Ecal face. 

Image

Image

We expect a greater spread in y since it is the more uncertain direction.

Image

Image

.github/validation_samples/reduced_ldmx/config.py
```diff
@@ -68,6 +68,46 @@
...
+uSmearing = 0.006       #mm #could bump up to 10 micron if we want
+vSmearing = 0.000001    #mm #~unused
+
+# Smearing Processor - Recoil
+digiRecoil = tracking.DigitizationProcessor("DigitizationProcessorRecoil")
+digiRecoil.hit_collection = "RecoilSimHits"
+digiRecoil.out_collection = "DigiRecoilSimHits"
+digiRecoil.merge_hits = True
+digiRecoil.sigma_u = uSmearing
+digiRecoil.sigma_v = vSmearing
```
```suggestion
digiRecoil.sigma_u = 0.006
digiRecoil.sigma_v = 0.000001
```

.github/validation_samples/reduced_ldmx/config.py
```diff
@@ -68,6 +68,46 @@
...
+from LDMX.Tracking import geo
+from LDMX.Tracking import dqm
+
+from LDMX.Tracking.geo import TrackersTrackingGeometryProvider as trackgeo
+trackgeo.get_instance().setDetector(det)
+
+#smearings
+uSmearing = 0.006       #mm #could bump up to 10 micron if we want
+vSmearing = 0.000001    #mm #~unused
+
```
```suggestion
```

Detectors/test/reducedLDMX_eGun_config.py
```diff
@@ -26,11 +26,11 @@
 p.termLogLevel = 0
```
tvami: This file is somewhat historical, and was made before we made it into a CI. I'd not add the tracking stuff in here (and actually would delete this file to avoid future confusion)

Tracking/exampleConfigs/test_reduced_config.py
```diff
@@ -0,0 +1,123 @@
...
+# Smearing Processor - Recoil
+digiRecoil = tracking.DigitizationProcessor("DigitizationProcessorRecoil")
+digiRecoil.hit_collection = "RecoilSimHits"
+digiRecoil.out_collection = "DigiRecoilSimHits"
+digiRecoil.merge_hits = True
+digiRecoil.sigma_u = uSmearing
+digiRecoil.sigma_v = vSmearing
+
+#Assuming rLDMX-v1 geometry. These should be updated if one changes the position of the recoil layers
+#TODO: is there some way to get this directly from the gdml?
```
tvami: There isnt :(

Tracking/exampleConfigs/test_reduced_config.py
```diff
@@ -0,0 +1,123 @@
...
+layer23_mid = (15.5+24.5)/2.
+layer34_mid = (24.5+30.5)/2.
+
+truth_tracking = tracking.LinearTruthTracking("LinearTruthTracking")
+truth_tracking.input_hit_collection = "DigiRecoilSimHits"
+truth_tracking.input_recHits_collection = "EcalRecHits"
+truth_tracking.out_track_collection = "LinearRecoilTruthTracks"
+truth_tracking.layer12_midpoint = layer12_mid
+truth_tracking.layer23_midpoint = layer23_mid
+truth_tracking.layer34_midpoint = layer34_mid
```
tvami: Why is this part not in the CI script? 
Also I have similar comments of having this info in the CI and not having yet another config

Tracking/include/Tracking/Event/Measurement.h
```diff
@@ -92,6 +92,8 @@ class Measurement {
   /// @return The hit time in ns.
   [[nodiscard]] float getTime() const { return t_; };
 
+  [[nodiscard]] float getEdep() const { return edep_; };
+
```
tvami: where is this used in the new code?

Tracking/include/Tracking/Event/StraightTrack.h
```diff
@@ -0,0 +1,187 @@
...
+  int trackID_{-1};
+  // Truth probability
+  double truthProb_{0.};
+  // pdgID (truth value)
+  int pdgID_{0};
+
+  std::vector<double> trk_cov_;
+
+  /// Class declaration needed by the ROOT dictionary.
+  ClassDef(StraightTrack, 3);
```
tvami:
```suggestion
  ClassDef(StraightTrack, 1);
```

Tracking/include/Tracking/Event/StraightTrack.h
```diff
@@ -0,0 +1,187 @@
...
+  // Covariance vector has 10 elements arranged as a vector
+  // mxmx mxbx mxmy mxby
+  //      bxbx bxmy bxby
+  //           mymy myby
+  //                byby
+  // m = slope, b = intercept
+  void setCov(const std::vector<double>& cov) { trk_cov_ = cov; }
+  std::vector<double> getCov() const { return trk_cov_; }
+
+ protected:
```
tvami, line 155: please rename the variables to `snake_case`, or for the member variables `snake_case_`.
Generally here are the style rules: https://ldmx-software.github.io/developing/coding-rules.html

Tracking/include/Tracking/Reco/LinearSeedFinder.h
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * Constructor.
+   *
+   * @param name The name of the instance of this object.
+   * @param process The process running this producer.
+   */
+  LinearSeedFinder(const std::string &name, framework::Process &process);
+
+  /// Destructor
+  ~LinearSeedFinder();
```
tvami: See III.3 https://ldmx-software.github.io/developing/coding-rules.html#iii-code-quality

Tracking/include/Tracking/Reco/LinearSeedFinder.h
```diff
@@ -0,0 +1,144 @@
...
+
+  /**
+   *
+   */
+  void onProcessStart() override;
+
+  /**
+   *
+   */
+  void onProcessEnd() override;
```
tvami:
```suggestion
```

Tracking/include/Tracking/Reco/LinearSeedFinder.h
```diff
@@ -0,0 +1,144 @@
...
+   * indicate if a charge particle can be found by the recoil tracker.
+   *
+   * @param event The event to process.
+   */
+  void produce(framework::Event &event) override;
+
+ protected:
+  // Function to find seeds based on 2 Recoil points and 1 EcalRecHit
+  ldmx::StraightTrack SeedTracker(
+      const std::tuple<std::array<double, 3>, ldmx::Measurement,
```
tvami, line 64: can we make that "3" a variable? If in some lucky case we get to have another tracking layer?

Tracking/include/Tracking/Reco/LinearSeedFinder.h
```diff
@@ -0,0 +1,144 @@
...
+
+  // Calculate chi2 of the fit
+  double globalChiSquare(const std::array<double, 3> &firstSensor,
+                         const std::array<double, 3> &secondSensor,
+                         const std::array<double, 3> &ecalHit, double ax,
+                         double ay, double bx, double by);
+
+  // Function to find the number of unique layers hit (to determine if we have
+  // enough points to fit)
+  int uniqueLayersHit(const std::vector<ldmx::Measurement> &digiPoints);
```
tvami: I dont see this used anywhere?
```suggestion
```

Tracking/include/Tracking/Reco/LinearSeedFinder.h
```diff
@@ -0,0 +1,144 @@
...
+
+  /// The name of the output collection of seeds to be stored.
+  std::string out_seed_collection_{"LinearRecoilSeedTracks"};
+  /// The name of the input hits collection to use in finding seeds..
+  std::string input_hits_collection_{"DigiRecoilSimHits"};
+  /// The name of the tagger Tracks (only for Recoil Seeding)
+  std::string input_recHits_collection_{"EcalRecHits"};
+
+  double ecal_uncertainty_{3.87};
+  double ecal_distance_threshold_{
```
tvami:
```suggestion
// Max distance from RecHit for valid track
  double ecal_distance_threshold_{10.0};  
```

Tracking/include/Tracking/Reco/LinearSeedFinder.h
```diff
@@ -0,0 +1,144 @@
...
+      10.0};  // Max distance from RecHit for valid track
```
tvami:
```suggestion
```
if you just accepted the above suggestion this line needs to be removed

Tracking/python/eventDisplay.py
```diff
@@ -0,0 +1,265 @@
+from LDMX.Framework import EventTree
```
tvami: I wonder if this would be better under https://github.com/LDMX-Software/ldmx-sw/tree/trunk/EventDisplay

Tracking/python/tracking.py
```diff
@@ -44,6 +44,105 @@ def __init__(self, instance_name="DigitizationProcessor"):
         self.hit_collection = 'TaggerSimHits'
         self.out_collection = 'OutputMeasurements'
                 
+class LinearSeedFinder(Producer):
```
tvami: all these might be better in its own file `Tracking/python/reducedTracking.py`

Tracking/src/Tracking/Event/StraightTrack.cxx
```diff
@@ -0,0 +1,11 @@
+#include "Tracking/Event/StraightTrack.h"
+
+#include <iostream>
+
+ClassImp(ldmx::StraightTrack)
+
+    namespace ldmx {
+  void StraightTrack::Print() const {
+    std::cout << "print straight track" << std::endl;
```
tvami: Hmm but you are not printing it out (yet)

Tracking/src/Tracking/Reco/LinearTruthTracking.cxx
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  trk.setAllSensorPoints(points);
+
+  trk.setChi2(globalChiSquare(points, mx, my, bx, by));
+  trk.setCov(truthTrk_cov);
+
+  if (ecalPoints.size() > 0) {
+    double ecal_firstLayer_z =
+        ecalPoints[0][0];  // Z position from the first point in ecalPoints
```
tvami: put the comment above here too

tvami (changes requested):
This is an amazing work @fdelzanno thanks! I have a few requests, nothing too big, but I didnt always point to it, so if the same thing is true in a different file, please change it there too

Tracking/include/Tracking/Reco/LinearSeedFinder.h
fdelzanno: The function is used in line 63 of the LinearSeedFinder.cxx. It's just a check to make sure we actually have enough distinct layers to look for a track.

Detectors/test/reducedLDMX_eGun_config.py
fdelzanno: I agree. I removed the file from the branch.

Tracking/exampleConfigs/test_reduced_config.py
fdelzanno: How sad!

Tracking/python/tracking.py
fdelzanno: I created a reducedTracking.py to fit these declarations, and updated the config to reflect the change.

Tracking/include/Tracking/Event/Measurement.h
fdelzanno: Good catch. I'm not sure where this came from, but I removed it because it shouldn't be there.

Tracking/src/Tracking/Event/StraightTrack.cxx
fdelzanno: I added this file because I was following the logic of the track class, which has a Track.cxx in which the only line is to print the track (although I've never used it). If this file is unnecessary I can remove it from the branch (?).

Tracking/python/eventDisplay.py
fdelzanno: I think it would make sense to put it in the EventDisplay, but in the README for the EventDisplay directory, it says that the directory was abandoned. I'm not sure it makes sense to store usable code in an abandoned directory (where people might not think to look for/use it)

Tracking/include/Tracking/Reco/LinearSeedFinder.h
fdelzanno: I think I resolved this according to the coding rule. Please check again and let me know if I misinterpreted it.

Tracking/include/Tracking/Event/StraightTrack.h
fdelzanno: I went through and renamed all the variables in each file (except the eventDisplay.py code) according to the style rules. Hopefully I didn't miss anything...

Tracking/include/Tracking/Reco/LinearSeedFinder.h
fdelzanno: The "3" in this case doesn't relate to the number of layers we have to fit. It just signifies that the array contains three position coordinates z, x, y.

I structured the function to build a recoil tuple that contains: an array<double, 3> which encodes the sensor position, an ldmx::Measurement that contains the actual ldmx reference to the first layer that was used, and an std::optional<ldmx::Measurement>, which is either the second layer that was used or a nullptr if there was no other layer with which to do the weighted averaging. The sensor position is passed directly to the fitting function along with the other recoil sensor position (two total) and the EcalRecHit position, while the two ldmx::Measurements are used to retrieve truth information from the Sim level.

If we were to have another tracking layer, all we would need to do is the add a third recoil tuple as a parameter and switch the third element in the fitting function from the Ecal position to the third recoil sensor position (as well as make some small adjustments to the uncertainties that are assigned and probably edit the StraightTracks class to save a third sensor position). I can add those references now but they would be unused in the current implementation.

tvami: It's ok let's keep this as is

Tracking/include/Tracking/Reco/LinearSeedFinder.h
tvami: Ahh I see it, github did not expand that file in default

Tracking/src/Tracking/Reco/LinearTrackFinder.cxx
```diff
@@ -0,0 +1,163 @@
...
+namespace tracking {
+namespace reco {
+
+LinearTrackFinder::LinearTrackFinder(const std::string& name,
+                                     framework::Process& process)
+    : TrackingGeometryUser(name, process) {}
+
+void LinearTrackFinder::onProcessStart() {}
+
+void LinearTrackFinder::configure(framework::config::Parameters& parameters) {
```
tvami, line 19: You should have the pass name in here too

Tracking/src/Tracking/Reco/LinearTrackFinder.cxx
```diff
@@ -0,0 +1,163 @@
...
+#include <fstream>
+
+namespace tracking {
+namespace reco {
+
+LinearTrackFinder::LinearTrackFinder(const std::string& name,
+                                     framework::Process& process)
+    : TrackingGeometryUser(name, process) {}
+
+void LinearTrackFinder::onProcessStart() {}
```
tvami: I dont think this is used 
```suggestion
```

Tracking/src/Tracking/Reco/LinearTruthTracking.cxx
```diff
@@ -0,0 +1,307 @@
...
+
+LinearTruthTracking::LinearTruthTracking(const std::string& name,
+                                         framework::Process& process)
+    : TrackingGeometryUser(name, process) {}
+
+void LinearTruthTracking::onProcessStart() {
+  truth_matching_tool_ = std::make_shared<tracking::sim::TruthMatchingTool>();
+}
+
+void LinearTruthTracking::configure(framework::config::Parameters& parameters) {
```
tvami, line 12: Pass name here too

Tracking/src/Tracking/Reco/LinearTruthTracking.cxx
```diff
@@ -0,0 +1,307 @@
...
+
+  const std::vector<ldmx::Measurement> recoil_hits =
+      event.getCollection<ldmx::Measurement>(input_hits_collection_);
+  const std::vector<ldmx::EcalHit> ecal_rec_hit =
+      event.getCollection<ldmx::EcalHit>(input_rec_hits_collection_);
+
+  std::vector<std::array<double, 3>> first_layer_ecal_rec_hits;
+
+  for (const auto& x_ecal : ecal_rec_hit) {
+    if (x_ecal.getZPos() < 250) {
```
tvami: that 250 should be not hardcoded

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
```diff
@@ -0,0 +1,386 @@
+#include "Tracking/dqm/StraightTracksDQM.h"
+
+#include <algorithm>
+#include <iostream>
+
+#include "Tracking/Sim/TrackingUtils.h"
+
+namespace tracking::dqm {
+
+void StraightTracksDQM::configure(framework::config::Parameters& parameters) {
```
tvami, line 10: pass name here too

tvami (changes requested):

Tracking/python/eventDisplay.py
tvami: It's true that the c++ in there is abandoned, but you can edit the README to point out that this file is working. People would also not look for event display under the Tracking :D

Tracking/src/Tracking/Event/StraightTrack.cxx
tvami: Well at one point it should be cleaned in the Track.cxx too, but let's just not add it first

Tracking/python/eventDisplay.py
fdelzanno: Haha fair enough. I will move the file to event display and update the README.

Tracking/src/Tracking/Event/StraightTrack.cxx
fdelzanno: Ok, I removed the file from the branch.

Tracking/src/Tracking/Reco/LinearTruthTracking.cxx
fdelzanno: Added a ecal_firstLayerZ_threshold variable and set it's value to 250, so that we can easily change the condition if we move the Ecal.

Tracking/src/Tracking/Reco/LinearTrackFinder.cxx
fdelzanno: You're right. I removed the function.

Tracking/src/Tracking/Reco/LinearTruthTracking.cxx
tvami: that's good, except that the it should be `ecal_first_layer_z_threshold` according to the `snake_case` haha

Detectors/test/reducedLDMX_eGun_config.py
tvami: I dont think this went thru, did you do `git rm Detectors/test/reducedLDMX_eGun_config.py` ?

Tracking/include/Tracking/Reco/LinearTrackFinder.h
```diff
@@ -0,0 +1,84 @@
...
+  /// Destructor
+  virtual ~LinearTrackFinder() = default;
+  /**
+   *
+   */
+  void onProcessStart() override;
+  /**
+   *
+   */
+  void onProcessEnd() override;
```
tvami:
```suggestion
```

Tracking/src/Tracking/Reco/LinearTruthTracking.cxx
fdelzanno: True, fixed!

Tracking/src/Tracking/Reco/LinearSeedFinder.cxx
```diff
@@ -0,0 +1,414 @@
...
+    if (x_ecal.getZPos() < ecal_first_layer_z_threshold_) {
+      first_layer_ecal_rec_hits.push_back(
+          {x_ecal.getZPos(), x_ecal.getXPos(), x_ecal.getYPos()});
+    }  // if first layer of Ecal
+  }    // for positions in ecalRecHit
+
+  // Check if we would fit empty seeds, if so: end tracking
+  if (recoil_hits.size() < 2 || first_layer_ecal_rec_hits.empty() ||
+      uniqueLayersHit(recoil_hits) < 2) {
+    n_missing_++;
```
tvami: I ran into parenthesis issues in the past for sure this is safer
```suggestion
  if ((recoil_hits.size() < 2) || (first_layer_ecal_rec_hits.empty()) ||
      (uniqueLayersHit(recoil_hits) < 2)) {
    n_missing_++;
```

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
```diff
@@ -0,0 +1,388 @@
...
+      "truth_collection", "LinearRecoilTruthTracks");
+  title_ = parameters.getParameter<std::string>("title", "recoil_lin_trk_");
+  track_prob_cut_ = parameters.getParameter<double>("trackProb_cut", 0.5);
+  subdetector_ = parameters.getParameter<std::string>("subdetector", "Recoil");
+  measurement_collection_ = parameters.getParameter<std::string>(
+      "measurement_collection", "DigiRecoilSimHits");
+  input_pass_name_ =
+      parameters.getParameter<std::string>("input_pass_name", "");
+
+  ldmx_log(info) << "Track Collection " << track_collection_ << std::endl;
```
tvami: You dont need the `<< std::endl` with the `ldmx_log`. I think you removed it in every other file, except for this one. Please remove it here (in all 15 occurrences)

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
```diff
@@ -0,0 +1,388 @@
...
+      else {
+        duplicate_tracks.push_back(sorted_tracks[i]);
+      }
+    }  // else (a real track)
+  }    // loop on sorted tracks
+
+  // The total number of elements in the uniqueTracks and duplicateTracks
+  // vectors should be equal to the number of elements in the original tracks
+  // vector
+  if (unique_tracks.size() + duplicate_tracks.size() + fake_tracks.size() !=
```
tvami:
```suggestion
  if ((unique_tracks.size() + duplicate_tracks.size() + fake_tracks.size()) !=
```

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
```diff
@@ -0,0 +1,388 @@
...
+      tracks.size()) {
+    std::cerr << "Error: unique and duplicate tracks vectors do not add up to "
```
tvami:
```suggestion
    ldmx_log(error) << "Unique and duplicate tracks vectors do not add up to "
```

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
```diff
@@ -0,0 +1,388 @@
...
+    std::cout << "Duplicate tracks:" << std::endl;
+    for (const ldmx::StraightTrack& track : duplicate_tracks) {
+      std::cout << "Track ID: " << track.getTrackID()
+                << ", Truth Prob: " << track.getTruthProb() << std::endl;
+    }
+    std::cout << "Fake tracks:" << std::endl;
+    for (const ldmx::StraightTrack& track : fake_tracks) {
+      std::cout << "Track ID: " << track.getTrackID()
+                << ", Truth Prob: " << track.getTruthProb() << std::endl;
+    }
```
tvami: ok this is super new, I didnt even get to add it in the documentation, but we introduced `ldmx_log(trace)` for things that are more verbose than `ldmx_log(debug)`, so I suggest to remove the `debug_` boolean and move all this stuff to `ldmx_log(trace)` [and I'll update the documentation about this now]

tvami (changes requested):
I got to the DQM file only now. 
But with that I'm done, thanks for following up on a multi-stage review from my side

Tracking/src/Tracking/Reco/LinearSeedFinder.cxx
fdelzanno: Fixed in Commit 675f482.

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
fdelzanno: Fixed in Commit 675f482.

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
fdelzanno: Fixed in Commit 675f482.

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
fdelzanno: Fixed in Commit 675f482.

Tracking/src/Tracking/dqm/StraightTracksDQM.cxx
fdelzanno: Ok, basically I removed the debug_ boolean from both the .cxx and .h dqm, and turned these statements into ldmx_log(trace). Otherwise the style and content of the statements is identical.

Addressed in Commit 675f482.

tvami (approved):

tvami:
Tests are running in https://github.com/LDMX-Software/ldmx-sw/actions/runs/13546924629/job/37860769076?pr=1635

Tracking/python/reducedTracking.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+
+    def __init__(self, instance_name="LinearSeedFinder"):
+        super().__init__(instance_name, 'tracking::reco::LinearSeedFinder', 'Tracking')
+        self.input_hits_collection = 'DigiRecoilSimHits'
+        self.input_rec_hits_collection = 'EcalRecHits'
+        self.out_seed_collection = 'LinearRecoilSeedTracks'
+        self.recoil_uncertainty = [0.006, 0.12]
+        self.ecal_uncertainty = 3.87
+        self.ecal_first_layer_z_threshold = 250
```
tvami:
```suggestion
        self.ecal_first_layer_z_threshold = 250.0
```

Tracking/python/reducedTracking.py
```diff
@@ -0,0 +1,104 @@
...
+        The z coordinate of the midpoint between Recoil layers 3 and 4
+    """
+
+    def __init__(self, instance_name="LinearTruthTracking"):
+        super().__init__(instance_name, 'tracking::reco::LinearTruthTracking', 'Tracking')
+        self.input_hits_collection = 'DigiRecoilSimHits'
+        self.input_rec_hits_collection = 'EcalRecHits'
+        self.out_track_collection = 'LinearRecoilTruthTracks'
+        self.recoil_truth_uncertainty = [0.006, 0.12]
+        self.ecal_first_layer_z_threshold = 250
```
tvami, line 92:
```suggestion
        self.ecal_first_layer_z_threshold = 250.0
```

tvami (changes requested):
Move ecal_first_layer_z_threshold to be a double

tvami:
@fdelzanno I think the `Local to global` printout should go into `ldmx_log(trace)` since it adds 7 lines per hit. Otherwise it seem to ran fine

fdelzanno:
> @fdelzanno I think the `Local to global` printout should go into `ldmx_log(trace)` since it adds 7 lines per hit. Otherwise it seem to ran fine

Ah yes, that actually could be a relic from the days (over the summer) in which we were debugging the ACTS transformations. I can covert it to a `ldmx_log(trace)`, but it looks like the info it outputs may not be super relevant -- so I could also just remove those output lines in the code.

tvami:
> @fdelzanno I think the `Local to global` printout should go into `ldmx_log(trace)` since it adds 7 lines per hit. Otherwise it seem to ran fine

Nvm, that's not in this PR (it's the digitizer), so that's something to change in a next PR

tvami (approved):

fdelzanno:
The first big change is to the LinearTruthTracking code, which has been modified to better represent a "truth track" in the rLDMX. The highlights are:
- I changed the hit collection to RecoilSimHits. Initially, I used DigiRecoilSimHits as the truth hits, but these hits are already "smeared" with respect to the actual RecoilSimHits (so they are already slightly different from the true sim-level hits).
-  I removed uncertainty from the line-fitting algorithm. The sim hits provide an exact position, so it didn't make sense to assign the realistic sensor uncertainty to them. The fitting is now a basic least-squares regression on sim hits with trackID == 1, and the chi2 of the track is calculated without uncertainty.
- The new fitting algorithm also handles cases where we only have 2 or 3 sim hits with trackID == 1. Basically, as long as we have enough points to construct a line (>= 2) from the recoil electron hits, we will output a truth track. 
- Rather than constructing combinations of hits on the recoil layers and looking for the combination with trackID == 1 and truthProb == 1 (i.e. all 4 layer hits are trackID == 1), the new code just looks directly in the collection for the points with trackID == 1. This speeds up computation and removes wasteful code.

bloodyyugo (approved):

fdelzanno:
The second change affects how the LinearSeedFinder creates the sensor points on which to do the track fitting. 

In the ideal case, these sensor points are created by taking the first axial/stereo pair (layers 1 and 2 of the Recoil) and combining their local measurements to produce a 3D space point. We would then do the same with the second axial/stereo pair (layers 3 and 4 of the Recoil) to produce another 3D space point. We would fit a straight line to these two points, combined with an EcalRecHit for an extra degree of freedom. Since the code to combine the axial/stereo pairs is a work in progress, the temporary solution is just to take advantage of the ability of the strip sensors at each layer to accurately measure the x-coordinate:
- We still average the z-position and x-position taken from the DigiRecoilSimHits collection to find the midpoint between the axial and stereo pairs, but we now manually set the y_position to 0 (since it is not actually measured).
- We still assign an uncertainty of sigma_x = 0.006 mm (6 microns) to the x_position, but now assign a large uncertainty of sigma_y = 20/sqrt(12), where 20 is the length of the strip above/below 0.

tvami:
Let's run the tests and then we can merge this

fdelzanno:
I re-ran a 5,000 event sample using a 4GeV eGun in the reduced-ldmx-v1 geometry to show the changes in tracking from the initial comment at the beginning of the PR. Here are the statistics:

There are 5000 events in the sample.
We find 5076 tracks / 4708 events with tracks + 292 empty events in which we don't find a track = 5076 tracks / 5000 events.
We find 4862 truth tracks / 4862 events with truth tracks + 138 empty events in which we don't fit a truth track = 4862 tracks / 5000 events

Out of the tracks in the sample, 4785 tracks have trackID = 1 + 291 tracks have a different trackID number = 5076 tracks

Out of the tracks with trackID = 1 in the sample:
- there are 4544 that have truthProb = 1 (i.e. all 4 hits used in tracking correspond to the recoil electron)
- 152 tracks that have truthProb = 0.75 (i.e. 3 hits / 4 used in tracking correspond to recoil electron)
- 88 tracks that have truthProb = 0.5 (i.e. 2 hits / 4 used in tracking correspond to recoil electron)
- 292 tracks with another truthProb value

We reconstruct 4951 electron tracks + 124 positron tracks + 1 pion track = 5076 tracks.

Here are the summary plots. In blue are all the reconstructed tracks. In green are the truth tracks. In red is the data only from reconstructed tracks with trackID = 1 (i.e. reconstructed recoil electron tracks).

eGun_rLDMX_numTrack_COMPARISON_v1

eGun_rLDMX_chi2_COMPARISON_v1

The chi2 is even lower now, which makes sense since we are fitting well in x, while the y coordinate has a large uncertainty (making it contribute less to the chi2).

eGun_rLDMX_dRecHit_COMPARISON_v1

This is the distance between the track extrapolated to the first Ecal layer and the RecHit used to create the track. The threshold value is 15 mm, so there is no reconstructed track with dRecHit > 15 mm. The radius of an Ecal cell is approximately 4 mm, meaning any track with dRecHit < 4 mm is likely to be in the correct cell. The distance from RecHit shrunk with the new tracking, probably because it is now our only reference value in y -- allowing the fitting to tune a track to fit this point well. 

I also have the pulls of the reconstructed track beam (theta) and azimuthal (phi) angle (these come from the dqm, and are for unique tracks):

pull_phi

eGun_rLDMX_phi_COMPARISON_v1

Pull phi and the values of phi themselves are reasonable.

pull_theta

eGun_rLDMX_theta_COMPARISON_v1

Without any y information we will seriously struggle to determine a meaningful theta.

For events in which we construct a truth track, I plotted the difference between the location of the trackID = 1 reconstructed track and the truth track at both the Target and Ecal face.

eGun_rLDMX_target_locX_diff_v1

eGun_rLDMX_target_locY_diff_v1

We cannot meaningfully fit in y, so we expect the spread in y at the target to be random (based on how well the fit is able to track the EcalRecHit).

eGun_rLDMX_ecal_locX_diff_v1

eGun_rLDMX_ecal_locY_diff_v1

Again, I believe the y position difference is behaving like we expect, since the tracking is only doing anything meaningful in y.

Overall, I think that the updated tracking is working properly and conforming to our expectations for tracks constructed without a real y measurement in the Recoil layers. Here are two example event displays to show how our tracking reconstructs tracks:

rLDMX_v1_eventDisplay_eventID_10

rLDMX_v1_eventDisplay_eventID_1883

This second one clearly shows how we will struggle to calculate theta without a y measurement.

jmmans:
Wandering in at a late point -- has it been considered to make use of the information from the TS?  It doesn't have _great_ resolution in Y, but it has some...

tvami:
> Wandering in at a late point -- has it been considered to make use of the information from the TS? It doesn't have _great_ resolution in Y, but it has some...

We tried this, unfortunately it just added "noise" to the fit

fdelzanno:
> > Wandering in at a late point -- has it been considered to make use of the information from the TS? It doesn't have _great_ resolution in Y, but it has some...
> 
> We tried this, unfortunately it just added "noise" to the fit

Here are some example EventDisplays from when we were trying this. 

rLDMX_v2_tsTrack_eventDisplay_eventID_4880

rLDMX_v2_tsTrack_eventDisplay_eventID_4841

At the time we did conclude that the TS was just adding noise, but our fitting was also generally better pre-TS since we had y information in the fit. Perhaps it could be worth doing a quick check to see how this new tracking performs with the TS reference point in y.

tvami:
> Perhaps it could be worth doing a quick check to see how this new tracking performs with the TS reference point in y.

Yeah I agree. But we can do it in a next PR as an improvement to what you have here.

tvami:
This already ran fine:
Screenshot 2025-03-07 at 08 57 00

So feel free to merge it @fdelzanno (when you squash make sure the final commit msg makes sense and doesnt have all the back and forth msgs from all the 56 commits)

PR #1633, opened 2025-02-23
Iss1376 geant4 object pointer for comparison

LiamBrennan-UCSB:
Fixing Issue #1286 and Issue #1376. Adds functionality to call a Geant4 pointer to compare to another pointer within Simcore, Biasing, etc such that a pointer to pointer comparison can be made rather than completing string to string as was done previously.

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments
- [x] I ran my developments and the following shows that they are successful.

PR #1640, opened 2025-02-28
patches to ldmx-sw following compiler update

tomeichlersmith:

This is following up on https://github.com/LDMX-Software/dev-build-context/pull/103 . I will maintain an checklist here of patches that I need to make before I feel comfortable merging.

## configure-quick
The `configure-quick` build does not enable as many warnings, but I still get the big ones so its a helpful place to start. After I patch these up, I can move to a configuration that enables more warnings (e.g. our standard `configure`).

- [x] `std::iterator` is deprecated, ~need to find replacement for `HgcrocDigiCollection`~ no replacement, just remove it and the callers deduce if the iterator class has the required functions
- [x] update configure Python for the specific python version included in Ubuntu 24.04
- [x] overflow potential in `recon::PFTruthProducer` when using `ldmx::SimParticle::getParents`
- [ ] the `HcalDigiPipelineTest` is now failing 812 / 1000 of the test energies. All of the test hits being reconstructed return a `daq_pe` of 0 rather than something close to the real value. The tests below 40PE are passing because 0 is within 40 of the true value. I don't think this is as apocalyptic as it seems because the PR validations all pass and those include distributions on non-zero PE hits in the HCal. I am guessing it has to do with how the test is creating the hits or how Catch2 is working with the new compiler.
- [x] we got a new Boost, so maybe https://github.com/LDMX-Software/ldmx-sw/blob/169ee858c76d1114e16397127ac77ce8d9a590ba/DetDescr/src/DetDescr/DetectorIDBindings.cxx#L2-L3 can be removed

## configure
The regular `configure` build enables additional warnings and clang tidy.

- [x] remove `std::move` in `Process::getDummy` - compiler can use copy elision to place the construction where it is first used
- [x] potential null pointer dereference in copy constructor of `RunHeader` which is inlined from `importRunHeaders`
- [x] dangling reference to temporary `channel` and `msg_level` in Logger
- [x] dangling pointer to unnamed temporary from using `{}` as default for `logRules`
- [x] some warnings from Catch2 internals, probably should upgrade Catch2 version we are using...
    - still there after locally updating to Catch2 3.8.0, its only in the NtupleManagerTest and comes from calling some `match` code `NtuplemanagerTest.cxx:124-132`
    - originates from using `Catch::Matchers::UnorderedEquals`, not sure if I'm misusing it though
- [x] `nPos` set but not used in `EventReadoutProducer.cxx:64`
- [x] taking absolute value of a bool in `TrigScintTrackProducer.cxx:333`
- [x] the address of `overlayEvent_` will never be NULL
- [x] potential null pointer dereference from G4PrimaryVertex.hh
- [x] potential null pointer dereference in `Kaonphysics::DumpDecayDetails` (lots of them)
- [x] `particleIndex` set but not used in `LHEReader` and `LHEPrimaryGenerator`
- [x] memory use after freed inside Geant4 in EcalSD if certain conditions are met
- [ ] potential null pointer dereference using `G4String` copy constructor
- [ ] potential null pointer dereference calling `GetLogicalVolume` in a lot of the user actions
    - I'm touching a lot of the same places that #1633 is touching so I'm not going to address this warning right now.
- [x] `TaggerHitFilter` same memory used after free as `EcalSD`
- [x] potential null pointer derefernece `Recon/DBScanClusterBuilder.cxx:132`, would be null dereference if size is zero


tvami:
Can you add https://github.com/LDMX-Software/ldmx-sw/blob/169ee858c76d1114e16397127ac77ce8d9a590ba/DetDescr/src/DetDescr/DetectorIDBindings.cxx#L2
to the TODO list?

tomeichlersmith:
I can't figure out the Hcal testing right now (I was hoping the additional warnings would point me towards something but they did not) and I want to wait on the nullptr dereferencing related to other developments, so I am marking this as Ready for Review.

The changes are all "small" in the sense that they are not expected to affect the physics. The previous PR Validation run did not show any differences and I am expecting the same again.

Framework/include/Framework/Configure/Python.h
```diff
@@ -0,0 +1,44 @@
...
+ * can get up to a whole lot of shenanigans that can help them
+ * make their work more efficient.
+ *
+ * @param[in] full pythonic path to the object to kickoff extraction
+ * @param[in] pythonScript Filename location of the python script.
+ * @param[in] args Commandline arguments to be passed to the python script.
+ * @param[in] nargs Number of commandline arguments, assumed to be >= 0
+ */
+Parameters run(const std::string& root_object, const std::string& pythonScript,
+               char* args[], int nargs);
```
tomeichlersmith, line 40: I'm sneaking in a refactor of the Configure/Python ecosystem that does a few things:

1. Removes the extraneous `ConfigurePython` class in favor of a function that runs the python script and returns a `Parameters` of the configuration extracted.
2. Updates `Parameters` to have shorter `add` and `get` methods.
3. Promotes the "root object" that is used to extract the configuration to the argument so that _hypothetically_ other programs could be written that use the same configuration system but with a different python object (e.g. https://github.com/LDMX-Software/ldmx-tb-online/tree/main/reformat could be brought in as a submodule with its own executable if we desired).

TrigScint/src/TrigScint/TrigScintTrackProducer.cxx
```diff
@@ -330,8 +330,8 @@ void TrigScintTrackProducer::produce(framework::Event &event) {
 
         // no need to start pulling constituents from tracks that are
         // ridiculously far apart
-        if (fabs(track.getCentroid() - nextTrack.getCentroid() <
-                 3 * maxDelta_)) {
+        if (fabs(track.getCentroid() - nextTrack.getCentroid()) <
+            3 * maxDelta_) {
```
tomeichlersmith, line 334: @bryngemark I think this is just a misplaced closing parentheses since the newer compiler warned me that `fabs(bool)` doesn't make any sense.

Recon/src/Recon/DBScanClusterBuilder.cxx
```diff
@@ -129,7 +129,7 @@ void DBScanClusterBuilder::fillClusterInfoFromHits(
     // skip fits for 'vertical' clusters
     std::vector<float> sortedZ = raw_zvals;
     std::sort(sortedZ.begin(), sortedZ.end());
-    if (sortedZ[sortedZ.size() - 1] - sortedZ[0] > 1e3) {
+    if (sortedZ.size() > 2 and sortedZ[sortedZ.size() - 1] - sortedZ[0] > 1e3) {
```
tvami:
```suggestion
    if ((sortedZ.size() > 2) && (sortedZ.back() - sortedZ.front() > 1e3)) {
```

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -205,12 +205,6 @@ void OverlayProducer::produce(framework::Event &event) {
       }
     }
 
-    // get event wherever nextEvent()  left us
-    if (!&overlayEvent_) {
```
tvami, line 209: why did you delete this?

tvami (commented):
Looks really good, I have a question and a small suggestion

Recon/src/Recon/OverlayProducer.cxx
tomeichlersmith: It is a useless check. `overlayEvent_` is a class member variable that is not a pointer and so will never be null. We already check for a successful event when calling `nextEvent` below.

tvami (approved):

TrigScint/src/TrigScint/TrigScintTrackProducer.cxx
bryngemark: i agree with this change

bryngemark (approved):
Looks good to me. Not sure what to think about the HCal digi errors, ask an HCal code expert if you want a real opinion on that

PR #1643, opened 2025-03-01
Introduce TrackerVetoProcessor

### LINKED ISSUE #1642: Include a tracking veto

Similarly as we have ecal and hcal vetos, we should have a tracking veto as well

Introduce a tracking veto, similar to the rest

**Additional context**

I'd like to make a skim of signal events that fail the tracking but pass the sim level acceptance filter `RecoilFiducialityProcessor`

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1642

## Check List

I added it to the CI, and it runs fine

### DIFF
--- a/Tracking/include/Tracking/TrackerVetoProcessor.h
+++ b/Tracking/include/Tracking/TrackerVetoProcessor.h
@@ -1,0 +1,98 @@
+/**
+ * @file TrackerVetoProcessor.h
+ * @brief Class that flags events if they pass the tracker veto
+ * @author Tamas Almos Vami, UCSB
+ */
+
+#ifndef TRACKING_TRACKERVETOPROCESSOR_H_
+#define TRACKING_TRACKERVETOPROCESSOR_H_
+
+#include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
+#include "Tracking/Event/Track.h"
+#include "Tracking/Event/TrackerVetoResult.h"
+
+namespace tracking {
+
+/**
+ * @class TrackerVetoProcessor
+ * @brief Flags events that pass the tracker veto by applying specific selection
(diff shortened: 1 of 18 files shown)


tvami:
Everything is passing tests as expected

EBerzin:
I'm not sure if we want to add this here, but would it make sense to include any track quality selections? Like a cut on chi2 or hits on track?

tvami:
Yeah I think the chi2 should be added indeed! For the number of hits, that actually seems like a limiting factor already, so I'd prob not add that.

Tracking/include/Tracking/TrackerVetoProcessor.h
bloodyyugo, line 1: Do we want this file to live in include/Tracking or make a include/Tracking/Veto or put it in existing directory (Reco probably)

tvami: I did think about this, but then I was like it's not really Reco as it doesnt do any new reconstruction. Then I look at the structures in ECAL and HCAL and the just live up in the tree like I ended up doing here. I could make a `include/Tracking/Veto` if you prefer that, please let me know!

bloodyyugo: If that's how it's done in other systems, it's fine.

bloodyyugo (approved):

tomeichlersmith (approved):

tvami:
> Yeah I think the chi2 should be added indeed! For the number of hits, that actually seems like a limiting factor already, so I'd prob not add that.

@EBerzin 
I added both after all, and kept them pretty loose. They are configurable, so we can play with them later if we want to.

EBerzin (approved):

PR #1649, opened 2025-03-11
only fetch acts and compile it if not found in image

tomeichlersmith:

This does not have a corresponding issue here. I am following up with https://github.com/LDMX-Software/dev-build-context/pull/115 while preparing for a v5 release of the ldmx/dev image.

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments. I test compiled with the newer image that has Acts locally. The CI will check that these developments work with the v4.2.2 image that does not have Acts.

The CI checks that an image without Acts can still compile and run.
I can show that this works with the newer image candidate that has Acts built in.
```
tom@appa:~/code/ldmx/ldmx-sw$ just use ldmx/dev:5.0.0-rc1
denv config image ldmx/dev:5.0.0-rc1
tom@appa:~/code/ldmx/ldmx-sw$ just configure
git fetch --tags && git describe --tags | cut -f 1 -d '-' > VERSION
denv cmake -B build -S . -DADDITIONAL_WARNINGS=ON -DENABLE_CLANG_TIDY=ON 
[ INFO ]: Building all modules.
[ INFO ]: Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.83.0/BoostConfig.cmake (found version "1.83.0") found components: log 
[ INFO ]: Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.83.0/BoostConfig.cmake (found version "1.83.0") found components: system log filesystem thread chrono atomic regex 
[ INFO ]: Deduced git SHA: 06e0b4dcde035ef689ab83b104b0f5f89d212c49
[ INFO ]: Building ROOT dictionary LinkDef
Looking for libboost_python
[ INFO ]: Building the modules necessary for the reconstruction.
[ INFO ]: Found ONNX Runtime version 
[ INFO ]: Found Acts 36.0.0
[ INFO ]: Found Geant4 version 10.2.3
[ INFO ]: Building the modules necessary for the simulation.
[ INFO ]: Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.83.0/BoostConfig.cmake (found version "1.83.0") found components: log 
[ INFO ]: Found Boost: /usr/lib/x86_64-linux-gnu/cmake/Boost-1.83.0/BoostConfig.cmake (found suitable version "1.83.0", minimum required is "1.68") found components: iostreams 
[ INFO ]: Installing all field map files.
-- Configuring done (0.7s)
-- Generating done (0.5s)
-- Build files have been written to: /home/tom/code/ldmx/ldmx-sw/build
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "docs/doxygen.conf/doxygen-awesome-css"]
 	path = docs/doxygen.conf/doxygen-awesome-css
 	url = https://github.com/jothepro/doxygen-awesome-css
-[submodule "Tracking/acts"]
-	path = Tracking/acts
-	url = https://github.com/acts-project/acts
 [submodule "Trigger/ruckus"]
 	path = Trigger/ruckus
 	url = https://github.com/slaclab/ruckus
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -33,10 +33,9 @@ list(APPEND CMAKE_MODULE_PATH ${LDMX_SW_SOURCE_DIR}/cmake/)
 # is set to NOTFOUND.
 include(BuildMacros RESULT_VARIABLE build_macros_found)
 
-# Adding ROOT, ACTS, HLS arb prec as "SYSTEM" 
+# Adding ROOT as "SYSTEM" 
 # which will silence compiler warnings from these 3rd party softwares
 include_directories(SYSTEM /usr/local/include/root)
-include_directories(SYSTEM Tracking/acts/Core/include)
 
 # FunctionalCoreTest.cxx doesnt comply with clang-tidy but that's ok
 set_source_files_properties(Framework/test/FunctionalCoreTest.cxx PROPERTIES COMPILE_OPTIONS "-Wno-null-dereference")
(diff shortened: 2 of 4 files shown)


tvami:
Isn't the solution to move those virtual destructors to be non virtual (even tho they are supposed to be virtual but given that ACTS didnt do them virtual...)?

tomeichlersmith:
That would be the solution, but that would involve editing Acts and I do not want to do that. There are also other warnings (turned errors) from Acts that are seen by the compiler that we are ignoring with the same `SYSTEM` strategy. I'm just localizing the `SYSTEM` include to be more specific.

https://github.com/LDMX-Software/ldmx-sw/blob/79fcb61262f9fc8e81ee58e03acaefd6fcc80d80/CMakeLists.txt#L39

tvami:
Yeah I agree that editing ACTS is beyond what we should do, but the earlier warning I saw in your original commit were about the inherited class which we have control over. It's true that I didnt look at the full list, so maybe there is some stuff that we have no control over... in that case going back to the SYSTEM approach is fine

tvami (approved):

PR #1650, opened 2025-03-12
Add missing pass names, part 1

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
This is part 1 of https://github.com/LDMX-Software/ldmx-sw/issues/1645
(The part I need immediatly, I'll let the rest be dealt with by @LiamBrennan-UCSB )

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [x] I ran my developments and the following shows that they are successful.

PR #1629, opened 2025-02-21
Move ECAL veto to fully rely on tracking when possible

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1549

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments
- [ ] I ran my developments and the following shows that they are successful.

PR #1652, opened 2025-03-12
Introduce WAB Processor

### LINKED ISSUE #1499: Introduce WABProcessor
During early 2025 running, we are planning on conducting an analysis of WAB events that we reconstruct with the reduced detector.

To create the pertinent distributions for WABs, I am introducing a Processor specifically for WABs which will automatically create the required kinematic distributions during the reconstruction step.

We could create the distributions after the fact using a python file that loops over the ECalRecHits. However, we would like to run on the inclusive samples and we think it'd be easier/faster to incorporate this into ldmx-sw and redo the reco step of the inclusive sample with the reduced geometry and WABProcessor.

SanjitMasanam:

This resolves issue #1499


Ecal/include/Ecal/EcalWABRecProcessor.h
```diff
@@ -0,0 +1,76 @@
...
+#include "DetDescr/EcalID.h"
+#include "Ecal/Event/EcalHit.h"
+#include "Ecal/Event/EcalWABResult.h"
+#include "Eigen/Dense"
+#include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
+#include "Tools/ONNXRuntime.h"
+
+// ROOT (MIP tracking)
+#include "TVector3.h"
```
tvami: Is this really used?

Ecal/include/Ecal/EcalWABRecProcessor.h
```diff
@@ -0,0 +1,76 @@
...
+
+// LDMX
+#include "DetDescr/EcalGeometry.h"
+#include "DetDescr/EcalID.h"
+#include "Ecal/Event/EcalHit.h"
+#include "Ecal/Event/EcalWABResult.h"
+#include "Eigen/Dense"
+#include "Framework/Configure/Parameters.h"
+#include "Framework/EventProcessor.h"
+#include "Tools/ONNXRuntime.h"
```
tvami: I'm pretty certain this is not used here

Ecal/include/Ecal/EcalWABRecProcessor.h
```diff
@@ -0,0 +1,76 @@
...
+ public:
+  EcalWABRecProcessor(const std::string& name, framework::Process& process)
+      : Producer(name, process) {}
+
+  virtual ~EcalWABRecProcessor() = default;
+
+  void onProcessStart() override;
+
+  void onProcessEnd() override;
+
```
tvami: you can delete these, since they dont seem to do anything extra

Ecal/include/Ecal/EcalWABRecProcessor.h
```diff
@@ -0,0 +1,76 @@
...
+  std::pair<Eigen::VectorXd, Eigen::VectorXd> polyfitXYvsZ(
+      const std::vector<double>& x, const std::vector<double>& y,
+      const std::vector<double>& z, int degree);
+
+  /** Name of the collection which will contain the results. */
+  std::string collection_name_{"EcalWABRec"};
+
+  /// handle to current geometry (to share with member functions)
+  const ldmx::EcalGeometry* geometry_;
```
tvami: Does this really have to be a global variable? I only see it used under the `analyze()`

Ecal/include/Ecal/Event/EcalWABResult.h
```diff
@@ -0,0 +1,225 @@
...
+  double recThetaDiffElectronPhoton_;
+  double recPhiDiffElectronPhoton_;
+  double trueRecThetaDiffElectron_;
+  double trueRecPhiDiffElectron_;
+  double trueRecThetaDiffPhoton_;
+  double trueRecPhiDiffPhoton_;
+  float trueElectronShowerEnergy_;
+  float truePhotonShowerEnergy_;
+  double recElectronShowerEnergy_;
+  double recPhotonShowerEnergy_;
```
tvami: I think all these should be `float`s, I'm sure we dont need the `double` precision, right? And then we can save some space. This will mean downstream changes that I'm not pointing out rn.

Also the rule I.3. 
https://ldmx-software.github.io/developing/coding-rules.html

Ecal/src/Ecal/EcalWABRecProcessor.cxx
```diff
@@ -0,0 +1,847 @@
...
+#include <iomanip>
+#include <numbers>  // For std::numbers::pi
+#include <numeric>
+#include <tuple>
+#include <vector>
+
+// ROOT (MIP tracking)
+#include "TDecompSVD.h"
+#include "TMatrixD.h"
+#include "TVector3.h"
```
tvami: The includes (that you actually use, which is for sure not all these) should be in the header file

Ecal/src/Ecal/EcalWABRecProcessor.cxx
```diff
@@ -0,0 +1,847 @@
...
+
+namespace ecal {
+
+// 68% Electron Radii of Containment for various theta ranges
+const std::vector<double> radius_68_theta_0_to_10 = {
```
tvami: These dont need to be `double`s too right?

Ecal/src/Ecal/EcalWABRecProcessor.cxx
```diff
@@ -0,0 +1,847 @@
...
+const std::vector<double> radius_68_theta_30_to_90 = {
+    22.50983127, 26.44537503, 58.24642887, 90.59076279, 130.0592014,
+    157.4611392, 184.2187293, 202.6994588, 225.3488816, 243.3454167,
+    269.2456428, 280.6119298, 303.8591523, 322.0522722, 335.1780181,
+    350.3398234, 353.7763544, 373.9942362, 382.9453608, 401.9703438,
+    441.6281859, 432.5241826, 455.2878243, 492.2888656, 502.6653722,
+    480.2334627, 566.5438302, 505.7032783, 556.1650321, 596.9112032,
+    616.1614593, 635.4117154, 654.6619715, 673.9122276};
+
+void EcalWABRecProcessor::onProcessStart() {}
```
tvami: del

Ecal/src/Ecal/EcalWABRecProcessor.cxx
```diff
@@ -0,0 +1,847 @@
...
+
+  // Get the collection of ecal_rec_hits and tracks
+  const std::vector<ldmx::EcalHit> ecal_rec_hits =
+      event.getCollection<ldmx::EcalHit>(rec_coll_name_, rec_pass_name_);
+  const std::vector<ldmx::StraightTrack> linear_tracks =
+      event.getCollection<ldmx::StraightTrack>(track_coll_name_,
+                                               track_pass_name_);
+
+  // Define variables to save recoil electron/photon information (SP)
+  std::vector<double> recoil_e_p, recoil_y_p;
```
tvami, line 91:
```suggestion
  std::vector<float> recoil_e_p, recoil_y_p;
```

Ecal/src/Ecal/EcalWABRecProcessor.cxx
```diff
@@ -0,0 +1,847 @@
...
+
+  if (event.exists("TargetScoringPlaneHits")) {
+    //
+    // Loop through all of the sim particles and find the recoil
+    // photon/electron.
+    //
+
+    // Find Target SP hit for recoil photon/electron
+    const std::vector<ldmx::SimTrackerHit> target_sp_hits =
+        event.getCollection<ldmx::SimTrackerHit>("TargetScoringPlaneHits");
```
tvami: Add pass names like in https://github.com/LDMX-Software/ldmx-sw/pull/1650/files

Ecal/src/Ecal/EcalWABRecProcessor.cxx
```diff
@@ -0,0 +1,847 @@
...
+          true_photon_shower_energy = sp_hit.getEnergy();
+          recoil_y_pos = sp_hit.getPosition();
+          photon_p_zmax = recoil_y_p[2];
+        }
+      }
+    }
+
+    // Calculating true theta values using SP hit parameters
+    if (recoil_y_p.size() == 3 &&
+        std::sqrt(std::pow(recoil_y_p[0], 2) + std::pow(recoil_y_p[1], 2) +
```
tvami: the `std:pow` is actually pretty costly, so instead of `std::pow(x, 2` it's more efficient to do x*x

Ecal/src/Ecal/EcalWABRecProcessor.cxx
tvami: Also since these are also 3D I'm about to move them to arrays in https://github.com/LDMX-Software/ldmx-sw/pull/1629/files

Ecal/src/Ecal/EcalWABRecProcessor.cxx
tvami: if you move `recoil_y_p` to be an array then the size check is not necessary anymore

tvami (changes requested):
Nice work Sanjit! 
I have a few comments but nothing major, mostly to make the code more efficient. I didnt comment everywhere but please move to `float`s when you can

Ecal/include/Ecal/EcalWABRecProcessor.h
SanjitMasanam: I have a line of code that runs `onProcessEnd()` but I deleted `void onProcessStart() override;`

Ecal/src/Ecal/EcalWABRecProcessor.cxx
SanjitMasanam:
```
recoil_e_p = sp_hit.getMomentum();
```
When I reference `recoil_e_p`, I set it equal to an object of type `std::vector<double>`. I could rewrite the code to be as shown below,
```
std::vector<double> momentum = sp_hit.getMomentum();
          std::array<float, 3> recoil_e_p = {
              static_cast<float>(momentum[0]),
              static_cast<float>(momentum[1]),
              static_cast<float>(momentum[2])
          };
```
Would you prefer this? It seems like a lot of code just to modify how the data is stored.

tvami: OK keep it as it is, we can fix it later

tvami (approved):
No further comments, thanks!

tomeichlersmith (approved):

Issue #1660, opened 2025-03-25
EcalWABRecProcessor does not find ReducedTracks


tvami:
**Describe the bug**
https://github.com/LDMX-Software/ldmx-sw/actions/runs/14065137112/job/39387124993#step:6:420
```
[ LinearRecoilTracksDQM ] 1 debug: Track Monitoring on duplicates and fakes
  [ fire ] 1 fatal: [ProductNotFound] : No product found for name 'ReducedTracks'
    at /home/runner/work/ldmx-sw/ldmx-sw/Framework/include/Framework/Event.h:293 in getObject
  [ fire ] 1 debug:     0 std::vector<ldmx::StraightTrack, std::allocator<ldmx::StraightTrack> > const& framework::Event::getObject<std::vector<ldmx::StraightTrack, std::allocator<ldmx::StraightTrack> > >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const + 7960 
      1 ecal::EcalWABRecProcessor::produce(framework::Event&) + 1618 
```

**Desired behavior**

I think this is just a renaming issue.

tvami:
@SanjitMasanam can you have a look?

I'm not sure how we missed this in the PR @tomeichlersmith

tvami:
@tomeichlersmith something is going wrong with the github checkout step I think, bc in the PR
https://github.com/LDMX-Software/ldmx-sw/actions/runs/13850377142/job/38757152245
this did pass

tomeichlersmith:
I don't think the GitHub checkout is doing anything wrong. 

I think its more likely that the GitHub file diff view tries to be "smart" and avoid showing "unrelated changes". GitHub checks out **the branch** and does not apply the changes listed in the "files changed" tab to trunk.

For example, I restored the `iss1499-introduce-wabProcessor` branch and checked it out locally.
```
git fetch
git switch iss1499-introduce-wabProcessor
```
The `git diff` is more interesting than you'd expect. With this branch merged as the latest commit on trunk, I'd naively expect there to be no differences at all.
```
tom@appa:~/code/ldmx/ldmx-sw$ git diff trunk --name-status
M       .github/workflows/asan_ubsan_build.yml
M       .github/workflows/basic_test.yml
M       .github/workflows/compile_clang_lto.yml
M       .github/workflows/docs.yml
[...]

PR #1662, opened 2025-03-25
only copy over golden histograms/logs

tomeichlersmith:
don't override config and don't bother doing this copy for PRs where GitHub handles the merging for us

related to discussion in #1660

### DIFF
--- a/.github/workflows/pr_validation.yml
+++ b/.github/workflows/pr_validation.yml
@@ -63,7 +63,11 @@ jobs:
       shell: bash
 
     - name: Git Gold Histograms from trunk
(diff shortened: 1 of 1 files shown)


tvami (approved):

tomeichlersmith:
The copy step is skipped when triggered by PRs.

image

And now to see if it works correctly when triggered by comment /run-validation

tomeichlersmith:
The comment-triggered workflow uses the workflow on `trunk` meaning that the un-patched version was run.

image

I'll merge this and then open a dummy PR which changes a `config.py` in order to make sure that it isn't overwritten.

PR #1668, opened 2025-03-27
Technical improvements to EcalVeto

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
This is some dev that was done in https://github.com/LDMX-Software/ldmx-sw/pull/1629/
but since it was technical, I factor it out now (especially to understand why there are some unexpected diffs in that PR).

In this PR 
 * I move the position and momentum vectors to be floats: we really dont need the double precision
 * I store 3-element "containers" in an array instead of a vector: this should be faster, and also these should never have more than 3 elements anyway
 * removed verbosity, moved all that stuff under `trace` logging
 * moved some of the pow(A,2) to be A*A [there is way more but I got bored]

I also added a commented out example for the ecal pn CI config just so if people want to switch this on quickly.

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [x] I ran my developments and the following shows that they are successful.

PR #1671, opened 2025-03-31
Prep for ldmx/dev v5 with some more compiler patches

### LINKED ISSUE #1656: prep for v5 release of ldmx/dev image
The v5 release of the ldmx/dev image comes with a new compiler version (amongst other dependency updates) which propagates to ldmx-sw as more/different warnings. These additional warnings need to be patched in order to maintain our strict building rules and keep the CI tests passing.

## Testing Set Up
```
just clean
just use ldmx/dev:5.0.0-rc2
mkdir build
unbuffer just configure |& tee build/configure.log
unbuffer just build 1 |& tee build/build.log
# unbuffer is from the `expect` package and is used to trick just and its child commands into using colors
```

[build.log](https://github.com/user-attachments/files/19540862/build.log)
[configure.log](https://github.com/user-attachments/files/19540863/configure.log)

(view with `less -R` otherwise you will see the ANSI color codes written into the files)

## To Do List
I'll keep editing this as I make progress with testing.

- [x] https://github.com/LDMX-Software/ldmx-sw/pull/1640
- [x] https://github.com/LDMX-Software/ldmx-sw/pull/1649
- [x] the HcalDigiPipelineTest is now failing 812 / 1000 of the test energies <details>
All of the test hits being reconstructed return a daq_pe of 0 rather than something close to the real value. The tests below 40PE are passing because 0 is within 40 of the true value. I don't think this is as apocalyptic as it seems because the PR validations all pass and those include distributions on non-zero PE hits in the HCal. I am guessing it has to do with how the test is creating the hits or how Catch2 is working with the new compiler.
I am now unable to replicate this with the centrally-built image and a from-scratch build. Going to assume I did not fully clean the build or something when originally seeing this issue.
</details>

- [x] use after free warning in EcalSD, TaggerHitFilter <details>
This comes from the `touchableHandle` being a reference-counted object that is de-allocated after being accessed. I think we can avoid this issue by constructing a _reference_ rather than a copy.
</details>

- [x] potential null pointer dereference using G4String copy constructor <details>
This originates in many user actions in Biasing in many similar places as the GetLogicalVolume dereference below. I think both can be resolved by adding explicit checks that the volumes are not nullptr.
</details>

- [x] potential null pointer dereference calling GetLogicalVolume in a lot of the user actions <details>
Resolving this issue by inserting more checks requires editing a lot of the same places that https://github.com/LDMX-Software/ldmx-sw/pull/1633 is touching so I'm going to wait for that PR before patching further.
</details>

- [x] (clang-lto) using integer absolute value function `abs` when argument is a float (EcalGeometry.cxx:98, MyClusterWeight.h:42, TrackDeDxMassEstimator.cxx:63, TrackerVetoProcessor.cxx:43,45,62, TrigElectronProducer.cxx:158,159)

tomeichlersmith:

This resolves #1656 

## Check List

The CI checks backwards compatibility with ldmx/dev:4.2.2.
Using ldmx/dev:5.0.0-rc2...
- a `configure-clang-lto-fail-on-warning` build succeeded 
[clang-lto-fail-on-warning.log](https://github.com/user-attachments/files/19541478/clang-lto-fail-on-warning.log)
- a `configure-force-error` build succeeded [configure-force-error.log](https://github.com/user-attachments/files/19553674/configure-force-error.log)

### Roadmap

After merging this PR, I want to do all of the following in rather quick succession, probably in the same day, to avoid confusion. ~Hopefully that day can be this week.~
Going to wait for #1629 and #1668 for the minor release but still merge this PR before the next patch.

- Double check that tests are passing on trunk with ldmx/dev:5.0.0 locally
- Release ldmx/dev:5.0.0
- Release ldmx-sw 4.3.0
  - new release so the golden histograms are generated with 5.0.0 (good test too)
  - new minor version so its easier to state where the line of "easy" support for 5.0.0 is

### DIFF
--- a/Biasing/include/Biasing/DeepEcalProcessFilter.h
+++ b/Biasing/include/Biasing/DeepEcalProcessFilter.h
@@ -74,10 +74,8 @@ class DeepEcalProcessFilter : public simcore::UserAction {
   double ecal_min_Z_{400.};
   /// Require that the hard brem photon originates from the target
   bool require_photon_fromTarget_{false};
-  /// Enable logging
-  enableLogging("DeepEcalProcessFilter")
-      /// member used to help tag events where the photon comes from the target
-      bool photonFromTarget_{false};
+  /// member used to help tag events where the photon comes from the target
+  bool photonFromTarget_{false};
   /// member used to help tag events that have a deep-ecal process ocurr
   bool hasDeepEcalProcess_{false};
 };  // DeepEcalProcessFilter
--- a/Biasing/include/Biasing/EcalProcessFilter.h
+++ b/Biasing/include/Biasing/EcalProcessFilter.h
@@ -64,9 +64,6 @@ class EcalProcessFilter : public simcore::UserAction {
   /// Process to filter
   std::string process_{""};
 
-  /// Enable logging
-  enableLogging("EcalProcessFilter")
-
 };  // EcalProcessFilter
 }  // namespace biasing
 
--- a/Biasing/include/Biasing/NonFiducialFilter.h
+++ b/Biasing/include/Biasing/NonFiducialFilter.h
@@ -65,9 +65,6 @@ class NonFiducialFilter : public simcore::UserAction {
   double recoil_max_p_{1500};  // MeV
   /// If turned on, this aborts fiducial events.
   bool abort_fiducial_{true};
-  /// Enable logging
-  enableLogging("NonFiducialFilter")
-
 };  // NonFiducialFilter
 }  // namespace biasing
 
--- a/Biasing/include/Biasing/PrimaryToEcalFilter.h
+++ b/Biasing/include/Biasing/PrimaryToEcalFilter.h
@@ -63,8 +63,6 @@ class PrimaryToEcalFilter : public simcore::UserAction {
   /// Energy [MeV] below which a primary should be vetoed.
   double threshold_;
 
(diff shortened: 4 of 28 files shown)


tomeichlersmith:
The PR Validation is failing because of my updates to the UserAction logging infrastructure https://github.com/LDMX-Software/ldmx-sw/pull/1671/commits/c47eb30840c9228c9eaae5795681aa8af104f58f. I am using the UserAction's name (rather than class) as the name of the logging channel which, for the `EcalProcessFilter` changes the number of characters causing the `ecal_pn` and `kaon_enhanced` validation samples to fail the tests.

tomeichlersmith:
I just realized that the updated OS comes up with an updated `clang-format` which still makes changes even on a branch that is already passing the clang format CI check. I am going to push a commit that is just clang-formatting _with the updated clang-format_ to see if the new and old clang-format "compete" with each other or not. Hopefully, the new one is just additional formatting _on top_ so we don't have the CI making commits that undo itself.

tomeichlersmith:
Crap :/ the clang-formats compete.

That's annoying but not world-ending. I'm going to force-push to remove the last two commits on this branch and then it'll be ready to merge

tvami (approved):
Great thanks!

LiamBrennan-UCSB (approved):

PR #1678, opened 2025-04-03
Add input pass names + printout to existence checks

### LINKED ISSUE #1677: pass name needed to disambiguate existence check
**Describe the bug**
When there are several passes of a collection, `exists` will return `false` (I thought we fixed this...) and thus any processor that doesn't specify pass name in it's existence check will get bitten by this. For now I have identified this in the PF algorithms.

**To Reproduce**
1. Have a file with multiple passes handy. E.g., one with ecal reconstruction run both on signal hits and signal+pileup hits 
2. Run pf ecal clustering. Optionally set the input pass name (won't change a thing).
3. The output will not have any PF Ecal clusters in them as the existence check triggers `return` 
4. Optional, add a printout to log before the `return`; you will see that this is where the problem lies. (I have added this in my fix.)

snippet:
``` python
from LDMX.Recon import pfReco
ecalPF = pfReco.pfEcalClusterProducer()
#ecalPF.input_pass_name=[desiredPassHere] #optional
p.sequence.extend([ ecalPF ])
```

**Desired behavior**
1. a more intuitive behavior of framework::Event::exists() ?
2. that all processors include pass name in their checks

bryngemark:

This at least partly resolves #1677

## Check List
- [x] I ran my developments and the following shows that they are successful:
After running the recipe in #1677, I now see non-empty cluster and track collections: the processors no longer return if the specified combination of collection and pass name exists in the file. 
Screenshot 2025-04-03 at 15 41 16

### DIFF
--- a/Recon/src/Recon/PFEcalClusterProducer.cxx
+++ b/Recon/src/Recon/PFEcalClusterProducer.cxx
@@ -22,7 +22,7 @@ void PFEcalClusterProducer::configure(framework::config::Parameters& ps) {
 
 void PFEcalClusterProducer::produce(framework::Event& event) {
   if (!event.exists(hitCollName_, hitPassName_)) {
-    ldmx_log(fatal) << "Couldn't find input collection " << hitCollName_;
+    ldmx_log(fatal) << "Couldn't find input collection " << hitCollName_ << " with pass name " << hitPassName_;
     return;
   }
   const auto ecalRecHits =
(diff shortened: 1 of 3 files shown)


Recon/src/Recon/PFEcalClusterProducer.cxx
```diff
@@ -21,7 +21,10 @@ void PFEcalClusterProducer::configure(framework::config::Parameters& ps) {
 }
 
 void PFEcalClusterProducer::produce(framework::Event& event) {
-  if (!event.exists(hitCollName_)) return;
+  if (!event.exists(hitCollName_, hitPassName_)) {
+    ldmx_log(fatal) << "Couldn't find input collection " << hitCollName_;
```
tvami: Shouldnt this be?
```suggestion
    ldmx_log(fatal) << "Couldn't find input collection " << hitCollName_ << " with pass name " << hitPassName_;
```

tvami: Same comment for the other two occurrence too

tvami (approved):

therwig (approved):

Issue #1681, opened 2025-04-07
Fix SimpleDigiAndRecProducer.cxx and implement Side Hcal Algorithm


tillrue:
**Describe the bug**
When using the SimpleDigiAndRecProducer.cxx to produce RecHits, I noticed that these do not resemble the SimHits very well. Further, the algorithm for the Side Hcal was not implemented yet and used the center of each bar (in all directions) where the SimHits were located.

**To Reproduce**
1. Use the SimpleDigiAndRecProducer to obtain RecHits
2. Inspect RecHits and SimHits and notice a discrepancy

**Desired behavior**
The SimpleRecHits should give a reconstructed picture of the SimHits for each event.

**Screenshots**
This screenshot shows the SimHits of a wide scattering event with a shower in the Side Hcal. Note that the left front side represents the ecal/side hcal front on which the beam is incident.

<img width="573" alt="Image" src="https://github.com/user-attachments/assets/feee9f04-7e66-4a9a-bd1f-143bc5d3ecb4" />

This screenshot shows the current SimpleRecHits of the same event in the Side Hcal (when removing the restriction to the center of each bar in the side hcal). If we compare this to the simhits, we see that they are much further spread out. The size & colour of the boxes represent the deposited energy.

<img width="558" alt="Image" src="https://github.com/user-attachments/assets/a7c728c6-63bc-4b65-abc0-c79164ce0e40" />

This screenshot shows the new/fixed SimpleRecHits of the same event in the Side Hcal without smearing. Comparing this to the Simhits (first picture), we can see that they are much better resembled.

<img width="551" alt="Image" src="https://github.com/user-attachments/assets/7cef36b1-754b-4764-81a9-2695374e797d" />

EinarElen:
Yep this is a straight-up bug from when #1591 was merged where the priority I think was getting energy resolution... 

If any of the regular ldmx-sw developers is able to, it would be good to include some checks (in addition to the  figures above of course) either when merging this (if @tillrue/others has no issue working on their local version) or as a separate issue (if @tillrue/others needs it). I would recommend the latter :)

tillrue:
Sure, I have no issue with working on my local version to continue my project. Just wanted to mention this issue and the "relatively simple" fix as it might be the source of problem for other people as well. My changes could probably be improved as I'm just starting to develop and should definitely be checked by someone with more experience!

EinarElen:
[...]

PR #1682, opened 2025-04-07
Update HcalSimpleDigiAndRecProducer.cxx

tillrue:
In the energy weighted position average, the cumulative energy was used instead of the individual hit energy. This resulted in the odd distribution of RecHits that you can see in issue #1681 . The energy weighted time average is affected as well.

Further, I have implemented the algorithm for the side Hcal bars but I am happy to receive feedback on how to make the code more efficient or on which parts could be left out which is why I open this PR as a draft.

This resolves issue #1681 . There you can see how the current code is producing RecHits that are much further spread out than the SimHits which might be due to using the cumulative deposited energy instead of the individual one for each hit to do the weighting. On the issue page, you can also see the same event with the implemented code.

## Check List
- [✅] I successfully compiled _ldmx-sw_ with my developments.
- [✅] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [✅] I ran my developments and the following shows that they are successful.
See issue #1681 for the result of the fix.

### DIFF
--- a/Hcal/src/Hcal/HcalSimpleDigiAndRecProducer.cxx
+++ b/Hcal/src/Hcal/HcalSimpleDigiAndRecProducer.cxx
@@ -123,29 +123,27 @@ void HcalSimpleDigiAndRecProducer::produce(framework::Event& event) {
 
       // Checks orientation of side Hcal bars
       if (orientation ==
-        ldmx::HcalGeometry::ScintillatorOrientation::horizontal) {
-          xpos += (*position_resolution_smear_)(rng_);
-          ypos = stripCenter.y();
-          zpos = stripCenter.z();
-        }
-      else if (orientation ==
-        ldmx::HcalGeometry::ScintillatorOrientation::vertical) {
-          xpos = stripCenter.x();
-          ypos += (*position_resolution_smear_)(rng_);
-          zpos = stripCenter.z();
-        }
-      else if (orientation ==
-        ldmx::HcalGeometry::ScintillatorOrientation::depth) {
-          xpos = stripCenter.x();
-          ypos = stripCenter.y();
-          zpos += (*position_resolution_smear_)(rng_);
-        }
+          ldmx::HcalGeometry::ScintillatorOrientation::horizontal) {
+        xpos += (*position_resolution_smear_)(rng_);
+        ypos = stripCenter.y();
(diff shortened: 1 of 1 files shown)


Hcal/src/Hcal/HcalSimpleDigiAndRecProducer.cxx
```diff
@@ -119,10 +120,30 @@ void HcalSimpleDigiAndRecProducer::produce(framework::Event& event) {
...
+                 ldmx::HcalGeometry::ScintillatorOrientation::depth) {
+        xpos = stripCenter.x();
+        ypos = stripCenter.y();
+        zpos += (*position_resolution_smear_)(rng_);
+      }
+      // This might be irrelevant
+      else {
+        xpos = stripCenter.x();
+        ypos = stripCenter.y();
+        zpos = stripCenter.z();
```
EinarElen, line 144: Probably worth adding a warning here so that if someone makes some new hcal model and things change, they'd be alerted

EinarElen (approved):
Looks correct :)

Hcal/src/Hcal/HcalSimpleDigiAndRecProducer.cxx
tillrue: Good idea and probably better than exception. I'm not sure how to create a warning within ldmx-sw though but maybe one of the reviewers can help with that?

bryngemark: I think an exception is ok until someone makes a change that allows us to end up here somehow, and then that should be dealt with accordingly. 

An example of printing stuff to the logger (includes terminal messages; use this instead of `cout`) can be found here: https://github.com/LDMX-Software/ldmx-sw/blob/3a1f52b2207aa75cc81ffc62c6a82402f716c58c/Hcal/src/Hcal/HcalVetoProcessor.cxx#L138-L139 
in your case select a level of more severity, like warning or fatal

tillrue: I see. Will implement this. Thanks for the help!

EinarElen (approved):

tvami (approved):

PR #1683, opened 2025-04-08
Move ECAL veto to fully rely on tracking

### LINKED ISSUE #1549: Update to latest tracking in CIs and use it in ECAL projections

Follow-up to https://github.com/LDMX-Software/ldmx-sw/issues/1517
So far the ECAL projections use truth information, we added some code already to rely on ACTS tracking, but the defaults were not changed yet.

Move the default for the ECAL projections to rely on the state-of-the art tracking, updated in https://github.com/LDMX-Software/ldmx-sw/pull/1526

No alternative, we have to do this in real data.

tvami:

This is instead of https://github.com/LDMX-Software/ldmx-sw/pull/1629/
Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1549

## Check List

### DIFF
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -66,7 +66,6 @@ from LDMX.DQM import dqm
 
 # Load ecal veto and use tracking in it
 ecal_veto = ecal_vetos.EcalVetoProcessor()
-ecal_veto.recoil_from_tracking = True
 
 # Load hcal veto
 import LDMX.Hcal.hcal as hcal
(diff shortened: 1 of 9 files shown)


tvami:
here are the failed plots for the ecal PN
[ecal_pn_recon_validation_plots_April8.pdf](https://github.com/user-attachments/files/19661104/ecal_pn_recon_validation_plots_April8.pdf)


I think this looks pretty good! The only real looking diff is in the MIP tracking tails. Funny thing is that I argued that the falling distribution makes more sense in https://github.com/LDMX-Software/ldmx-sw/pull/1668#issuecomment-2767407147
Now however we are back to the shape as we had it in 4.2.17. I guess at this point the only thing I'll say is that I dont have a good intuition on how this should look like, but given that this is how it's been in all releases except for 4.2.18, I guess I'll go with the consistency argument...

tomeichlersmith (approved):
Looks good :)

JYoo001 (approved):
Looks good to me!

PR #1688, opened 2025-04-10
Move to simple HCAL reco in all CIs

### LINKED ISSUE #1686: Move to the "new" simple HCAL reco in CI

There are several problems with the "old" HCAL reco

Chatting with @bechenard I understand we should really just move to the simplified reco introduced (back) in https://github.com/LDMX-Software/ldmx-sw/pull/1591

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1686

## Check List

### DIFF
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -37,6 +37,7 @@ import LDMX.Ecal.vetos as ecal_vetos
 import LDMX.Hcal.HcalGeometry
 import LDMX.Hcal.hcal_hardcoded_conditions
 import LDMX.Hcal.digi as hcal_digi
+hcal_digi_reco = hcal_digi.HcalSimpleDigiAndRecProducer()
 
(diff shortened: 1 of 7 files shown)


tomeichlersmith (approved):

EinarElen:
Wouldn't it be possible to run both versions with different pass name for the digi + reco one? So we still get info about anything happening on the ecal Digi/reco side affecting the hcal? Also useful whenever we pick hcal reco back up.

Hcal reco is an extremely small part of the runtime so doubling it seems fine to me

tvami:
> Wouldn't it be possible to run both versions with different pass name for the digi + reco one? 

I think this is relevant for private use, for the CI, we need one collection that can go into the DQM. Unless you are suggesting to double the DQM too, but that will be pretty big change in the number of DQM histograms

cmantill (approved):

bechenard (approved):

PR #1699, opened 2025-05-05
Add CLUE clustering

### LINKED ISSUE #1411: Update Ecal clustering with the CLUE algorithm

tomeichlersmith:
I am copying over the CLUE clustering, additional analyzer and update to the SimCalorimeterHit that was originally developed in Summer 2024 to study ECal clustering at Lund by Ella Viirola. The visualization code based on Phoenix I plan to move into its own repository and the JSON-producers that prepare an event to be visualized I plan to put into EventDisplay.

This resolves #1411 

## Check List

@Lysarina originally did these developments and includes evidence of their success in the original PR #1416 

## To Do
- [x] is overlay of the sim hits breaking now? See https://github.com/LDMX-Software/ldmx-sw/pull/1416#issuecomment-2757530651


tomeichlersmith:
@bryngemark The fact that the PR Validations pass completely leads me to believe that the issues you were observing before are not present (at least within the clean CI environment.) The total energy is appropriately shifted high and is not the discrete distribution you've shown before.

image


Since this distribution is cut off at higher energies, I'm re-running the `it_pileup` job locally so that I can make sure it still works above the 12GeV line.

tvami (changes requested):
I skimmed thru quickly, I'll have a 2nd read-thru later, but two general comments:
 * the variable names are not according to our coding rules
 * there are `cout`s and `debug` flags that should go away into the logging system (everything under debug should be `trace` in our logging system)

tomeichlersmith:
@bryngemark I have confirmed that, locally **and running from scratch**, the IT overlay is working as expected. As a double check, I just looked at the `EcalVeto_overlay.summedDet_` branch like you have and I see a normal distribution.

image

Without being able to replicate your issue, I would like to ask you to provide more detail on the environment and inputs you used to produce the plot in the original issue. What version of ldmx-sw produced the overlay events? What version produced the "main" events? What version ran overlay? Those sort of questions. If the environment is on SLAC's S3DF, I'd be happy to look at it directly as well if you give me the directory (and I have access to read that directory).

bryngemark:
that's great, @tomeichlersmith. i was getting hopeful when the validation pileup test went through. 

i ran on ella's dev branch (that the PR was made from) locally on my laptop. i confirmed that it was behaving differently than the then-current `trunk` on the same machine (which showed reasonable energies with overlay); that was sometime early fall. i don't think we have any reason to believe whatever happened is present here (in fact we have reasons to believe it isn't) so i suggest we merge and then if i reproduce the bug with new `trunk`, i make an issue.

tomeichlersmith:
@tvami I _have not_ moved all of the variables in the CLUE implementation to `lower_case`  (amongst other violations of our coding rules) because I fear breaking the implementation when I'm doing it manually. I am currently trying to figure out how to use `clang-tidy` to try to have it done algorithmically, but I also feel like that is outside the scope of this PR. Can we merge this despite its failure to completely follow the coding rules? I can put my `clang-tidy` research into an issue and maybe we can have a workflow that enforces more of our coding rules like `clang-format` does.

Ecal/exampleConfigs/cluster.py
```diff
@@ -0,0 +1,13 @@
...
+import sys
+p.inputFiles = sys.argv[1:]
+p.outputFiles = [ 'clusters.root' ]
+p.histogramFile = 'h_clusters.root'
+
+from LDMX.Ecal.ecalClusters import *
+p.sequence = [
+    EcalClusterProducer(),
+    EcalClusterAnalyzer()
+]
```
tvami, line 12: Should we add this to the CI, so it actually tests this?

Ecal/include/Ecal/EcalClusterAnalyzer.h
```diff
@@ -0,0 +1,52 @@
+/**
+ * @file EcalClusterAnalyzer.h
+ * @brief Analysis of cluster performance
+ * @author Ella Viirola, Lund University
+ */
+
+#ifndef DQM_ECALCLUSTERANALYZER_H
```
tvami: Should this not be under DQM (as the def name also suggests?)

Ecal/include/Ecal/WorkingEcalCluster.h
```diff
@@ -0,0 +1,44 @@
...
+
+#include <iostream>
+#include <vector>
+
+#include "Ecal/Event/EcalHit.h"
+#include "TLorentzVector.h"
+
+namespace ecal {
+
+class WorkingEcalCluster {
```
tvami: Do we really want to call this "Working"? I'm not even sure I understand what this is supposed to mean, if to keep it maybe some more description in a comment would be good

SimCore/src/SimCore/SDs/EcalSD.cxx
```diff
@@ -112,8 +112,19 @@ G4bool EcalSD::ProcessHits(G4Step* aStep, G4TouchableHistory*) {
     if (compressHitContribs_ and contrib_i != -1) {
       hit.updateContrib(contrib_i, edep, time);
     } else {
-      hit.addContrib(getTrackMap().findIncident(track_id), track_id, pdg, edep,
-                     time);
+      int origin{-1};
+      auto map{getTrackMap()};
+      auto incident{map.findIncident(track_id)};
+      for (int i{1}; i < 6; ++i) {
```
tvami: this "6" should be not hard coded here

Lysarina:
Hi! This might already be resolved or might not even be an issue, but as I wrote in my PR, the initial cluster producer used the WorkingCluster class, that took in EcalHits as pointers. However when I started out I was kind of a pointer noob so I created WorkingEcalCluster, which does essentially the same thing (with some additional improvements), but takes in references instead of pointers in the methods. I never had time to fix this (going back to pointers), so I just wanted to point this out again, in case this would cause some issues later down the line.

Thanks again for taking this on @tomeichlersmith !

Ecal/include/Ecal/WorkingEcalCluster.h
tomeichlersmith: The "working" clusters are supposed to be the intermediate clusters that are not finished yet. As Ella pointed out, this class should be removed and replaced with using the `WorkingCluster` class which uses pointers to try to avoid as much copying of the EcalHit data.

tvami: > The "working" clusters are supposed to be the intermediate clusters that are not finished yet. 

OK I understood this like it's working (i.e. functioning) now, as opposed to that it didnt work in the past. Should this not be better called as `TempCluster`, or `IntermediateCluster`, `InProgressCluster`, `ClusterInProgress` or something else that doesnt make it sound like this is the final product?

tomeichlersmith: Sure that makes sense, I like Intermediate so I'll go with that one.

tvami (approved):
Awesome, thanks a lot!

PR #1704, opened 2025-05-12
Factory updates and unification

### LINKED ISSUE #1465: LTO prevents SimCore Factory from creating objects
**Describe the bug**
The SimCore plugin factory is found unaware of declared plugins when LTO is enabled.

**To Reproduce**
```
just configure-clang-lto
just build
just fire SimCore/test/basic.py
```
and then you see
```
[ fire ] 0 fatal: [SimFactory] : An object named simcore::TrackerSD has not been declared.
  at /home/tom/code/ldmx/ldmx-sw/SimCore/include/SimCore/Factory.h:274 in make
```

**Desired behavior**
It'd be cool to be able to enable LTO.

**Additional context**
So, there are actually two different plugin factories in ldmx-sw.
- [Framework/PluginFactory](https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Framework/include/Framework/PluginFactory.h) is a C-style plugin manager that forces registration to happen at library-load-time using compiler attributes (like `__attribute(init_priority(500))` and `__attribute__(constructor(1000))`)
- [SimCore/Factory](https://github.com/LDMX-Software/ldmx-sw/blob/trunk/SimCore/include/SimCore/Factory.h) is a templated C++17 style factory that uses cleverly-placed `static` variables to force certain behaviors at library-load-time.

The Framework/PluginFactory appears to be working when LTO is enabled since we get to attempting to create the TrackerSD. This implies that the Framework/PluginFactory was already used to declare and create the simulation processor. This causes me to focus attention on SimCore/Factory which works out since I originally designed it. I placed some printouts including the address of the Factory so we can observe the separate Factories being created for the different types. This is what I see.
```
denv fire SimCore/test/basic.py 
Factory(0x7ff2238562b8)
Factory(0x7ff2238562b8): declare simcore::BertiniAtLeastNProductsModel
Factory(0x7ff2238562b8): declare simcore::BertiniModel
Factory(0x7ff2238562b8): declare simcore::BertiniNothingHardModel
Factory(0x7ff2238562b8): declare simcore::BertiniSingleNeutronModel
Factory(0x7ff2238562b8): declare simcore::NoPhotoNuclearModel
Factory(0x7ff22383d4b0)
Factory(0x7ff22383d4b0): declare simcore::EcalSD
Factory(0x7ff22383d4b0): declare simcore::HcalSD
Factory(0x7ff22383d4b0): declare simcore::ScoringPlaneSD
Factory(0x7ff22383d4b0): declare simcore::TrackerSD
Factory(0x7ff22383d4b0): declare simcore::TrigScintSD
Factory(0x7ff223816538)
Factory(0x7ff223816538): declare simcore::generators::GeneralParticleSource
Factory(0x7ff223816538): declare simcore::generators::LHEPrimaryGenerator
Factory(0x7ff223816538): declare simcore::generators::MultiParticleGunPrimaryGenerator
Factory(0x7ff223816538): declare simcore::generators::ParticleGun
Factory(0x7ff2270bcfe0)
Factory(0x7ff2270bcfe0): make simcore::TrackerSD
[ fire ] 0 fatal: [SimFactory] : An object named simcore::TrackerSD has not been declared.
  at /home/tom/code/ldmx/ldmx-sw/SimCore/include/SimCore/Factory.h:274 in make
~Factory(0x7ff2270bcfe0)
~Factory(0x7ff223816538)
~Factory(0x7ff22383d4b0)
~Factory(0x7ff2238562b8)
```
So LTO _does not_ prevent the declaration process from happening (which is what I assumed was going wrong), but - for some reason - a _new_ Factory is created when attempting to create the TrackerSD instead of using the one that was already created during library-load for declaration. My guess is that I'll need to read-up on the actual meaning of `static` variable in the context of a `static` class member function. Oofdah.

I'll probably try to create a separate testing area since it takes so long to recompile after changing the templated Factory.h header. See [my gist](https://gist.github.com/tomeichlersmith/017ae50b9c1c35afa227a4bebde933b9) about this Factory for more links and probably where I'll put results.

### LINKED ISSUE #1698: improve `UnableToCreate` error message
https://github.com/LDMX-Software/ldmx-sw/blob/1b469b698636f4d2c11240662cefd67c84201cef/Framework/src/Framework/Process.cxx#L85-L91

This error message could have more detail about the necessary ingredients for the Framework to be able to create a processor.
- class needs to inherit from `Producer` or `Analyzer`
- needs to be a declaration macro in the implementation file (e.g. `DECLARE_PRODUCER_NS`)
- class name including namespace(s) needs to be given in the configuration

This could also be done at the same time as https://github.com/LDMX-Software/ldmx-sw/issues/1465 if I ever figure that out.

tomeichlersmith:

This resolves #1465 and resolves #1698 

Besides fixing the issue with the SimCore/Factory I also did some refactoring to simplify/unify the dynamic factory code.
- Move SimCore/Factory into Framework and remove Framework/PluginFactory (and SimCore/Geo/ParserFactory)
- Refactor EventProcessor/Produer/Analyzer to function with templated Factory
- New registration macros do not require separating namespace and class name so I've removed the `*_NS` macros
- When I was updating the Factory documentation, I noticed that `doxygen` was running over all the files in the `build` subdirectory as well, so I am also updating the doxyfile so that it properly excludes the generated/copied files.
- The Factory does not throw its own exceptions, instead it returns `nullptr` if it is unable to create an object by name leaving it up to the caller to decide how to handle the unable-to-create situation (and what to write into the exception). I chose to do this because some exceptions need to have more detail than others.

## Check List
- [x] I have added `just test` into the clang-lto workflow which includes running the basic simulation. Running the basic simulation fails on `trunk` as documented in #1465 but it doesn't anymore.
- [x] I manually ran a asan-ubsan build locally to see if I introduced any new memory leaks and I found none
- [x] Look at how the new error message appears in terminal
- [x] Make sure single-file processors still functions. I noticed when testing this that the single-file compilation complains about the C++ standard so I have added the C++20 standard to the single-file compilation command `-std=c++20`.
- [x] Undo library/file movement in SimCore? I don't think its necessary and I'm unsure if its even helpful from a code-understandability perspective. _Reply_: I'm going to leave it as is. I don't think reverting it makes any more sense anyways.


tvami (approved):
This is awesome work, thanks Tom!

EinarElen:
+1 on Tamas.

Regarding returning a nullptr, could we not return a std::optional? Nullptrs have a habit of being ignored (std::expected would be better but that's c++23 😔)

Framework/include/Framework/ConditionsObjectProvider.h
```diff
@@ -72,7 +70,7 @@ class ConditionsObjectProvider {
   /**
    * Class destructor.
    */
-  virtual ~ConditionsObjectProvider() { ; }
+  virtual ~ConditionsObjectProvider() = default;
```
EinarElen, line 74: Underappreciated patch in the middle of all of this

Framework/include/Framework/EventProcessor.h
```diff
@@ -312,6 +310,20 @@ class Analyzer : public EventProcessor {
...
+  virtual void process(Event &event) final { analyze(event); }
+
+  /**
+   * Don't allow Analyzers to add parameters to the run header
+   *
+   * We make a `final` override of beforeNewRun to be empty so
+   * that any derived Analyzer will not be able to compile an
+   * override of the beforeNewRun callback.
+   */
+  virtual void beforeNewRun(ldmx::RunHeader &runHeader) final {}
```
EinarElen, line 325: Really nice

Framework/src/Framework/Process.cxx
```diff
@@ -80,14 +93,20 @@ Process::Process(const framework::config::Parameters &configuration)
...
+                      "The EventProcessor Factory was unable to create " +
+                          instanceName + " of type " + className +
+                          ". Did you inherit from framework::Producer or "
+                          "framework::Analyzer? "
+                          "Did you DECLARE_PRODUCER or DECLARE_ANALYZER in the "
+                          "implementation (.cxx) file? "
+                          "Did you use the class's full name (including "
+                          "namespaces) in the Python configuration class? "
+                          "Does the Python configuration class reference the "
+                          "correct library it is a part of?");
```
EinarElen, line 109: 🔥

EinarElen (approved):

tomeichlersmith:
The reason I did not use `std::optional` is because then the returned object would have three possible states.

1. `std::nullopt`
2. `std::optional::has_value` is `true` but the pointer is `nullptr`
3. `std::optional::has_value` is `true` and the pointer is valid

I didn't like the confusion of having two different ways of reflecting that the returned pointer is invalid. I also don't like having to trust that we check for `nullptr`, but I think (hope?) that our additional warnings and the fact that the Factory code is not interacted with much will keep it from being an issue.

EinarElen:
Yeah if we were to return optional we would have to return a value type or make the API promise that optional.has_value() for this part always implies a valid pointer

tomeichlersmith:
Alright, I'm convinced. I was just looking into SimCore a bit more and found a few places where I forgot to check if the returned value was non-null. I'm going to change the return type of `Factory::make` to `std::optional<PrototypePtr>` and mark it as `[[nodiscard]]`.

EinarElen (approved):

tvami (approved):

PR #1705, opened 2025-05-13
Add Hit Reconstruction from Axial and Stereo Tracking Layers to rLDMX Tracking

### LINKED ISSUE #1672: Update 3D Hit Construction in reduced-ldmx Tracking

Currently, the reduced tracking scripts only create 3D points with x and z ldmx coordinates, while the y coordinate is always taken to be 0 (since this is the insensitive direction along the strips).

The tracking should take advantage of the axial-stereo pair geometry to construct the y coordinate and make a representative 3D spatial point in the tracking geometry. This would be similar to the hit construction used in HPS.

We initially tried using an energy-weighted average of the global position of (digitized) hits in the tracker, but this is not really a realistic approach to tracking.

fdelzanno:

This resolves #1672 by adding a method to LinearSeedFinder.cxx that reconstructs 3D hits from all combinations of an axial and a stereo measurement. It also updates the uncertainties associated with the x and y coordinates of the reconstructed hit. 

## Check List

Performance plots are shown in #1672.

### DIFF
--- a/.github/validation_samples/reduced_ldmx/config.py
+++ b/.github/validation_samples/reduced_ldmx/config.py
@@ -16,7 +16,7 @@ myGun.pdgID = 11
 myGun.enablePoisson = False #True   
 
 mySim = sim.simulator( "mySim" ) # Build simulator object
-det = 'ldmx-reduced-v1'
+det = 'ldmx-reduced-v2'
 mySim.setDetector(det, True )
 mySim.beamSpotSmear = [20.,80.,0.]
 mySim.description = 'Reduced ECal Electron Gun Test Simulation'
@@ -113,7 +113,6 @@ rSeedTracking.layer23_midpoint = layer23_mid
 rSeedTracking.layer34_midpoint = layer34_mid
 rSeedTracking.recoil_uncertainty = [0.006, 0.085]
 rSeedTracking.ecal_distance_threshold = 15.0
-rSeedTracking.buildHistograms()
 
 rTracking = reducedTracking.LinearTrackFinder("LinearTrackFinder")
 rTracking.seed_collection = "LinearRecoilSeedTracks"
--- a/.github/validation_samples/reduced_ldmx/config.py
+++ b/.github/validation_samples/reduced_ldmx/config.py
@@ -16,7 +16,7 @@ myGun.pdgID = 11
 myGun.enablePoisson = False #True   
 
 mySim = sim.simulator( "mySim" ) # Build simulator object
-det = 'ldmx-reduced-v2'
+det = 'ldmx-reduced-v1'
 mySim.setDetector(det, True )
 mySim.beamSpotSmear = [20.,80.,0.]
 mySim.description = 'Reduced ECal Electron Gun Test Simulation'
@@ -111,7 +111,9 @@ rSeedTracking.out_seed_collection = "LinearRecoilSeedTracks"
 rSeedTracking.layer12_midpoint = layer12_mid
 rSeedTracking.layer23_midpoint = layer23_mid
 rSeedTracking.layer34_midpoint = layer34_mid
+rSeedTracking.recoil_uncertainty = [0.006, 0.085]
 rSeedTracking.ecal_distance_threshold = 15.0
+rSeedTracking.buildHistograms()
 
 rTracking = reducedTracking.LinearTrackFinder("LinearTrackFinder")
 rTracking.seed_collection = "LinearRecoilSeedTracks"
--- a/EventDisplay/python/rLDMX_EventDisplay.py
+++ b/EventDisplay/python/rLDMX_EventDisplay.py
@@ -27,9 +27,10 @@ def addBranch(tree, ldmx_class, branch_name):
 
     return branch
     
-def trackPlotter(tree, event_number, tag, save):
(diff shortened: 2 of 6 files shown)


EventDisplay/python/rLDMX_EventDisplay.py
```diff
@@ -102,14 +113,15 @@ def trackPlotter(tree, event_number, tag, save):
...
+    targetSP_hits = np.column_stack((targetSP_z, targetSP_x, targetSP_y))
     first_sensor_pos = np.column_stack((first_sensor_z, first_sensor_x, first_sensor_y))
     second_sensor_pos = np.column_stack((second_sensor_z, second_sensor_x, second_sensor_y))
     
     fig = plt.figure(figsize=(8, 8))
     ax = fig.add_subplot(111, projection='3d')
 
     x_sensor_uncertainty = 0.006 #mm
-    y_sensor_uncertainty = 20./np.sqrt(12) # mm
+    y_sensor_uncertainty = 0.085 # mm
```
tvami, line 127: can you add a comment on how you got this number?

Tracking/src/Tracking/Reco/LinearSeedFinder.cxx
```diff
@@ -43,74 +43,196 @@ void LinearSeedFinder::configure(framework::config::Parameters& parameters) {
...
+            else {
+                second_sensor_point = {second_combo_3d_point, std::get<ldmx::Measurement>(second_layer_one), std::nullopt};
+            } //...we are taking only one layer as the measurement (axial or stereo)
+            
+            for (const auto& rec_hit : first_layer_ecal_rec_hits) {
+                // Do fitting on 2 sensor + 1 recHit combinations = 1 degree of freedom for linear fit
+                ldmx::StraightTrack seed_track = SeedTracker(first_sensor_point, second_sensor_point, rec_hit);
+                
+                // Seed passed RecHit distance check, add it
+                if (seed_track.getChi2() > 0.0) {
```
tvami: shouldnt this be a harder cut? or that comes later, if we want it?

tvami (changes requested):
You'll have to rebase so that the code compiles, and then I had two small points:

Tracking/src/Tracking/Reco/LinearSeedFinder.cxx
fdelzanno: The recHit_distance cut of <= 15 mm has always been rather arbitrary. I could cut at 10 mm without much difference. Once we reach cuts of ~5 mm, we will likely start to lose events. This is the only place in the code in which the recHit distance cut is applied. Later, the code just relies on selecting the minimum chi-2 value.

EventDisplay/python/rLDMX_EventDisplay.py
fdelzanno: Ok, added a comment. Here is a picture of the calculation:
uncertainty_calc

tvami (approved):
Great! Please also edit the PR title

bloodyyugo (commented):
Looks good!

bloodyyugo (approved):

tvami:
Let's see if the rLDMX tests succeed in the CI too

tvami:
So the CI sees this issue of
```
 Histogram LinearSeedFinder_sensor1_measured_x-sim_stereo_layer1_x_ALL not found in pool.
```

But that also makes me think that these plots should be under the DQM directory not directly in the Producer. Technically even it should have it's own Analyzer c++ package. @fdelzanno Maybe we should decouple that to another PR?

.github/validation_samples/reduced_ldmx/config.py
```diff
@@ -16,7 +16,7 @@
 myGun.enablePoisson = False #True   
 
 mySim = sim.simulator( "mySim" ) # Build simulator object
-det = 'ldmx-reduced-v2'
+det = 'ldmx-reduced-v1'
```
```suggestion
det = 'ldmx-reduced-v2'
```
I think we should keep this on v2

.github/validation_samples/reduced_ldmx/config.py
```diff
@@ -111,7 +111,9 @@
 rSeedTracking.layer12_midpoint = layer12_mid
 rSeedTracking.layer23_midpoint = layer23_mid
 rSeedTracking.layer34_midpoint = layer34_mid
+rSeedTracking.recoil_uncertainty = [0.006, 0.085]
 rSeedTracking.ecal_distance_threshold = 15.0
+rSeedTracking.buildHistograms()
```
```suggestion
rSeedTracking.title = ""
rSeedTracking.buildHistograms()
```
if we just want to resolve the issue

tvami:
CI plots look according to the expectations now:
[reduced_ldmx_recon_validation_plots.pdf](https://github.com/user-attachments/files/20212345/reduced_ldmx_recon_validation_plots.pdf)

tvami:
@fdelzanno feel free to squash and merge this

tvami:
> CI plots look according to the expectations now: [reduced_ldmx_recon_validation_plots.pdf](https://github.com/user-attachments/files/20212345/reduced_ldmx_recon_validation_plots.pdf)

@SanjitMasanam FYI you should look at the WAB plot changes

fdelzanno:
Ok, I'll wait for the last 3 tests to conclude and then merge.

PR #1706, opened 2025-05-14
Fix PFTrack producer to find electrons in signal with overlay

### LINKED ISSUE #1680: PFTrack producer unable to find electrons in signal with overlay
**Describe the bug**
I can only get half the electron tracks (the signal ones) when I have a signal+PU sample

**To Reproduce**
Run pf track reco on a signal + PU sample, check size of track collection: it s 1

**Desired behavior**
Both pileup electron and recoil electron are found

bryngemark:

This resolves #1680

## Check List
- [x] I ran my developments and the following shows that they are successful: track multiplicity after fix is on average 2 for 2e (signal + pu overlay) events
nTrk


Recon/src/Recon/PFTrackProducer.cxx
```diff
@@ -28,14 +29,39 @@ void PFTrackProducer::produce(framework::Event& event) {
...
-        continue;
       if (spHit.getPdgID() == 22 || spHit.getPdgID() == 2112) continue;
-      pfTracks.push_back(spHit);
-      break;
-    }
-  }
+      if (fabs(240 - spHit.getPosition()[2]) > 0.1) continue;
+      if (doElectronTracking_) {  // only select electron SP hits
+        if (spHit.getPdgID() != 11) continue;
+        if (spHit.getTrackID() < 2 && spHit.getMomentum()[2] > 2500) {
```
tvami: Why is this 2500? This should be configurable, and then have it the incoming energy dependent, right?

Recon/src/Recon/PFTrackProducer.cxx
```diff
@@ -28,14 +29,39 @@ void PFTrackProducer::produce(framework::Event& event) {
...
+          // this is almost guaranteed to be a pileup beam electron! keep it
+          pfTracks.push_back(spHit);
+          ldmx_log(debug) << "Added beam electron SP hit: trackID="
+                          << spHit.getTrackID()
+                          << ", pz = " << spHit.getMomentum()[2];
+        } else if (spHit.getTrackID() <= 30 && spHit.getMomentum()[2] > 5) {
```
tvami: What's special about track ID 30?

Recon/src/Recon/PFTrackProducer.cxx
bryngemark: maybe it should be energy dependent. i pseudo-optimised this by eye, having a sample with both signal and pileup in it, and 2500 allows both electrons to be caught while screening out secondaries. i suspect this also depends on some cut in signal generation. since this is all using truth, one could have had a cut on parents/history, but since this looks different between signal and other samples, i didn't pursue that direction. i can make it configurable but i'm not sure precisely how it relates to beam energy. i suspect signal mass point dependence to be larger for us going forward tbh

Recon/src/Recon/PFTrackProducer.cxx
bryngemark: nothing except the distribution of trackIDs for good electrons naturally falling off at that point. this also shows how it's not completely trivial to select the beam electron based on for instance trackID in a pileup sample. imagine we have 1+4e... 
overall, i expect this whole procedure to abandon SP info for recoil tracking in the near future

bryngemark:
build fails after unrelated compile errors in the `EcalVetoResult` like 
```
/home/runner/work/ldmx-sw/ldmx-sw/Ecal/src/Ecal/Event/EcalVetoResult.cxx:113:67: error: 'recoilP' has incomplete type
  113 |     std::vector<float> EcalLayerEdepReadout, std::array<float, 3> recoilP,
      |                                              ~~~~~~~~~~~~~~~~~~~~~^~~~~~~
```

tvami (approved):

tvami:
> build fails after unrelated compile errors

I think this should be gone if you rebased to trunk

Recon/src/Recon/DBScanClusterBuilder.cxx
```diff
@@ -124,6 +126,7 @@ void DBScanClusterBuilder::fillClusterInfoFromHits(
   cl->setHitValsY(raw_yvals);
   cl->setHitValsZ(raw_zvals);
   cl->setHitValsE(raw_evals);
+  cl->addHits(constituentHits);  // associate used hits to cluster
```
therwig, line 129: Can we make it optional to store all hits? I see it could definitely be useful, but its also nice to be able to make lighter-weight events with just the cluster attributes themselves.

tvami:
This error OTOH is connected to this PR
```
/home/runner/work/ldmx-sw/ldmx-sw/Recon/src/Recon/PFEcalClusterProducer.cxx: In member function 'virtual void recon::PFEcalClusterProducer::produce(framework::Event&)':
/home/runner/work/ldmx-sw/ldmx-sw/Recon/src/Recon/PFEcalClusterProducer.cxx:42:21: error: no matching function for call to 'recon::DBScanClusterBuilder::runDBSCAN(std::vector<const ldmx::CalorimeterHit*>&, bool)'
   42 |         cb.runDBSCAN(ptrs, false);
      |         ~~~~~~~~~~~~^~~~~~~~~~~~~
In file included from /home/runner/work/ldmx-sw/ldmx-sw/Recon/src/Recon/PFEcalClusterProducer.cxx:3:
/home/runner/work/ldmx-sw/ldmx-sw/Recon/include/Recon/DBScanClusterBuilder.h:29:59: note: candidate: 'std::vector<std::vector<const ldmx::CalorimeterHit*> > recon::DBScanClusterBuilder::runDBSCAN(const std::vector<const ldmx::CalorimeterHit*>&)'
   29 |   std::vector<std::vector<const ldmx::CalorimeterHit *> > runDBSCAN(
      |                                                           ^~~~~~~~~
/home/runner/work/ldmx-sw/ldmx-sw/Recon/include/Recon/DBScanClusterBuilder.h:29:59: note:   candidate expects 1 argument, 2 provided
```

therwig (approved):
Thanks for this nice update and for the bug fixes!  My only request would be to make hit-saving optional but it's a minor point. Let me know what you think about this.  in any case, happy to approve..

bryngemark:
> This error OTOH is connected to this PR
> 
> ```
>    29 |   std::vector<std::vector<const ldmx::CalorimeterHit *> > runDBSCAN(
>       |                                                           ^~~~~~~~~
> [...]
> /home/runner/work/ldmx-sw/ldmx-sw/Recon/include/Recon/DBScanClusterBuilder.h:29:59: note:   candidate expects 1 argument, 2 provided
> ```

should be fixed now

Recon/src/Recon/DBScanClusterBuilder.cxx
bryngemark: we could, but right now, there are no configurable parameters at all in `DBScanClusterBuilder`. can we make a separate issue for making parameters configurable (set up a python function etc) where we can add a boolean to choose to keep hits or not? 

the reason i am not keen on doing this now is i started a rebase and got a series of new compile errors related to the CLUE merge that i don't really have time to understand. while i can do simple fixes from the browser i would prefer to start from a clean slate to set up larger things like this properly, to make sure stuff compiles.

tvami (approved):

tvami:
Looks good!

Recon/src/Recon/DBScanClusterBuilder.cxx
therwig: Fine by me - I will defer to the person who actually has to do the work 😉 
If it becomes an issue for me, I can always make a PR (for you to review 😃) later.

PR #1708, opened 2025-05-16
Add pass name to every get collection instance

### LINKED ISSUE #1645: Add pass name to every getCollection() occurance

With the resim sample there are 2 SPs and none of the code is consuming pass names for the SPs. Which made me look where else we are missing the pass names.

Add pass name to every occurance of `getCollection()`
https://github.com/search?q=repo%3ALDMX-Software%2Fldmx-sw%20getCollection&type=code

None, we enforce this for any future code anyway.

**Additional context**

Missing momentum analysis with the reSimmed ecalPN sample

LiamBrennan-UCSB:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1645
Pass names are added to every occurrence of getCollection(), getMap(), getObject(), and exists(). This updates previous code such that they match the formatting of the newer code.

## Check List

### DIFF
--- a/Framework/include/Framework/Event.h
+++ b/Framework/include/Framework/Event.h
@@ -397,19 +397,6 @@ class Event {
   //    const std::string &passName) const {
   //  return getObject<std::vector<ContentType> >(collectionName, passName);
   //}
-  
-  //template <typename ContentType>
-  //const std::vector<ContentType>& getCollection(
-  //    const std::string& collectionName,
-  //    const std::string& passName) const {
-  //  return getObject<std::vector<ContentType>>(collectionName, passName);
-  //}
-  //template <typename ContentType>
-  //const std::vector<ContentType>& getCollection(
-  //    const std::string& collectionName,
-  //    const std::string& passName) const {
-  //  return getObject<std::vector<ContentType>>(collectionName, passName);
-  //}
 
   template <typename ContentType>
   const std::vector<ContentType>& getCollection(
--- a/Framework/include/Framework/Event.h
+++ b/Framework/include/Framework/Event.h
@@ -132,7 +132,11 @@ class Event {
    * false if allowing for one or more matching objects
    * @return True if the object or collection exists in the event.
    */
-  bool exists(const std::string &name, const std::string &passName = "",
+  bool exists(const std::string& name) const {
+    return exists(name, "", true);
+  }   
+
+  bool exists(const std::string &name, const std::string &passName,
               bool unique = true) const;
 
   /**
@@ -274,9 +278,12 @@ class Event {
    * @param passName name of pass you want
    * @return const reference to requested object
    */
+  //template <typename T>
+  //const T &getObject(const std::string &collectionName,
+  //                   const std::string &passName) const {
   template <typename T>
-  const T &getObject(const std::string &collectionName,
-                     const std::string &passName = "") const {
+  const T& getObject(const std::string& collectionName,
+                    const std::string& passName) const {  
     // get branch name
     std::string branchName;
     if (collectionName == ldmx::EventHeader::BRANCH) {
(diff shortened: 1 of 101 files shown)


Framework/include/Framework/Event.h
```diff
@@ -274,9 +278,12 @@ class Event {
    * @param passName name of pass you want
    * @return const reference to requested object
    */
+  // template <typename T>
+  // const T &getObject(const std::string &collectionName,
+  //                    const std::string &passName) const {
```
tvami:
```suggestion
```

Framework/include/Framework/Event.h
```diff
@@ -132,7 +132,11 @@ class Event {
    * false if allowing for one or more matching objects
    * @return True if the object or collection exists in the event.
    */
-  bool exists(const std::string &name, const std::string &passName = "",
+  // bool exists(const std::string& name) const {
+  //   return exists(name, "", true);
+  // }  just
+
```
tvami:
```suggestion
```

Framework/src/Framework/Event.cxx
```diff
@@ -6,6 +6,11 @@ namespace framework {
 
 Event::Event(const std::string& thePassName) : passName_(thePassName) {}
 
+// void Event::configure(framework::config::Parameters &parameters) {
+//   nextEvent_passName_ =
+//   parameters.getParameter<std::string>("nextEvent_passName");
+// }
+
```
tvami:
```suggestion
```

Framework/include/Framework/Event.h
```diff
@@ -518,6 +523,7 @@ class Event {
    */
   ldmx::EventHeader eventHeader_;
 
+  const std::string eventHeaderPassName_{"eventPassName"};
```
tvami:
```suggestion
  const std::string eventHeaderPassName_{""};
```
I think we should keep the default to be an empty string

Tracking/include/Tracking/Digitization/SiElectrodeDataCollection.h
```diff
@@ -25,12 +25,18 @@ class SiElectrodeDataCollection {
 
   std::map<int, int> getChargeMap() const;
 
-  std::map<int, SiElectrodeData> getCollection() const { return collection_; };
+  // std::map<int, SiElectrodeData> getCollection() const { return collection_;
+  // };
```
tvami:
```suggestion
```

tvami (changes requested):
Thanks Liam, this seems like a lot of work/time! I had some cleaning requests. And then there is a question about the header's pass name, which I think we should keep the ""

tvami:
the CI fails with
```
Configuration Error [Config] : Parameter 'tracker_hit_passname' does not exist in list of parameters.
```

Framework/include/Framework/Event.h
tomeichlersmith: Indeed, the pass name of the event header will be ignored since we never want to append a pass name to it.

https://github.com/LDMX-Software/ldmx-sw/blob/6aca7d804cc0f5fb555b4808fe10f01fc2dd41f7/Framework/include/Framework/Event.h#L181-L186

I would remove this member entirely and hardcode the empty string into the `nextEvent` source code.

Framework/src/Framework/Event.cxx
```diff
@@ -149,7 +149,8 @@ void Event::setInputTree(TTree* tree) {
 }
 
 bool Event::nextEvent() {
-  eventHeader_ = getObject<ldmx::EventHeader>(ldmx::EventHeader::BRANCH);
+  eventHeader_ = getObject<ldmx::EventHeader>(ldmx::EventHeader::BRANCH,
+                                              eventHeaderPassName_);
```
tomeichlersmith:
```suggestion
  eventHeader_ = getObject<ldmx::EventHeader>(ldmx::EventHeader::BRANCH, "");
```

tomeichlersmith (commented):
Just looking at Framework stuff

I think the overall goal of forcing a passname is a good idea and the work across all of ldmx-sw is appreciated. Thank you.

tvami (approved):
Looks good thanks again Liam!

tomeichlersmith (approved):
Thank you again :raised_hands:

PR #1713, opened 2025-05-20
Move prod image to use metadata-action

### LINKED ISSUE #1438: align image tagging with development image
In the development image, we use docker's metadata action to deduce the tags and labels for the image.

https://github.com/LDMX-Software/docker/blob/1c729e21b606ac9400e2e29c184a645e14c39338/.github/workflows/ci.yml#L36-L41

but here, we use some scripting I threw together since this action wasn't available at the time.

https://github.com/LDMX-Software/ldmx-sw/blob/001807987bd3a23444c2260d6d17118a8bb88a14/.github/workflows/build_production_image.yml#L44-L76

I'd like to update to docker's metadata action for the following reasons:
- The dev image does not put a `v` prefixing the image version while the pro image does. This is confusing and switching to the metadata action would resolve this.
- The pro image just inherits the labels from the dev image which, most confusingly, includes a version label which actually corresponds to the dev image version and not the ldmx-sw version. Switching to the metadata action would resolve this.
- The pro image doesn't have any arm builds. We could attach the image building runner at UMN to this repository as well and use it here to make sure the pro image also has arm builds.

- [x] update tag deduction to use docker's metadata action
- [ ] see about building an ARM-architecture production image - the production image build takes 30min which means the ARM-architecture build requiring emulation would take about 4.25hr (the ARM build takes about 8.5 times as long as the AMD build for the dev image) which is under the 6hr time limit for GitHub hosted runners, so I think we can start by doing both builds on GitHub hosted runners. Folks will have to wait a bit in order for the production image to be posted, but it'll still be posted in under a day.
- [x] use `just prep-version` before building so the compiled ldmx-sw has the right version

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1438

## Check List


.github/workflows/build_production_image.yml
```diff
@@ -42,38 +47,24 @@ jobs:
...
-      shell: bash
+      name: Docker meta
+      id: meta
+      uses: docker/metadata-action@v5
+      with:
+        # Default image name for push actions
+        images: ${{ github.event.inputs.repo || 'ldmx/pro' }}
+        # Tag strategies based on the event type
+        tags: |
+          just prep-version
```
tomeichlersmith: This is _not_ spawned into a shell, this is just providing configuration to the metadata-action. You need to
```yaml
    - run: just prep-version
```
below this and before the `build-push` action.

tomeichlersmith: I'm not sure if this will work out of the box though. `docker/build-push-action` will checkout a fresh copy of the repository if it doesn't have one yet, so we may need to `actions/checkout` above this as well.

tvami:
Cross referencing that this is the same strategy we will do with https://github.com/LDMX-Software/dark-brem-lib-gen/pull/28 and https://github.com/LDMX-Software/dev-build-context/pull/126

.github/workflows/build_production_image.yml
tvami: So now the order is
- actions/checkout
- docker/metadata-action
- just prep-version
- docker/build-push-action

Not sure how to test this besides merging and looking on dockerhub

tomeichlersmith (commented):
This looks good, we just need to decide what types of images we want to allow to be built.

The workflow is triggered on pushes to `trunk`, on tags, and on `workflow_dispatch` which allows users to ask for a specific branch to be built into the image.

https://github.com/LDMX-Software/ldmx-sw/blob/df24fa804c1540e50732260be8757c61a9c948c3/.github/workflows/build_production_image.yml#L4-L20

If we want to enable `workflow-dispatch`, we need to update the inputs to docker/metadata-action:
1. change `images` to be either `ldmx/pro` or `github.event.inputs.repo` if workflow_dispatch. Note: without additional scripting, this would allow any `workflow_dispatch` to push to `ldmx/pro` which is probably error prone.
2. Add something like `type=raw,tag=${{ github.event.inputs.tag }},enable=${{ github.event.name == "workflow_dispatch" }}` to the metadata-action `tags`

Another option is just to remove `workflow_dispatch` from the `on` section and have people focus on getting their updates into `trunk` in order to get an `ldmx/pro` build (or - for advanced folks - build it themselves locally).

I'm more in favor of this second option to be honest. I initially implemented the `workflow_dispatch` option for myself, but I don't believe anyone else has used it after looking into the CI history.
4.

tvami:
> I'm more in favor of this second option to be honest. I initially implemented the workflow_dispatch option for myself, but I don't believe anyone else has used it after looking into the CI history.

I agree, I did it in: https://github.com/LDMX-Software/ldmx-sw/pull/1713/commits/94ffcb8ad3a3cc4f090d0c1bdde41d1387b437e8

.github/workflows/build_production_image.yml
tomeichlersmith: I think merging and watching is a reasonable choice. We can check the `sha-XXXXX` tag and the build logs to see if the `VERSION` was correctly included in the build before making another ldmx-sw release.

tomeichlersmith (approved):

.github/workflows/build_production_image.yml
tvami: OK I'll cut a new release now then

Issue #1702, opened 2025-05-08
Update all filter defaults to 8 GeV values


bryngemark:
**Describe the bug**
Most of the process filter defaults in Biasing/filters.py still have 4 GeV defaults.

**To Reproduce**
Look at the code [here](https://github.com/LDMX-Software/ldmx-sw/blob/1b469b698636f4d2c11240662cefd67c84201cef/Biasing/python/filters.py) 

**Desired behavior**
8 GeV defaults (in most cases, just making thresholds a factor 2 larger).

tvami:
This is a dupplicate of  https://github.com/LDMX-Software/ldmx-sw/issues/1696 ?

bryngemark:
great, better to do it that way

PR #1715, opened 2025-05-22
Update filters to take gen energy for 8 GeV Update

### LINKED ISSUE #1696: Better handle filter thresholds (8 GeV vs 4 GeV)

Right now we have moved a lot of things to 8 GeV, but the filter thresholds are behind many times, e.g.
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Biasing/python/filters.py#L191
The user has to remember to change this at production, which leads to the situation like

Image

I think we should rewrite these so they take the generator as input and then all the thresholds are the function of the beam energy, i.e. this needs to be changed once only.

Change the defaults

**Additional context**

Design report

JYoo001:

Resolves #1696 
Also a solution to #1702 

## Check List

### DIFF
--- a/Biasing/python/filters.py
+++ b/Biasing/python/filters.py
@@ -113,7 +113,7 @@ class TargetENFilter(simcfg.UserAction) :
         Maximum energy recoil electron is allowed to have [MeV]
     """
 
-    def __init__(self,recoil_thresh = 5000.) :
+    def __init__(self,recoil_thresh) :
         super().__init__("target_en_process_filter","biasing::TargetENProcessFilter")
 
         from LDMX.Biasing import include
--- a/Biasing/python/filters.py
+++ b/Biasing/python/filters.py
@@ -29,7 +29,7 @@ class TargetBremFilter(simcfg.UserAction):
         Should we kill the recoil electron track for a worst case scenario?
     """
 
-    def __init__(self,recoil_max_p = 1500.,brem_min_e = 2500.) :
+    def __init__(self,recoil_max_p = 3000.,brem_min_e = 5000.) :
         super().__init__("target_brem_filter", "biasing::TargetBremFilter")
 
         from LDMX.Biasing import include
@@ -50,7 +50,7 @@ class NonFiducialFilter(simcfg.UserAction):
         If true, aborts fiducial events. Otherwise, the fiducial events will be only tagged
     """
 
-    def __init__(self,recoil_max_momentum = 1500., abort_fid_event = True) :
+    def __init__(self,recoil_max_momentum = 3000., abort_fid_event = True) :
         super().__init__("nonfiducial_filter", "biasing::NonFiducialFilter")
 
         from LDMX.Biasing import include
@@ -113,7 +113,7 @@ class TargetENFilter(simcfg.UserAction) :
         Maximum energy recoil electron is allowed to have [MeV]
     """
 
-    def __init__(self,recoil_thresh = 2500.) :
+    def __init__(self,recoil_thresh = 5000.) :
         super().__init__("target_en_process_filter","biasing::TargetENProcessFilter")
 
         from LDMX.Biasing import include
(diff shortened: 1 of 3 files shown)


tvami:
```
    File "config.py", line 9, in <module>
      mySim = ecal.photo_nuclear(det,gen.single_8gev_e_upstream_tagger())
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/runner/work/ldmx-sw/ldmx-sw/install/python/LDMX/Biasing/ecal.py", line 73, in photo_nuclear
      filters.TaggerVetoFilter(generator = generator),
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/runner/work/ldmx-sw/ldmx-sw/install/python/LDMX/Biasing/filters.py", line 197, in __init__
      self.threshold = thresh
                       ^^^^^^
  NameError: name 'thresh' is not defined
  ```
  @JYoo001 did you see this in the CI?

**Comment by JYoo001** [comment]
Yes, I'll go update it right now!

**Comment by JYoo001** [comment]
ignore for now, just rebasing

**Comment by JYoo001** [comment]
I've updated ecal.py, eat.py, and target.py to calculate the run parameters to give to filters.py. I've also updated the default values in filters.py, and passed in inputs in some of the other files that relied on the defaults. I've left the dark brem biasing untouched as well.

File: Biasing/python/filters.py
Comment (tvami) [inline Biasing/python/filters.py:1]: I'm wondering if in this file, we should just not have defaults at all. That will force the user to have the values passed wherever these are used, and that kinda ensures that we never end up in this situation of forgetting to overwrite certain settings. Like I think at this point Jihoon did add the explicit values in all occurrences, so why not just remove the defaults?

File: Biasing/python/filters.py [hunk 1]
```diff
@@ -29,7 +29,7 @@ class TargetBremFilter(simcfg.UserAction):
         Should we kill the recoil electron track for a worst case scenario?
     """
 
-    def __init__(self,recoil_max_p = 1500.,brem_min_e = 2500.) :
+    def __init__(self,recoil_max_p = 3000.,brem_min_e = 5000.) :
```
tvami:
```suggestion
    def __init__(self,recoil_max_p,brem_min_e) :
```

Biasing/python/filters.py
```diff
@@ -50,7 +50,7 @@ class NonFiducialFilter(simcfg.UserAction):
         If true, aborts fiducial events. Otherwise, the fiducial events will be only tagged
     """
 
-    def __init__(self,recoil_max_momentum = 1500., abort_fid_event = True) :
+    def __init__(self,recoil_max_momentum = 3000., abort_fid_event = True) :
```
tvami:
```suggestion
    def __init__(self,recoil_max_momentum, abort_fid_event = True) :
```

Biasing/python/filters.py
```diff
@@ -113,7 +113,7 @@ class TargetENFilter(simcfg.UserAction) :
         Maximum energy recoil electron is allowed to have [MeV]
     """
 
-    def __init__(self,recoil_thresh = 2500.) :
+    def __init__(self,recoil_thresh = 5000.) :
```
tvami:
```suggestion
    def __init__(self,recoil_thresh) :
```

Biasing/python/filters.py
```diff
@@ -188,7 +188,7 @@ class TaggerVetoFilter(simcfg.UserAction):
         Also veto events where the primary particle misses the tagger region
     """
     
-    def __init__(self,thresh=3800., reject_events_missing_tagger=True) :
+    def __init__(self,thresh=7600., reject_events_missing_tagger=True) :
```
tvami:
```suggestion
    def __init__(self,thresh, reject_events_missing_tagger=True) :
```

Biasing/python/filters.py
tomeichlersmith: I second this idea.

tomeichlersmith (approved):

tvami (approved):

PR #1731, opened 2025-06-19
Allow NTupling of both data and simulated HCalDigis

erik-wallin:
Spill number/timing information and data quality measures are not produced by the simulation, and are only present in data from the 2022 test beam. With these changes, NtuplizeHgcrocDigiCollection checks whether we are dealing with a simulation, and then fakes some timing information and says that the link quality is perfect. 

Now, the correct solution would be to actually simulate the spill structure in MC, and store the data quality measures good_bxheader and good_trailer in a proper class. But, since this is so specific to the 2022 test beam, let's not complicate thing upstream in the simulation.  

## Check List

Now you can make NTuples out of simulated digis: 
image

### DIFF
--- a/DQM/src/DQM/NtuplizeHgcrocDigiCollection.cxx
+++ b/DQM/src/DQM/NtuplizeHgcrocDigiCollection.cxx
@@ -80,7 +80,7 @@ void NtuplizeHgcrocDigiCollection::analyze(const framework::Event& event) {
 
   const ldmx::EventHeader* header =
       const_cast<framework::Event&>(event).getEventHeaderPtr();
-  bool is_simulation_ = not (header->isRealData());
+  bool is_simulation_ = not(header->isRealData());
 
   if (not is_simulation_) {
     if (already_aligned_) {
@@ -104,10 +104,10 @@ void NtuplizeHgcrocDigiCollection::analyze(const framework::Event& event) {
   std::vector<bool> good_trailer;
 
   if (not is_simulation_) {
-    good_bxheader = event.getCollection<bool>(
-        input_name_ + "GoodLinkHeader", input_pass_);
-    good_trailer = event.getCollection<bool>(
-        input_name_ + "GoodLinkTrailer", input_pass_);
+    good_bxheader =
+        event.getCollection<bool>(input_name_ + "GoodLinkHeader", input_pass_);
+    good_trailer =
+        event.getCollection<bool>(input_name_ + "GoodLinkTrailer", input_pass_);
   }
 
   auto const& digis{
--- a/DQM/src/DQM/NtuplizeHgcrocDigiCollection.cxx
+++ b/DQM/src/DQM/NtuplizeHgcrocDigiCollection.cxx
@@ -80,7 +80,7 @@ void NtuplizeHgcrocDigiCollection::analyze(const framework::Event& event) {
 
   const ldmx::EventHeader* header =
       const_cast<framework::Event&>(event).getEventHeaderPtr();
-  bool is_simulation_ = !(header->isRealData());
+  bool is_simulation_ = not (header->isRealData());
 
   if (not is_simulation_) {
     if (already_aligned_) {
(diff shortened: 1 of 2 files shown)


DQM/src/DQM/NtuplizeHgcrocDigiCollection.cxx
```diff
@@ -77,38 +77,65 @@ void NtuplizeHgcrocDigiCollection::analyze(const framework::Event& event) {
       hcal::HcalDetectorMap::CONDITIONS_OBJECT_NAME)};
 
   ldmxsw_event_ = event.getEventNumber();
-  if (already_aligned_) {
-    aligned_ = event.getObject<bool>(input_name_ + "Aligned", input_pass_);
+
+  const ldmx::EventHeader* header =
+      const_cast<framework::Event&>(event).getEventHeaderPtr();
+  bool is_simulation_ = !(header->isRealData());
```
tomeichlersmith: Not a member variable so it should not have the trailing underscore. (also change `!` to `not` since it compiles to the same code and I think it reads better but thats more of a personal preference)

```suggestion
  bool is_simulation = not (header->isRealData());
```

DQM/src/DQM/NtuplizeHgcrocDigiCollection.cxx
```diff
@@ -77,38 +77,65 @@ void NtuplizeHgcrocDigiCollection::analyze(const framework::Event& event) {
...
-  const auto& good_trailer{
-      event.getCollection<bool>(input_name_ + "GoodLinkTrailer", input_pass_)};
+  std::vector<bool> good_bxheader;
+  std::vector<bool> good_trailer;
+
+  if (not is_simulation_) {
+    good_bxheader = const_cast<framework::Event&>(event).getCollection<bool>(
+        input_name_ + "GoodLinkHeader", input_pass_);
+    good_trailer = const_cast<framework::Event&>(event).getCollection<bool>(
+        input_name_ + "GoodLinkTrailer", input_pass_);
```
tomeichlersmith: Is this `const_cast` necessary? I think you should be able to just `getCollection` and the returned `std::vector` will be copied.

The `const auto&` stuff is just there to avoid this copy on a per-event basis.

tomeichlersmith: Hell, you could get fancy with it and use `std::optional<const std::vector<bool>&>` but I don't know if that would avoid the copy or not tbh. I think it would be I'd also need to check.

tomeichlersmith (changes requested):
Looks good. I agree with you about not trying to emulate the messiness of the 2022 testbeam links. I just have some technical/format edits.

DQM/src/DQM/NtuplizeHgcrocDigiCollection.cxx
erik-wallin: Right, since .getCollection() is const:ed I don't need the const_cast here, like I do for the event header above. Fixed it.

tomeichlersmith: Can you just add a `const EventHeader& getEventHeader() const` to the Framework/Event?

I want to avoid `const_cast` as much as possible and I think it makes sense to have `const` access to the event header so analyzers can get any parameters stored within it. You would just copy this function:

https://github.com/LDMX-Software/ldmx-sw/blob/862b1421b32cb45f3080720394221cbdb22e23de/Framework/include/Framework/Event.h#L55-L59

but then add `const` before `EventHeader&` and after `()`. The compiler will prefer `const` access if possible and will only allow `const` access for analyzers who only have a `const Event&`.

Then your `isRealData` access would be
```
bool is_simulation = not event.getEventHeader().isRealData()
```

tomeichlersmith: I wasn't sure if this would work so I did a quick test locally and it compiled. I can push the commit if you want.

tvami (approved):
I have nothing else to add on top of Tom's comments

DQM/src/DQM/NtuplizeHgcrocDigiCollection.cxx
erik-wallin: Thanks @tomeichlersmith, feel free to push it

erik-wallin:
Thanks Tom, just wanted to confirm that things still look good after your last commit. You can merge this now

tomeichlersmith (approved):
Sorry! Totally forgot to click the button.

PR #1736, opened 2025-06-23
Re-Organized the SimCore Libraries

tomeichlersmith:

This does not directly address an issue, but it does make sure the `User*Information` is accessible to the Generators which is necessary for the GENIE generator. https://github.com/LDMX-Software/ldmx-sw/pull/1488#issuecomment-2993596099

Along the way, I also
- attempted to patch a new use-after-free warning that popped up after this re-ordering
- patched up Packing/test use of parameters
- patched up asan-ubsan compiler warning in Trigger with cleaner code :tada:

## Check List

~The basic CI tests and PR Validation shows the base functionality. I was worried about the use-after-free warning patch accidentally leaking memory so I did a short test run of the BertiniEventTopologyProcess with the config copied below. I cannot find any reference to G4DynamicParticle allocated from the Bertini cascade in the log [test-betp.log](https://github.com/user-attachments/files/20868721/test-betp.log), so I'm concluding the patch is successful.~

After discussing with @EinarElen on slack, we agreed that the `delete` call was correct and calling the allocator directly would just be more confusing, so I have instead remove clang-analyzer from operating on those two lines of code to avoid that warning turned error.

<details>
<summary>test-betp.py</summary>

```python
p = ldmxcfg.Process("test")
p.maxEvents = 10
p.logFrequency = 1
p.termLogLevel = 0
p.run = 9001
p.outputFiles = [ '/dev/null' ]
import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry
mySim.setDetector( 'ldmx-det-v14-8gev' , True )

from LDMX.SimCore import photonuclear_models as pn 
from LDMX.Biasing import particle_filter 
myModel = pn.BertiniSingleNeutronModel()
myModel.hard_particle_threshold=200. # Count particles with >= 200 MeV as "hard"
myModel.zmin = 0 # Apply the model for any nucleus
myModel.emin = 2500. # Apply the model for photonuclear reactions with > 2500 MeV photons
myModel.count_light_ions=True # Don't disregard deutrons, tritons, 3He, or 4He ions when counting hard particles 
# Change the default model to the single neutron model
mySim.photonuclear_model = myModel
myFilter = particle_filter.PhotoNuclearTopologyFilter.SingleNeutronFilter()
# Default values 
myFilter.hard_particle_threshold = 200. # MeV, use the same as for the model 
myFilter.count_light_ions = True # As above
# Add the filter at the end of the current list of user actions. 
mySim.actions.extend([myFilter])
# Get a pre-written generator
mySim.generators.append( gen.single_8gev_e_upstream_tagger() )
p.sequence.append( mySim )
```

<details>


tvami (approved):

SimCore/src/SimCore/PhotoNuclearModels/BertiniEventTopologyProcess.cxx
```diff
@@ -5,9 +5,14 @@ void BertiniEventTopologyProcess::cleanupSecondaries() {
...
   for (int i{0}; i < secondaries; ++i) {
+    // disable clang's use-after-free warning for this delete
+    // as far as we (@tomeichlersmith and @EinarElen) can tell,
+    // this is correct and clang is just getting confused since
+    // Geant4's inlined delete method is written in a wacky way
+#ifndef __clang_analyzer__
     auto secondary{theParticleChange.GetSecondary(i)->GetParticle()};
-    pDynamicParticleAllocator->FreeSingle(secondary);
-    pDynamicParticleAllocator->destroy(secondary);
+    delete secondary;
```
tvami, line 14: why do we create `secondary` just so we can delete it in the next line? :D (sorry for my ignorance)

tomeichlersmith: Its not creating, we are getting the _pointer_ and then deleting the _object_.

This function is called when we are disposing of an event that is not accepted by the filter and so we need to cleanup the particles that were created by the hadronic process. Usually, the particles created by the hadronic process would continue propagating through the rest of the sim like other particles created by other processes.

tvami: I see, thanks for the explanation (if you push again, maybe add this text as an in-line comment to clarify?)

PR #1738, opened 2025-06-24
Break apart Mip Tracking from EcalVetoProcessor

### LINKED ISSUE #1306: Break-Up Ecal Veto
_Edit_: I've given this more thought as I chatted with @tvami on slack.

Currently, the `EcalVetoProcessor` is very hefty. Moreover, a lot of the variables calculated by the veto processor can be used by other analyses. With this in mind, my proposal is to break up the current veto processor into different processors that create different event bus objects.

1. EcalHitStatistics - calculate various statistics of the hit collection that don't require other inputs. Things like energy-weighted average. These are relatively fast to calculate and don't require other inputs so could be just attached onto the end of the ECal reconstruction chain.
2. SingleElectronShowerFeatures - the shower-shape features for the single-electron BDT that require additional inputs. Mainly things that depend on knowledge of the incident electron direction like containment. These are not slow themselves, but would require track reconstruction within the Recoil and thus should be separated so that we can pre-select the data on simpler statistics from (1) before taking the time to do track reconstruction.
3. EcalMIPTracking - reconstruct straight and linear-regression tracks. These are helpful for other analyses (e.g. EaT), but could also be helpful for multi-electron anaylses. In addition, keeping it separate would allow us to avoid running it on already-veto'ed tracks (saving compute time) and enable the potential to develop an acts solution that can be swapped in-and-out as simple as a python config change.
4. SingleElectronTargetBDT - BDT for separting target-based dark brem from ecal PN. This is to separate it from other ML that could be focused on ecal-based dark brem (EaT) or another ML method (DNN) or other anlaysis channels (multi-electron). Instead of taking a single, partially-formed class as input, it would take in several event objects (pretty much all of the objects produced by the above processors) and then create a new "result" object (or set of simpler objects) that only contain the decision and other metrics (like the inference value).

Hopefully, breaking up the ecal veto in this way will make it more maintain-able. This will also be cause for some additions to the event model, but I can hold off from removing the `EcalVetoResult` object until we are comfortable with breaking on-disk backwards compatibility.

JYoo001:

Resolves #1306 

The linear regression stage of the Mip Variables takes up most of the time in the cutflow, and to save time, we've moved part of this to the new EcalMipProcessor

## Check List


Ecal/include/Ecal/EcalMipTrackingProcessor.h
```diff
@@ -0,0 +1,155 @@
...
+#include <iostream>
+#include <map>
+#include <memory>
+#include <utility>
+
+// ROOT (MIP tracking)
+#include "TDecompSVD.h"
+#include "TMatrixD.h"
+#include "TObject.h"
+#include "TVector3.h"
```
tvami: Can you check which of these could be removed? For example I'm pretty sure the ONNX runtime is not neeed

Ecal/include/Ecal/EcalMipTrackingProcessor.h
```diff
@@ -0,0 +1,155 @@
...
+ private:
+  int nevents_{0};
+  double processing_time_{0.};
+
+  std::map<std::string, double> profiling_map_;
+
+  double linreg_radius_{0};
+
+  int nEcalLayers_{0};
+  int nReadoutHits_{0};
```
tvami: make this snake case

Ecal/include/Ecal/EcalVetoProcessor.h
```diff
@@ -17,7 +18,7 @@
 #include "Tools/ONNXRuntime.h"
 
 // ROOT (MIP tracking)
-#include "TVector3.h"
+#include "Ecal/Event/EcalTrajectoryInfo.h"
```
tvami: please alpha order the includes and now that it's not ROOT include also remove the comment

tvami (changes requested):

Ecal/src/Ecal/EcalMipTrackingProcessor.cxx
```diff
@@ -0,0 +1,490 @@
...
+    for (auto &hitData : trackingHitList) {
+      TVector3 hitPos = hitData.pos;
+      TVector3 hitPrime = hitPos - origin;
+      if (hitPrime.Dot(gToe) <= 0) {
+        photon_territory_hits_++;
+      }
+    }
+    ldmx_log(trace) << "Photon territory hits: " << photon_territory_hits_;
+  } else {
+    photon_territory_hits_ = nReadoutHits_;
```
JYoo001: Should we add nReadoutHits_ to the EcalTrajectoryInfo object? This is only calculated in the VetoProcessor as of now

JYoo001: Ah nvm I'll just get it from EcalVetoResult

tvami: But that is in the EcalVetoResults already, no? So we could just read it from it directly

tvami:
Screenshot 2025-06-25 at 07 43 40

we need to fix the straight tracking

Ecal/include/Ecal/EcalVetoProcessor.h
```diff
@@ -156,15 +148,10 @@ class EcalVetoProcessor : public framework::Producer {
   float avgLayerHit_{0};
   float stdLayerHit_{0};
   float ecalBackEnergy_{0};
-  // MIP tracking
+  // MIP tracking (Keeping here for now for BDT variables)
   /// Number of "straight" tracks found in the event
```
tvami:
```suggestion
```

Ecal/include/Ecal/EcalVetoProcessor.h
```diff
@@ -156,15 +148,10 @@ class EcalVetoProcessor : public framework::Producer {
...
-  int nStraightTracks_{0};
-  /// Number of "linreg" tracks found in the event
-  int nLinregTracks_{0};
-  /// First ECal layer in which a hit is found near the photon
-  int firstNearPhLayer_{0};
-  /// Number of hits near the photon trajectory
-  int nNearPhHits_{0};
+  int n_straight_tracks_{0};
```
tvami:
```suggestion
```

tvami (changes requested):

.github/validation_samples/reduced_ldmx/config.py
```diff
@@ -46,6 +46,7 @@
 ecalVeto.num_ecal_layers = 6
 ecalVeto.beam_energy = 4000.
 ecalVeto.recoil_from_tracking = False
+ecalMip = ecal_vetos.EcalMipProcessor()
```
```suggestion
ecalMip = ecal_vetos.EcalMipProcessor()
ecalMip.num_ecal_layers = 6

```

tvami:
Perfect, no visible change!
Screenshot 2025-06-25 at 15 35 38

Also BDT for the bkg has no visible change. There is some change in the signal, but it's expected since we now dont have the MIP in the SegMIP. So all of this looks good, and we'll train a new BDT soon

tvami (approved):

Ecal/include/Ecal/EcalMipTrackingProcessor.h
```diff
@@ -0,0 +1,147 @@
...
+#include "TDecompSVD.h"
+#include "TMatrixD.h"
+#include "TVector3.h"
+
+namespace ecal {
+class EcalMipTrackingProcessor : public framework::Producer {
+ public:
+  typedef std::pair<ldmx::EcalID, float> CellEnergyPair;
+
+  typedef std::pair<float, float> XYCoords;
```
tomeichlersmith: No reason to repeat yourself, we can pull in the type from the event object.
```suggestion
using XYCoords = ldmx::XYCoords;
```

Ecal/include/Ecal/EcalMipTrackingProcessor.h
```diff
@@ -0,0 +1,147 @@
...
+  void configure(framework::config::Parameters& parameters) override;
+
+  void produce(framework::Event& event) override;
+
+  // MIP tracking:  Class for storing hit information for tracking in a
+  // convenient way
+  struct HitData {
+    int layer;
+    TVector3 pos;
+  };
```
tomeichlersmith: Similar to above

```suggestion
  using HitData = ldmx::HitData;
```

tomeichlersmith (changes requested):
Looks good! Only thing I noticed was some repetition we should remove to avoid confusion.

tomeichlersmith (approved):

tvami:
I'll wait to merge this since it changes gold, and we just cut one and so would be good to have a bit more PRs in

PR #1740, opened 2025-06-26
Optionally store full pulse shape information from the HGCROC emulation

erik-wallin:
Before producing digitized hits, the HgcrocEmulator emulates an analog pulse from the incoming simhits. This includes combining pulses from many different simhits into one pulse shape, as well as a crude simulation of the analog electronics on the chip. If it was possible to save truth information from this intermediate analog step of the emulation, we could:

- More easily debug and improve the HGCROC emulator, by not just looking at the digitized output data

- There is some time structure to our incoming simhits, especially in the HCal's long scintillator bars. It would be interesting to study the effect on the pulse shapes from this. 

- As a substitute for a complete simulation of charge injections and LED flashes, this would be useful for calibration. E.g. if you want to make an ADC-TOT cross-calibration based on simulated data, without looking into the condition table. Similar to what #1475 achieves.

This MR adds a class HgcrocPulseTruthCollection that stores the full pulse shape produced by the HgcrocEmulator, as well an (as of yet) barebones DQM analyzer that makes some simple plots with this information.

All of this can be enabled in the HcalDigiProducer by setting
`digi.savePulseTruthInfo = True` 
but it is false by default. 

## Check List
- [X ] I successfully compiled _ldmx-sw_ with my developments.
- [ ] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

Here's a plot from the DQM, showing the peak voltage in the pulse against the sum of ADCs in the digi:
vpeak_sumADC

## Notes

I put both the HgcrocPulseTruthCollection class and CompositePulse in Recon/Event/, but I suppose we could add a Tools/Event/ folder. Is that a good idea? I'm not very CMake savvy.

The HgcrocPulseTruth class doesn't have a lot of functionality yet, so right now it just looks like a wrapper around the CompositePulse. But one could imagine saving more interesting information here that isn't suitable to store in CompositePulse itself, so I wanted to open up that possibility.  

The DQM plots are not histograms, but TGraphs, so I have to write them to the output TFile manually. If no one is against it, could I add a _buildTGraph()_ function to the EventProcessor?

### DIFF
--- a/DQM/include/DQM/HgcrocPulseTruth.h
+++ b/DQM/include/DQM/HgcrocPulseTruth.h
@@ -32,9 +32,6 @@ class HgcrocPulseTruth : public framework::Analyzer {
   std::string input_digi_pass_;
   std::string input_truth_name_;
   std::string input_truth_pass_;
-
-  TGraph* vpeak_sumADC_graph_;
-  TGraph* vpeak_TOT_graph_;
 };
 
 }  // namespace dqm
--- a/DQM/include/DQM/HgcrocPulseTruth.h
+++ b/DQM/include/DQM/HgcrocPulseTruth.h
@@ -1,0 +1,43 @@
+#ifndef _DQM_HGCROCPULSETRUTH_H_
+#define _DQM_HGCROCPULSETRUTH_H_
+
+/*~~~~~~~~~~~~~~~*/
+/*   Framework   */
+/*~~~~~~~~~~~~~~~*/
+#include "Framework/Configure/Parameters.h"
+#include "Framework/Event.h"
+#include "Framework/EventProcessor.h"
+#include "Tools/AnalysisUtils.h"
+
+#include "TGraph.h"
+
+namespace dqm {
+
+class HgcrocPulseTruth : public framework::Analyzer {
+public:
+  HgcrocPulseTruth(const std::string& name, framework::Process& process)
+    : framework::Analyzer(name, process) {}
+
+  ~HgcrocPulseTruth() {}
+
+  void configure(framework::config::Parameters& parameters) override;
+
+  void onProcessStart() override;
+
+  void onProcessEnd() override;
+
+  void analyze(const framework::Event& event) override;
+
+private:
+  std::string input_digi_name_;
+  std::string input_digi_pass_;
+  std::string input_truth_name_;
+  std::string input_truth_pass_;
+
+  TGraph* vpeak_sumADC_graph_;
+  TGraph* vpeak_TOT_graph_;
+};
+
+} // namespace dqm
+
+#endif
--- a/DQM/include/DQM/HgcrocPulseTruth.h
+++ b/DQM/include/DQM/HgcrocPulseTruth.h
@@ -7,16 +7,15 @@
 #include "Framework/Configure/Parameters.h"
 #include "Framework/Event.h"
 #include "Framework/EventProcessor.h"
-#include "Tools/AnalysisUtils.h"
-
 #include "TGraph.h"
+#include "Tools/AnalysisUtils.h"
 
 namespace dqm {
 
 class HgcrocPulseTruth : public framework::Analyzer {
-public:
+ public:
   HgcrocPulseTruth(const std::string& name, framework::Process& process)
-    : framework::Analyzer(name, process) {}
+      : framework::Analyzer(name, process) {}
 
   ~HgcrocPulseTruth() {}
 
@@ -28,7 +27,7 @@ public:
 
   void analyze(const framework::Event& event) override;
 
-private:
+ private:
   std::string input_digi_name_;
   std::string input_digi_pass_;
   std::string input_truth_name_;
@@ -38,6 +37,6 @@ private:
   TGraph* vpeak_TOT_graph_;
 };
 
-} // namespace dqm
+}  // namespace dqm
 
 #endif
--- a/DQM/src/DQM/HgcrocPulseTruth.cxx
+++ b/DQM/src/DQM/HgcrocPulseTruth.cxx
@@ -12,30 +12,10 @@ void HgcrocPulseTruth::configure(framework::config::Parameters& ps) {
   input_truth_pass_ = ps.getParameter<std::string>("input_truth_pass");
 }
 
-void HgcrocPulseTruth::onProcessStart() {
-  getHistoDirectory();
 
-  vpeak_sumADC_graph_ = new TGraph();
-  vpeak_sumADC_graph_->SetMarkerSize(1);
-  vpeak_sumADC_graph_->SetMarkerStyle(5);
-  vpeak_sumADC_graph_->SetLineWidth(0.0);
-  vpeak_sumADC_graph_->SetTitle(
-      "All channels;Truth pulse peak voltage;Digi sum of ADC");
-  vpeak_sumADC_graph_->SetName("vpeak_sumADC_all_ch");
+void HgcrocPulseTruth::onProcessStart() {}
 
-  vpeak_TOT_graph_ = new TGraph();
-  vpeak_TOT_graph_->SetMarkerSize(1);
-  vpeak_TOT_graph_->SetMarkerStyle(5);
-  vpeak_TOT_graph_->SetLineWidth(0.0);
-  vpeak_TOT_graph_->SetTitle("All channels;Truth pulse peak voltage;Digi TOT");
-  vpeak_TOT_graph_->SetName("vpeak_TOT_all_ch");
-}
-
-void HgcrocPulseTruth::onProcessEnd() {
-  getHistoDirectory();
-  vpeak_sumADC_graph_->Write();
-  vpeak_TOT_graph_->Write();
-}
+void HgcrocPulseTruth::onProcessEnd() {}
 
 void HgcrocPulseTruth::analyze(const framework::Event& event) {
   auto digis{event.getObject<ldmx::HgcrocDigiCollection>(input_digi_name_,
@@ -55,9 +35,9 @@ void HgcrocPulseTruth::analyze(const framework::Event& event) {
           for (int i = 0; i < d.size(); i++) {
             sumADC += d.at(i).adc_t();
           }
-          vpeak_sumADC_graph_->AddPoint(vpeak, sumADC);
+          histograms_.fill("vpeak_sumADC", vpeak, sumADC);
         } else {
-          vpeak_TOT_graph_->AddPoint(vpeak, d.tot());
+          histograms_.fill("vpeak_TOT", vpeak, d.tot());
         }
         break;
       }
--- a/DQM/src/DQM/HgcrocPulseTruth.cxx
+++ b/DQM/src/DQM/HgcrocPulseTruth.cxx
@@ -19,7 +19,8 @@ void HgcrocPulseTruth::onProcessStart() {
   vpeak_sumADC_graph_->SetMarkerSize(1);
   vpeak_sumADC_graph_->SetMarkerStyle(5);
   vpeak_sumADC_graph_->SetLineWidth(0.0);
-  vpeak_sumADC_graph_->SetTitle("All channels;Truth pulse peak voltage;Digi sum of ADC");
+  vpeak_sumADC_graph_->SetTitle(
+      "All channels;Truth pulse peak voltage;Digi sum of ADC");
   vpeak_sumADC_graph_->SetName("vpeak_sumADC_all_ch");
 
   vpeak_TOT_graph_ = new TGraph();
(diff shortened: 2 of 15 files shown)


Recon/include/Recon/Event/CompositePulse.h
```diff
@@ -29,7 +29,7 @@ class CompositePulse {
 
   CompositePulse() = default;
 
-  virtual ~CompositePulse() {};
+  virtual ~CompositePulse(){};
```
tomeichlersmith:
```suggestion
  virtual ~CompositePulse() = default;
```

Recon/include/Recon/Event/HgcrocPulseTruth.h
```diff
@@ -1,42 +1,41 @@
...
-public:
-  HgcrocPulseTruth(unsigned int id, ldmx::CompositePulse p) : id_(id), compositePulse_(p) {};
+class HgcrocPulseTruth {
+ public:
+  HgcrocPulseTruth(unsigned int id, ldmx::CompositePulse p)
+      : id_(id), compositePulse_(p){};
 
-  HgcrocPulseTruth() {}; // Default constructor needed, because ROOT will call std::vector<HgcrocPulseTruth>::resize
+  HgcrocPulseTruth(){};  // Default constructor needed, because ROOT will call
+                         // std::vector<HgcrocPulseTruth>::resize
```
tomeichlersmith:
```suggestion
  /// default constructor needed for std::vector::resize
  HgcrocPulseTruth() = default;
```

Recon/include/Recon/Event/HgcrocPulseTruth.h
```diff
@@ -1,42 +1,41 @@
...
 
-  virtual ~HgcrocPulseTruth() {};
+  virtual ~HgcrocPulseTruth(){};
```
tomeichlersmith:
```suggestion
  virtual ~HgcrocPulseTruth() = default;
```

Recon/src/Recon/Event/CompositePulse.cxx
```diff
@@ -2,37 +2,38 @@
 
 ClassImp(ldmx::CompositePulse)
```
tomeichlersmith: adding a semi-colon after this macro (even though its not required) will help avoid confusing clang format.
```suggestion
ClassImp(ldmx::CompositePulse);
```

Recon/src/Recon/Event/HgcrocPulseTruth.cxx
```diff
@@ -2,29 +2,28 @@
 
 ClassImp(ldmx::HgcrocPulseTruth)
```
tomeichlersmith:
```suggestion
ClassImp(ldmx::HgcrocPulseTruth);
```

Tools/src/Tools/HgcrocEmulator.cxx
```diff
@@ -221,8 +221,8 @@ bool HgcrocEmulator::digitize(
     }  // TOT or ADC Mode
   }  // sampling baskets
 
-
-  if(savePulseTruthInfo_) pulseTruthColl_->push_back(ldmx::HgcrocPulseTruth(channelID, pulse));
+  if (savePulseTruthInfo_)
+    pulseTruthColl_->push_back(ldmx::HgcrocPulseTruth(channelID, pulse));
```
tomeichlersmith, line 225: So, I _think_ (haven't checked) that this vector of HgcrocPulseTruth never gets cleared. Am I wrong? If not, we will want to clear it at the start of each event so that there aren't pulses from previous events in the collection for a given event.

tomeichlersmith (changes requested):
I have some formatting stuff. The only technical piece is about clearing the `pulseTruthColl_` member of the HgcrocEmulator. I don't think it is being cleared between events right now which is a problem.

tomeichlersmith:
To answer your questions...

- I think your placement of HgcrocTruthPulse and CompositePulse makes sense
- I think wrapping CompositePulse also makes sense, maybe we could store the final samples in it?
- Personally, I'm opposed to having a `buildTGraph` option of EventProcessor mailny due to size issues. Having a TGraph with number of points equal to the number of events is pretty large and mostly duplicates information that is in the event tree. Taking a quick look at your DQM, while the CompositePulse does use TGraph, what you are doing could also be described as a 2D histogram to check for linearity. The [EcalDigiVerifier](https://github.com/LDMX-Software/ldmx-sw/blob/trunk/DQM/src/DQM/EcalDigiVerifier.cxx) has a couple of these 2D histograms comparing the simulated and reconstructed energy.

Tools/src/Tools/HgcrocEmulator.cxx
erik-wallin: The collection is declared inside the producer, so it's recreated/cleared for each event.

erik-wallin:
Thanks Tom. I opted for a TGraph because I wanted to make fits to these points, but you're right, I should just do that with the full information that is stored in the event tree anyway. I'll change them to histograms.

erik-wallin:
Uproot really dislikes deserializing the std::vector<std::pair<double, double>> in CompositePulse. 

```
import uproot4 as uproot
f = uproot.open('mu-Simslow_4.00GeV_3_ldmx-hcal-prototype-v2.0.root')
events = f['LDMX_Events/HcalPulseTruth_sim.compositePulse_.hits_'].arrays()
```
throws
```
  File "/home/ewallin/Programming/tot_digitization_python/loop.py", line 5, in <module>
    events = f['LDMX_Events/HcalPulseTruth_sim.compositePulse_.hits_'].arrays()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/behaviors/TBranch.py", line 825, in arrays
    return self.parent.arrays(
           ^^^^^^^^^^^^^^^^^^^
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/behaviors/TBranch.py", line 888, in arrays
    _ranges_or_baskets_to_arrays(
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/behaviors/TBranch.py", line 3198, in _ranges_or_baskets_to_arrays
    uproot.source.futures.delayed_raise(*obj)
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/source/futures.py", line 38, in delayed_raise
    raise exception_value.with_traceback(traceback)
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/behaviors/TBranch.py", line 3134, in basket_to_array
    basket_array = interpretation.basket_array(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/interpretation/objects.py", line 175, in basket_array
    output = self.basket_array_forth(
             ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/interpretation/objects.py", line 261, in basket_array_forth
    output = self._discover_forth(
             ^^^^^^^^^^^^^^^^^^^^^
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/interpretation/objects.py", line 421, in _discover_forth
    output[i] = self._model.read(
                ^^^^^^^^^^^^^^^^^
  File "/home/ewallin/Programming/tot_digitization_python/.local/lib/python3.11/site-packages/uproot/containers.py", line 727, in read
    raise NotImplementedError(
NotImplementedError: memberwise serialization of AsArray
in file mu-Simslow_4.00GeV_3_ldmx-hcal-prototype-v2.0.root
```

f.file.show_streamers() gives:
```
ldmx::CompositePulse (v1)
    hits_: vector<pair<double,double> > (TStreamerSTL)
    pulseFunc_: TF1 (TStreamerObject)
    gain_: double (TStreamerBasicType)
    pedestal_: double (TStreamerBasicType)

pair<double,double> (v1)
    first: double (TStreamerBasicType)
    second: double (TStreamerBasicType)
```
which doesn't look strange to my non-expert eyes.

tomeichlersmith:
What version of `uproot` are you using? I'm also fine with changing the definition of CompositePulse before we merge it into the event dictionary so that it is more `uproot` friendly. We could have the `std::pair` be a small `struct` that more clearly defines the two pieces that define a "hit" in the pulse.

erik-wallin:
```
ewallin@ludde:~/Programming/tot_digitization_python$ denv pip show uproot
Name: uproot
Version: 5.6.2
```
The latest version unfortunately. 

I'll experiment with this a bit. Maybe there are more classes from ldmx-sw that uproot fails to open? Not sure how often people read our root files this way.

tvami:
I think @EBerzin figured out how to deal with this in uproot for tracking stuff. I do everything in the ldmx c++ custom analyzer way where this is not a problem.

erik-wallin:
After some more unsuccessful attempts and some reverse engineering of uproot:  I'm not convinced you can have an STL container member variable where the elements are anything but a fundamental built-in type, if you want uproot to be able to deserialize it. No matter if the elements are a complicated class, a small simple struct, or just an std::pair, the branch will get this (undocumented) bit "kStreamedMemberWise" set and uproot can not read it. 

First I looked at how HgcrocDigiCollection solves this, but all its vector member variables only contain built-in types.

Then I found HgcrocTrigDigiCollection
```
typedef std::vector<HgcrocTrigDigi> HgcrocTrigDigiCollection;
```  
which clearly is a vector of some non-fundamental type, but uproot has no problem reading it? It's different somehow.

-----

Anyway, I can of course split the std::vector<std::pair<double,double>> in CompositePulse.h into two vectors, but it's more inconvenient to work with. 

I don't even want to use uproot for what I'm doing, I'm going to write a proper analyzer. But it makes us ask the question, do we want to guarantee that every class in ldmx-sw works well with uproot?

tomeichlersmith:
We've never made the guarantee that every class in ldmx-sw works well with uproot before, so I'm not worried about that. Thank you for debugging it to see if there was an easy solution, but I suspect that our Framework is configuring ROOT writing to do some slightly abnormal streaming that is not easily handled (https://github.com/scikit-hep/uproot5/issues/38).

erik-wallin:
Ok. If we're fine with this class not playing well with uproot, then this MR can be merged as is. Thanks for the help!

tomeichlersmith (approved):

DQM/include/DQM/HgcrocPulseTruth.h
```diff
@@ -0,0 +1,39 @@
...
+/*~~~~~~~~~~~~~~~*/
+#include "Framework/Configure/Parameters.h"
+#include "Framework/Event.h"
+#include "Framework/EventProcessor.h"
+#include "TGraph.h"
+#include "Tools/AnalysisUtils.h"
+
+namespace dqm {
+
+class HgcrocPulseTruth : public framework::Analyzer {
```
tvami: Can we call this class `HgcrocPulseTruthAnalyzer`? Or something that does not repeat the object name in Recon/include/Recon/Event/HgcrocPulseTruth.h

DQM/include/DQM/HgcrocPulseTruth.h
```diff
@@ -0,0 +1,39 @@
...
+ public:
+  HgcrocPulseTruth(const std::string& name, framework::Process& process)
+      : framework::Analyzer(name, process) {}
+
+  ~HgcrocPulseTruth() {}
```
tvami: default this

DQM/src/DQM/HgcrocPulseTruth.cxx
```diff
@@ -0,0 +1,49 @@
...
+void HgcrocPulseTruth::configure(framework::config::Parameters& ps) {
+  input_digi_name_ = ps.getParameter<std::string>("input_digi_name");
+  input_digi_pass_ = ps.getParameter<std::string>("input_digi_pass");
+  input_truth_name_ = ps.getParameter<std::string>("input_truth_name");
+  input_truth_pass_ = ps.getParameter<std::string>("input_truth_pass");
+}
+
+void HgcrocPulseTruth::onProcessStart() {}
+
+void HgcrocPulseTruth::onProcessEnd() {}
```
tvami: delete & then dont need to override in the header

DQM/include/DQM/HgcrocPulseTruth.h
erik-wallin: The constructor can not be defaulted here. There needs to be some function body after the `framework::Analyzer(name, process)` initialization list, even if it just a couple of empty curly-braces like here. 

The destructor could be defaulted though (by just removing that line).

tvami: I meant the destructor indeed. Tho I thought to default here & then remove in the cxx, but I guess it's fine like you did it.

tvami:
Let's run the tests & then I'll sign this

tvami (approved):

PR #1743, opened 2025-06-27
Add CI for GENIE target EN simulations

### LINKED ISSUE #1739: Add CI with GENIE

Now that https://github.com/LDMX-Software/ldmx-sw/pull/1488 is merged it's time to have a CI that tests it

Add a CI with GENIE config

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1739

## Check List

Ran on 10 events, it produces events

### DIFF
--- a/.github/validation_samples/target_genie/config.py
+++ b/.github/validation_samples/target_genie/config.py
@@ -42,7 +42,7 @@ p.sequence = [ mySim ]
 import os
 import sys
 
-p.maxEvents = int(os.environ['LDMX_NUM_EVENTS'])
+p.maxEvents = int(int(os.environ['LDMX_NUM_EVENTS']) * 0.7)
 p.run = int(os.environ['LDMX_RUN_NUMBER'])
 
 p.histogramFile = f'hist.root'
--- a/.github/validation_samples/target_genie/config.py
+++ b/.github/validation_samples/target_genie/config.py
@@ -1,0 +1,160 @@
+from LDMX.Framework import ldmxcfg
+p = ldmxcfg.Process('test')
+
+p.maxTriesPerEvent = 10000
+
+from LDMX.Biasing import ecal
+from LDMX.SimCore import generators as gen
+from LDMX.SimCore import simulator as sim
+det = 'ldmx-det-v14-8gev'
+mySim = sim.simulator('sim')
+mySim.setDetector(det, True)
+genie = gen.genie(name=f'genie_G18_02a_02_11b',
+                        energy = 8.0,
+                        targets = [ 1000741820, 1000741830, 1000741840, 1000741860 ],
+                        target_thickness = 0.3504,
+                        abundances = [ 0.2650, 0.1431, 0.3064, 0.2843 ],
+                        time = 0.0,
+                        position = [0.,0.,0.],
+                        beam_size = [ 20., 80. ],
+                        direction = [0.,0.,1.],
+                        tune='G18_02a_02_11b',
+                        spline_file=f'gxspl_emode_GENIE_v3_04_00.xml',
+                        message_threshold_file=f'Messenger_ErrorOnly.xml')
+
+mySim.generators = [ genie ]
+
+from LDMX.SimCore import genie_reweight
+genie_rw = genie_reweight.GenieReweightProducer(name='genie_reweight')
+genie_rw.hepmc3CollName = "SimHepMC3Events"
+genie_rw.hepmc3PassName = ""
+genie_rw.var_types = ["GENIE_INukeTwkDial_MFP_pi","GENIE_INukeTwkDial_MFP_N"]
+genie_rw.verbosity = 0
+
+
+p.sequence.append(genie_rw)
+
+p.sequence = [ mySim ]
+
+##################################################################
+# Below should be the same for all sim scenarios
+
+import os
+import sys
+
+p.maxEvents = int(os.environ['LDMX_NUM_EVENTS'])
+p.run = int(os.environ['LDMX_RUN_NUMBER'])
+
+p.histogramFile = f'hist.root'
+p.outputFiles = [f'events.root']
+
+# Load the full tracking sequance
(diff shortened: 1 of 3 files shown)


tomeichlersmith (commented):
I love the idea of including a GENIE CI job _and_ I am nervous about including such large files again. How often will these XML and ROOT files change?

tvami:
I did worry about that too.
 * the XML file includes cross section information, so I think that's only expected to change when we or DUNE or some other experiment makes much better xsection measurements, right @wesketchum 
 * the ROOT file I actually have no idea where it's coming from, but also it's needed for the trigger and not for GENIE per se, maybe @therwig can chime in about that processor?

tomeichlersmith:
- For the XML files, that makes sense. They are also text which can maybe be helped if we are only changing a few xsec numbers at a time.
- For the ROOT file, I actually don't see it referenced in the config at all, so is it actually being used? Maybe its nice to apply these production weights when doing physics analyses but do we need to do this weighting when we are just testing that it runs like expected in CI?

.github/validation_samples/target_genie/config.py
```diff
@@ -0,0 +1,151 @@
...
+count.input_pass_name = ''
+
+en_trigger = [
+        ecal_trig_digi.EcalTrigPrimDigiProducer(),
+        hcal_trig_digi.HcalTrigPrimDigiProducer(),
+        trigger_energy_sums.EcalTPSelector(),
+        trigger_energy_sums.TrigEcalEnergySum(),
+        trigger_energy_sums.TrigHcalEnergySum(),
+        trigger_energy_sums.TrigEcalClusterProducer(),
+        trigger_energy_sums.TrigElectronProducer('propagationMap.root'),
```
JYoo001, line 103: I'm not sure if this is what you mean, but this is where the propagationMap is used

therwig:
The root file is needed for the calculation of the electron PT at trigger level. Basically, it maps out the expected trajectory of a charge particle using the magnetic field map information. It is just a single histogram, so it should be pretty lightweight, and never change for a given detector geometry.
I don’t have a strong opinion about whether or not to upload it, or find another solution, but just wanted to chime in to explain the purpose here. In principle, we might not need to run this particular trigger processor for a CI test, or maybe though it would be nice to have.

tvami:
Thanks @therwig ! The ROOT file is really small, 130 KB

tvami:
>  CI test, or maybe though it would be nice to have.

Yes in that sense this CI would serve to test all the trigger developments, and also the PF reco is included here (which is not included in other CIs)

therwig:
Agreed - that would be great!

tvami:
```
Error in <TFile::TFile>: file /home/runner/work/ldmx-sw/ldmx-sw/ci-data/target_genie/gold.root does not exist
    File "/home/runner/work/ldmx-sw/ldmx-sw/./.github/actions/validate/compare.py", line 209, in <module>
      compare(sys.argv[1], sys.argv[2], sys.argv[3], sys.argv[4])
    File "/home/runner/work/ldmx-sw/ldmx-sw/./.github/actions/validate/compare.py", line 153, in compare
      gold = HistogramFile(gold_f,gold_label,ROOT.kRed ,False )
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/runner/work/ldmx-sw/ldmx-sw/./.github/actions/validate/compare.py", line 76, in __init__
      self.__file = ROOT.TFile.Open(f)
                    ^^^^^^^^^^^^^^^^^^
    File "/usr/local/lib/ROOT/_pythonization/_tfile.py", line 104, in _TFileOpen
      raise OSError('Failed to open file {}'.format(str(args[0])))
  OSError: Failed to open file /home/runner/work/ldmx-sw/ldmx-sw/ci-data/target_genie/gold.root
```

@tomeichlersmith what's the plan here? Do I add a `histo.root` in the `ci-data` repo?

.github/validation_samples/target_genie/config.py
tomeichlersmith: that is what I meant, I guess I was just reading too quickly

tomeichlersmith:
The next release should add the new sample's gold to ci-data for you, so I don't think its a problem. If that doesn't work, I'd consider it a bug with the ci-data update workflow. In that case, we'll add them manually and then make an issue.

tomeichlersmith (approved):

wesketchum:
Right, so for the XML file, as Tamas said it's holding xsec spline information, and will change only if we add new models, new targets, or have an update in GENIE. That should not happen _that_ often, and for purposes of a CI test, don't have to be changed often at all.

But, I'll ask: is there a way that we can do some additional command as part of the CI? E.g. we could git clone a repo that had these spline files in it. I suppose we could do that directly from the config.py ... is there a suggestion on if/how to do something like this? That may be the most robust way.

tomeichlersmith:
Yes, if the `init.sh` script exists within the validation sample directory, that script will be run before the `config.py` is run.

For example, the signal sample generates the dark brem event library with MadGraph before using it in the config: https://github.com/LDMX-Software/ldmx-sw/tree/trunk/.github/validation_samples/signal

tvami:
OK so I'll upload the splines to ci-data and think about how to pull it in when needed using the sh approach.
It also takes 140 min to run his, I'll scale the events to have it at 100 - 120 min like the other CIs

tomeichlersmith:
The ci-data repo specifically is already cloned by the workflows for the golden files so you probably don't need to add any `init.sh` and instead just update the paths in the config.

therwig (approved):

tvami:
Added the GENIE files to the ci-data in https://github.com/LDMX-Software/ci-data/pull/1

tvami:
OK it finished in 95 min, I think that's good.
In the end I did the init bc the ci-data checkout only happens at comparison time which is after the running.

PR #1753, opened 2025-07-02
Add mip target CIs

tvami:

Merged the MIP tracking PR before merging the new CIs, so it was missing, now it's added


tvami:
K it works nicely

PR #1767, opened 2025-07-11
Put logging messages into logger format: Part 1

### LINKED ISSUE #1319: Put SimCore logging messages into logger format

tvami:

This is part 1 of https://github.com/LDMX-Software/ldmx-sw/issues/1319

## Check List

Code compiles, should lead to no changes in the logs

### DIFF
--- a/Biasing/src/Biasing/EcalDarkBremFilter.cxx
+++ b/Biasing/src/Biasing/EcalDarkBremFilter.cxx
@@ -9,11 +9,6 @@
 
 #include "Biasing/EcalDarkBremFilter.h"
 
-#include "G4DarkBreM/G4APrime.h"              //checking if particles match A'
-#include "G4DarkBreM/G4DarkBremsstrahlung.h"  //checking for dark brem secondaries
-#include "G4LogicalVolumeStore.hh"            //for the store
-#include "SimCore/G4User/UserTrackInformation.h"  //make sure A' is saved
-
 namespace biasing {
 
 EcalDarkBremFilter::EcalDarkBremFilter(
@@ -38,21 +33,14 @@ EcalDarkBremFilter::EcalDarkBremFilter(
   }
 
   if (G4RunManager::GetRunManager()->GetVerboseLevel() > 0) {
-    std::cout << "[ EcalDarkBremFilter ]: " << "Looking for A' in: ";
-    for (auto const& volume : volumes_) std::cout << volume->GetName() << ", ";
-    std::cout << std::endl;
+    ldmx_log(trace) << "Looking for A' in: ";
+    for (auto const& volume : volumes_) {
+      ldmx_log(trace) << "\t" << volume->GetName() << ", ";
+    }
   }
 }
 
 void EcalDarkBremFilter::BeginOfEventAction(const G4Event*) {
-  /* Debug message
-  std::cout << "[ EcalDarkBremFilter ]: "
-      << "(" <<
-  G4EventManager::GetEventManager()->GetConstCurrentEvent()->GetEventID() << ")
-  "
-      << "Beginning event."
-      << std::endl;
-  */
   foundAp_ = false;
   return;
 }
@@ -61,12 +49,9 @@ G4ClassificationOfNewTrack EcalDarkBremFilter::ClassifyNewTrack(
     const G4Track* aTrack, const G4ClassificationOfNewTrack& cl) {
   if (aTrack->GetParticleDefinition() == G4APrime::APrime()) {
     // there is an A'! Yay!
-    /* Debug message
-    std::cout << "[ EcalDarkBremFilter ]: "
-              << "Found A', still need to check if it originated in requested
-    volume."
-              << std::endl;
-    */
+    ldmx_log(trace) << "Found A', still need to check if it originated in "
+                       "requested volume.";
+
     if (not foundAp_ and aTrack->GetTotalEnergy() > threshold_) {
       // The A' is the first one created in this event and is above the energy
       // threshold
@@ -88,12 +73,9 @@ void EcalDarkBremFilter::NewStage() {
 }
 
 void EcalDarkBremFilter::PostUserTrackingAction(const G4Track* track) {
-  /* Check that generational stacking is working
-  std::cout << "[ EcalDarkBremFilter ]:"
-      << track->GetTrackID() << " " <<
-  track->GetParticleDefinition()->GetPDGEncoding()
-      << std::endl;
-  */
+  // Check that generational stacking is working
+  ldmx_log(trace) << track->GetTrackID() << " "
+                  << track->GetParticleDefinition()->GetPDGEncoding();
 
   const G4VProcess* creator = track->GetCreatorProcess();
   if (creator and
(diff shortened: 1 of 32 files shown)


tvami:
All checks passed!

Hcal/include/Hcal/HcalDoubleEndRecProducer.h
```diff
@@ -31,13 +40,7 @@ class HcalDoubleEndRecProducer : public framework::Producer {
...
   double clock_cycle_;
 
- public:
-  HcalDoubleEndRecProducer(const std::string& n, framework::Process& p)
-      : Producer(n, p) {}
-
-  virtual ~HcalDoubleEndRecProducer() = default;
-  void configure(framework::config::Parameters& p) override;
-  void produce(framework::Event& event) override;
+  enableLogging("HcalDoubleEndRecProducer")
```
tomeichlersmith: You do not need to `enableLogging` in event processors. The base class already creates `theLog_`.

Biasing/src/Biasing/EcalDarkBremFilter.cxx
```diff
@@ -130,11 +112,11 @@ bool EcalDarkBremFilter::inDesiredVolume(const G4Track* track) const {
 
 void EcalDarkBremFilter::AbortEvent(const std::string& reason) const {
   if (G4RunManager::GetRunManager()->GetVerboseLevel() > 1) {
```
tomeichlersmith: I think we should remove these Geant4 verbose level guards now that the messages are in the logging. I want to avoid having to do two things in order to see a message.

tomeichlersmith (changes requested):

Hcal/include/Hcal/HcalDoubleEndRecProducer.h
tvami: done in https://github.com/LDMX-Software/ldmx-sw/pull/1767/commits/6b869b0b771b66170058a7f34f31c68e7f24fa95

Biasing/src/Biasing/EcalDarkBremFilter.cxx
tvami: I agree, done in done in https://github.com/LDMX-Software/ldmx-sw/pull/1767/commits/6b869b0b771b66170058a7f34f31c68e7f24fa95

Biasing/src/Biasing/EcalDarkBremFilter.cxx
```diff
@@ -38,21 +33,14 @@ EcalDarkBremFilter::EcalDarkBremFilter(
   }
 
   if (G4RunManager::GetRunManager()->GetVerboseLevel() > 0) {
```
tomeichlersmith: Another `GetVerboseLevel` guard

SimCore/src/SimCore/ReSimulator.cxx
```diff
@@ -31,22 +31,21 @@ void ReSimulator::produce(framework::Event& event) {
   const auto eventNumber{eventHeader.getEventNumber()};
   if (skip(event)) {
     if (verbosity_ > 1) {
```
tomeichlersmith: another verbosity guard

SimCore/src/SimCore/ReSimulator.cxx
```diff
@@ -31,22 +31,21 @@ void ReSimulator::produce(framework::Event& event) {
...
-      std::cout << "Skipping event: " << eventNumber
-                << " since it wasn't part of the requested events..."
-                << std::endl;
+      ldmx_log(trace) << "Skipping event: " << eventNumber
+                      << " since it wasn't part of the requested events...";
     }
     this->abortEvent();  // get out of processors loop
     return;
   }
   if (verbosity_ > 0) {
```
tomeichlersmith: verbosity guard

SimCore/src/SimCore/ReSimulator.cxx
```diff
@@ -31,22 +31,21 @@ void ReSimulator::produce(framework::Event& event) {
...
-    std::cout << "Resimulating " << eventNumber << std::endl;
+    ldmx_log(trace) << "Resimulating " << eventNumber;
   }
 
   std::istringstream iss(eventHeader.getStringParameter("eventSeed"));
   G4Random::restoreFullState(iss);
   runManager_->ProcessOneEvent(eventNumber);
   if (verbosity_ > 1) {
```
tomeichlersmith: verbosity guard

tomeichlersmith (changes requested):
Found some more `verbose` guards that can be removed. If this is too much, we could also just put it into another issue, but I think its applicable to this move to the logger.

tvami:
I did those, but kinda lead deeper, I guess there wont be too much for Part 2, let's hope it works: I wasnt sure about the case when we have the SimulationBase class which then has derived classes

tvami:
Now I get a lot of msg like
```
G4Material WARNING: duplicate name of material G4_AIR
G4GDML: Reading solids...
G4GDML: Reading solids...
G4GDML: Reading solids...
G4GDML: Reading solids...
G4GDML: Reading solids...
G4GDML: Reading structure...
G4GDML: Reading structure...
G4GDML: Reading structure...
G4GDML: Reading structure...
G4GDML: Reading structure...
G4GDML: Reading setup...
```

Some of this is pretty useful, like the `G4_AIR` but it would be good to silence the rest of them

tvami:
Oh maybe it's here https://github.com/LDMX-Software/ldmx-sw/blob/3283af5fa8f873bce2c7a84eddec280e038b6f8f/SimCore/src/SimCore/SimulatorBase.cxx#L202

tomeichlersmith:
We can update [G4Session](https://github.com/LDMX-Software/ldmx-sw/blob/trunk/SimCore/src/SimCore/G4Session.cxx) which is given to Geant4 in order to decide how the messages are routed. This configuration is done in 

https://github.com/LDMX-Software/ldmx-sw/blob/3283af5fa8f873bce2c7a84eddec280e038b6f8f/SimCore/src/SimCore/SimulatorBase.cxx#L150-L163

which may need to be called earlier so that our updated G4Session is used for all G4 messages.
Currently, the G4Session just forwards messages either to nowhere (`BatchSession`) or into a file `LoggedSession`. We could probably drop both of those and just always forward messages to `trace` or `debug` and have our implementation of `G4Session` hold a log whose name is `Geant4` so that our logs show that its a message from Geant4.

tvami:
OK so I went back so the GEANT4 msgs are not printed out by default, so now the changes in the logs should really just be moving some stuff into the logging system that originally was there just as a cout.

Let's do the G4Session in the Part 2 of this issue

tomeichlersmith (approved):

JYoo001 (approved):

tvami:
The diffs look as expected

PR #1783, opened 2025-07-22
Move from TVector3 to XYZVector

### LINKED ISSUE #1780: Update deprecated `TVector3` to `XYZVector`

TVector is a legacy ROOT class
https://root.cern.ch/doc/master/classTVector3.html

However, there are 18 occurances of
https://github.com/search?q=repo%3ALDMX-Software%2Fldmx-sw%20TVector3&type=code

Move them to `XYZVector`

No alternatives, `TVector3`  will be not supported at one point

**Additional context**

https://github.com/LDMX-Software/ldmx-sw/pull/1778#pullrequestreview-3038322231

JYoo001:

Resolves #1780 

ROOT's TVector3 class is outdated and should be moved to the newer XYZVector. Some functionality from the TVector3 is no longer used, so I've replaced them with the XYZVector equivalent when needed.

## Check List


Ecal/src/Ecal/EcalMipTrackingProcessor.cxx
```diff
@@ -371,18 +373,26 @@ void EcalMipTrackingProcessor::produce(framework::Event &event) {
...
+
+        // Compute mean in an indexable array
+        double mean_arr[3] = {(p0.X() + p1.X() + p2.X()) / 3.0,
+                              (p0.Y() + p1.Y() + p2.Y()) / 3.0,
+                              (p0.Z() + p1.Z() + p2.Z()) / 3.0};
+
+        // Put back into h_mean (XYZVector)
+        h_mean.SetX(mean_arr[0]);
+        h_mean.SetY(mean_arr[1]);
+        h_mean.SetZ(mean_arr[2]);
```
tomeichlersmith: I feel like we should be able to do the mean within the vector as well.

Does the following not work?
```
ROOT::Math::XYZVector h_mean = (p0 + p1 + p2)/3;
```

Ecal/include/Ecal/EcalMipTrackingProcessor.h
```diff
@@ -83,7 +83,8 @@ class EcalMipTrackingProcessor : public framework::Producer {
    * @param[in] w2 A second, distinct point on line w
    * @returns Closest distance of approach of lines u and v
    */
-  float distTwoLines(TVector3 v1, TVector3 v2, TVector3 w1, TVector3 w2);
+  float distTwoLines(ROOT::Math::XYZVector v1, ROOT::Math::XYZVector v2,
+                     ROOT::Math::XYZVector w1, ROOT::Math::XYZVector w2);
```
tomeichlersmith: Since this function (and `distPtToLine`) are copied between this processor and the `EcalVetoProcessor`, could you move them into their own header? (Similar to `Ecal/include/Ecal/TrackPropagator.h`)

tomeichlersmith (changes requested):

Ecal/include/Ecal/EcalMipTrackingProcessor.h
tvami: I was actually thinking about having the `Ecal/include/Ecal/TrackPropagator.h` called as a `Ecal/include/Ecal/EcalHelper.h` and host more functions there, what do you think about that @tomeichlersmith ?

EventDisplay/include/EventDisplay/EveShapeDrawer.h
```diff
@@ -16,7 +16,7 @@
 #include "TGeoMatrix.h"
 #include "TGeoShape.h"
 #include "TGeoTube.h"
-#include "TVector3.h"
+#include "TVector3D.h"
```
tvami: I think this missed the Math, right?

tomeichlersmith: This isn't even compiled. Its the old-old TEve-based event display that I abandoned when developing the v12 geometry. We could delete these files to avoid confusion and direct people to use and contribute to https://github.com/LDMX-Software/event-vis

Ecal/include/Ecal/EcalMipTrackingProcessor.h
tomeichlersmith: Yea I think that makes sense. If that header gets too big we can break it up, but for now it just has these few geometry/utility functions.

EventDisplay/include/EventDisplay/EveShapeDrawer.h
tvami: Should the whole `EventDisplay` directory be deleted? (Probably another PR tho)

tomeichlersmith: So I think the `EventDisplay` directory should instead be the processors that create the JSON that could be fed into the pheonix-based event-vis repo linked above. There's also some other Python/Matplotlib vis tools that are in the `EventDisplay/python` directory that I think could be kept around.

tvami:
@JYoo001 I think you need to do a `git rm Ecal/TrackPropagator.h` too (and push after)

Ecal/src/Ecal/EcalMipTrackingProcessor.cxx
JYoo001: Yes, it does! I was just trying to follow the previous as closely as possible. I've made the changes!

tvami (approved):
Looks good, thanks for the quick turnaround on this

tomeichlersmith (approved):

PR #1789, opened 2025-07-30
Move Validation module to use hist package, rename to ComparePlots

### LINKED ISSUE #1729: Add error bars in the plots for Validation module

We went thru this already in the CI comparison plots, we should show error bars on the histograms that we are comparing. 

Add histo error bars in the Validation module

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1729

## Check List


tomeichlersmith (commented):
I pushed a couple patches I had time to do today.

I think transitioning to `hist` motivates a deeper refactoring since `hist` provides more utilities not only for plotting, but also filling and manipulating histograms. My general goal would be to move all plotting into the `Differ` class. The `File` would be focused on getting a histogram from an input file whether that is defining and filling a histogram using an event file or loading a histogram into memory from a ROOT file. The plotting in `Differ` could then also have callbacks for manipulating the histogram if desired.

- [x] load `hist.Hist` from ROOT using `uproot` (calling `to_hist()`). This is nice because then it retrieves the axis tick labels for us and we don't have to re-set them at plotting time.
- [x] update minimum Python version so there isn't a ceiling on the numpy version
- [x] move plotting out of `File` and into `Differ`
- [x] remove histogram filling entirely? Does anybody use that? Frankly, I think filling histograms is a different beast and so I think it could make sense to have this python area just be making comparison plots from pre-filled histograms.
- [x] check visuals for a wider variety of plots
- [x] remove packaging - I think it is more ergonomic to instead run the files "from source" instead of building a package. Still would require a `pip install` step to ensure dependencies are installed, but would remove the `setup.py` and sub-directory nonsense in favor of specifying the directory location after `python3 -m` on the command line. Opens door to using `uv` as project manager in future.

---

On a separate note, should we consider just moving this python plotting/comparison tool outside of ldmx-sw? It is entirely Python and there are better Python environment management tools than the `denv`+container image nonsense I cooked up (e.g. [uv](https://docs.astral.sh/uv/) lets us pin Python versions and package versions across machines and manages the Python venv for us so we don't have to worry about remembering to install the package or activate the venv). I'm just thinking that this code can/should evolve at a different rate than ldmx-sw C++, so it can/should reside in a different repository. One could argue that we want the DQM histograms that are filled in ldmx-sw also be plotted here, but my counter-point is we could just have the code scan the histograms in the ROOT file(s) provided like the current CI PyROOT script does.

tvami:
> Remove histogram filling entirely? 

I think we should be good to do this

Validation/pyproject.toml
```diff
@@ -4,13 +4,10 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "Validation"
-version = "0.0.1"
+version = "0.0.2"
 description = "LDMX Data Validation"
 readme = "README.md"
-requires-python = ">=3.7"
+requires-python = ">=3.9"
```
tvami: this didnt fix it, at least I see 
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
numba 0.61.2 requires numpy<2.3,>=1.24, but you have numpy 2.3.2 which is incompatible.
```

tvami:
I run a test on the CLUE DQM:
Screenshot 2025-08-03 at 17 05 05


Screenshot 2025-08-03 at 17 05 18


Screenshot 2025-08-03 at 17 05 30


Screenshot 2025-08-03 at 17 05 56

tvami:
>  check visuals for a wider variety of plots

I checked this, as this seems to have some diversity in the plot (e.g. alphanum labels, small binnings, big bins, etc)

Validation/pyproject.toml
tomeichlersmith: What is your setup for using this module? Are you using a python venv or the ldmx-sw dev container?

I kinda want to not use the dev container since it evolves at C++ pace instead of at Python pace (which is usually faster).

tvami: In the dev container. I agree that having a max version number is not great for the further future. But I prose the following: let's have this compatible with the dev container upto the point where we have it under ldmx-sw, and then later if we move it out of it, this issue will be solved with the min requirement on the python version

tomeichlersmith: Makes sense to me :+1:

Validation/src/Validate/hcal.py
```diff
@@ -39,10 +39,6 @@ def dqm(d: Differ, out_dir=None):
     d.plot1d('HcalInefficiencyAnalyzer/HcalInefficiencyAnalyzer_efficiency',
              'Hcal part involved in veto',
              'Efficiency',
-             tick_labels=['', 'Back', 'Top', 'Bottom',
-                          'Right', 'Left', 'Any',
-                          'Both', 'Back only', 'Side only',
-                          'Neither', ''],
```
tomeichlersmith, line 45: @tvami Should we remove the `tick_labels` option? As shown by this example plot, the `hist`+`uproot` solution is able to load the axis labels from ROOT file, so we don't need to add labels if they are defined when filling (which they should be to avoid confusion).

Validation/pyproject.toml
tvami: Do you still want to do the "remove packaging " in this PR?

Validation/src/Validate/hcal.py
tvami: Oh but in the example I did have the labels: https://github.com/LDMX-Software/ldmx-sw/blob/4de10d49d7bd2e81b88dfac69b924466fcb2ed22/Validation/src/Validate/ecal.py#L123-L127

Let me try to see what happens when I remove that

Validation/pyproject.toml
tomeichlersmith: I would like to, but it could also be left for future work when/if we move this to somewhere outside ldmx-sw.

tvami: OK, good for me if you do it here, thanks!

tvami:
Seems to work nicely with removing the labels, I'll push it in second
Screenshot 2025-08-04 at 08 56 47

Validation/src/Validate/hcal.py
tvami: Done in https://github.com/LDMX-Software/ldmx-sw/pull/1789/commits/e092c4a15f853e04697c24eeac986c96658ff850

tomeichlersmith (approved):

Validation/README.md
```diff
@@ -18,46 +21,27 @@ from other python packages that may be on your system.
...
 ## Usage
 
 ### CLI
 The Validation module is constructed to do some common tasks quickly on the command line.
-Printing out its help message shows how to run it and gives you the details on what
-parameters to provide.
+Running it can be accessed via `just` as well
 ```
-denv python3 -m Validate -h
+just validate -h
```
tvami: now that you made it a "just" command, I kinda wonder if we should rename this... it doesnt "validate" anything per se, right? can we call it like `just compare-plots`? or `overlay-plots` or `hists` instead of `plots`

tomeichlersmith: I like `compare-plots`, I'll change it to that.

tvami: thanks!

tomeichlersmith (approved):

tvami:
OK great, I'll test it one more time with the new just commands

tvami:
OK works nicely!

PR #1806, opened 2025-08-11
Implement Birks law in the TS SD

### LINKED ISSUE #1088: Implement Birks' law for the trigger scintillator in framework
We need to account for Birks' law in the TS light yield calculation, especially for LYSO. Looking at how this is implemented for HCal, it seems to be a rather straightforward addition. 

Our implementation could be based on [what's done in the HCalSD.cxx](https://github.com/LDMX-Software/SimCore/blob/0656170b61f39dd972c6b8fd01fd9707391001d5/src/SimCore/SDs/HcalSD.cxx#L31-L88) 
  
In addition we should have a configurable flag `isLYSO_`/`isPlastic_` so we know which version to apply. I also note that in the HCal implementation, Birks' law is not applied to gammas and neutrons, while I think the most correct approach would be to only apply it to charged particles?

You could change this inside G4 but that would mean having to have custom stuff ported whenever we change G4 version, and we can access each G4 step from within the framework, so doing it in `ldmx-sw` should be equivalent and much easier.

tvami:

This resolves https://github.com/LDMX-Software/ldmx-sw/issues/1088

In default, it's switched off for the TS, but switched on for the LYSO target.

## Check List

### DIFF
--- a/SimCore/include/SimCore/SDs/TrigScintSD.h
+++ b/SimCore/include/SimCore/SDs/TrigScintSD.h
@@ -65,6 +65,12 @@ class TrigScintSD : public SensitiveDetector {
   std::string vol_name_;
   /// the ID number for the module we are gathering hits from
   int module_id_;
+  /// Whether to use Birks law for energy deposition
+  bool use_birks_law_;
+  /// Birks law constants c1
+  double birks_const_one_;
+  /// Birks law constants c2
+  double birks_const_two_;
 };
 
 }  // namespace simcore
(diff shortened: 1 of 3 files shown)


tomeichlersmith (approved):
Looks good to me, let's see how the LYSO DQM changes

tvami:
This is going in the direction we wanted it to go:
Screenshot 2025-08-11 at 13 38 30

mghrear (commented):
Looks good to me! Quick question about birks_const_one and birks_const_two: I see that the defaul values are from https://github.com/LDMX-Software/SimCore/blob/0656170b61f39dd972c6b8fd01fd9707391001d5/src/SimCore/SDs/HcalSD.cxx#L31-L88. Should these be modified for the TS and LYSO?

tvami:
@mghrear I dont think so... but they are configurable, so in case somebody wants to study how changing them will effect anything, they can do it easily the way I implemented it

mghrear (approved):

PR #1820, opened 2025-08-18
move denv into ldmx sw for new `init`s

tomeichlersmith:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
This resolves #1472 by moving newly initialized denv's into ldmx-sw. While I was there, I also noticed we can get [`git worktree`](https://git-scm.com/docs/git-worktree) to work for us by dropping the directory crawling in Framework/CMakeLists.txt and using `git rev-parse` to retrieve the commit sha. I decided this was okay because (a) I've wanted to use `git workspace` for a long time and (b) we already require a pre-cmake step `prep-version` where I can run `git rev-parse`.

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).
- [x] I ran my developments and the following shows that they are successful. See https://github.com/LDMX-Software/ldmx-sw/issues/1472#issuecomment-3128405292
- [x] Updated compatibility comments on docs site: https://github.com/LDMX-Software/ldmx-software.github.io/pull/54

## Side Effects
There are two major side effects of this change that I want to document. These will only affect people who have already been developing ldmx-sw.

### Access Files in Parallel Directories
When the denv is in the parent diretory of ldmx-sw, you are able to easily access files that are "parallel" to ldmx-sw itself.
```
tom@appa:~/code/ldmx$ mkdir parallel-eg
tom@appa:~/code/ldmx$ cd parallel-eg/
tom@appa:~/code/ldmx/parallel-eg$ touch file-on-disk
tom@appa:~/code/ldmx/parallel-eg$ cd ../
# ldmx-sw is on trunk
tom@appa:~/code/ldmx/ldmx-sw$ denv ls ../parallel-eg/
file-on-disk
tom@appa:~/code/ldmx/ldmx-sw$ cd ../1472-move-denv-into-ldmx-sw/
# 1472-... is on this branch
tom@appa:~/code/ldmx/1472-move-denv-into-ldmx-sw$ denv ls ../parallel-eg/
ls: cannot access '../parallel-eg/': No such file or directory
```
[...]

PR #1824, opened 2025-08-20
Remove BLAS and fix warning in HgcrocPulseTruth

### LINKED ISSUE #1823: Issues when building an image on Apple Si
```
500.1 gmake[2]: *** No rule to make target '/usr/lib/x86_64-linux-gnu/libblas.so.3', needed by 'SimCore/libSimCore_Generators.so'.  Stop.
500.1 gmake[1]: *** [CMakeFiles/Makefile2:1835: SimCore/CMakeFiles/SimCore_Generators.dir/all] Error 2
```

and I also see a warning with
```
11.84 [  7%] Building CXX object ReconEvent/CMakeFiles/Recon_Event.dir/src/Recon/Event/HgcrocPulseTruth.cxx.o
12.12 In file included from /usr/include/c++/13/bits/stl_algo.h:61,
12.12                  from /usr/include/c++/13/functional:67,
12.12                  from /usr/local/include/root/TF1.h:25,
12.12                  from /code/Recon/include/Recon/Event/CompositePulse.h:6,
12.12                  from /code/Recon/include/Recon/Event/HgcrocPulseTruth.h:8,
12.12                  from /code/Recon/src/Recon/Event/HgcrocPulseTruth.cxx:1:
12.12 /usr/include/c++/13/bits/stl_heap.h: In function 'constexpr void std::__adjust_heap(_RandomAccessIterator, _Distance, _Distance, _Tp, _Compare) [with _RandomAccessIterator = __gnu_cxx::__normal_iterator<pair<double, double>*, vector<pair<double, double> > >; _Distance = long int; _Tp = pair<double, double>; _Compare = __gnu_cxx::__ops::_Iter_comp_iter<ldmx::HgcrocPulseTruth::getMax() const::<lambda(const pair<double, double>&, const pair<double, double>&)> >]':
12.12 /usr/include/c++/13/bits/stl_heap.h:224:5: note: parameter passing for argument of type 'std::pair<double, double>' when C++17 is enabled changed to match C++14 in GCC 10.1
12.12   224 |     __adjust_heap(_RandomAccessIterator __first, _Distance __holeIndex,
12.12       |     ^~~~~~~~~~~~~
```

The first one is the real issue, and it seems it's manually added in the GENIE here
https://github.com/LDMX-Software/ldmx-sw/blob/65c6b5c984707aee650121ca9316e7f21d6c0889/cmake/BuildMacros.cmake#L136-L138

We either add the BLAS to the Docker, or remove it's use.

The second one seems to be solved by moving to `ranges::sort`

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1823

## Check List

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,3 +31,4 @@ EventDisplay/include/EventDisplay/EventDisplayLinkDef.h
 .bash*
 .profile
 .denv
+.apptainer
--- a/Recon/src/Recon/Event/HgcrocPulseTruth.cxx
+++ b/Recon/src/Recon/Event/HgcrocPulseTruth.cxx
@@ -9,7 +9,7 @@ void HgcrocPulseTruth::clear() {}
 double HgcrocPulseTruth::getMax() const {
   auto hits = compositePulse_.hits();
 
-  std::sort(
+  std::ranges::sort(
       hits.begin(), hits.end(),
       [](const std::pair<double, double> &a,
          const std::pair<double, double> &b) { return a.second < b.second; });
--- a/cmake/BuildMacros.cmake
+++ b/cmake/BuildMacros.cmake
@@ -133,10 +133,6 @@ macro(setup_genie_target)
     add_genie_target(GRwIO)
     add_genie_target(GRwClc)
     
-    add_library(blas SHARED IMPORTED GLOBAL) # or STATIC instead of SHARED
-    set_target_properties(blas PROPERTIES
-      IMPORTED_LOCATION "/usr/lib/x86_64-linux-gnu/libblas.so.3")
-    
     set(GENIE_LIBS
        log4cpp
        gsl
--- a/cmake/BuildMacros.cmake
+++ b/cmake/BuildMacros.cmake
@@ -193,8 +193,7 @@ macro(setup_genie_target)
        Genie::GTlGeo
        Genie::GTlFlx
        ROOT::MathCore
-       ROOT::MathMore
-       blas)
+       ROOT::MathMore)
      message(STATUS "Found GENIE")
   endif()
 endmacro()


tvami:
OK it builds
https://github.com/LDMX-Software/ldmx-sw/actions/runs/17104993738/job/48511457779?pr=1824

.gitignore
```diff
@@ -8,6 +9,9 @@ __pycache__/
 .project
 .settings
 .pydevproject
+.bash*
+.profile
+.denv
```
tomeichlersmith: These were added to gitignore by #1820 so maybe you need to `git pull` when on `trunk` since I just merged that this morning?

tomeichlersmith (approved):
Besides the double-change in gitignore (which isn't a big deal), I'm happy to merge this, but I really want @wesketchum to weigh in to avoid us barreling forward through a known necessity. **Does the genie CI test run check if blas is needed?** If yes, then it running right now confirms we can remove it.

tvami:
>  Does the genie CI test run check if blas is needed? If yes, then it running right now confirms we can remove it.

Yeah, that was my idea too, it really should test it

.gitignore
tvami: Interesting, I just did a git clone like just before opening this PR, so it should have had this changes, not sure why

tvami: Yeah it really got just merged after my clone. I'll rebase after the CI tests finish

tvami:
For our other issue @tomeichlersmith : The signal passed!
https://github.com/LDMX-Software/ldmx-sw/actions/runs/17105328931/job/48513529800?pr=1824

tvami:
GENIE ran all fine too:
https://github.com/LDMX-Software/ldmx-sw/actions/runs/17105328931/job/48513529807?pr=1824

wesketchum (approved):
I think the test should be checking for this, and as far as I can see looking through GENIE code and documentation, it's pulling in CBLAS from GSL, so shouldn't need system BLAS.

PR #1832, opened 2025-08-27
Shift TS pad-3 to match LYSO in the default geometry

### LINKED ISSUE #1759: Shift TS Pad3 slightly upstream to seamlessly fit LYSO
The LYSO target is thicker than the W target. Thus, TS, which is right upstream of the target, needs to shift a bit to accommodate two different targets with a single upstream geometry.

Shift TS Pad3 up by something like 1.3738 mm. Leave Pad1 and 2 where they are.

Two positions, according to target. Easy in simulation, less fun in hardware.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1759

## Check List

This shows that LYSO and now the default in v15 has the TS at the same Z:
dark_brem_z_comparison

### DIFF
--- a/ComparePlots/__main__.py
+++ b/ComparePlots/__main__.py
@@ -26,7 +26,7 @@ from ._plotter import plotter
 
 # guard incase someone imports this somehow
 if __name__ == '__main__' :
-    parser = argparse.ArgumentParser("python3 -m Validation",
+    parser = argparse.ArgumentParser("just compare-plots",
         description="""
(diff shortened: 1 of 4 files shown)


JYoo001 (approved):

tomeichlersmith (approved):

tvami:
Screenshot 2025-08-27 at 08 49 20
Here is also the new range for the same DQM plot

PR #1833, opened 2025-09-02
Add deep Ecal gun as CI

### LINKED ISSUE #1516: Introduce new CI that has plots after trigger cuts

Right now all the CI show plots before any selection.

I'd like to have another CI that has it after the trigger

**Additional context**

In a recent PR we concluded that the changes would prob be washed out after cuts, so that brought up that it would make sense to have a version of a CI that does that

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1516
Altho not with the original intention. We are kinda of getting around it and a sample that by design has high trigger eff. 
So it's more of an approximation and having an effective hard brem for cases when this happens deep in the Ecal.

To me this also gives us the chance to have an example of a gun that shoots two particles in the CI. 

## Check List

Here is the comparison of this gun vs a trigger skimmed hard-brem sample.
Please note there is no reecoil tracking like this, so the plots that would rely on that info make no sense in this comparrison: 

[Henry_TriggeredVsGun.pdf](https://github.com/user-attachments/files/22102493/Henry_TriggeredVsGun.pdf)

### DIFF
--- a/.github/validation_samples/deep_ecal_gun/config.py
+++ b/.github/validation_samples/deep_ecal_gun/config.py
@@ -1,0 +1,129 @@
+from LDMX.Framework import ldmxcfg
+p = ldmxcfg.Process('test')
+
+p.maxTriesPerEvent = 10000
+
+from LDMX.Biasing import ecal
+from LDMX.SimCore import generators as gen
+from LDMX.SimCore import simulator as sim
+det = 'ldmx-det-v15-8gev'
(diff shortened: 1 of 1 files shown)


tomeichlersmith (approved):
Some of the ECal veto variables that rely on the from-target assumption will be meaningless in this sample, but it looks like you and Henry have already observed this (from the slides).

JYoo001 (approved):

PR #1835, opened 2025-09-03
CLUE clustering improvements, fixing multilayer option

### LINKED ISSUE #1834: CLUE does not work with multiple layers
**Describe the bug**
Segfault if nbr_of_layers is more than 1

**To Reproduce**
Set this
https://github.com/LDMX-Software/ldmx-sw/blob/58f78d7851f8a743a42f447badd54674a0177951/Ecal/python/ecalClusters.py#L37
to more than 1 and it segfaults.

**Desired behavior**

Being able to work with more than 1 layers

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1834

## Check List

### DIFF
--- a/DQM/python/dqm.py
+++ b/DQM/python/dqm.py
@@ -1065,7 +1065,7 @@ class EcalClusterAnalyzer(ldmxcfg.Analyzer) :
 
         self.build1DHistogram("number_of_clusters_first_layer", "Number of CLUE clusters on the first layer", 5, -0.5, 4.5)
         self.build1DHistogram("number_of_clusters_per_layer", "Number of CLUE clusters per layer", 5, -0.5, 4.5)
-        self.build1DHistogram("number_of_clusters", "Total number of CLUE clusters", 51, -0.5, 50)
+        self.build1DHistogram("number_of_clusters", "Total number of CLUE clusters", 51, -0.5, 50.5)
         self.build1DHistogram("correctly_predicted_events", "Correct Cluster Count", 3, 0., 3.)
 
         # Need to mod for more than two electrons
--- a/DQM/python/dqm.py
+++ b/DQM/python/dqm.py
@@ -1063,7 +1063,9 @@ class EcalClusterAnalyzer(ldmxcfg.Analyzer) :
         self.ecal_sp_hits_passname = ''
 
 
-        self.build1DHistogram("number_of_clusters", "Number of CLUE clusters", 5, -0.5, 4.5)
+        self.build1DHistogram("number_of_clusters_first_layer", "Number of CLUE clusters on the first layer", 5, -0.5, 4.5)
+        self.build1DHistogram("number_of_clusters_per_layer", "Number of CLUE clusters per layer", 5, -0.5, 4.5)
+        self.build1DHistogram("number_of_clusters", "Total number of CLUE clusters", 51, -0.5, 50)
         self.build1DHistogram("correctly_predicted_events", "Correct Cluster Count", 3, 0., 3.)
 
         # Need to mod for more than two electrons
@@ -1080,7 +1082,8 @@ class EcalClusterAnalyzer(ldmxcfg.Analyzer) :
         self.build2DHistogram("total_energy_vs_hits", "Total energy (edep) [MeV]", 30, 0, 150, "Hits in cluster", 20, 0, 200)
         self.build2DHistogram("total_energy_vs_purity", "Total energy (edep) [MeV]", 30, 0, 150, "Energy purity %", 21, 0, 105)
         self.build2DHistogram("sp_ele_distance_vs_purity", "SP ele distance in xy-plane [mm]", 50, 0, 250, "Energy purity %", 21, 0, 105)
-        
+        self.build2DHistogram("sp_clue_distance_vs_layer", "CLUE centroid to SP ele distance in xy-plane [mm]", 125, 0., 250., "Layer", 33, -0.5, 32.5)
+
         self.build1DHistogram("sp_clue_distance", "CLUE centroid to SP ele distance in xy-plane [mm]", 125, 0., 250.)
         self.build1DHistogram("sp_clue_x_residual", "CLUE centroid X - SP ele X [mm]", 250, -250., 250.)
         self.build1DHistogram("sp_clue_y_residual", "CLUE centroid Y - SP ele Y [mm]", 250, -250., 250.)
--- a/Ecal/src/Ecal/CLUE.cxx
+++ b/Ecal/src/Ecal/CLUE.cxx
@@ -5,27 +5,23 @@
 
 namespace ecal {
 
-// Euclidean distance between two points
-// Note that x*x is much faster than std::pow(x,2)
-float CLUE::dist(double x1, double y1, double x2, double y2) {
+/**
+ * Euclidean distance between two points
+ * Using x*x and std::sqrt since they are more specialized and faster than
+ * std::pow
+ * @tparam T floating point type
+ */
+template <typename T>
+T CLUE::dist(T x1, T y1, T x2, T y2) {
   auto delta_x = x1 - x2;
   auto delta_y = y1 - y2;
   auto r_square = delta_x * delta_x + delta_y * delta_y;
   return std::sqrt(r_square);
 }
 
-// Euclidean distance between two points (float version)
-// Overloaded for 2D and 3D
-float CLUE::floatDist(float x1, float y1, float x2, float y2) {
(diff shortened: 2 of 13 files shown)


Ecal/src/Ecal/EcalClusterProducer.cxx
```diff
@@ -14,57 +14,68 @@ EcalClusterProducer::EcalClusterProducer(const std::string& name,
...
+    }
 
     std::vector<ldmx::EcalCluster> ecal_clusters;
+    ldmx_log(debug) << "Filling " << wc_vec.size()
+                    << " clusters into ecal_clusters";
     for (int a_wc = 0; a_wc < wc_vec.size(); a_wc++) {
       ldmx::EcalCluster cluster;
 
       cluster.setEnergy(wc_vec[a_wc].centroid().E());
-      cluster.setCentroidXYZ(wc_vec[a_wc].centroid().Px(),
```
tvami, line 62: I dont understand why this was on the momentum in the first place, that's not something we can measure with the ECal

tvami:
This is not perfect, but at least it works.
I can open another issue with next TODOs:
 * CLUE3D doesnt work because `seed_` is not filled anywhere
 * The layer-wise CLUE now works, but it uses a different `rho_c` than the first-layer-only on: 

a) I made this tunable by the same external parameter, so this is just an optimization at this point, which I can have one of my students do

b) I also persist now which layer the given cluster came from, and this is now used in the DQM as well

c) Obv this leads to much more clusters, so I introduced a few layer avg quantities in DQM

Ecal/src/Ecal/EcalClusterProducer.cxx
bryngemark: this is a LorentzVector artefact afaik

bryngemark:
@tvami can you help me find what you did to fix the multiple layers bug explicitly? this PR is a jungle of formatting and name changes on top of the algorithmic work. (i think it will be a nightmare at this point to contribute the stuff i was working on in late July...)

Ecal/src/Ecal/CLUE.cxx
```diff
@@ -391,131 +423,149 @@ std::vector<std::vector<const ldmx::EcalHit*>> CLUE::clustering(
...
-                << next_layer[j]->total_energy_;
+        for (const auto& next_seed : next_layer) {
+          auto distance_2d_next = floatDist(current_seed->x_, current_seed->y_,
+                                            next_seed->x_, next_seed->y_);
+          auto dz_next = std::abs(current_seed->z_ - next_seed->z_);
+          ldmx_log(trace) << "      DeltaXY to index " << next_seed->index_
+                          << ": " << std::setprecision(4) << distance_2d_next;
+          ldmx_log(trace) << "      DeltaZ to index_" << next_seed->index_
+                          << ": " << std::setprecision(4) << dz_next;
+          if (next_seed->total_energy_ > current_seed->total_energy_ &&
```
bryngemark, line 498: i suppose this (avoiding index lookup) is the equivalent of my fix? what i found was a simple confusion of `i` and `j`
```
         for (int j = 0; j < nextLayer.size(); j++) {
           auto d = floatDist(currentLayer[i]->x, currentLayer[i]->y,
                              nextLayer[j]->x, nextLayer[j]->y);
-          auto dz = std::abs(currentLayer[i]->z - nextLayer[i]->z);
+          auto dz = std::abs(currentLayer[i]->z - nextLayer[j]->z);
```

tvami:
@bryngemark I think it's two things: (1) the layers are now not determined based on their Z position anymore but rather the hit's own knowledge of what layer they are on, (2) the indexes are gone and everything is range based, avoiding any possible typos

Ecal/src/Ecal/CLUE.cxx
```diff
@@ -1,22 +1,39 @@
...
 
+// Euclidean distance between two points (float version)
+// Overloaded for 2D and 3D
 float CLUE::floatDist(float x1, float y1, float x2, float y2) {
-  return powf(powf(x1 - x2, 2) + powf(y1 - y2, 2), 0.5);
+  auto delta_x = x1 - x2;
+  auto delta_y = y1 - y2;
+  auto r_square = delta_x * delta_x + delta_y * delta_y;
+  return std::sqrt(r_square);
 }
```
tomeichlersmith: We could avoid repetition here either by using position vectors (like the ones from ROOT or Eigen) that already have the subtraction and mangitude operators defined, _or_ if we want to avoid this larger refactor we can template over the argument types.

```suggestion
/**
 * Euclidean distance between two points
 * Using x*x and std::sqrt since they are more specialized and faster than std::pow
 * @tparam T floating point type
 */
template<typename T>
T dist(T x1, T y1, T x2, T y2) {
  auto delta_x = x1 - x2;
  auto delta_y = y1 - y2;
  auto r_square = delta_x * delta_x + delta_y * delta_y;
  return std::sqrt(r_square);
}
```

This isn't super necessary, just trying to practice DRY.

Ecal/src/Ecal/CLUE.cxx
```diff
@@ -391,131 +423,149 @@ std::vector<std::vector<const ldmx::EcalHit*>> CLUE::clustering(
...
-  for (const auto& vec : clusters) {
-    auto c = IntermediateCluster();
-    auto fc = IntermediateCluster();
-    for (const auto& hit : vec) {
-      c.add(hit);
-      // if hit is in first layer_, add to first layer_ cluster
-      if (hit->getZPos() < first_layer_max_z_) fc.add(hit);
+  for (const auto& cluster : clusters) {
+    auto intermediate_cluster = IntermediateCluster();
+    auto intermediate_cluster_first_layer = IntermediateCluster();
```
tomeichlersmith: I think this is a relic of pointers-past.

```suggestion
    IntermediatCluster intermediat_cluster{}, intermediat_cluster_first_layer{};
```

Ecal/src/Ecal/CLUE.cxx
```diff
@@ -80,107 +97,114 @@ float CLUE::floatDist(float x1, float y1, float z1, float x2, float y2,
...
-      layer_rho_c_.push_back(highest_energy / rhoc_factor);
-      layer_tag++;
-      ldmx_log(trace) << "    Highest energy: " << highest_energy << "\n"
-                      << "  Creating layer_ " << layer_tag;
-      highest_energy = 0.;
+  ldmx_log(trace) << "Number of layers: " << nbr_of_layers_;
+
+  // vector of layers, each layer having a vector of hits
+  std::vector<std::vector<const ldmx::EcalHit*>> hits_per_layer;
+  hits_per_layer.resize(nbr_of_layers_);
```
tomeichlersmith: You can construct a vector of a certain size "in place" by providing the size as the argument to the constructor. WIthout a second argument, it just default-constructs all of the elements (which is what you want in this case).

```suggestion
  std::vector<std::vector<const ldmx::EcalHit*>> hits_per_layer(nbr_of_layers_);
```

There are other instances of this throughout this file, but its not super important. It probably doesn't have any performance difference, it just helps make the code shorter (but maybe less simple?).

tomeichlersmith (changes requested):
Just some C++ technical stuff as usual. I've certainly missed some things, but hopefully there are some incremental improvements.

tvami:
> Just some C++ technical stuff as usual. I've certainly missed some things, but hopefully there are some incremental improvements.

Done in https://github.com/LDMX-Software/ldmx-sw/pull/1835/commits/158e64374c539869700ebc8f214c06197c74ff7b

tomeichlersmith (approved):

bryngemark (approved):

DQM/python/dqm.py
```diff
@@ -1063,7 +1063,9 @@ def __init__(self,name='EcalClusterAnalyzer') :
         self.ecal_sp_hits_passname = ''
 
 
-        self.build1DHistogram("number_of_clusters", "Number of CLUE clusters", 5, -0.5, 4.5)
+        self.build1DHistogram("number_of_clusters_first_layer", "Number of CLUE clusters on the first layer", 5, -0.5, 4.5)
+        self.build1DHistogram("number_of_clusters_per_layer", "Number of CLUE clusters per layer", 5, -0.5, 4.5)
+        self.build1DHistogram("number_of_clusters", "Total number of CLUE clusters", 51, -0.5, 50)
```
tvami:
```suggestion
        self.build1DHistogram("number_of_clusters", "Total number of CLUE clusters", 51, -0.5, 50.5)
```

PR #1846, opened 2025-09-15
More ecal layer stack updates

tomeichlersmith:

No issue, just some updates to the model that are focused on bringing the layer stack model of the slice prototype in sync with what we are able to build. The change from attempting to construct a PCB material to just using layers of copper and fiberglass affects the layer weights but at a really small level, so I'm not re-copying them over.


tvami (approved):

Detectors/util/ecal_layer_stack.py
```diff
@@ -305,6 +327,12 @@ def tungsten(t) :
     def pcb() :
         return Layer('PCB',1.666)
 
+    def pcb_unpacked():
+        return [
+            Layer('Cu', 0.170),
+            Layer('FR4', 1.030)
+        ]
+
```
tomeichlersmith: replace `pcb` with `pcb_unpacked` and drop the `PCB` material from the material mapping to avoid confusion.

Detectors/util/pcb-layers.nbt
```diff
@@ -4,16 +4,21 @@ fn _hexagon_area(side: Length) -> Area = 3*sqrt(3)/2*side*side
 fn _copper_layer_weight(n: Scalar) = ((n-2)*(0.5 oz / ft^2) + 2*(1.0 oz / ft^2))
 
 let hexagon_side : Length = 85 mm
-let hexagon_thick : Length = 1.66 mm
+let pcb_thick : Length = 1.2 mm
```
tomeichlersmith: While I like numbat, I think that it is more transparent to future developers to have this calculation inside of `ecal_layer_stack.py` as well. So I'm going to remove `pcb-layers.nbt` and do the estimate of the copper thickness in a PCB layer "live" in the Python script.

PR #1854, opened 2025-09-25
improve friendliness of histogram pool

### LINKED ISSUE #1381: Friendliness of Histogram pooling
Since filling histograms is a common task especially for new developers of ldmx-sw, I think focusing on improving its friendliness is helpful. First and foremost, the error being thrown from `get` is hard to parse for people new to ldmx-sw and ROOT files, so I'd like to reword this to help more.

https://github.com/LDMX-Software/ldmx-sw/blob/613a9d10b87235eb4ad3a6d3fc8229752692ad98/Framework/src/Framework/Histograms.cxx#L46-L53

## Necessity of Singleton Pool
I can't remember why we need a singleton pool. Can we get away with a pool for each processor? This would allow us to avoid adding the processor's name as a prefix to the histogram name (I think) which is another source of confusion/annoyance. The goal here would get to a point where the histogram file could look like
```
hist.root:
  analyzer1/
    histogram
  analyzer2/
    histogram
```
so both analyzer1 and analyzer2 could share histogram names but the two histograms are distinct both in memory and in the output `hist.root` file.

## Creation Helpers for Categorical Axes
I've written some helpers that create categorical axes that I think we could use within our own DQM analyzers. Maybe they don't get used immediately within the DQM in order to avoid breaking anything, but having them around for writing new categorical histograms may be helpful.

```cpp
  std::tuple<std::size_t,double,double> category_bins(
      const std::vector<std::string>& categories,
      int offset = 0
  ) {
    std::size_t n_categories = categories.size();
    double min = offset - 0.5;
    double max = offset + n_categories + 1.5;
    return std::make_tuple(n_categories, min, max);
  }
  void label_axis(TAxis* axis, const std::vector<std::string>& categories) {
    for (std::size_t ibin{1}; ibin <= categories.size(); ibin++) {
      axis->SetBinLabel(ibin, categories[ibin-1].c_str());
    }
  }
  void create_categorical_histogram(
      const std::string& name,
      const std::vector<std::string>& categories
  ) {
    auto [nbin, xmin, xmax] = category_bins(categories);
    histograms_.create(name, "", nbin, xmin, xmax);
    auto h{histograms_.get(name)};
    label_axis(h->GetXaxis(), categories);
  }
  void create_categorical_histogram(
      const std::string& name,
      const std::vector<std::string>& categories,
      const std::string& ylabel, int nybin, double ymin, double ymax
  ) {
    auto [nxbin, xmin, xmax] = category_bins(categories);
    histograms_.create(name, "", nxbin, xmin, xmax, ylabel, nybin, ymin, ymax);
    label_axis(histograms_.get(name)->GetXaxis(), categories);
  }
  void create_categorical_histogram(
      const std::string& name,
      const std::vector<std::string>& xcategories,
      const std::vector<std::string>& ycategories
  ) {
    auto [nxbin, xmin, xmax] = category_bins(xcategories);
    auto [nybin, ymin, ymax] = category_bins(ycategories);
    histograms_.create(name, "", nxbin, xmin, xmax, "", nybin, ymin, ymax);
    label_axis(histograms_.get(name)->GetXaxis(), xcategories);
    label_axis(histograms_.get(name)->GetYaxis(), ycategories);
  }
```

tomeichlersmith:

This resolves #1381 

- abandon a histogram singleton shared across all processors, instead each processor has their own pool
- the pool makes sure to call the processor's `getHistoDirectory()` and `cd` to that directory in the output file whenever `histograms_.create` is called. This also patches the annoying runtime issue of forgetting to call `getHistoDirectory()` will just silently not create an output histogram file.
- add categorical axis creations into the histogram pool
- add support to enable weighting of histograms (Sumw2)
  - before, we were weighting the histograms, but without tracking the Sumw2 separately, the errors would not be estimated appropriately
- add testing to make sure pool is functioning as designed

> [!WARNING]
> I expect this to fail *all* of the PR Validation tests. The histograms are compared _by name_ and since this changes their full name in the output file, the auto-testing in the CI will not be able to line them up. I can manually do the comparison by modifying the script locally after the CI runs and puts the histogram files onto GitHub as artifacts.

## Check List
- [x] I ran my developments and the following shows that they are successful. (see new `Framework/test/HistogramPoolTest.cxx`)
- [x] update `Framework/python/histogram.py` to include categorical axis support (probably also requiring updates to `Framework/src/Framework/EventProcessor.cxx` where the interpretation of the python config is done)
- [x] convert some DQM histograms to new categorical axes as examples
- [x] document usage in `HistogramPool` class documentation
- [x] update https://ldmx-software.github.io/using/analysis/ldmx-sw.html with comment about old (`proc/proc_hist`) vs new (`proc/hist`) naming scheme, direct reader to `HistogramPool` class documentation for detail usage. https://github.com/LDMX-Software/ldmx-software.github.io/commit/da2c322006221cb82c6a66511bbca29b5373d3f7


Framework/src/Framework/HistogramPool.cxx
```diff
@@ -0,0 +1,299 @@
...
+            " already exists in histogram pool."
+            "\nMake sure to use distinct names for your different histograms!"
+            "\nYou can use `histograms_.get` to retrieve a pointer to a "
+            "specific histogram "
+            "if you want to do some other customizations besides filling.");
+  }
+
+  get_directory_()->cd();
+
+  auto h = factory();
```
tvami: can we have this as `histo` or something to avoid the one-letter variable names?

Framework/src/Framework/HistogramPool.cxx
```diff
@@ -0,0 +1,299 @@
...
+    }
+  }
+}
+
+void HistogramPool::create(const std::string& name, const std::string& x_label,
+                           const std::vector<std::string>& categories,
+                           bool weighted) {
+  insert(
+      name,
+      [&]() {
```
tvami, line 109: can you add a comment somewhere about why we are doing this lambda approach here, and then why we pass nothing in the body of it?

Framework/test/HistogramPoolTest.cxx
```diff
@@ -0,0 +1,180 @@
...
+  const char* test_file = "/tmp/test_histogram_pool.root";
+
+  SECTION("exception on request") {
+    HistogramPool p{cantCreateDir};
+    // raise exception from get_directory
+    REQUIRE_THROWS_WITH(p.create("dne", "foo", 10, 0.0, 1.0), "NOFILEGIVEN");
+  }
+
+  SECTION("creation on request") {
+    TFile* f{nullptr};
```
tvami: can we avoid the one-letter variable names here too?

Framework/test/HistogramPoolTest.cxx
```diff
@@ -0,0 +1,180 @@
...
+    auto open_on_request = [&f, &test_file]() -> TDirectory* {
+      if (f == nullptr) {
+        f = TFile::Open(test_file, "recreate");
+      }
+      return f->mkdir("histo_directory");
+    };
+    HistogramPool p{open_on_request};
```
tvami: can we avoid the one-letter variable names here too?

tvami (commented):
Nice work! I have a few cosmetic + documentation points:

Framework/src/Framework/HistogramPool.cxx
tomeichlersmith: A few comments.

The "body" of the lambda function is between the curly brackets `{ }` and is what creates the histogram. The _arguments_ of the lambda function are between the parenthesese `()` and this factory function does not accept any arguments.

I need it to create the histogram (`new TH1F(...)`) _after_ we `cd` into the output directory. This is the only way I've found to be able to get the histograms into the correct location in the output file. This means we cannot pass the already-created histogram pointer into the `insert` function. We _could_ put the `cd` at the top of all `create` functions, but - in the interest of DRY - I factored out the common stuff into `insert` and have the `create` functions build a "factory function" that constructs the histogram in the correct way. The factory function is then called at the correct time (after `cd()`) so that the histogram is in the right directory. I'll add this to the documentation of `insert`.

DQM/python/dqm.py
```diff
@@ -118,7 +118,7 @@ def __init__(self,name="hcal_dqm", pe_threshold=8, section=0, max_hit_time = 50.
         self.build1DHistogram('along_x', 'x', 1200, -3000,3000)
         self.build1DHistogram('along_y', 'y', 1200, -3000,3000)
         self.build1DHistogram('along_z', 'z', 1200, 0,6000)
-        # Per hit
+#Per hit
```
tomeichlersmith: The removal of these tabs is leftover from me accidentally running `clang-format` on this file. Need to undo.

DQM/python/dqm.py
```diff
@@ -998,21 +1040,43 @@ def __init__(self, name='SampleValidation') :
         self.sim_particles_passname = ''
         self.target_scoring_plane_passname = ''
 
-        # primary histograms
-        self.build1DHistogram("primaries_pdgid", "ID of primary particles", 20, 0, 20)
+        pdgid_bin_labels = [
+            "e^{+}",                   # 1
```
tomeichlersmith: Update comment with correct index - current indices are above the correct index by 1.

DQM/src/DQM/SampleValidation.cxx
```diff
@@ -128,45 +128,8 @@ int SampleValidation::pdgidLabel(const int pdgid) {
     label =
         18;  // dark photon, need pdg id for other models like ALPs and SIMPs
 
-  return label;
+  return label - 1;
```
tomeichlersmith: Update rest of the function to avoid future confusion.

tomeichlersmith:
Interesting, the PR Validations all _passed_ (because no histograms were ever compared!). I believe the ECal PN sample contains all the different types of histograms, so I'm going to run that locally and check if things break.

tomeichlersmith:
Sweet, this looks good. The only plots that fail the KS test are the categorical ones where I changed the binning (shifted by 0.5 or 1.5 depending on the categories).

Some examples...
PN_event_type
SampleValidation_primaries_pdgid

tvami (approved):
Great work, thanks Tom!

PR #1856, opened 2025-10-02
PF hit association

bryngemark:

This allows association back to hits from clusters that were associated with PF tracks in the PFlow algorithm. It also includes a first go at a pileup electron identifier processor, that subtracts hits associated to high-momentum tracks from the ecal rechit collection (or rather, makes a new collection without those hits).

## Check List

Plot showing 10 2e inclusive events, ecal rec hits _x_: black is all hits, red is after pileup removal. This is not at all optimized yet (clustering needs improvements) but the principle works. 

ecalRecHits_2e_allBlack_noPUred


Ecal/src/Ecal/CLUE.cxx
```diff
@@ -34,64 +34,66 @@ T CLUE::dist(T x1, T y1, T z1, T x2, T y2, T z2) {
...
-//       firstLayerClusters[j].centroid().Px(),
-//       firstLayerClusters[j].centroid().Py());
-//       ldmx_log(trace) << "Dist to cluster " << j << ": " << d;
-//     }
-//   }
-// }
+void CLUE::electronSeparation(std::vector<ldmx::EcalHit> hits) {
+  std::vector<double> layerThickness = {2.,   3.5,  5.3,  5.3, 5.3, 5.3,
+                                        5.3,  5.3,  5.3,  5.3, 5.3, 10.5,
+                                        10.5, 10.5, 10.5, 10.5};
```
tomeichlersmith: Worried about this hard-coded geometry stuff especially since its indexed by bi-layer instead of silicon-layer (2 silicon layers per bi-layer).

I think the solution would be to share the `EcalGeometry` condition with `CLUE` but that is technical and I'm willing to just make a new issue and merge this as-is.

Recon/include/Recon/Event/PFCandidate.h
```diff
@@ -104,6 +110,21 @@ class PFCandidate {
...
+  // void setEcalHits(const std::vector<const ldmx::EcalHit*> hits) { ecal_hits_
+  // = hits; }
+
+  /**
+   * Take in the hcal hits that make up the candidate.
+   * @param hit The digi hit's entry number in the events digi
+   * collection.
+   */
+  //  void setHcalHits(const std::vector<const ldmx::HcalHit*> hits) {
+  //  hcal_hits_ = hits; }
```
tomeichlersmith: Cleanup

```suggestion
```

Recon/include/Recon/Event/PFCandidate.h
```diff
@@ -161,6 +186,20 @@ class PFCandidate {
...
+   */
+  // std::vector<const ldmx::EcalHit*> getEcalHits() { return ecal_hits_; }
+
+  /**
+   * Take in the hcal hits that make up the candidate.
+   * @param hit The digi hit's entry number in the events digi
+   * collection.
+   */
+  // std::vector<const ldmx::HcalHit*> getHcalHits() { return hcal_hits_; }
+
```
tomeichlersmith: cleanup
```suggestion
```

tomeichlersmith (changes requested):
This is looking good, I have a few cleanup requests and one comment about sharing EcalGeo information with CLUE.

Ecal/src/Ecal/CLUE.cxx
bryngemark: I agree that's a nicer solution but I really haven't been working on this part. A lot of stuff Ella introduced was touched by Tamas and differences are sure to show up here since I had to patch a lot when merging. 
So yeah in short a separate issue for that sounds great.

bryngemark:
The failing histogram tests are caused by a name change `clusterless` --> `unclustered` (which in turn is no biggie, just reflects the difference in perspective between a more seasoned particle physicist and a happy algorithm developer)

tomeichlersmith (approved):

DQM/src/DQM/EcalClusterAnalyzer.cxx
```diff
@@ -166,6 +214,22 @@ void EcalClusterAnalyzer::analyze(const framework::Event& event) {
...
+  ldmx_log(debug) << "Got " << n_mixed << " mixed hits, a fraction of "
+                  << (float)n_mixed / ecal_rec_hits.size();
+
+  if (ecal_clusters.size() >= 2) {
+    float d_r = std::sqrt(std::pow((ecal_clusters[0].getCentroidX() -
+                                   ecal_clusters[1].getCentroidX()),
+                                  2) +
+                         std::pow((ecal_clusters[0].getCentroidY() -
+                                   ecal_clusters[1].getCentroidY()),
+                                  2));
```
tvami: I suggest to do `a*a` instead of `pow(a,2)`, it's much faster

Ecal/src/Ecal/CLUE.cxx
```diff
@@ -34,64 +34,66 @@ T CLUE::dist(T x1, T y1, T z1, T x2, T y2, T z2) {
...
-//       firstLayerClusters[j].centroid().Px(),
-//       firstLayerClusters[j].centroid().Py());
-//       ldmx_log(trace) << "Dist to cluster " << j << ": " << d;
-//     }
-//   }
-// }
+void CLUE::electronSeparation(std::vector<ldmx::EcalHit> hits) {
+  std::vector<double> layer_thickness = {2.,   3.5,  5.3,  5.3, 5.3, 5.3,
+                                        5.3,  5.3,  5.3,  5.3, 5.3, 10.5,
+                                        10.5, 10.5, 10.5, 10.5};
```
tvami: This is not true for v15 anymore

Recon/python/pileupFinder.py
```diff
@@ -0,0 +1,24 @@
+"""Configuration for pfReco producers
+
+Sets all parameters to reasonable defaults.
+
+Examples
+--------
+    from LDMX.Recon.pfReco import pileupFinder
+    p.sequence.append( pileupFinder )
```
tvami, line 8: I'd suggest to add this to the 2e CI

bryngemark: does it make sense to do this without also developing a corresponding DQM analyzer?

Ecal/src/Ecal/CLUE.cxx
bryngemark: that's right, and tom pointed out that we should look this up instead of hardwiring it. i agree, and, i didn't introduce this part myself and don't really have time to work on it in the near future. would you be ok with making a new issue for this instead?

DQM/src/DQM/EcalClusterAnalyzer.cxx
bryngemark: done

Ecal/src/Ecal/CLUE.cxx
tvami: sure, makes sense!

Recon/python/pileupFinder.py
tvami: I'd say it still makes a technical check on it, we can see how the timing changes, etc. The DQM should anyway be it's own processor that we can add whenever it's developed

bryngemark: added the needed CLUE, PFlow and pileup finding processors. ok like this?

tvami: CLUE was already included in 
https://github.com/LDMX-Software/ldmx-sw/blob/801c25f7e07b570640566f675505a0ff64cba16b/.github/validation_samples/it_pileup/config.py#L239
so one of them should be removed.

Did you modify the values from the defaults in 
https://github.com/LDMX-Software/ldmx-sw/blob/801c25f7e07b570640566f675505a0ff64cba16b/.github/validation_samples/it_pileup/config.py#L120-L131
? The ones that are the same as defaults, I'd suggest not to repeat in the config.

bryngemark: I did change some of them. At this point, I am no longer sure which ones really needed changing 😄 I removed everything that was set to default.

I removed the doubling of CLUE and kept my instance as it needs some name setting to play nicely with the rest of the pileup workflow.

.github/validation_samples/it_pileup/config.py
```diff
@@ -104,6 +104,43 @@
...
+trackPF.doElectronTracking=True
+# reference info
+truthPF = pfReco.pfTruthProducer()
+    
+# CLUE     
+import LDMX.Ecal.ecalClusters as cl
+cluster = cl.EcalClusterProducer()
+cluster.seedThreshold = 350.                                                                                                                                                                             cluster.dc = 0.3
+cluster.nbrOfLayers = 1
+cluster.nbr_of_layers = cluster.nbrOfLayers #just here to highlight confusion
```
tvami: We dont anymore have `nbrOfLayers` parameter
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Ecal/python/ecalClusters.py#L37
```suggestion
cluster.nbr_of_layers = 1
```

Recon/include/Recon/Event/PFCandidate.h
```diff
@@ -9,6 +9,9 @@
 
 // ROOT
 #include "TObject.h"  //For ClassDef
+// ldmx-sw objects
+// #include "Ecal/Event/EcalHit.h"
+// #include "Hcal/Event/HcalHit.h"
```
tvami:
```suggestion
```

tvami:
```
Configuration Error [Config] : Parameter 'ecal_sp_hits_passname' does not exist in list of parameters.
```
we need to have this with a default in the python file

.github/validation_samples/it_pileup/config.py
bryngemark: great

tvami:
The CI failed with
```
Configuration Error [Config] : Parameter 'ecal_sp_hits_passname' does not exist in list of parameters.
```

bryngemark:
> The CI failed with
> 
> ```
> Configuration Error [Config] : Parameter 'ecal_sp_hits_passname' does not exist in list of parameters.
> ```

Sorry @tvami can you tell me which python config/test you are referring to? There are thousands of histogram errors in the failed test output.

tvami:
The code is expecting
```
ecal_sp_hits_passname_ = ps.get<std::string>("ecal_sp_hits_passname");
```
i.e. `ecal_sp_hits_passname` but you renamed it to `ecal_sp_hits_pass_name`
Screenshot 2025-12-19 at 09 08 17

So one of the two needs to be updated

tvami:
Screenshot 2026-01-09 at 08 29 19

This is failing bc this branch got out of sync with the trunk, so to fix it, this branch should be rebased

bryngemark:
Rebasing is a potentially disruptive pain. Three weeks ago, all validation jobs were successful. Since then, I have only removed an unused variable in an analyzer class, but I am not surprised if gold has been updated at the pace we change things. So unless there is something this PR introduces that interferes with the clang tests, I argue that rebasing can be skipped and this branch merged.

This PR would have been merged months ago if it wouldn't have been stuck on nice-to-haves that could have been separate feature request issues. Deferring to making new issues for such requests will be my policy from now on as it is the only way that I can get code improvements out the door before the rest of the code base has shifted.

tvami:
If you look here
https://github.com/LDMX-Software/ldmx-sw/actions/runs/20858439581/job/59932576195?pr=1856
```
    File "config.py", line 112, in <module>
      trackPF.input_pass_name=inputPassName
                              ^^^^^^^^^^^^^
  NameError: name 'inputPassName' is not defined
  Configuration Error [Python] : Execution of python script failed.
```

If we merge this as is, that will lead to a broken `trunk`

tvami (approved):

.github/validation_samples/it_pileup/config.py
```diff
@@ -104,6 +104,41 @@
 hcal_veto = hcal.HcalVetoProcessor()
 hcal_veto.input_hit_pass_name = thisPassName
 
+# Load and configure  particle flow sequence. 
+# Here we use PF "tracking" and CLUE Ecal clustering 
+from LDMX.Recon import pfReco
+trackPF = pfReco.pfTrackProducer()
+trackPF.inputTrackCollName=trackPF.inputTrackCollName+overlayStr #"EcalScoringPlaneHitsOverlay" #                                                                                                                   
+trackPF.input_pass_name=inputPassName
```
tvami:
```suggestion
trackPF.input_pass_name=thisPassName
```

tvami:
> If we merge this as is, that will lead to a broken trunk

I committed the fix, should be good now

bryngemark:
> > If we merge this as is, that will lead to a broken trunk
> 
> I committed the fix, should be good now

Thanks!

tvami:
The CI pointed out another issue that I fixed in https://github.com/LDMX-Software/ldmx-sw/pull/1856/changes/a93e9ba0b40b32917d3913c1fb6c93ba624ac358

tomeichlersmith:
alright, looks like the CI fully completed, so I'm going to merge this and then we can do further patches off the updated `trunk`.

PR #1857, opened 2025-10-02
Schema Evolution Phase I: LinkDef refactor

### LINKED ISSUE #1815: Implement Schema Evolution for Event Objects
The `clang-tidy` renaming of event object members means that ldmx-sw >= 4.4.8 is unable to read the files that were generated with ldmx-sw <= 4.4.7. This is annoying and a surprise to many of us since our schema evolution (in the past) has been able to be handled auto-magically by ROOT's internal schema evolution detecting algorithm.

### Error Message
If you run into this, you'll see _many_ errors from `TBranchElement::InitializeOffsets` complaining about specific member variables "missing" when constructing the branch. For example:
```
Info in <TBranchElement::InitializeOffsets>: TTree created with an older schema, some data might not be copied in 'slow-cloning' mode; fast-cloning should have the correct result. 'trackID_' is missing when constructing the branch 'TrigScintScoringPlaneHits_eat'.
```
In the short term, you should drop back to using v4.4.7.

## Solution
The solution is to actually implement schema evolution and the documentation has improved since the last time I looked at it, giving me more hope.

Main Reference: https://root.cern/manual/io/#dealing-with-changes-in-class-layouts-schema-evolution

This reference gives us many details that are helpful, but in short, we need to update the LinkDef file that specifies which classes are to be treated as event objects with additional "reading rules" that customize how the object currently in memory is to be updated with the information from the object residing on disk. There are other options, but the documentation discourages using the C++ API for Schema Rules, custom TClass, or custom TStreamer because they are more complicated for most use cases (I assume).

### Plan Outline
Currently, we have CMake generate a single LinkDef file after "registering" classes in various CMakeLists.txt. I don't want to update our CMake "registration" with this custom rule stuff and so I'm instead going to abandon CMake generating the LinkDef and have the LinkDefs committed into the repository. I actually have done this before (when I was experimenting with HDF5 writing [fire](https://github.com/LDMX-Software/fire/tree/trunk)), so this is not that big of a hurdle. Additionally, it will remove the very confusing double-pass nature of our current CMake infrastructure which few (if any) other developers are aware of. Completed by #1857 

Once the LinkDef files are in their Event folders and things are working in that mode, I can then go through the `git` history to find the source evolution of the event objects. This will inform my decision on minimum-supported event object versions (mapped to minimum-readable ldmx-sw version used to generate) and then I can start writing these reading rules. The good news is that re-naming variables is probably the simplest case of schema evolution that is not able to be handled automatically.

## Alternative Solution
We could _never_ do schema evolution and instead have v4.4.8 be a backwards-compatibility-breaking release. I don't like this since I think we will want to continue to evolve the schema and - upon further reflection and talking with other developers - I think maintaining readability of past samples is very important for validation and comparison to newer samples. I also think we will have other schema evolutions that are more complicated and thus more worthy of being backwards-breaking.

tomeichlersmith:

This is the first (of two) phases of #1815 where I refactor the event dictionary and CMake infrastructure:
- no more double-pass over the module CMakeLists.txt, that allows a pretty drastic simplification but means **we have to specify the event libraries that we depend on** (`SimCore::Event`) instead of inheriting from them being linked to framework.
  - This specification is particularly apparent for "stand-alone" processors as described in https://ldmx-software.github.io/using/analysis/ldmx-sw.html where users need to add `needs = 'Ecal_Event'` to their python config.
- I needed to refactor how GENIE was being configured in the CMake to get this to work. This was my whole day today (sorry @wesketchum for my grumpy messages!) but I persevered!
- While refactoring CMake, I also did some other cleanup work that I hope makes the CMake infrastructure easier to understand in the future.
  - updated how the Geant4 target was made, trying to make the `BuildMacros.cmake` file less all-encompassing.
  - I dropped support for Reco-only or Sim-only builds. I don't think folks use this much and it is mainly leftover from a time before the container images. It would be a pain to maintain since many of the "Reco" modules still depend on the simulation module via the event objects.
  - moved executables into `app` directories so they are not compiled into the library and as an executable
  - fix-up documentation comments at top of `cmake/Find` scripts

## Check List
- [x] Compiling, linking, and running basic simulation
- [x] ~Have a `GENIE::Reweight` and `GENIE::Generator` targets in CMake? They are two different projects...~ Not doing this because we don't use them separated. Our `GenieReweightProducer` uses stuff from both Reweight and Generator, so it needs both anyways. Separating them provides no help to us at this time.
- [x] Run validation samples from development build
- [x] Double check a production image build works. I mocked this by `mv build hide-build` so that it appeared like the build wasn't there (since it is not copied into the production image).
- [x] Remove `link off` lines in LinkDef files that are ignored in our version of ROOT (and future versions)
- [x] Update documentation of "stand-alone" processors https://ldmx-software.github.io/using/analysis/ldmx-sw.html


tomeichlersmith:
- [x] `ecal_pn`, `inclusive`, `reduced_ldmx`, `kaon_enhanced`, `signal` failing because the `TrigScint` library is not linked to the `TrigScint_Event` library
- [x] `target_genie` failing because `SimCore_Reweight` is not linked to `SimCore_Event`
- `deep_ecal_gun` completes and fails because of different character counts logs. The different linking appears to affect the last few digits in some of the numbers we are printing.

cmake/FindGENIE.cmake
```diff
@@ -0,0 +1,107 @@
...
+# (other targets are defined for specific GENIE libraries)
+#
+# GENIE::GENIE
+#
+
+if (GENIE_FOUND)
+  return()
+endif()
+
+set(GENIE_INCLUDE_DIRS "/usr/local/include/GENIE")
```
tomeichlersmith: Do we want to update this to not include the `GENIE` subdirectory? This would then force all tine `#include` statements to have `GENIE` in them when using `GENIE` stuff (which might be helpful?)

tomeichlersmith: While this is "Outdated" because I replaced the hard-coded `/usr/local` with the variable `GENIE_DIR` defined as `/usr/local` for consistency with other CMake Config scripts, my question remains. Looking for insight from @wesketchum or another developer.

cmake/FindGENIE.cmake
```diff
@@ -0,0 +1,108 @@
+#
+# FindGENIE.cmake
+#
+# Find an installation of GENIE. Highly tuned to the installation
+# of GENIE that is built into the ldmx/dev image.
+#
+# This will define the following variables
```
bryngemark, line 7: and GENIE_DIR? picky documentation comment :D

bryngemark (approved):
this is at the level that if you tell me this works, I believe you. out of curiosity, what was the deal with the versions in all the CMakeLists projects?

tomeichlersmith:
We started adding the `project` CMake statements (with the `VERSION` keyword) when the modules were pulled out into their own GitHub repositories with the idea that one could keep track of which versions of each of the modules they were using.

Without the modules residing in their own repositories, it does not make sense to me to keep track of their versions separately. We could even remove the `project` statements, but I left those in because some CMake configurations use the `PROJECT_*` variables which are defined only after calling the `project` function.

cmake/FindGENIE.cmake
tomeichlersmith: Actually nevermind, I tried this and GENIE's own headers use the include path without the `GENIE` prefix, so we need it in the include directories anyways. I'm not interested in updating the CMake config to try to make that be private, so I'm leaving this as-is.

tomeichlersmith:
Waiting to merge because I'm currently seeing an annoying runtime error when attempting to do a stand-alone analysis run.

```
tom@appa:~/code/ldmx/website/src/using/analysis$ just ../../../../ldmx-sw/fire gen.py                                 
denv_workspace="/home/tom/code/ldmx/ldmx-sw" denv fire gen.py                                                         
[ SimulatorBase ] 10  warn: Started 10 events to produce 10 events.                                                   
tom@appa:~/code/ldmx/website/src/using/analysis$ just ../../../../ldmx-sw/fire ana-cfg.py                             
denv_workspace="/home/tom/code/ldmx/ldmx-sw" denv fire ana-cfg.py                                                     
Processor source file /home/tom/code/ldmx/website/src/using/analysis/MyAnalyzer.cxx is newer than its compiled library
 /home/tom/code/ldmx/website/src/using/analysis/libMyAnalyzer.so (or library does not exist), recompiling...          
done compiling /home/tom/code/ldmx/website/src/using/analysis/MyAnalyzer.cxx                                          
Warning in <TClass::Init>: no dictionary for class pair<int,ldmx::SimParticle> is available                           
Warning in <TClass::Init>: no dictionary for class ldmx::SimParticle is available                                     
Warning in <TClass::Init>: no dictionary for class ldmx::SimTrackerHit is available                                   
Warning in <TClass::Init>: no dictionary for class ldmx::SimCalorimeterHit is available                               
Warning in <TClass::Init>: no dictionary for class ldmx::HgcrocDigiCollection is available                            
Warning in <TClass::Init>: no dictionary for class ldmx::EcalHit is available                                         
Warning in <TClass::Init>: no dictionary for class ldmx::CalorimeterHit is available     

... proceeds to seg fault on getEnergy ...

**Comment by tomeichlersmith** [comment]
I needed to load the libraries from with `Process` at runtime instead of just attempting to link them to the stand-alone processor's library.

I _could_ look more into appropriate link mechanics, but I don't want to. Now this is working and I've updated the stand-alone processor tutorial with a warning about linking issues potentially causing seg faults if a dictionary failed to load.

**Comment by tvami** [comment]
sorry @tomeichlersmith I'm a little confused about the scope...
I ran on an old ecal PN file with this branch and I get a lot like
```
[ Process ] 0  info: Opening file v3.3.3_ecalPN_tskim_resim_mc_v14-8gev-8.0GeV-1e-ecal_photonuclear_222.root
Info in <TBranchElement::InitializeOffsets>: TTree created with an older schema, some data might not be copied in 'slow-cloning' mode; fast-cloning should have the correct result. 'eventNumber_' is missing when constructing the branch 'EventHeader'. 
Info in <TBranchElement::InitializeOffsets>: TTree created with an older schema, some data might not be copied in 'slow-cloning' mode; fast-cloning should have the correct result. 'isRealData_' is missing when constructing the branch 'EventHeader'. 
Info in <TBranchElement::InitializeOffsets>: TTree created with an older schema, some data might not be copied in 'slow-cloning' mode; fast-cloning should have the correct result. 'intParameters_' is missing when constructing the branch 'EventHeader'. 
Info in <TBranchElement::InitializeOffsets>: TTree created with an older schema, some data might not be copied in 'slow-cloning' mode; fast-cloning should have the correct result. 'floatParameters_' is missing when constructing the branch 'EventHeader'. 
Info in <TBranchElement::InitializeOffsets>: TTree created with an older schema, some data might not be copied in 'slow-cloning' mode; fast-cloning should have the correct result. 'stringParameters_' is missing when constructing the branch 'EventHeader'. 
```

I guess this was not expected to work yet?

tomeichlersmith:
Yes, this is just the first preparation phase. I have not updated the ROOT dictionaries to allow for this to actually read files with old names.

tvami:
OK, let's run the CI one more time and then I'm ready to sign this

tvami (approved):
Tests pass, the PNet scores are little different, but not so that the KS would flag it

PR #1860, opened 2025-10-24
clang tidy ci patch

### LINKED ISSUE #1859: clang-tidy CI job patches
**Describe the bug**
Right now, the `clang-tidy` job fails during certain cases that are confusing because they don't have anything to do with `clang-tidy`.
- Failure to push refs: e.g. https://github.com/LDMX-Software/ldmx-sw/actions/runs/18229176974/job/51908027143
  - Either because the developer pushed another commit faster than `clang-tidy` took to run or because `clang-format` pushed a commit
- Attempt to push even though there were no changes: e.g. https://github.com/LDMX-Software/ldmx-sw/actions/runs/18106593744/job/51522585062
- Fail when no changes to C++ were made: e.g. https://github.com/LDMX-Software/ldmx-sw/actions/runs/17768972876/job/50499641618

## Proposed Solution
Merge the `clang-tidy` and `clang-format` CI workflows so that we only attempt to push commits to a branch once. This way the failing-to-push errors will only happen if a developer pushes new commits faster than the format+tidy workflow can run. Opening the door to tidy checks that do not have a auto-fix option is left to the future, probably just failing the workflow and writing the tidy report to `GITHUB_SUMMARY`.

Besides this bigger change, I'd want to
- add a check to make sure that we only run `clang-tidy` and `clang-format` when there are C++ changes
- run `clang-tidy` on all C++ files on `trunk` periodically (checking how long this takes now...)
- check if we need to build ldmx-sw or if we just need it to be configured, this could save ~11min

tomeichlersmith:

This resolves #1859 

I've updated the `justfile` with some `format` and `tidy` recipes that analyzie all of the C++ files (`*-all`) or just the files that differ between the current branch and `origin/trunk` (`*-diff`). These recipes are then used in the merged workflow.

The "clang-tools" workflow runs clang-tidy and clang-format:
- on every push to a PR, only looking at C++ files that differ with trunk
- on any workflow_dispatch (i.e. folks can launch it on their branch from the Actions tab)
- weekly on all C++ files on trunk. In this case, no commits are pushed, but we can hopefully notice the failure and remedy it from there. Probably just as easy as opening a new branch for the workflow to push to.

## Check List
- [x] both clang tools being run together
  - https://github.com/LDMX-Software/ldmx-sw/actions/runs/18844277145/job/53764142096?pr=1860
  - https://github.com/LDMX-Software/ldmx-sw/actions/runs/18844406985/job/53764589933?pr=1860
- [x] quickly skipping this branch since it has no changes to C++ files https://github.com/LDMX-Software/ldmx-sw/actions/runs/18843177551/job/53760309315
- [x] `clang-tidy` can run without compiling ldmx-sw, just needs to be configured: https://github.com/LDMX-Software/ldmx-sw/actions/runs/18789272617/job/53615270958
- [x] can push to non-`trunk` branches https://github.com/LDMX-Software/ldmx-sw/actions/runs/18843885989/job/53762786804?pr=1860


tomeichlersmith:
The `ecal_pn` is failing only because the last few digits of the particle net discriminator is changing and sometimes that means the number's trailing zero is trimmed and the character count in the logs change.

tvami (approved):
Thanks Tom!

PR #1866, opened 2025-11-03
Add skimming option to EcalClusterAnalyzer

### LINKED ISSUE #1865: Add filtering option for CLUE clusters in EcalClusterAnalyzer

I'd like to use `EcalClusterAnalyzer` as a skimming decision input, similar to https://github.com/LDMX-Software/ldmx-sw/blob/trunk/DQM/src/DQM/EcalDigiVerifier.cxx#L151-L155

```
  if (!inverse_skim_) {
    if (n_ecal_clusters > 0) {
      setStorageHint(framework::HINT_SHOULD_KEEP);
    } else {
      setStorageHint(framework::HINT_SHOULD_DROP);
    }
  } else {
    if (n_ecal_clusters > 0) {
      setStorageHint(framework::HINT_SHOULD_DROP);
    } else {
      setStorageHint(framework::HINT_SHOULD_KEEP);
    }
  }
```

This is not really needed as one could do it in an analysis too.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1865

## Check List

### DIFF
--- a/DQM/src/DQM/EcalClusterAnalyzer.cxx
+++ b/DQM/src/DQM/EcalClusterAnalyzer.cxx
@@ -17,6 +17,7 @@ void EcalClusterAnalyzer::configure(framework::config::Parameters& ps) {
   cluster_pass_name_ = ps.get<std::string>("cluster_pass_name");
 
   ecal_sp_hits_passname_ = ps.get<std::string>("ecal_sp_hits_passname");
+  inverse_skim_ = ps.get<bool>("inverse_skim", false);
   return;
 }
 
@@ -279,6 +280,20 @@ void EcalClusterAnalyzer::analyze(const framework::Event& event) {
   histograms_.fill(
       "clusterless_hits_percentage",
       100. * (ecal_rec_hits.size() - clustered_hits) / ecal_rec_hits.size());
+
+  if (!inverse_skim_) {
+    if (n_ecal_clusters > 0) {
+      setStorageHint(framework::HINT_SHOULD_KEEP);
+    } else {
+      setStorageHint(framework::HINT_SHOULD_DROP);
+    }
+  } else {
+    if (n_ecal_clusters > 0) {
+      setStorageHint(framework::HINT_SHOULD_DROP);
+    } else {
+      setStorageHint(framework::HINT_SHOULD_KEEP);
+    }
+  }
 }
 
 }  // namespace dqm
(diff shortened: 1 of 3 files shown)


JYoo001 (approved):

DQM/src/DQM/EcalClusterAnalyzer.cxx
```diff
@@ -17,6 +17,7 @@ void EcalClusterAnalyzer::configure(framework::config::Parameters& ps) {
   cluster_pass_name_ = ps.get<std::string>("cluster_pass_name");
 
   ecal_sp_hits_passname_ = ps.get<std::string>("ecal_sp_hits_passname");
+  inverse_skim_ = ps.get<bool>("inverse_skim", false);
```
tomeichlersmith: Maybe have this variable named more clearly? I didn't know that the "skim" that the "EcalClusterAnalyzer" does is "usually" for zero clusters (and the "inverse" would be for non-zero clusters).

Maybe name this `skim_for_zero_clusters_` and default it to `true`?

DQM/src/DQM/EcalClusterAnalyzer.cxx
```diff
@@ -279,6 +280,20 @@ void EcalClusterAnalyzer::analyze(const framework::Event& event) {
   histograms_.fill(
       "clusterless_hits_percentage",
       100. * (ecal_rec_hits.size() - clustered_hits) / ecal_rec_hits.size());
+
+  if (!inverse_skim_) {
```
tomeichlersmith: Why start with the `not` branch when you are doing both branches anyways? I suggest flipping them to make the code easier to reason about.

```cpp
if (skim_for_zero_clusters_) {
  if (n_ecal_clusters > 0) {
    setStorageHint(framework::HINT_SHOULD_DROP);
  } else {
    setStorageHint(framework::HINT_SHOULD_KEEP);
  }
} else {
  if (n_ecal_clusters > 0) {
    setStorageHint(framework::HINT_SHOULD_KEEP);
  } else {
    setStorageHint(framework::HINT_SHOULD_DROP);
  }
}

tomeichlersmith (changes requested):

DQM/src/DQM/EcalClusterAnalyzer.cxx
tvami: The default should be for the case when we keep the events with clusters, and `inverse_skim` is actually used elsewhere too in ldmx-sw, so I'd like to keep that. I also realized that other numbers might be interesting too so I'll add a `skim_threshold` variable and check if the `n_ecal_clusters` is higher or lower than `skim_threshold`. I'm open to other names instead of `skim_threshold`  tho!

DQM/src/DQM/EcalClusterAnalyzer.cxx
tvami: OK sure, I can do this

tvami: Done in https://github.com/LDMX-Software/ldmx-sw/pull/1866/commits/a02b74d791635d51f7bc619a37d8f3f79de7b39f

tomeichlersmith (approved):

PR #1869, opened 2025-11-11
Add TS light pipes to v15 geometry

### LINKED ISSUE #1760: Add light pipes to TS geometry
The TS scintillation light will be guided to the SiPMs in light pipes. 

I'd like the light pipe material be there, potentially as Sensitive Detectors, but at least so we can check the impact of additional material along the side of the incoming beam. Some details: 
- The SiPMs will all be located on the back of the tracker mounting structure, at the same position in _x_, while the TS pads shift in _x_ along _z_ to follow the incoming beam trajectory. Thus, the light pipes will be of slightly different lengths `L_at_pad3 + offset_at_pad_i` which can all be found from the `v14` geometry
- LYSO will also use light pipes for readout, but the LYSO bars are longer since they make up the target 
- The material is likely poly(methyl methacrylate) (PMMA), the UV transmitting kind (UVT) -- @awhitbeck will follow up 

We could wait, but we will need a complete decetor geometry description eventually. 

**Additional context**

Layan-Sarayra:

## Check List
- [ ] I successfully compiled _ldmx-sw_ with my developments.
- [ ] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).


bryngemark:
This would close #1760

Detectors/data/ldmx-det-v15-8gev/detector.gdml
```diff
@@ -36,6 +36,13 @@
          x = y = z = 0 mm.     
     -->
     <position name="trig_scint_pos" x="0.0" y="0.0" z="trig_scint_pos_z"/>
+
+    <!-- uncomment to add light pipes-->
+    <!-- might be helpful to turn on geometry checker to make sure there are no overlaps -->
```
bryngemark: have you done this?

Detectors/data/ldmx-det-v15-8gev/detector.gdml
```diff
@@ -94,11 +101,22 @@
...
         <positionref ref="magnet_pos"/> 
         <rotationref ref="identity"/> 
       </physvol>
+    <!-- uncomment to add light pipes-->
+      <!-- <physvol copynumber="9">
+        <file name="light_pipes_US.gdml"/>
+        <positionref ref="LP_pos_us"/>
+        <rotationref ref="identity"/>
+      </physvol>  
+      <physvol copynumber="10">
```
bryngemark: i'm not sure i understand the logic of having two different volumes for upstream and downstream (right?) light pipes. likewise for copy numbers. if you look at the TS, the trigscint volume (pad1 and 2) has one copy number (3).

Detectors/data/ldmx-det-v15-8gev/detector.gdml
```diff
@@ -94,11 +101,22 @@
         <positionref ref="hadron_calorimeter_pos"/> 
         <rotationref ref="identity"/> 
       </physvol>
-      <physvol>
+      <physvol copynumber="8">
         <file name="magnet.gdml"/> 
         <positionref ref="magnet_pos"/> 
         <rotationref ref="identity"/> 
       </physvol>
+    <!-- uncomment to add light pipes-->
```
bryngemark: i think we do want to uncomment these, unless you wanted to us the continuous integration tests to validate the geometry without lightpipes first (which might be a good idea).

bryngemark: i see now that the CI test fails only on the hcal validation test, so i think we can uncomment and see what happens.

bryngemark (commented):
this looks like a good start, i have a few questions about the implementation.

@tvami do we know why the hcal validation test is failing? i'm not sure i understand the plots to boot (in gold), especially the TS simhit ones. the test is using a gun upstream of the ecal, isn't it? i wondered why we don't see hits centered at x=0 in pad3 (at the target), but perhaps this just means all the TS hits are backscattered junk that is swept sideways by the magnet?

bryngemark:
@Layan-Sarayra also as a general practice, please follow the steps in the checklist at the top of the PR, and check the items once you have completed them. adding (to the PR discussion) any plots or printouts you have that show why you think your work is solving the issue is super helpful for the review, and is a great way to document your work for posterity.

Detectors/data/ldmx-det-v15-8gev/detector.gdml
Layan-Sarayra: It's because we added a 4th TS for the new geometry (will be called ldmx-reduced-v3), that complicated things when I added 4 light pipes - so I found that the easiest thing to do was to split into 2 + 2 (upstream & downstream) modules. I should've paid attention when I copied files here that here we have 3 pads and so should've made one file with 3 light pipes. Apologies! I will coordinate fixes with CJ.

tvami:
> Do we know why the hcal validation test is failing? i'm not sure i understand the plots to boot (in gold), especially the TS simhit ones. the test is using a gun upstream of the ecal, isn't it? i

Yes, it's shooting neutrons upstream of the ECAL face (220 mm). But I dont understand why it fails, I could imagine possible overlaps to cause issues, but that should be there for all cases. Let's try to test again

tvami:
changes in the hcal are unrelated, I see a big list of KS failing in https://github.com/LDMX-Software/ldmx-sw/actions/runs/19351298324/job/55364065010?pr=1874 and that PR really doenst change anything in the DQM plots

tvami:
So funny HCAL passed all good now
https://github.com/LDMX-Software/ldmx-sw/actions/runs/19352192552/job/55366685464?pr=1869

tvami (commented):
I was wondering if we should spell out what DS and US stand for, and if TS should also be explicitly added to the name since like as is it could be connected to any detector elements

tvami:
Retesting with new golds

tvami:
Closing this, as I understand https://github.com/LDMX-Software/ldmx-sw/pull/1888 does the same + the sensitive detectors setup

PR #1871, opened 2025-11-12
Elaborate log diff in check.sh

### LINKED ISSUE #1870: remove clutter in CI log comparisons
I'd like to remove some of the noise from the CI log comparisons, to make important changes stand out more. 
- Not comparing the timestamps is quite easy, and would remove a large number of lines from the diff (--> less scrolling). 
- That the reconstruction time that some producers log differs is often not useful but could be a very good warning. The way it is now, I doubt that anyone actually checks these numbers.

- some bash magic to skip timestamps in the diff
- we could consider removing printing of reco times to a lower log level but keep track of the total run time and print a warning if it blows up (can be caused by a one-off fluctuation but warrants investigation before it's ignored)

haven't really thought about it all that long

**Additional context**

bryngemark:

I added a sed replace to skip the HH:MM:SS timestamp at the beginning of a line in the log comparisons. For the reconstruction time comparison part, some more discussion is needed (comments on #1870 welcome).

This partially resolves #1870

## Check List
- [x] I ran my developments and the following shows that they are successful: CI tests work as intended

### DIFF
--- a/.github/actions/check/check.sh
+++ b/.github/actions/check/check.sh
@@ -27,7 +27,8 @@ __main__() {
   # unpack the logs so we can compare them
   tar xzf ${_archive} gold.log output.log
 
-  if ! diff gold.log output.log > log.diff; then
+  # use sed replace (by blank) to run the diff without the initial HH:MM:SS timestamp
+  if ! diff  -I '^#' <(sed -e 's/^[0-9]\{2\}:[0-9]\{2\}:[0-9]\{2\}/ /g' gold.log) <(sed -e 's/^[0-9]\{2\}:[0-9]\{2\}:[0-9]\{2\}/ /g' output.log) > log.diff; then
     # do not error out (don't set rc here) if diff is non-zero, the timestamps printed out
     # by some processors prevent a full text diff so we do the character
     # count check below to look for big changes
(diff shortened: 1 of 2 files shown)


tomeichlersmith (approved):
I love this. Thank you.

bryngemark:
so funnily enough, now the diffs don't show any timestamp differences, so the validation test that failed was **the log character count in ecalPN** -- i checked this in detail and often, there is a rounding difference where trailing zeroes would have preserved the character count. which one has more kinda flows back and forth; in the end, one of the logs ended up having 5 more characters than the other.

this brings up the idea to stop printing the ecal veto discriminator value to the log in the CI tests (and rely on the histogram KS test to catch differences in the BDT, which again, honestly is more likely to be caught). in practice, decrease the logger level of the ecalveto processor.

tvami:
It's the ParticleNet score not the BDT score right? But anyway, sure we can have that output demoted to `debug` from `info` and that will fix this issue,

tvami (approved):

bryngemark:
ok. thanks, @tvami, since we're at it i went ahead and demoted the particlenet printout to `debug` level. the CI tests will fail again because of that, of course.

once the tests are done i'm merging this.

tvami:
Thanks @bryngemark , looks good!

bryngemark:
i confirm failure caused by character count discrepancy in many tests. this change removes several tens of thousands of lines from our CI logfiles. pruning a few MB of data 🙂

PR #1878, opened 2025-11-18
Added reduced-v3 ECal changes

### LINKED ISSUE #1877: Create reduced_v3 Ecal geometry to match mini-LDMX geometry
For the upcoming mini-LDMX test run, we need a simulation geometry to match the mini-LDMX geometry.

A new geometry (`ldmx-reduced-v3`) that matches the mini-LDMX geometry.

N/A

**Additional context**
@SanjitMasanam will complete the reduced ECal but other detector components may need to be updated to match the true mini-LDMX geometry. This issue will establish a baseline that can be modified post-installation/post-beam

SanjitMasanam:

This resolves #1877

## Check List

Important ECal changes for reduced-v3 vs. reduced-v2: 
1. Only keeping central silicon module for each layer, rather than the entire flower
2. Removing all absorbers

To test these changes, I ran two simulations (and a check):
1. A straight 8 GeV muon gun sim. with vertex [0, 0, 239] mm (right in front of ECal) and smear of [1000, 1000] mm. The expectation for this simulations is that hits will register for the entire flower in v2 but only in the central module in v3. This would indicate only the central module is present in v3.
2. An straight 8 GeV electron gun sim. with vertex [0, 0, 239] mm and smear of [200, 200] mm. The expectation for this simulation is the # of hits will increase wrt the z-axis for v2 and hold constant for v3. This would indicate there are no absorbers in v3 because if there were, the electrons would brem and produce showers (i.e. # of hits would increase wrt to z).
3. The digi - rec hit z position histogram should primarily be a 0 for v3 which would indicate the layer positions in EcalGeometry.py are correct.

v2 results:

Image

Image

Image

v3 results:

image

image

image

Conclusion:
The results of both simulations (and the check) match expectation.


tvami:
@SanjitMasanam FYI at this point it's certain that we wont have 3 double layers... The 2 double layers are likely, but then reading all of them out is still in the air, but I guess we can deal with that part upstream of the GDML

SanjitMasanam:
> @SanjitMasanam FYI at this point it's certain that we wont have 3 double layers... The 2 double layers are likely, but then reading all of them out is still in the air, but I guess we can deal with that part upstream of the GDML

Ok I'll edit the code to be only 2 double layers

SanjitMasanam:
@tomeichlersmith @tvami, this PR is (approx.) ready. I was not able to go from 3 double layers to 2 because the GDML loops ([see most relevant loop for this instance](https://github.com/LDMX-Software/ldmx-sw/blob/a179e57e3bef12ce58582157db2778ebfee4f2d5/Detectors/data/ldmx-reduced-v3/ecal.gdml#L742)) don't like just having one iteration. The 3 double layers mean this loop iterates twice. If either of you know a workaround (other than making the loop stand-alone code which I considered but that would require rewriting many other loops as standalone which is more work than it's worth I think), it'd be great to have that workaround implemented. Otherwise, the current reduced_v3 has the first 2 double layers obviously so this subset would be accurate to the test run geometry.

tomeichlersmith (approved):
I think this is a good first approximation at least from an Ecal-perspective.

tvami:
Is it really hard to just move it out of the loop?

SanjitMasanam:
> Is it really hard to just move it out of the loop?

A bit yeah but doable so it has been done.

tvami:
Thanks @SanjitMasanam ! Please edit the CI to use this, and also I think we can delete v2

DetDescr/python/EcalGeometry.py
```diff
@@ -175,5 +175,19 @@ def reduced_v2() :
         eg.layer_shift_x = 2*eg.moduleMinR / eg.nCellRHeight
         return eg
 
+    def reduced_v3() :
+        eg = EcalGeometry(detectors_valid = ["ldmx-reduced-v3"],
+                gap = 1.5,
+                layerZPositions = [
+                      7.582, 16.062, 32.226, 40.206, 55.37, 63.35
```
tvami: I talked to Lincoln about the CAD model, there we have
```
86.00, 92.23, 136.8, 143.03 mm
```
I think we should edit that in the GDML too @SanjitMasanam

SanjitMasanam: Sure, those distances are from the target? Also for editing that in the GDML, do you want me to add empty air of suitable thickness between each layer to achieve those distances?

tvami: no this is from the front of the ECAL box. We are about 26 inch away from the "target" (there is no target), but that is to be measured tomorrow (fingers crossed).

Sounds good about the air, yeah! Thanks

SanjitMasanam: So 92.23-86=6.23 which is less than 16.062-7.582=8.48. How did that end up happening? I can add air for the other layers but I'll need some guidance on what to remove for the 2nd layer I think.

Detectors/util/ecal_silicon_pos_calc.py
```diff
@@ -22,12 +23,12 @@
...
-bilayer_absorber_cumulative = [0.0, 3.0, 8.0, 15.1, 22.2, 29.3, 36.4, 43.5, 50.6, 57.7, 64.8, 71.9, 85.9, 99.9, 113.9, 127.9, 141.9]
-front_tungsten_dz = [1.0, 2.0, 3.5, 3.5, 3.5, 3.5, 3.5, 3.5, 3.5, 3.5, 3.5, 7.0, 7.0, 7.0, 7.0, 7.0]
-cooling_tungsten_dz = [1.0, 1.5, 1.8, 1.8, 1.8, 1.8, 1.8, 1.8, 1.8, 1.8, 1.8, 3.5, 3.5, 3.5, 3.5, 3.5]
+bilayer_absorber_cumulative = [0.0, 0.0, 0.0, 15.1, 22.2, 29.3, 36.4, 43.5, 50.6, 57.7, 64.8, 71.9, 85.9, 99.9, 113.9, 127.9, 141.9]
+front_tungsten_dz = [0.0, 0.0, 0.0, 3.5, 3.5, 3.5, 3.5, 3.5, 3.5, 3.5, 3.5, 7.0, 7.0, 7.0, 7.0, 7.0]
+cooling_tungsten_dz = [0.0, 0.0, 0.0, 1.8, 1.8, 1.8, 1.8, 1.8, 1.8, 1.8, 1.8, 3.5, 3.5, 3.5, 3.5, 3.5]
 
 # Calculated constants
-Flower_dz = PCB_dz+Glue_dz+Si_dz+GlueThick_dz+CarbonBasePlate_dz
+Flower_dz = PCB_dz+Glue_dz+Si_dz+Si_extra_dz+GlueThick_dz+CarbonBasePlate_dz
```
tomeichlersmith: These changes appear to be specific to `reduced-v3` so when (inevitably) someone wants to update the full ecal geometry, they will attempt to use this script and be very confused. Can you add a `print` mentioning this is for `reduced-v3` and leave the old arrays in the file commented-out?

tomeichlersmith (changes requested):
I think the GDML looks good, I just want to update the helper script so that it is less confusing for a future user.

DetDescr/python/EcalGeometry.py
tvami: Sorry @SanjitMasanam  I dont get why you'd need to add Air? 6.23 < 8.48, no? So you'd need to remove stuff. And I think that's the `CarbonBasePlate_dz` we dont have a carbon base plate at all, but let's bother with that point later, for now just reduce it's size to things add up. Thanks!

SanjitMasanam: @tvami Do you know what the ecalFrontZ is actually (in software it's 240 mm)? You said about 26 in but do you have the exact measurement?

tvami: 26 inch

tvami: well I was too fast, 26 inch was the distance to the "target". That is 660.4 mm. So the 240 is either in cm, or it's not measured from the same point, since I'd expect the reduced geometry to have a smaller number than the full geometry

cjbarton151: Is there a place where the z measurements between each of the subsystems have been recorded? We need these numbers also for the HCal and for the TS pads eventually. Perhaps we could even have a fine-tuning of the z values once all of the components are implemented, and just drop them in so that they fit for the time being.

SanjitMasanam:
```
<!-- 
    Distance from target to the ECal parent volume.
    The calorimeter is an additional .5 mm downstream at 240.5 mm.
-->
<constant name="ecal_front_z" value="24.0*cm"/>
```
It looks like the ECal is 24 cm away currently. Perhaps it is measured from a different point but I'm just going to leave it as is and then we can fine-tune z-values once all components are implemented as mentioned.

SanjitMasanam:
@tomeichlersmith @tvami This PR is ready to be reviewed/merged.

tomeichlersmith (approved):

tvami:
```
2026-01-21T04:54:28.0180564Z [ fire ] 3434 fatal: [BadConf] : Relative cell coordinates (-77.32, 39.20) mm derived from world coordinates (39.20, 77.32) mm with layer = 3 and module = 0 are outside module hexagon
2026-01-21T04:54:28.0182360Z   at /home/runner/work/ldmx-sw/ldmx-sw/DetDescr/src/DetDescr/EcalGeometry.cxx:158 in getID
2026-01-21T04:54:28.0183638Z [ fire ] 3434 debug:     0 /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libDetDescr.so(+0x335d0) [0x7f29832f75d0]
2026-01-21T04:54:28.0184890Z     1 simcore::EcalSD::ProcessHits(G4Step*, G4TouchableHistory*) + 469 
2026-01-21T04:54:28.0185772Z     2 G4SteppingManager::Stepping() + 492 
2026-01-21T04:54:28.0186706Z     3 G4TrackingManager::ProcessOneTrack(G4Track*) + 284 
2026-01-21T04:54:28.0193186Z     4 G4EventManager::DoProcessing(G4Event*) + 1098 
2026-01-21T04:54:28.0193781Z     5 G4RunManager::ProcessOneEvent(int) + 43 
2026-01-21T04:54:28.0194366Z     6 simcore::Simulator::produce(framework::Event&) + 433 
2026-01-21T04:54:28.0195036Z     7 framework::Process::process(int, int, framework::Event&) const + 174 
2026-01-21T04:54:28.0195627Z     8 framework::Process::run() + 10687 
2026-01-21T04:54:28.0196056Z     9 fire(+0x412c) [0x56477cd6b12c]
2026-01-21T04:54:28.0196574Z    10 /lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca) [0x7f298c0fe1ca]
```

Not looking great...

tvami:
OK adding back the shift in the odd layers fixes it; I think this is good enough for now

tvami (approved):

PR #1888, opened 2025-12-01
New features for v15 geometry (TSPads and associated parts)

### LINKED ISSUE #1760: Add light pipes to TS geometry

### LINKED ISSUE #1761: Add TS SiPMs as G4 volumes
As a step leading up to radiation studies for the TS, we should add SiPMs as volumes. The idea is that we can export Geant4 geometries to FLUKA for these studies. 

Add a piece of silicon corresponding to a 2x3 (? dimensions TBC) SiPM at the far end of each light pipe. If we make them Sensitive Detectors, we could even study what hits them already in `ldmx-sw` (especially check for neutrons).

N/A

**Additional context**

cjbarton151:
TS Pad components added as active volumes under the TrigScintSD type. TS Pad component loops removed from other files and now generated automatically as assembly files by new script "makeassemblyfiles.py". GDML styles converted from online to local access.

_This (hopefully) resolves #1760 and #1761_

There may be some adjustments to the dimensions of the lightpipes later on, but that's not as important as the other changes introduced in this PR.

## Check List

image
image


tvami:
Screenshot 2025-12-01 at 18 51 24

I think the geometry couldnt be constructed. At least the CI ends with a segfault. @cjbarton151 has anything changed since you tested it locally?

tvami:
> has anything changed since you tested it locally?

and more: when you tested did you do the full chain or just upto SimHits?

tvami:
I can reproduce the issue (at SLAC), I also increased the log level to see where it dies:
```
[ SimulatorBase ] 0 trace: Reading in geometry from '/sdf/group/ldmx/users/tamasvami/v15-tscomps-sensitive/ldmx-sw/install/data/detectors/ldmx-det-v15-8gev/detector.gdml'
Aborted (core dumped)
```

Detectors/data/ldmx-det-v15-8gev/detector.gdml
```diff
@@ -3,7 +3,7 @@
 <!ENTITY constants SYSTEM "constants.gdml">
 <!ENTITY materials SYSTEM "materials.gdml">
 ]>
-<gdml xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://service-spi.web.cern.ch/service-spi/app/releases/GDML/schema/gdml.xsd">
+<gdml xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="gdml.xsd">
```
tvami, line 6: I think when we do the compile this needs to be moved to the install directory, we do that in
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Detectors/CMakeLists.txt
maybe if you add `*xsd` to the lines where the `*gdml` is ?

cjbarton151: I suspected it was related to trying to make the XSD files local. This change isn't necessarily part of the v15 geometry update, but rather a QOL feature to make `g4-vis` work. I could remove this change and separate it out into another PR, but I'll try your fix first

tvami:
I guess this didnt fix it :(

cjbarton151:
Well, I tried a fresh pull of this branch on cosmos, and tried to make sure I wasn't "cheating" with any extra setup in my local environment, and multiple validation tests still executed just fine (I didn't check the plots though). I guess the next step is to revert the changes making the XSD files local and see if we still have a problem.

tvami:
Instead of merge you should rebase @cjbarton151

tvami:
@cjbarton151 
```
[ GEANT4 ] 0 error: 
  -------- EEEE ------- G4Exception-START -------- EEEE -------
  *** G4Exception : ReadError
        issued by : G4GDMLReadStructure::GetVolume()
  Referenced volume '' was not found!
  *** Fatal Exception *** core dump ***
  -------- EEEE -------- G4Exception-END --------- EEEE -------
```

Detectors/data/ldmx-det-v15-8gev/target.gdml
```diff
@@ -63,34 +63,23 @@
...
-      <solidref ref="trigger_bar_box"/>
-      <auxiliary auxtype="VisAttributes" auxvalue="TriggerPadVis"/>
-      <auxiliary auxtype="DetElem" auxvalue="TriggerPad"/>
-    </volume>
+    <assembly name="tspad3_assembly">
+          <physvol>
+            <file name="tspad3_assembly.gdml" />
+            <rotationref ref="identity" />
+	  </physvol>
+    </assembly>
```
tvami: Maybe the volume is missing here?

Detectors/data/ldmx-det-v15-8gev/makeTSassemblyfiles.py
tvami, line 1: Can you rename this so it reflects that this is for the TS assembly only?

tvami:
The ASAN failure seems completely random, let's test again
https://github.com/LDMX-Software/ldmx-sw/actions/runs/20017170405/job/57397002315?pr=1888

cjbarton151:
🤔 if it failed randomly, it did so in a similar way both times... Maybe this is something I can test locally as well.

cjbarton151:
With the recent changes to trunk, it looks like the ASAN is happy.

Would you like me to remove the XSD files/offline functionality for the GDML specifications and move them to another issue/PR, as we briefly discussed in-person? Or would you rather keep them in this PR?

tvami:
> Or would you rather keep them in this PR?

Let's just keep them in here but connect the issue to this PR, this seems like the easiest solution

Detectors/data/ldmx-det-v15-8gev/materials.gdml
```diff
@@ -16,30 +15,43 @@
...
-  <MEE unit="eV" value="173"/>
-  <D unit="g/cm3" value="2.32999877404956"/>
-  <fraction n="1" ref="Si"/>
-</material>
-<material name="Vacuum" state="gas">
-  <MEE unit="eV" value="19.2"/>
-  <D unit="g/cm3" value="9.99999473841014e-09"/>
-  <fraction n="1" ref="H"/>
-</material>
+
```
tvami, line 130: After https://github.com/LDMX-Software/ldmx-sw/pull/1899 goes in this part you'll have to rebase

cjbarton151:
Once the validation checks finish, I think this should be ready for review and hopefully merging.

bryngemark:
@cjbarton151 what happened when you got the PF hit association branch development tagged onto this work?

cjbarton151:
Somehow the commits got mixed up while I was rebasing. It should be fixed now.

Detectors/data/ldmx-det-v15-8gev/makeTSassemblyfiles.py
```diff
@@ -0,0 +1,215 @@
+import sys
```
bryngemark, line 1: most of your longer comments have your voice 🙂 add your name as author to the top so we know who's speaking (and to credit, and where to direct questions, this is a novel enough addition to how we do these things).

SimCore/python/sensitive_detectors.py
```diff
@@ -172,6 +176,24 @@ def pad2() :
     def pad1() :
         return TrigScintSD(2,'Pad1','trigger_pad1_bar_volume')
 
+    def pad1lg() :
```
bryngemark, line 187: the case convention we've been using suggests that these should be `Pad1LightGuide`, `Pad1SiPM`... i think

bryngemark (commented):
This looks sensible. But judging from code only is hard; thanks for the screen shots, they are promising, and I'd like some more plots if you don't mind:
- capturing the placement of the light guides and the SiPM, perhaps relative to the TS pads or something (so we won't have to remember what _x_ they're at), 
- seeing that the number of hits is reasonable (I expect fewer hits in them as they are outside the beam spot)
- verifying that the assembly gives the same placement of bars as the old copy-paste way
- if you do have `g4-vis` working again, a visualization would be great! 

Including plots like these with interpretation is also our best way of documenting the reasoning behind changes we do to our code base for posterity. Hang in there, this is looking close to being done.

cjbarton151:
[PR1888-spotchecks.pdf](https://github.com/user-attachments/files/24781832/PR1888-spotchecks.pdf)

I've attached a PDF with some (very) basic images from g4-vis and plots or partial plots from an output of a PN background script I was using for something else.

All of the g4-vis images have 2 views: one facing along the direction of the beamline, and a more isometric view of the same. Each set of images use the same zoom and angle, except for the SiPMs-only images, which include a 3/4 view with no raising and a side view, since these components are harder to visualize.

You may notice an extra white rectangle in the new geometry visualizations. This is an extra envelope introduced for each of the TSPad assemblies. It's tightly wrapped in the x and z direction (calculated from the dimensions of the objects inside), while the y direction is a little looser. The z direction shrink-wrapping is to prevent interference with other envelopes along the beamline, such as the tagger envelope, and the x direction shrink-wrapping would allow for the eventual introduction of the aluminum frame that will attach the SiPMs to their electronic readouts and hold the TSPads/taggers in place.

The plots are not representative of any particular physics result, and meant as a quick hit-counting sanity check. As expected, there are significantly more hits in the TSPad scintillator bars than in the light guides, and the light guides in TSPad3 have significantly more hits than the other light guides due to increased scattering off the target or other extra material further down the beamline. Also as expected, the SiPMs have significantly fewer hits than even the light guides, with the most being in the Pad3 SiPMs with 22 counts (the Pad1 SiPMs had just 5 counts).

If there are any other checks or details that I could expand on, don't hesitate to ask!

cjbarton151:
I'll also add a new commit with authorship and PR reference at the top of the new `makeTSassemblyfiles.py` script, an extra bit of compatibility with the reduced geometries that I overlooked previously, and updated naming conventions on the new collections.

Detectors/CMakeLists.txt
```diff
@@ -5,12 +5,23 @@ set(detector_install_path ${CMAKE_INSTALL_PREFIX}/data/detectors)
...
   if(IS_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/data/${detector})
     #this is a directory, assume it is a detector we need to update and install
-    #  go through the gdml files here and replace relative paths with full paths
     set(detector_dir ${CMAKE_CURRENT_SOURCE_DIR}/data/${detector})
+    #If directory contains any python scripts, execute those first
+    #This is intended for use to generate GDML files that need some extra parameters
+    #but can be extended to any additional scripts which need to run before building
+    file(GLOB py_scripts RELATIVE ${detector_dir} ${detector_dir}/*.py)
+    foreach(py_script ${py_scripts})
+      execute_process(#execute_process is always called PRE-build
```
tvami, line 20: @tomeichlersmith are you OK with this?

tvami:
I'm not sure what happened here, this is supposed to take about 55 min, not more than 360 min

Screenshot 2026-01-22 at 07 24 00

Detectors/CMakeLists.txt
tomeichlersmith: Yea, I think it makes sense. I think we could use a similar tactic with other subsystems as well since there is a lot of repetition.

tvami:
I restarted the signal CI. Otherwise I looked thru the 800 plots we have KS differences in and they all look compatible within the uncertainties, so after the signal finishes, I'm good to approve this, thanks for your work CJ!

tvami:
Screenshot 2026-01-22 at 11 21 37

Something is going wrong here; and other PRs dont have this issue so likely coming from this PR

cjbarton151:
This was a tricky one... While playing around with the validation script locally using 100 events, I wasn't receiving any output at all, and it was taking too long per event (about 10 events/minute, way too slow to get through the 20000 events in the CI test).

Re-examining the logs from the attempt in the CI test, I noticed that there was only one event output printed to terminal every ~100-200 events! Considering there is a maximum of 10000 attempts at a valid electron per event, this means 1-2 million electrons were being sampled before having a single valid track.

After more digging, I tracked down a [single-line change](https://github.com/LDMX-Software/ldmx-sw/commit/0c0e6815d4ad4d80aa492c7ba708581a7c139921) between v14 and v15 in `target.gdml`:

image

Here's where I have to admit that my spot-checks weren't thorough enough... I did front-view and isometric checks of `trig_scint.gdml`, `target.gdml`, and both together. I also did a side-view check of `trig_scint.gdml`, but I didn't do a side-view check of `target.gdml`. Here it is:

image

This extra few mm of space has caused the target and the TSPad3 to overlap in my implementation. Here it is without the change:

image (1)

This explains needing to sample 1-2 million electrons to get any tracks reaching the detectors - my guess is that they would need to scatter in such a way as to bend around the messed up part of the geometry. What doesn't make sense, though, is why this only occurs in the signal CI test? Shouldn't it occur in all of the tests, or in none of them?

Anyways, would it be OK to revert the target air gap to the v14 definition, or should I find a way to make the TSPad3 copacetic with the v15 definition?

tvami:
We need the gap so we are can shove in the LYSO. So the idea is that in `ldmx-det-v15-8gev` we have a gap, and then in `ldmx-lyso-r4-v15-8gev` that gap is filled with LYSO (which actually means the changes in this PR should be repeated on that geometry too -- this is for later tho)

I agree it's super weird only the signal CI picks this up... The signal goes thru the `G4DarkBreM` which is a difference from the rest of them, but still I would have expected this to come up in all cases. Maybe @tomeichlersmith has an idea.

cjbarton151:
OK, the full explanation is nuanced and not really worth discussing in detail here, but it turns out that the TargetDarkBremFilter has been removing almost every event in the signal CI test. When I moved the TSPads to assemblies, I assigned all the pads + their components to the 'trig_scint' region of the geometry. This was incorrect, because TSPad3 is  considered part of the 'target' region. I believe that, with this setup, an electron had to make it to the target without entering TSPad3 to be accepted, hence the abysmally low rate. I'm still working on this, but now that the hard part of identifying a cause is over, I'm hoping the solution will be easy.

tomeichlersmith:
The signal sample is the only validation sample currently that biases and filters for a specific process to occur within the target *region*. (The Ecal PN wants a hard brem, but that does not require biasing and is easy to check without referencing the region.)

@cjbarton151 has found the correct issue - while we logically can distinguish the TS and Target from each other, the TS Pad3 which is immediately upstream of the target, _cannot_ be reliably distinguished in a real physics event. If we want the TS and Target to be in separate regions, that's fine but then the biasing/filtering code needs to be updated. I would prefer to leave TS Pad 3+Target in the same region to make the C++ easier. TS Pad 1 and Pad 2 could be in a new region with the tagger tracker, maybe named "tagger"? Or not in any region - we don't use production cuts and so the only reason to use regions is to make biasing/filtering easier.

tvami (approved):
Thanks a lot @cjbarton151 ! The CI tests all pass, they show a lot of small changes due to the GEANT seed differences, I think all this looks good! Thanks again for your excellent work!

Detectors/data/ldmx-det-v15-8gev/constants.gdml
```diff
@@ -43,8 +43,11 @@
 <constant name="trigger_pad_dim_x"          value="target_dim_x" />
 <constant name="trigger_pad_dim_y"          value="target_dim_y" />
 <constant name="trigger_bar_dx"             value="30"/>
-<constant name="trigger_bar_dy"             value="3.05"/>   
+<constant name="trigger_bar_dy"             value="3.15"/>
```
bryngemark: verifying: this is the most recent agreement on what we'll have (and not just what we had in the ESA modules)?

Detectors/data/ldmx-det-v15-8gev/constants.gdml
```diff
@@ -67,6 +77,11 @@
 <constant name="trig_scint_area_envelope_z" 
           value="-trigger_pad1_z + trigger_pad_thickness + 2*clearance"/>
 
+<!-- Light-pipe bar dimensions for TS (2 x 2 x 36.24 mm) -->
+<constant name="light_pipe_dx"        value="36.24*mm"/>
+<constant name="light_pipe_dy"        value="2*mm"/>
+<constant name="light_pipe_thickness" value="2*mm"/>
+
```
bryngemark, line 99: In comments you say you copy the TS bar dimensions, but these look like the SiPM dimensions and some bar length. 

I think we really do want them to be the same dimensions as the scintillator bars in _y_ and _z_. For _x_, we need different lightpipe lengths for each pad: they should reach all the way from the SiPM cards mounted on the back of the tracker+TS support plate, out to the individual pad _x_ positions.

I thought I had mentioned that but perhaps it was only in Layan's reduced geometry PR. 
Majd will know a good number to put in for `Pad3`, and for the rest it's just adding the displacement in _x_.

bryngemark (commented):
Thanks @cjbarton151 , this is a big improvement over the clunky repeated loops. And good detective work -- both on you and @tvami for catching the subtle CI failure. 

For my education, are we using replicas/parametrized volumes anywhere? 

My main sticky point is the one about light guide dimensions. My take is that we might as well get the geometry of this added element right (as much as possible) from the start, but if you have a reason why we should modularize it, let me know.

Detectors/data/ldmx-det-v15-8gev/constants.gdml
cjbarton151: @mghrear would you kindly share a good number for the light pipe length on pad 3, i.e. the distance between the edge of the TSPad scintillator bars and the support plate?

mghrear: Hi CJ, it looks like you have the correct dimensions for the light pipes used in the beam test. Also, the length of the light guides (36.24 mm) is the distance between the SiPMs and the Scintillator bars, which is longer than the distance between the scintillator bars and support plate (24.19 mm).

cjbarton151: Hey Majd - so 36.24 mm is the distance between the SiPMs and the scintillator bars at the target? And then the scintillator bars for TSPad1 and TSPad2 will be longer?

mghrear: Yes, 36.24 mm is the distance between the SiPMS and scintillator bars in the beam test setup. This should be a good number for Pad3 in the v15 geometry. As LK states, it makes sense to add a displacement in x for Pad1 and Pad 2, which could be achieved by making the light pipes for Pad1 and Pad2 longer.

tvami:
About the indexes here. 
```
def pad3() :
        return TrigScintSD(3,'Pad3','trigger_pad3_bar_volume')
    def pad2() :
        return TrigScintSD(1,'Pad2','trigger_pad2_bar_volume')
    def pad1() :
        return TrigScintSD(2,'Pad1','trigger_pad1_bar_volume')
    def pad1lg() :
        return TrigScintSD(6,'Pad1LightGuide','tspad1_lightpipe_volume')
    def pad2lg() :
        return TrigScintSD(5,'Pad2LightGuide','tspad2_lightpipe_volume')
    def pad3lg() :
        return TrigScintSD(7,'Pad3LightGuide','tspad3_lightpipe_volume')
    def pad1sipm() :
        return TrigScintSD(9, 'Pad1SiPM','tspad1_sipm_volume')
    def pad2sipm() :
        return TrigScintSD(8, 'Pad2SiPM','tspad2_sipm_volume')
    def pad3sipm() :
        return TrigScintSD(10,'Pad3SiPM','tspad3_sipm_volume')
```

Sorry for my ignorance, but will a hit know if it actually hit the light guide, the SiPM or the Pad? I thought it wouldnt, and then this should be
```
    def pad3() :
        return TrigScintSD(3,'Pad3','trigger_pad3_bar_volume')
    def pad2() :
        return TrigScintSD(1,'Pad2','trigger_pad2_bar_volume')
    def pad1() :
        return TrigScintSD(2,'Pad1','trigger_pad1_bar_volume')
    def pad1lg() :
        return TrigScintSD(2'Pad1LightGuide','tspad1_lightpipe_volume')
    def pad2lg() :
        return TrigScintSD(1,'Pad2LightGuide','tspad2_lightpipe_volume')
    def pad3lg() :
        return TrigScintSD(3,'Pad3LightGuide','tspad3_lightpipe_volume')
    def pad1sipm() :
        return TrigScintSD(2, 'Pad1SiPM','tspad1_sipm_volume')
    def pad2sipm() :
        return TrigScintSD(1, 'Pad2SiPM','tspad2_sipm_volume')
    def pad3sipm() :
        return TrigScintSD(3,'Pad3SiPM','tspad3_sipm_volume')
```

cjbarton151:
As it turns out, there was a pair of bugs latent in the last commit. These bugs had no effect so long as the TSPad modules all remained identical dimensionally, but activated once I started varying the light pipe dimensions between pads. In the process of debugging this PR, I discovered latent bugs in other pieces of the geometry, but those will be characterized and fixed in a future PR.

Attached you can find some visualizations of the TSPad + target geometries. The SiPMs are all aligned along x now, with the light pipes extending from the scintillator bars to the SiPMs in their new positions. I verified that there are still SimHits being produced in the bars and the pipes, but didn't run enough events to see any hits in the SiPMs. The geometry _looks_ correct, but I haven't yet done verification of the x,y,z positions in the output data. Nevertheless, I wanted to push a commit by end of week, and I can follow up on validation on Monday.

[v15-geometry-checks_20-02-2026.pdf](https://github.com/user-attachments/files/25448434/v15-geometry-checks_20-02-2026.pdf)

Detectors/data/ldmx-det-v15-8gev/constants.gdml
cjbarton151: I missed this comment before -  I was indeed using the slice test specifications for the TSPads, thinking that they're identical to the final geometry. If this is something which will change, is there a source I can consult for the final design specs?

bryngemark: @cjbarton151 i've asked @latompkins , the bars used in ESA were actually 2.9mm to fit the SiPM spacing, but i'm not sure if this was a one-off thing

bryngemark:
> Sorry for my ignorance, but will a hit know if it actually hit the light guide, the SiPM or the Pad? I thought it wouldnt, 

@cjbarton151 we made these separate sensitive detectors, right? so then i assume the simhit knows where it hit. 

also i don't remember why the index ordering is swapped between pad1 and 2 but i _think_ it's an artifact of when we rearranged the pads such that the one downstream of the target was moved upstream of everything; this made comparisons of `v12` (`up`, `down`, `tag`,) and `v14` (`pad3`, `pad1`, `pad2`) easier. if it's not a big deal in the downstream processors i think we should make the index ordering more logically follow the number in the name.

bryngemark:
> Attached you can find some visualizations of the TSPad + target geometries. The SiPMs are all aligned along x now, with the light pipes extending from the scintillator bars to the SiPMs in their new positions. ... The geometry _looks_ correct, but I haven't yet done verification of the x,y,z positions in the output data. ...
>
> [v15-geometry-checks_20-02-2026.pdf](https://github.com/user-attachments/files/25448434/v15-geometry-checks_20-02-2026.pdf)

Thanks for making these! I agree it looks correct in some views but others are confusing; some places it looks like the bars are the closest to the support plate. Maybe it's caused by their location in negative `x`, this occasionally throws visualizations off. The view from `+y` with `+z` pointing upward (and `+x` to the left) should be, in exaggerated ASCII art that is not to scale: 

|._~
|
|
|
|.__~
|.___~
|

**Legend**: 
. SiPM
| support plate (not included in geometry and actually surrounding SiPMs)
_ light guide
~ TS bar 

If you confirm that this is what you see then I'm happy with this, pending confirmation on the bar `y` dimension.

Detectors/data/ldmx-det-v15-8gev/constants.gdml
cjbarton151: Updating this to 3.0 based on Slack communications

cjbarton151:
I've updated the y dimensions of the scintillator bars/light pipes and also straightened out the swapped ordering of the module IDs in the components.

> If you confirm that this is what you see then I'm happy with this, pending confirmation on the bar y dimension.

I'm as sure as I can reasonably be with the current visualization tools I'm using. In the near-middle term I may see if we can fix/update the visattributes so it's easier to differentiate various parts and to display things more unambiguously.

tvami:
>  we made these separate sensitive detectors, right? so then i assume the simhit knows where it hit.

The SimHit, sure, but what about the RecHit? I'm asking because the moduleIDs are determined from the index that's used in the SD here. See here https://github.com/LDMX-Software/ldmx-sw/blob/e5dbf6964d16eded2af3e2319f95f7f0cdc8cefa/SimCore/python/sensitive_detectors.py#L149 and then the RecHit just copies this until the end. Maybe it's fine to have this separated at the SimHit level and then "merge" at the RecHit level?

cjbarton151:
Ah, I see now you also asked in a previous comment about the full chain past the SimHits... I never intended to add any functionality past the SimHits stage, since the light pipes and SiPMs aren't really active detector components. My understanding was that we simply would like to know more about how much radiation or accidental activation we may have in the light guides/SiPMs during beam operation.

tvami:
Ok so I read this as in "this is good at SIM level as is, and the needed reco changes should be decoupled / to be dealt with in another PR"

bryngemark:
> My understanding was that we simply would like to know more about how much radiation or accidental activation we may have in the light guides/SiPMs during beam operation.

I confirm that this is what we intended. We just need to gauge radiation levels. And as you say @tvami, if we ever change our mind, we can add a reconstruction step (but likely if we ever wanted this, we'd pull in contributions from e.g. radiation on the SiPMs themselves into TS hit reconstruction directly with special treatment).

bryngemark (approved):
Looks good

tvami:
Shall we cut a new gold and compare this PR to a clean state? Or just merge it w/o any last final final testing?

tvami:
Looks good
https://github.com/LDMX-Software/ldmx-sw/actions/runs/22694482397/job/65798974991?pr=1888

PR #1898, opened 2025-12-09
Fast fix for out-of-range array in EcalClusterAnalyzer.cxx.

### LINKED ISSUE #1744: Bug in EcalClusterAnalyzer when running with en sample
**Describe the bug**

At around event 45 with seed = 1 when running the GENIE EN config I get

```
double free or corruption (out)
Aborted (core dumped)
```

**To Reproduce**
Just run the 
```
.github/validation_samples/target_genie/config.py
```
and add 
```
dqm.EcalClusterAnalyzer()
```
in the sequence and it will come up in the first 50 events.

**Desired behavior**

Smooth running. 

**Screenshots**
Image

**Environment:**
```
denv_name="ldmx"
denv_image="ldmx/dev:latest"
denv_mounts=""
denv_shell="/bin/bash -i"
denv_network="true"
apptainer version 1.2.5
LDMX_NUM_EVENTS=100
LDMX_RUN_NUMBER=1
DISPLAY=localhost:17.0
DENV_NAME=ldmx
DENV_RUNNER=apptainer
DENV_IMAGE=ldmx/dev:latest
```

cjbarton151:

When running with the address sanitizer enabled, a heap-buffer-overflow is detected for any event with 0 reconstructed electron tracks reaching the ECal. This is due to an array which has a size based on the number of reconstructed electron tracks going out of range during the clustering.

It's important to note that the clustering algorithm runs, even if there are no primary electrons reaching the ECal. I think this is correct, because other particles can still reach the ECal. A possible improvement would be to not run the ECalClusterAnalyzer (or any part of the ECal reco/DQM) for events with NO depositions in the ECal, but this should be a very small fraction of events and would negligibly impact runtime.

Editing to add: this (hopefully) resolves #1744 , although the error states look slightly different. Also, we may want to consider increasing the number of events in the asan-ubsan validation test to a significantly higher number.

## Check List

Before:
image
After:
image

### DIFF
--- a/DQM/src/DQM/EcalClusterAnalyzer.cxx
+++ b/DQM/src/DQM/EcalClusterAnalyzer.cxx
@@ -211,7 +211,7 @@ void EcalClusterAnalyzer::analyze(const framework::Event& event) {
     // for each cluster
     // total number of hits coming from electron, index = electron ID
     std::vector<double> n_hits_from_electron;
-    n_hits_from_electron.resize(nbr_of_electrons + 1);
+    n_hits_from_electron.resize(nbr_of_electrons + 2);
     // total number of energy coming from electron, index = electron ID
     std::vector<double> energy_from_electron;
     energy_from_electron.resize(nbr_of_electrons + 1);
--- a/DQM/src/DQM/EcalClusterAnalyzer.cxx
+++ b/DQM/src/DQM/EcalClusterAnalyzer.cxx
@@ -214,7 +214,7 @@ void EcalClusterAnalyzer::analyze(const framework::Event& event) {
     n_hits_from_electron.resize(nbr_of_electrons + 2);
     // total number of energy coming from electron, index = electron ID
     std::vector<double> energy_from_electron;
-    energy_from_electron.resize(nbr_of_electrons + 1);
+    energy_from_electron.resize(nbr_of_electrons + 2);
     double energy_sum = 0.;
     double n_sum = 0.;
 


tvami:
I'm a little confused why it doesn't fail on 
```
energy_from_electron.resize(nbr_of_electrons + 1);
```
now, it should be the same issue as in the line you fixed

tvami:
But also with this fix, now it says there is one ECAL CLUE cluster, where is that from? Are you running the ecal PN config? (so  we know it's a photon)

cjbarton151:
> it should be the same issue as in the line you fixed

Almost! One was inside of a loop over nelectrons (which is 0), the other was not.

image

I haven't looked deeply into how the analyzer is actually supposed to work, just looking at it from a code perspective, so I don't know whether or not they both belong in that loop or if neither belongs in the loop.

> But also with this fix, now it says there is one ECAL CLUE cluster, where is that from? Are you running the ecal PN config? (so we know it's a photon)


I've only been testing with the ecal_pn config, since that was the default for the sanitizer validation. Although now that you mention it, it might behoove us to do all of the validation tests (with higher statistics) for the sanitizers.

tomeichlersmith (approved):
Excellent and detailed study :) I'd be happy to merge this (although I've lost track on why the validation samples are failing due to testbeam craziness)

tvami:
I still think that the current way of doing it
```
    n_hits_from_electron.resize(nbr_of_electrons + 2);
    // total number of energy coming from electron, index = electron ID
    std::vector<double> energy_from_electron;
    energy_from_electron.resize(nbr_of_electrons + 1);
```
is asymmetric the least, and I'd make it uniform. Maybe @bryngemark has a comment too?

bryngemark:
i don't really have a comment, except that some of this code was always a little 2e-centric and it's no surprise we run into trouble in more challenging 1e-cases. i haven't spent enough time on @cjbarton151 's detailed analysis to understand why one needs changing and not the other, but sure, add 2 to both of them to keep it coherent.

bryngemark:
pending that potential fix i'd approve it but i'm sure @tvami will too once it's in.

tvami (approved):

Issue #1376, opened 2024-07-17
Use Geant4-object pointer comparisons instead of string-comparisons where possible


EinarElen:
We have a lot of code in SimCore and Biasing which does things like 
```C++

auto MyThing {track->GetThing()->GetThingName()};
if (MyThing.compare(SomeString)) {
   // Do the work here
}
```

Where thing can be something like
- A particular process, e.g. photoNuclear 
- A particular geometry region, e.g. "CalorimeterRegion"
- A particular physical/logical volume, e.g. "hadronic_calorimeter"

This has two big downsides 
1. It is way too easy to mess this up and introduce bugs. For example: 
https://github.com/LDMX-Software/ldmx-sw/blob/28fd69677fca6250588d02b182661c24156f1e60/Biasing/src/Biasing/EcalProcessFilter.cxx#L101-L111

This is a bug. There is no physical volume called `hcal_PV` so that condition will never trigger. The volume in question happens to be called "hadronic_calorimeter" which differs from all the others that have names like `tagger_PV` etc. If we are using pointers as our comparison tool, we can enforce at configuration time that these are initialized and throw otherwise (so issues like the hcal name here would be caught). 

2. String comparisons can be _really_ slow, especially for anything that is called ~at every Geant4 step (biasing operators and stepping actions). I made a quick and dirty comparison and there was a ~30% performance improvement by moving from string comparisons to pointer comparisons for EcalPN samples (!!!) 

I think the best way to do this would be to have a helper type in SimCore that can be used to get the various pointers people might need and make sure that things like biasing wrappers are handled correctly. 

```C++
class Geant4ObjectUtility // Someone please suggest a better name 
{ 
   const G4VProcess* GetPhotonuclearProcess() const {
   return GetProcess(G4Gamma::Definition(), "photonNuclear");
   }
   const G4VProcess* GetProcess(G4ParticleDefinition* particle, std::string processName) const {
         const auto manager {particle->GetProcessManager()};
      const auto processes {manager->GetProcessList()};
      for (int i{0}; i < processes->size(); ++i) {
        const auto process {(*processes)[i]};
        const auto name {process->GetProcessName()};
        if (name.contains(processName)) {
          return process;
        }
      }
      return nullptr;

   }
  G4Region* GetRegion(std::string name) {
  return G4RegionStore::GetInstance()->GetRegion(name);
  }
};
```
etc

tomeichlersmith:
[...]

PR #1899, opened 2025-12-10
Cleaned up material definitions+updated trackers to compare against v15

### LINKED ISSUE #1892: Clean up material definitions for v15 geom

When letting Geant4 complain (it does :P ) it gives this list
```
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Air
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Carbon
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Silicon
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Vacuum
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Air
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Carbon
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Silicon
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Vacuum
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Vacuum
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Tungsten
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Polyvinyltoluene
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Vacuum
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Polyvinyltoluene
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Air
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Carbon
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Silicon
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Vacuum
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material G4_Al
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material FR4
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Glue
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material G4_Si
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material CarbonEpoxyComposite
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material G4_AIR
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material G4_W
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material G4_C
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material G4_AIR
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Steel
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Scintillator
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Vacuum
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Fe
[ GEANT4 ] 0  warn: G4Material WARNING: duplicate name of material Cu
```

Clean these up.

Do nothing and risk some inconsistencies

**Additional context**

https://github.com/LDMX-Software/ldmx-sw/issues/1750

cjbarton151:

In v14, geometry definitions are kind of all over the place in more than one sense. They're located in various files, inconsistently named, and often re-defined. Materials exported from other software may even have slightly different material properties, like density, composition, or atomic mass.

Compounding this issue, many python scripts used for validation, tracking etc. are still pointing to 'ldmx-det-v14-8gev' or 'ldmx-det-v14-8gev-nocals'. I don't know if v15 is stable enough for all of the python scripts yet, but using a script which references v14 but then uses v15 causes ldmx-sw to load both geometries and then apply one. However, the material definitions are still loaded, causing all the warnings.

This (mostly) resolves #1892 . Read below if you want to know why it doesn't fully resolve it, or you can skip the next paragraph.

I don't consider this fully resolved for 3 reasons.
1) As you can see in the screenshots, the detector is still set twice in the ecal_pn test (I haven't run the other validation tests manually but I assume there will be similar behavior). This is related to the next point,
2) There is still exactly _one_ re-definition of each material. On the bright side, this is kind of nice because it lists all of the defined materials in the terminal output during runtime - but is still unintended. I wasted some time trying to track (pun intended) down this issue, and it seems to have something to do with the tracking, specifically `geo.py` and/or `full_tracking_sequence.py`. I don't want to deal with this (or at best deal with it in a separate issue) so I'm not investigating further.
3) There are other python scripts in the repo that are still hardcoded to use v14 instead of v15. Again, I don't feel like re-validating all of these scripts with v15, and that should be a separate issue anyway.

## Check List

Before:
image
image

After:
image

### DIFF
--- a/Detectors/data/ldmx-det-v15-8gev/ecal_motherboard5_assembly.gdml
+++ b/Detectors/data/ldmx-det-v15-8gev/ecal_motherboard5_assembly.gdml
@@ -275,7 +275,7 @@
 
   <structure>
     <volume name="nohole_motherboard5_assembly">
-      <materialref ref="FR40x27264f0"/>
+      <materialref ref="FR4"/>
       <solidref ref="nohole_motherboard5_assembly-SOL"/>
     </volume>
   </structure>
--- a/Detectors/data/ldmx-det-v15-8gev/hcal.gdml
+++ b/Detectors/data/ldmx-det-v15-8gev/hcal.gdml
@@ -43,122 +43,8 @@
                   -(ecal_side_dy/2.0-hcal_scintWidth/2.0)
                   ecal_side_dy/2.0-hcal_scintWidth/2.0"/>
   </define>
-    
-  <materials>
-    <isotope N="16" Z="8" name="O16">
-      <atom unit="g/mole" value="15.9949"/>
-    </isotope>
-    <isotope N="17" Z="8" name="O17">
-      <atom unit="g/mole" value="16.9991"/>
-    </isotope>
-    <isotope N="18" Z="8" name="O18">
-      <atom unit="g/mole" value="17.9992"/>
-    </isotope>
-    <element name="O">
-      <fraction n="0.99757" ref="O16"/>
-      <fraction n="0.00038" ref="O17"/>
-      <fraction n="0.00205" ref="O18"/>
-    </element>
-    <isotope N="12" Z="6" name="C12">
-      <atom unit="g/mole" value="12"/>
-    </isotope>
-    <isotope N="13" Z="6" name="C13">
-      <atom unit="g/mole" value="13.0034"/>
-    </isotope>
-    <element name="C">
-      <fraction n="0.9893" ref="C12"/>
-      <fraction n="0.0107" ref="C13"/>
-    </element>
-    <isotope N="1" Z="1" name="H1">
-      <atom unit="g/mole" value="1.00782503081372"/>
-    </isotope>
-    <isotope N="2" Z="1" name="H2">
-      <atom unit="g/mole" value="2.01410199966617"/>
-    </isotope>
-    <element name="H">
-      <fraction n="0.999885" ref="H1"/>
-      <fraction n="0.000115" ref="H2"/>
-    </element>
-    <material name="G4_C" state="solid">
-      <T unit="K" value="293.15"/>
-      <MEE unit="eV" value="81"/>
-      <D unit="g/cm3" value="2"/>
-      <fraction n="1" ref="C"/>
-    </material>
-    <isotope N="14" Z="7" name="N14">
-      <atom unit="g/mole" value="14.0031"/>
-    </isotope>
-    <isotope N="15" Z="7" name="N15">
-      <atom unit="g/mole" value="15.0001"/>
-    </isotope>
-    <element name="N">
-      <fraction n="0.99632" ref="N14"/>
-      <fraction n="0.00368" ref="N15"/>
-    </element>
-    <isotope N="36" Z="18" name="Ar36">
-      <atom unit="g/mole" value="35.9675"/>
-    </isotope>
-    <isotope N="38" Z="18" name="Ar38">
-      <atom unit="g/mole" value="37.9627"/>
-    </isotope>
-    <isotope N="40" Z="18" name="Ar40">
-      <atom unit="g/mole" value="39.9624"/>
-    </isotope>
-    <element name="Ar">
-      <fraction n="0.003365" ref="Ar36"/>
-      <fraction n="0.000632" ref="Ar38"/>
-      <fraction n="0.996003" ref="Ar40"/>
-    </element>
-    <material name="G4_AIR" state="gas">
-      <T unit="K" value="293.15"/>
-      <MEE unit="eV" value="85.7"/>
-      <D unit="g/cm3" value="0.00120479"/>
-      <fraction n="0.000124000124000124" ref="C"/>
-      <fraction n="0.755267755267755" ref="N"/>
-      <fraction n="0.231781231781232" ref="O"/>
-      <fraction n="0.0128270128270128" ref="Ar"/>
-    </material>
-    <isotope N="55" Z="25" name="Mn55">
-      <atom unit="g/mole" value="54.938"/>
-    </isotope>
-    <element name="Mn">
-      <fraction n="1" ref="Mn55"/>
-    </element>
-    <isotope N="54" Z="26" name="Fe54">
-      <atom unit="g/mole" value="53.9396"/>
-    </isotope>
-    <isotope N="56" Z="26" name="Fe56">
-      <atom unit="g/mole" value="55.9349"/>
-    </isotope>
-    <isotope N="57" Z="26" name="Fe57">
-      <atom unit="g/mole" value="56.9354"/>
-    </isotope>
-    <isotope N="58" Z="26" name="Fe58">
-      <atom unit="g/mole" value="57.9333"/>
-    </isotope>
-    <element name="Fe">
-      <fraction n="0.05845" ref="Fe54"/>
-      <fraction n="0.91754" ref="Fe56"/>
-      <fraction n="0.02119" ref="Fe57"/>
-      <fraction n="0.00282" ref="Fe58"/>
-    </element>
-    <material name="Steel" state="solid">
-       <T unit="K" value="293.15"/>
-       <D unit="g/cm3" value="7.87"/>
-       <fraction n="0.9843" ref="G4_Fe"/>
-       <fraction n="0.014" ref="G4_Mn"/>
-       <fraction n="0.0017" ref="G4_C"/>
-    </material>
-    <material name="Scintillator" state="solid">
-       <T unit="K" value="293.15"/>
-       <!-- <MEE unit="eV" value="64.7494480275643"/> -->
-       <D unit="g/cm3" value="1.032"/>
-       <fraction n="0.91512109"  ref="G4_C"/>
-       <fraction n="0.084878906" ref="G4_H"/>
-    </material>
-  </materials>
 
-
+  
   <solids>
   
     <!-- - - - - - - - - HCal solids - - - - - - - -  -->
@@ -240,7 +126,7 @@
       
       <!-- - - - - - - - -  FULL HCAL STRUCTURES - - - - - - - -  -->
     <volume name="hadronic_calorimeter">
-      <materialref ref="G4_AIR" />
+      <materialref ref="Air" />
       <solidref ref="hadron_calorimeter_Box" />
       <!-- BACK HCAL  -->
         <!-- Note the step size of 2  -->
--- a/Detectors/data/ldmx-det-v15-8gev/materials.gdml
+++ b/Detectors/data/ldmx-det-v15-8gev/materials.gdml
@@ -1,6 +1,36 @@
+<!--
+This file is organized in the following way:
+
+Element definitions
+Pure (one element) material definitions
+Composite material definitions
+
+We don't currently have any enriched elements in LDMX, but
+if we did, I would add those between the element definitions
+and the pure material definitions. And don't worry about the
+isotopic composition of the natural elements - GDML is at
+least smart enough to assign those abundances automatically
+
+Both the elements and the pure materials are organized by
+Z number in ascending order. Also, not all elements have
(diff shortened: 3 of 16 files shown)


Tracking/python/geo.py
```diff
@@ -46,7 +46,7 @@ def __init__(self):
             raise Exception('TrackersTrackingGeometryProvider is a singleton class and should only be retrieved using get_instance()')
         else: 
             super().__init__('TrackersTrackingGeometry', 'tracking::geo::TrackersTrackingGeometryProvider', 'Tracking')
-            self.setDetector('ldmx-det-v14-8gev-no-cals')
+            self.setDetector('ldmx-det-v15-8gev')
```
tvami: Let's keep it as it was, the point of this is that the tracker doenst need to see the calorimeters for the tracking geometry, so it made Matt's studies much faster, in the end we should make a `ldmx-det-v15-8gev-no-cals` but that's only the relevant when the tracker geometry will change

Detectors/data/ldmx-det-v15-8gev/hcal.gdml
```diff
@@ -240,7 +126,7 @@
       
       <!-- - - - - - - - -  FULL HCAL STRUCTURES - - - - - - - -  -->
     <volume name="hadronic_calorimeter">
-      <materialref ref="G4_AIR" />
+      <materialref ref="Air" />
```
tvami, line 129: I dont know what the difference is, did you move away from the G4 default on purpose?

Detectors/data/ldmx-det-v15-8gev/ecal_motherboard5_assembly.gdml
```diff
@@ -275,7 +275,7 @@
 
   <structure>
     <volume name="nohole_motherboard5_assembly">
-      <materialref ref="FR40x27264f0"/>
+      <materialref ref="FR4"/>
```
tvami, line 278: Is this just a simple rename? I'm worried if the FR4 existed somewhere else and the `FR40x27264f0` was done on purpose to make sure that it's unique. But it was before me, so maybe @tomeichlersmith can chime in

Detectors/data/ldmx-det-v15-8gev/hcal.gdml
cjbarton151: The GDML geometry builder doesn't have access to the Geant4 default materials intrinsically*
That material is named G4_AIR but is defined in GDML the same way every other material is:
image

Most likely scenario: someone made that segment of the HCal in a different G4 simulation, then exported it. The other definition of "Air" (which I just took from one of the other files) doesn't include the trace carbon, and the other parameters are different by about a factor of 0.1%


*although there may be a way I don't know about to invoke G4NistManager::FindOrBuildMaterial from a GDML material definition 🤔

Detectors/data/ldmx-det-v15-8gev/ecal_motherboard5_assembly.gdml
cjbarton151: This again looks like a material which was exported from another program. Ironically, that automatic naming convention on export is precisely to prevent duplicate material definitions... But this was the only place FR4 was defined in v14.

tvami:
OK did you see we have a segfault?
```
  
  ===========================================================
  There was a crash.
  This is the entire stack trace of all threads:
  ===========================================================
  #0  0x00007f8bcd2a07a7 in __GI___wait4 (pid=31, stat_loc=stat_loc
  entry=0x7ffd90c9cf28, options=options
  entry=0, usage=usage
  entry=0x0) at ../sysdeps/unix/sysv/linux/wait4.c:30
  #1  0x00007f8bcd2a08eb in __GI___waitpid (pid=<optimized out>, stat_loc=stat_loc
  entry=0x7ffd90c9cf28, options=options
  entry=0) at ./posix/waitpid.c:38
  #2  0x00007f8bcd1e85bb in do_system (line=<optimized out>) at ../sysdeps/posix/system.c:172
  #3  0x00007f8bcd983583 in TUnixSystem::StackTrace() () from /usr/local/lib/libCore.so
  #4  0x00007f8bcd982ef4 in TUnixSystem::DispatchSignals(ESignals) () from /usr/local/lib/libCore.so
  #5  <signal handler called>
  #6  0x00007f8bcaa36cdd in simcore::geo::GDMLParser::read() () from /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libSimCore_Geo.so
  #7  0x00007f8bcaa972a2 in simcore::SimulatorBase::buildGeometry() () from /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libSimCore.so
  #8  0x00007f8bcaa98a5f in simcore::SimulatorBase::configure(framework::config::Parameters&) () from /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libSimCore.so
  #9  0x00007f8bcdc7b34c in framework::Process::Process(framework::config::Parameters const&) () from /home/runner/work/ldmx-sw/ldmx-sw/install/lib/libFramework.so
  #10 0x0000555ef63cf0b9 in main ()
```

cjbarton151:
Boh, I accidentally typed <!~~ instead of <!-- when making a squashed commit where all I was trying to do was add a comment. Since all I did was add one comment block I didn't recompile and rerun before pushing. The typo didn't show up on my display because emacs doesn't enter xml editing mode automatically when editing materials.gdml (it does for all the other .gdml files, though...). Common emacs L. Should work now.

tvami:
> Common emacs L. Should work now.

One point to `vim` then :P

Detectors/data/ldmx-det-v15-8gev/ecal_motherboard5_assembly.gdml
tomeichlersmith: Can confirm, the auto-naming added this suffix and it was not left in out of anything besides laziness on my part.

tomeichlersmith (commented):
I think this looks good - one thing I'm worried about is that we often use the material and/or volume name within the biasing and filtering code to check if we are within a certain region of the detector. In a better world, we would actually using Geant4 regions more and update the biasing/filtering code, but that requires a lot of validation that folks have been avoiding.

Have you checked if the biasing/filtering looks for any of these material names? (I don't think it does, I think it looks at volumes but I'm uncertain now.)

cjbarton151:
Oof, good call... I didn't check for hardcoded material matches in the C++ or Python code. I've gone through now and traced all instances of `GetMaterial()` to check how they're used in the code (and I did a case-insensitive search so I could also find instances of `PtrRetrieval::getMaterial()` ). 

There was a warning in the `kaon_enhanced` validation test (which uses biasing) that I hadn't seen before:
image
Hardcoded volume matches (the majority of hardcoded matches in ldmx-sw, it seems) should be unaffected by the changes I've made in this PR, and I kept all material names the same except in two cases:
1) The material had a unique random string assigned to the end of its name (as in the case Tamas mentioned of FR4)
2) The material had a name like G4_AIR or G4_Si
In the first case, Geant4 is smart enough (don't know the details) to strip off the random string, so 'FR40x27264f0' actually becomes 'FR4' internally during runtime. Cleaning up these material names is more aesthetic than anything.
In the second case, though, there was the potential for issues.

I've updated the few instances of hardcoded material matching that were affected by this PR (changing 'G4_w' to 'Tungsten', for example), the warning is gone and AFAIK everything is running smoothly. Also, I reverted the geometry version in `geo.py` as Tamas suggested.

SimCore/src/SimCore/BiasOperators/PhotoNuclear.cxx
```diff
@@ -70,7 +70,7 @@ G4VBiasingOperation* PhotoNuclear::ProposeOccurenceBiasingOperation(
         std::max(em_xsec_unbiased + pn_xsec_unbiased_ - pn_xsec_biased_,
                  pn_xsec_unbiased_);
     static auto material_tungsten =
-        simcore::g4user::ptrretrieval::getMaterial("G4_W");
+        simcore::g4user::ptrretrieval::getMaterial("Tungsten");
```
tvami: With this change we loose backward compatibility, in this sense that we wont be able to run v14 geom anymore. Maybe rename Tungsten in v14 too? Or somehow make this code geometry version dependent? Probably easier to just rename W in v14 too

tvami (changes requested):

SimCore/src/SimCore/BiasOperators/PhotoNuclear.cxx
cjbarton151: Good catch. I made `material_tungsten` non-static and now, if PtrRetrieval can't find "Tungsten", it looks for "G4_W". This should keep it backwards compatible for all versions which use G4_W. I did similar for the other changes (involving G4_Si and G4_AIR).

tvami (approved):
Looks good! I'm not sure if we had a reason for the material to be `static` earlier but I think this should be good now

SimCore/src/SimCore/BiasOperators/PhotoNuclear.cxx
```diff
@@ -69,8 +69,11 @@ G4VBiasingOperation* PhotoNuclear::ProposeOccurenceBiasingOperation(
     double em_xsec_biased =
         std::max(em_xsec_unbiased + pn_xsec_unbiased_ - pn_xsec_biased_,
                  pn_xsec_unbiased_);
-    static auto material_tungsten =
-        simcore::g4user::ptrretrieval::getMaterial("G4_W");
+    auto material_tungsten =
+        simcore::g4user::ptrretrieval::getMaterial("Tungsten");
+    // For versions < v15
+    if (material_tungsten == nullptr)
+      material_tungsten = simcore::g4user::ptrretrieval::getMaterial("G4_W");
```
tomeichlersmith: Please leave the `static` in the declaration. The `getMaterial` function is slow since it has to do a string comparison across the materials loaded into memory but the materials do not change after the geometry is loaded so we only need to look up this pointer once and then we can use the pointer for the rest of the run.

**Context**
https://github.com/LDMX-Software/ldmx-sw/pull/1633
https://github.com/LDMX-Software/ldmx-sw/issues/1376

Detectors/data/ldmx-det-v15-8gev/materials.gdml
```diff
@@ -1,6 +1,36 @@
...
+as seen on the periodic table. Pure elemental materials
+should have the formal name listed on the periodic table
+(I guess we'll use the US standard named like 'aluminum'
+instead of 'aluminium' because Geant4 is US-based and
+uses those standards). Composite materials can have
+whatever name you want, but I would recommend using
+names that are coloquially understandable, like how we
+use "glue" instead of listing the chemical formula.
+
+-->
```
tomeichlersmith, line 30: this is awesome, thank you for such detailed work!!!

tomeichlersmith (commented):
Excellent work, just requesting to put the `static` back in for potential-performance related reasons.

SimCore/src/SimCore/BiasOperators/PhotoNuclear.cxx
cjbarton151: No problem, I reverted it

tomeichlersmith (approved):

PR #1904, opened 2025-12-18
Decoder for ESA TS data format

### LINKED ISSUE #1883: ESA slice test TS data format+decoder/encoder
The data format bundling the QIE output is slightly changed in the ESA slice test compared to the CERN 2022 testbeam. With the different beam clock approach, there is no longer a need for words like "time since spill", but a different time stamp is introduced.

We are still using the CERN testbeam data, so that functionality needs to be retained. I will transfer these (rename) to a specific set of classes. 

I have considered making word positions etc configurable, but my conclusion is that since the number of future users of TS data from the ESA slice test is larger than the number using the old data, the benefits are small compared to the confusion the possibility of changing data format in python configs could easily introduce.

bryngemark:

This mostly resolves #1883 . I have not written an encoder -- or rather, I started but it's not a priority [1] for reading ESA data. In the end I decided to leave the CERN testbeam related code intact (except for minor example config changes) and use new names for the ESA format.

`ZCCMOutput.h` sets up this data format, as communicated by @EBerzin 
Screenshot 2025-12-18 at 15 01 03

### Some notes on running:
- to strip off all the Rogue metadata and filter on the channel containing TS data, first run the extraction script I got from @tomeichlersmith, tweaked a little bit and added to (freshly created) `Recon/util`:
```
$ denv init ghcr.io/slaclab/rogue:v6.6.2
$ denv python3 rogue-extract.py data_20251208_225935_dark_current_kicker_trigger.dat -s 2 -c ts -n 1000 -o output_file_name.dat
```
TS is here selected by name as the correct `c`ontributor, `s`ubsystem is 2. `-n` for number of events.
- You will use the output file from this step, but first: there is a "handshake" frame of (event header+) one time sample from each lane at the start of run. This must be stripped, as this event is not the same length [2] as the others. Easiest way I've found is to just remove those bytes from the top of the file like so:
```
tail -c +$((0x101)) output_file_name.dat output_file_name_truncated.dat 
```

- Then, as there is no longer a setup script that defines an LDMX_BASE in the shell/session, I have taken to using a simple export:
```
 export LDMX_BASE=/path/to/ldmx-sw-mother-dir
```
but fixes to this that use `denv` properly are welcome.

- Another remainder from last testbeam setup is that the full run script runs over several files in a file list, if desired. So make one:
 ```
ls path/to/output_file_name_truncated.dat > fTest
```
and then it should be ready:
```
. /path/to/TrigScint/util/runReformatting_ESA2025sliceTest.sh -f fTest -e 500 -r 32 -c channelMap_4modules_14lanes.txt -m moduleMap_4modules_14lanes.txt 
```
where `-e` is for number of `e`vents and `-r` is for hit `r`econstruction start sample.

## Check List

I can decode 50k events without any errors or warnings, I can extract gains and pedestals, and get a tiny number of hits with hit reconstruction (7 samples wide starting at 32, which I have no confirmation, just very brief trial and error) in Pad 1, bar 14 + adjacent channels.

PE_pad1bar14_dec8_50kEvents
PEvsBarID_dec8_50kEvents

[1] I will make a separate issue for this, and I think it's a neat exercise, for someone who wants to use the simulated ESA geometry once it's available, to invert the decoder. 
[2] Current ROOT conversion expects fixed-length samples (to be improved in the new year).

### CODE REVIEW

File: TrigScint/include/TrigScint/Event/TrigScintQIEDigis.h [hunk 1]
```diff
@@ -107,6 +127,10 @@ class TrigScintQIEDigis {
   int chan_id_;
   /// channel ID
   int elec_id_{-1};
+  /// module ID (pad)
+  int module_id_;
+  /// lane ID (fiber)
+  int lane_id_{-1};
```
tomeichlersmith, line 155: Increase ClassDef version number since this changes the data stored in the class.

TrigScint/exampleConfigs/runRawUnpackerZCCM.py
```diff
@@ -0,0 +1,54 @@
...
+lenMessage=lenTSflagsAndLane+lenEmpty+lenADCandTDC
+# find the number of messages (one per lane)
+nLanes=nChannels/6     
+nBytes=lenHeader + nLanes*nTimeSamples*lenMessage
+
+
+from LDMX.Packing import rawio
+
+p.sequence = [
+    rawio.SingleSubsystemUnpacker(raw_file = RAWfileName, output_name = "ZCCMoutput", detector_name="ldmx-hcal-prototype-v1.0", num_bytes_per_event = int(nBytes))
```
tomeichlersmith, line 46: ah, I see - is this where the "fixed length" requirement comes from? If it is, that's good news. I think I can pretty easily avoid that.

Packing/util/rogue-extract.py
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python3
```
tomeichlersmith, line 1: Maybe move to `Packing/util`?

I also am a bit concerned about running this accidentally within the ldmx-sw dev image. Maybe add extra usage comments here and in the `description` of the `parser`?

TrigScint/exampleConfigs/runZCCMDecode.py
```diff
@@ -0,0 +1,61 @@
...
+p.outputFiles = [sys.argv[2]]
+inputPass=sys.argv[3]
+nSamp=int(sys.argv[4])
+logName= p.outputFiles[0].replace(".root", "_toLDMX.log")
+
+
+if len(sys.argv) > 5 :
+    mapFile=sys.argv[5]
+else :
+    mapFile="toyChannelMap_4modules_14lanes.txt"
```
tomeichlersmith, line 22: Instead of providing the mapping on the command line, you could have it installed and then the Python configuration can be configured to look into the install location. This is particularly helpful for these kinds of small(ish) data files that are relatively static. In the future, we should probably change this to be a "condition" so that it can be versioned and dependent on run number, but let's not make the perfect the enemy of the good.

Example: Ecal's BDT Models
- `setup_data` in `Ecal/CMakeLists.txt` to get the `Ecal/data` files copied to the install location
- `@CMAKE_INSTALL_PREFIX@` in `Ecal/python/makePath.py` which is replaced by the full install path when copying the python configurations into the install location

TrigScint/python/zccmFormat.py
```diff
@@ -0,0 +1,84 @@
...
+
+class ZCCMDecoder(ldmxcfg.Producer) :
+    """Configuration for ZCCM decoder"""
+    def __init__(self, mapFile, name = 'ZCCMDecode'): 
+        super().__init__(f'{name}', 'trigscint::ZCCMDecoder', "TrigScint")
+        self.name = name 
+        self.input_pass_name = ''
+        self.input_collection = 'ZCCMoutput'
+        self.output_collection = 'decodedZCCM'
+        self.channel_map_file = mapFile
```
tomeichlersmith, line 59: See `Ecal/python/vetos.py` for usage, but you could have the `mapFile` be refereneced by name and then the full path is deduced. Obviously users can still manually override the deduced path with a full path of their own if they desire.

TrigScint/util/runReformatting_ESA2025sliceTest.sh
```diff
@@ -0,0 +1,149 @@
...
+
+echo "trigger flags set to ${trigFlag}"
+
+echo "using identifier $id"
+
+#we will assume a few env variables are set: LDMX_BASE (from sourcing ldmx-env.sh) and the appropriate fire command (prepended by ldmx only when running locally) set by the calling script. 
+reformatDir=${PWD}
+configDir=${LDMX_BASE}/ldmx-sw/TrigScint/exampleConfigs
+utilDir=${LDMX_BASE}/ldmx-sw/TrigScint/util
+digiDir=${LDMX_BASE}/ldmx-sw/run
```
tomeichlersmith: If you are running this from inside the container, you can avoid manually setting `LDMX_BASE` by inspecting `HOME`.

See the code that deducees the ldmx-sw install location within the container: https://github.com/LDMX-Software/dev-build-context/blob/3316272fb6f18e1ab8f888c211c207060e7acd60/ldmx-env-init.sh#L13-L18

If this is being run outside the container, you could avoid requirement of this environment variable by using the location of this script as the reference. (Technically this would work inside the container as well but I like reminding folks that the container's HOME is set to a helpful value.)

```bash
# requires to be run by bash
ldmx_sw_dir="$( cd -- "$(dirname -- "${BASH_SOURCE[0]}")/../../" && pwd)"
```

tomeichlersmith (commented):
No real comments on the decoding, I'm hoping to update `Packing` with better support for the Rogue binary file format but I'm unwilling to commit to a timeline for that.

TrigScint/exampleConfigs/runRawUnpackerZCCM.py
```diff
@@ -0,0 +1,54 @@
...
+        ]
+
+p.maxEvents = int(nEv) # this HAS TO be set! single subsystem unpacker will abort all events when it runs out of data so it cannot simply be left running
```
EBerzin, line 49: Just to clarify--this means that we have to pre-specify the number of events in the file? Is there a way to just process all of the events?

TrigScint/include/TrigScint/Event/ZCCMOutput.h
```diff
@@ -0,0 +1,171 @@
...
+   * Store tdcs of all time samples
+   * @param tdc_ array of tdcs
+   */
+  void setTDC(const std::vector<int> tdc) { tdcs_ = tdc; }
+
+  /**
+   * Store cids of all time samples
+   * @param cid_ array of cids
+   */
+  void setCID(const std::vector<int> cid) { cids_ = cid; }
```
EBerzin, line 93: I think it might be useful to also store the information from the other flags somewhere. There was one QIE card that was misbehaving quite a bit during some of the later runs. The data is still useable, but I think will be misaligned by a few samples from the other lanes. This is something the BC0 or the CE could flag.

TrigScint/exampleConfigs/runRawUnpackerZCCM.py
tomeichlersmith: Yes, this is an (intentional) limitation of the Framework and one of the reasons I developed the `reformat` program for 2022 testbeawm. I have plans to either improve the handling of this in Framework or include an updated `reformat` in ldmx-sw. https://github.com/LDMX-Software/ldmx-sw/issues/1906

TrigScint/include/TrigScint/Event/TrigScintQIEDigis.h
bryngemark: done

Packing/util/rogue-extract.py
bryngemark: Moved, and added a comment to the top of the script + the description.

TrigScint/util/runReformatting_ESA2025sliceTest.sh
bryngemark: ooooh hadn't thought of using the path of the script itself, thanks!

TrigScint/include/TrigScint/Event/ZCCMOutput.h
bryngemark: oh nice catch, i had forgotten to implement this! i think originally i wanted to keep a boolean to flag misalignment or errors, but my holidays are fast approaching so i opted for just storing this the same way i do CIDs. improvements are welcome though once you have started using the code!

TrigScint/exampleConfigs/runRawUnpackerZCCM.py
bryngemark: yes this is a little annoying. but until we have a different reformatter, to process the entire file, one can get the number of events by just dividing the file size by the number of bytes per event. or spit out the final event count in the rogue parsing step and keep track of that somewhere. (in the cern testbeam run script i looked that up from a log of the first preprocessing (fiber data reordering) step.)

TrigScint/exampleConfigs/runZCCMDecode.py
bryngemark: Would you be ok with me making a separate issue for this? I think it's a neater way to do it, and tried it, but there are several places I'd need to change to this different style of doing it and I'd really like to have this code handed over before I am out for a few weeks... in like 30 minutes 😄

tomeichlersmith: Fine by me

EBerzin (approved):

tomeichlersmith (approved):

PR #1909, opened 2026-01-05
add script for multi-fire runs in parallel

### LINKED ISSUE #1863: Parallelization efforts for ldmx-sw

When submitting jobs at many of our site, we are constrained by the number of jobs. However these machines usually have 4-8 cores, so 100 jobs could theoretically run 400-800 threads, which would significantly increase our production time.

I was originally imagining multi-threading as a solution, but see alternative below:

At the sw meeting yesterday @tomeichlersmith suggested just to have a wrapper script that would run 4-8 instances of `fire`. That seems like a good alternative solution that certainly doesn't require as much dev time

**Additional context**

LDCS not working we need to be smart on how to produce samples manually

tomeichlersmith:
I am adding a shell script that runs multiple copies of `fire` with different arguments to the same configuration script. This is only intended for batch processing and should **not** be used in development of ldmx-sw.

This resolves #1863 

## Check List
- [x] I ran my developments and the following shows that they are successful. (below)
- [x] shellcheck
- [x] need to wait for updated dev image release that includes GNU parallel https://github.com/LDMX-Software/dev-build-context/releases/tag/v5.2.0
- [x] add documentation to doc site https://github.com/LDMX-Software/ldmx-software.github.io/issues/59
    - link to GNU parallel manual
    - point out that output is not "live" (job output only printed to terminal at end of job, saving to file slows everyone down since all jobs then are trying to "live" write their output to disk)
    - mention `hadd` for combining results
    - warnings about how there are no checks on data repetition (repeat run numbers or repeat input files)

## Testing
Using a simple inclusive simulation that takes the number of events and run number from the command line.
```python
p = ldmxcfg.Process('test')
mySim.setDetector('ldmx-det-v14-8gev')
mySim.generators.append( gen.single_8gev_e_upstream_tagger() )
p.sequence = [ mySim ]
import sys
import LDMX.Ecal.EcalGeometry
import LDMX.Hcal.HcalGeometry
p.run = int(sys.argv[2])
p.outputFiles = [f'events-run-{p.run}.root']
```

Generating 10 events with a single process.
```
tom@ty-lee:~/code/ldmx/1863-fire-parallel/test$ time just fire config.py 10 1                                                                                                                                                                 
# omit messages from inside process

real    0m11.901s                                          
user    0m0.150s                                           
sys     0m0.094s  
```

Generating 10 events with two processes that have different run numbers.
```
tom@ty-lee:~/code/ldmx/1863-fire-parallel/test$ time just fire-parallel config.py 5 {} ::: 2 3                         

real    0m8.528s
user    0m0.155s
sys     0m0.126s
```

which is about what I expected since the simulation-setup time dominates at this low number of events.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ EventDisplay/include/EventDisplay/EventDisplayLinkDef.h
 .denv
 .apptainer
 .wget-hsts
+.parallel
--- a/Framework/CMakeLists.txt
+++ b/Framework/CMakeLists.txt
@@ -56,6 +56,11 @@ set_target_properties(
 add_executable(fire ${PROJECT_SOURCE_DIR}/app/fire.cxx)
 target_link_libraries(fire PRIVATE Framework::Framework)
 install(TARGETS fire DESTINATION ${CMAKE_INSTALL_PREFIX}/bin)
+install(FILES app/fire-parallel
+  TYPE BIN
+  PERMISSIONS OWNER_READ OWNER_WRITE OWNER_EXECUTE
+              GROUP_READ GROUP_EXECUTE
+              WORLD_READ WORLD_EXECUTE)
 
 # Setup the test
 setup_test(dependencies Framework::Framework)
--- a/Framework/app/fire-parallel
+++ b/Framework/app/fire-parallel
@@ -16,11 +16,11 @@ usage() {
     --help|-h : print this help and exit
     --dry-run : print the jobs that would be run instead of running them
     --log-dir : specify where the logs for the jobs and parallel should be stored
-                if the special string '-', then the logs are not stored and instead
-                printed to the terminal. This '-' prevents '--follow' from being used.
-    --follow  : put the parallel run into the background and watch the parallel
-                job log while waiting for the jobs to complete
-                Cannot be used if '-' is used for '--log-dir'.
+                defaults to printing the output of the jobs to the terminal AFTER
+                the job fully completes (i.e. the parallel default).
+                If this is provided, we instead write the output of each job to
+                its own file and follow parallel's joblog file reporting when jobs
+                complete.
     --        : signal that the rest of the arguments should be passed to GNU parallel
 
   ARGUMENTS
@@ -48,8 +48,7 @@ HELP
 }
 
 dry_run=false
-log_dir="logs"
-follow=false
+log_dir=""
 config=""
 template_args=""
 while [ $# -gt 0 ]; do
@@ -60,9 +59,6 @@ while [ $# -gt 0 ]; do
       ;;
     --dry-run)
       dry_run=true
-      ;;
-    --follow)
-      follow=true
       ;;
     --log-dir)
       shift
@@ -110,11 +106,6 @@ if [ -z "${template_args}" ]; then
   template_args="{}"
 fi
 
-if ${follow} && [ "${log_dir}" = "-" ]; then
-  echo "error: cannot '--follow' when redirecting the logs to the terminal ('--log-dir -')"
-  exit 2
-fi
-
 if [ -z "${config}" ] || [ ! -f "${config}" ]; then
   echo "error: the config script argument is required"
   exit 3
@@ -122,7 +113,7 @@ fi
 
 cmd="parallel"
 if ${dry_run}; then
-  if [ "${log_dir}" != "-" ]; then
+  if [ -n "${log_dir}" ]; then
     if [ -d "${log_dir}" ]; then
       echo "warning: actual run would fail here because the output log directory already exists"
     fi
@@ -130,7 +121,7 @@ if ${dry_run}; then
   fi
   cmd="${cmd} --dry-run"
 else
-  if [ "${log_dir}" != "-" ]; then
+  if [ -n "${log_dir}" ]; then
     if [ -d "${log_dir}" ]; then
       echo "error: output log directory '${log_dir}' already exists"
       exit 4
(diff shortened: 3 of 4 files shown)


tvami:
Could you please also add a `just` command?

tvami:
I ran the ecal PN, each config with 2000 events
```
time just fire-parallel config.py  {} ::: 1 2 3 4 5
```
and I get
```
real	44m6.808s
user	217m44.044s
sys	0m31.145s
```

The 45min real time is 2x faster than what we have in the CI-based logs.
The user/real does match the 5x, altho I'm a little skeptical that SDF is another factor of 2.5 slower than the avg github action machine... I guess I can run a single core job overnight to have a real apple to apple comparison.

One more interesting thing: increasing the logging level does NOT significally change the time parallel takes. I was not expecting that. Here is with log level at 10.
```
real	47m24.971s
user	218m56.450s
sys	0m44.404s
```

The only printout like this is
```
21:33:11    GSF            WARNING   Performed steps with errors: MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1
21:33:11    GSF_PROP       ERROR     Step failed with MultiStepperError:4: Stepping error occurred in all components
21:38:41    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
21:39:55    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
21:41:27    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
21:45:04    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
21:46:04    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
21:46:14    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
21:51:43    GSF            WARNING   Performed steps with errors: MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1
21:51:43    GSF_PROP       ERROR     Step failed with MultiStepperError:4: Stepping error occurred in all components
21:51:56    GSF            WARNING   Performed steps with errors: MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1
21:51:56    GSF_PROP       ERROR     Step failed with MultiStepperError:4: Stepping error occurred in all components
21:54:27    GSF            WARNING   Performed steps with errors: MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1
21:54:27    GSF_PROP       ERROR     Step failed with MultiStepperError:4: Stepping error occurred in all components
21:56:05    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
21:59:24    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
21:59:59    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:01:57    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:04:25    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:06:30    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:07:28    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:07:44    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:10:52    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:14:58    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:17:31    PROP           ERROR     Step failed with MagneticFieldError:1: Interpolation out of bounds was requested
22:18:42    GSF            WARNING   Performed steps with errors: MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1 | MagneticFieldError:1
22:18:42    GSF_PROP       ERROR     Step failed with MultiStepperError:4: Stepping error occurred in all components
```
And I was expecting this to be much faster wrt log level at 1.

So that makes me think the `cout`-s we have are really bad for the parallel efforts

tomeichlersmith:
GNU Parallel does not do anything super smart with the logs - it has its own cache area (usually in `~/.parallel`) where it writes job logs while the jobs are running and then `cat`s those logs to the screen when the job is finished (if we don't capture the logs ourselves with `--log-dir`). This could be the source of the slow down if the different jobs are waiting for access to the disk to write their logs. **However**, even _if_ we didn't have any logs, we'd still have multiple jobs trying to write to different files on the same disk.

We could investigate if its the logs specifically by editing `fire-parallel` to redirect all logging messages from the jobs into `/dev/null` and do another time comparison. I suspect it to be slightly better because the data files seem to be smarter about caching/flushing which can help avoid waiting for each other on the same disk, but the amount of improvement may not be large. If the improvement is large, then we should document that `fire-parallel` should _only_ be used in batch situations where the user can have their `config.py` disable most/all logging.

For me, I think this solution with its ~2x speed increase is good enough. We could invest a lot more time into truly parallelizing the whole processor chain[^1] which is certainly an interesting and worthwhile project, but would also come with its own technical issues.

[^1]: I would still advocate for not parallelizing _within_ a specific event and instead just run events on different cores. This would simplify the handling of memory since each event would only be processed on one core and then the result would be copied back to the "writer" when the event is done.

tvami:
>  to redirect all logging messages from the jobs into

```
mkdir: cannot create directory '/dev/null': File exists
```

I'm thinking of adding another flag, `--no-log` to deal with this

tomeichlersmith:
I would not want `--no-log` to be an option unless there is a significant and reproducible performance improvement. I think the logs are more important than a < 5% speed up (2min / 45min) if that is the case.

But you are correct, the script itself will need development in order to check if throwing away the logs has a drastic impact.

tvami:
This is for 2000 events, single threaded:
```
time just fire config.py 11
real    52m59.610s
user    51m55.601s
sys     0m22.149s
```

Running 1250 events in 8 cores:
```
time just fire-parallel config.py  {} ::: 12 13 14 15 16 17 18 19
real    32m36.482s
user    243m3.433s
sys     1m0.726s
```

Now testing `--no-log`

```
time just fire-parallel --no-log config.py {} ::: 12 13 14 15 16 17 18 19
real    35m31.850s                         
user    266m2.279s
sys     0m59.150s
```

I think we dont need to add the  `--no-log` in the PR after all

tvami (approved):

tvami:
@tomeichlersmith I think we should merge this, the documentation goes to a different repo anyway, right?

PR #1911, opened 2026-01-13
Remove G4 warnings about no production cuts

### LINKED ISSUE #1891: G4 warning for not having specific production cuts in the G4regions

When I print out the G4 warnings I get this
```
[ GEANT4 ] 0  warn: Warning : Region <CalorimeterRegion> does not have specific production cuts,
[ GEANT4 ] 0  warn: Warning : Region <MagnetRegion> does not have specific production cuts,
[ GEANT4 ] 0  warn: Warning : Region <tagger> does not have specific production cuts,
[ GEANT4 ] 0  warn: Warning : Region <trig_scint> does not have specific production cuts,
[ GEANT4 ] 0  warn: Warning : Region <target> does not have specific production cuts,
```

I think in default this still applies the default production cuts. So I doubt this is a big issue, but we could / should double check.

Given the above, I dont think anything bad happens to have this in

cjbarton151:

Resolving #1891 with a quick fix.
Geant4 is complaining because we're creating G4Regions (to use for other purposes) but not setting explicit production cuts.
The default production cut for most physics lists is ~1 mm, plus or minus a few tenths of a mm, so I just set it to 1 mm.

## Check List

Before:
image

After:
image

### DIFF
--- a/SimCore/src/SimCore/Geo/AuxInfoReader.cxx
+++ b/SimCore/src/SimCore/Geo/AuxInfoReader.cxx
@@ -14,6 +14,7 @@
 #include "G4GDMLEvaluator.hh"
 #include "G4LogicalVolumeStore.hh"
 #include "G4ProductionCuts.hh"
+#include "G4ProductionCutsTable.hh"
 #include "G4Region.hh"
 #include "G4RegionStore.hh"
 #include "G4SDManager.hh"
@@ -246,10 +247,7 @@ void AuxInfoReader::createRegion(const G4String& name,
   auto region = new G4Region(name);
   region->SetUserInformation(region_info);
   // To get rid of those pesky G4 warnings
-  auto defaultcuts = new G4ProductionCuts;
-  defaultcuts->SetProductionCut(
-      1. * mm);  // Omitting index applies it to all particle types
-  region->SetProductionCuts(defaultcuts);
+  region->SetProductionCuts(G4ProductionCutsTable::GetProductionCutsTable()->GetDefaultProductionCuts());
 }
 // NOLINTEND
 
--- a/SimCore/src/SimCore/Geo/AuxInfoReader.cxx
+++ b/SimCore/src/SimCore/Geo/AuxInfoReader.cxx
@@ -13,9 +13,9 @@
 #include "G4FieldManager.hh"
 #include "G4GDMLEvaluator.hh"
 #include "G4LogicalVolumeStore.hh"
+#include "G4ProductionCuts.hh"
 #include "G4Region.hh"
 #include "G4RegionStore.hh"
-#include "G4ProductionCuts.hh"
 #include "G4SDManager.hh"
 #include "G4SystemOfUnits.hh"
 #include "G4UniformMagField.hh"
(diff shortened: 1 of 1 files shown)


SimCore/src/SimCore/Geo/AuxInfoReader.cxx
```diff
@@ -244,6 +245,11 @@ void AuxInfoReader::createRegion(const G4String& name,
   // NOLINTBEGIN
   auto region = new G4Region(name);
   region->SetUserInformation(region_info);
+  // To get rid of those pesky G4 warnings
+  auto defaultcuts = new G4ProductionCuts;
+  defaultcuts->SetProductionCut(
+      1. * mm);  // Omitting index applies it to all particle types
```
tomeichlersmith: _really_ picky but since its only five lines i have to find something

```suggestion
  // Omitting index applies it to all particle types
  defaultcuts->SetProductionCut(1. * mm);
```

tomeichlersmith (commented):
The only technical aspect I'm concerned about is if the `1 mm` production cut actually affects the physics at all. ~Is this a sensible default? Where does this number come from?~ Edit: After one more sip of coffee, I notice you comment that Geant4's own defaults are ~1mm for most physics lists. My below comment about "citing" that still stands so that future devs know where this magic number comes from.

If you have a link to public docs or presentation that we could put in this PR comments (or even as a comment in the code), that would be best.

SimCore/src/SimCore/Geo/AuxInfoReader.cxx
tomeichlersmith: (I'm moving the in-line comment above the code since the `clang-format` does a wrapping that's ugly in my opinion.)

tvami (approved):
Looks good, I agree with the comment from Tom to link where the 1 mm cut comes from

tvami:
I cut a new gold, testing with this will actually help us check any effect on the physics

cjbarton151:
In the v10.3 guide for application developers there are two possibly conflicting statements:
In section 2.4.2 [How to Specify Particles: Range Cuts](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/BackupVersions/V10.3/html/ch02s04.html#sect.HowToSpecParti.RangeCuts):
> The default implemetation of SetCuts() method provides a defaultCutValue member as the unique range cut-off value for all particle types. The defaultCutValue is set to 1.0 mm by default.

In section 5.2.1 [Physics processes - Production Cuts](https://geant4-userdoc.web.cern.ch/UsersGuides/ForApplicationDeveloper/BackupVersions/V10.3/html/ch05s02.html#sect.PhysProc.EleMag)
> Remember that production cuts for secondaries can be specified as range cuts, which are converted at initialisation time into energy thresholds for secondary gamma, electron, positron and proton production. The cut for proton is applied by elastic scattering processes to aal recoil ions.
A range cut value is set by default to 0.7 mm in Geant4 reference physics lists.

I interpret this to mean that for built-in physics lists, the default value is 0.7 mm. However, when constructing a new user-defined physics list, the default is set to 1.0 mm. Not sure why it's like this, and I could be misinterpreting. Anyways, to remove ambiguity I did what I should have done to begin with, and just let G4 reference the current default cut value from the physics list. This shouldn't introduce any additional overhead, because I just inserted the method that G4 was using every time it sent the warning message anyways.

cjbarton151:
Not sure if this has been already discussed in the past, but it could be interesting to play around with the cut values in the ECal to see how this affects the clustering and energy reconstruction.

tvami:
> , but it could be interesting to play around with the cut values in the ECal to see how this affects the clustering and energy reconstruction.

I agree this could be interesting, please make an issue about this!

tvami:
Logs are looking good! The diffs are there since this warning msg is now removed

tomeichlersmith (approved):

PR #1915, opened 2026-01-18
Move back to realistic reco in the CIs, add some misc HCAL improvements

### LINKED ISSUE #1344: Spurious PE entries in Hcal hits at around 817 PE
As noted by @jpasc27 and @mrsolt, we seem to have some odd entries in the HCal rechits with occasional photoelectrons > what you get at max ADC (see [Jessica's talk at the 2023 spring collaboration meeting ](https://indico.fnal.gov/event/58323/contributions/259798/subcontributions/8885/attachments/166149/221105/LDMX%20CM%20Hcal%20Studies%20Update.pdf) slide 5). 

As far as I understood it, Jessica does not take the `maxPE` hits from the veto processor so this should be unrelated to LDMX-Software/Hcal#58. It also looks like the hits always have the same value, which wouldn't make sense if it was LDMX-Software/Hcal#58. Unclear if it is related to LDMX-Software/ldmx-sw#1348 

I tested this quickly with a neutron gun while working on the HCal DQM today and can confirm that I'm getting the same thing.

Screenshot from 2023-04-24 18-04-38

All of the RecHits included in the figure above are from the Back HCal.

### LINKED ISSUE #1348: Large positions and energy in Hcal reconstruction
# Strange Hcal Reconstruction output

## Describe the bug

There seems to be two strange parts of the Hcal reconstruction output (at least with the prototype geometry).

1.  Some hits have positions that are wildly outside of the geometry
    
    This seems to happen when we have one of the TOA values (`TOA_posend` or `TOA_negend`) be negative and the other positive. I suspect that negative TOA values aren&rsquo;t good in the first place, but I&rsquo;m not sure.

2.  The distribution of the reconstructed energy is significantly different from
    the distribution of energy per bar from simulated hits. Specifically, the total reconstructed energy is significantly larger than the total simulated energy
    
    I haven&rsquo;t looked into this much but wanted to get the bug report up so I&rsquo;ll
    update the issue here. I&rsquo;m keeping them together because I suspect that they
    might be related. If they turn out not to be, ill make a separate issue for
    that.

## To reproduce

I don&rsquo;t think that this should be a problem that is specific to the prototype geometry, unless we have some part of the digitization code that relies on particularities of the regular detector. I haven&rsquo;t tested with the full detector yet but in principle, you&rsquo;d just have to change the corresponding line in the configuration file below.

Fastest way to see it is to just run a simulation with the configuration file below, and then just inspecting it either manually or with a script like below.

    ldmx fire iss21.py
    ldmx python3 checkIss21.py

    import ROOT
    from ROOT import gSystem
    from ROOT import TFile
    gSystem.Load('libFramework.so')
    file = TFile('iss21.root')
    tree = file.Get('LDMX_Events')
    # Position tests
    for i,event in enumerate(tree):
        recHits = event.HcalRecHits_iss21
        simHits = event.HcalSimHits_iss21
        for j,hit in enumerate(recHits):
            x = hit.getXPos()
            y = hit.getYPos()
            t = hit.getTime()
            if (x > 1000):
                print("X position {:.4f} out of bounds, event {}, hit {}".format(x,i+1,j))
            if (y > 1000):
                print("Y position {:.4f} out of bounds, event {}, hit {}".format(y,i+1,j))
    # Energy tests
    for i,event in enumerate(tree):
        recHits = event.HcalRecHits_iss21
        simHits = event.HcalSimHits_iss21
        totalSimEnergy = 0.
        totalRecEnergy = 0.
        for hit in simHits:
            totalSimEnergy += hit.getEdep()
        for hit in recHits:
            totalRecEnergy += hit.getEnergy()
        deltaEnergy = totalRecEnergy - totalSimEnergy
        print("Event: {}, E_{{sim}}: {:.2f}, E_{{rec}}: {:.2f}".format(i+1, totalSimEnergy, totalRecEnergy))
        print("E_{{rec}} - E_{{sim}}: {:.2f}".format(deltaEnergy))
        # Don't need all the output
        if (i+1 >= 5):
            break

this should output

    X position 8037.7466 out of bounds, event 3, hit 12
    Y position 8069.7676 out of bounds, event 53, hit 31
    X position 7619.1831 out of bounds, event 106, hit 49
    Y position 7347.0024 out of bounds, event 136, hit 5
    X position 8186.4165 out of bounds, event 169, hit 38
    Y position 7930.2466 out of bounds, event 193, hit 0
    Event: 1, E_{sim}: 113.80, E_{rec}: 187.65
    E_{rec} - E_{sim}: 73.85
    Event: 2, E_{sim}: 540.15, E_{rec}: 620.73
    E_{rec} - E_{sim}: 80.58
    Event: 3, E_{sim}: 454.35, E_{rec}: 579.76
    E_{rec} - E_{sim}: 125.41
    Event: 4, E_{sim}: 50.94, E_{rec}: 95.35
    E_{rec} - E_{sim}: 44.42
    Event: 5, E_{sim}: 95.47, E_{rec}: 176.14
    E_{rec} - E_{sim}: 80.67

With GDB inside of the container you can explore it as follows, which lets us see that the large positions comes from the TOA values. I&rsquo;m setting three breakpoints, you only need the first one. The second and third will break whenever one (but not two) of the TOA values is negative.

    ldmx use dev v3.2
    ldmx gdb fire
    # Note if it is a fresh session, you'll have to respond "yes" to creating a breakpoint despite the shared library not being loaded yet
    break HcalRecProducer.cxx:238 if position_bar > 1000
    
    break HcalRecProducer.cxx:217 if TOA_posend < 0 && TOA_negend > 0
    
    break HcalRecProducer.cxx:221 if TOA_negend < 0 && TOA_posend > 0
    
    
    run iss21.py
    
    # When one of the breakpoints triggers, the following commands can be used to check out the variables
    
    print TOA_posend
    print TOA_negend
    # You might have to press next a couple of times to get to where this is calculated
    print position_bar
    
    # First break has negative TOA_negend, but only ~-9 so position is within bar (258)
    continue
    # Second break is similar
    continue
    # Third triggers the position > 1000 trigger
    # Should output 8037.7464360643971
    # Should output -53.27899194934173
    # Should output 32.516454878479138

## Configuration file

    process = ldmxcfg.Process("iss21")
    process.logFrequency = 1
    process.termLogLevel = 1
    process.maxEvents = 200
    process.run = 0
    simulation = simulator.simulator("iss21")
    # simulation.setDetector('ldmx-det-v13')
    simulation.setDetector('ldmx-hcal-prototype-v1.0')
    gun = generators.gun('particle_gun')
    
    gun.particle  = 'proton'
    gun.energy    = 8.
    gun.position  = [0., 0., 0.]
    gun.direction = [0., 0., 1.]
    
    simulation.generators=[gun]
    
    import LDMX.Hcal.HcalGeometry
    import LDMX.Ecal.EcalGeometry
    
    from LDMX.Hcal import digi
    from LDMX.Hcal import hcal
    
    digis = digi.HcalDigiProducer()
    reco = digi.HcalRecProducer()
    
    from LDMX.Hcal import hcal_hardcoded_conditions
    process.sequence=[simulation,
                      digis,
                      reco,
                      ]
    process.outputFiles = ['iss21.root']

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1344

## Check List

Validation plots shown in https://github.com/LDMX-Software/ldmx-sw/issues/1344#issuecomment-3765600605

### DIFF
--- a/Hcal/src/Hcal/HcalDoubleEndRecProducer.cxx
+++ b/Hcal/src/Hcal/HcalDoubleEndRecProducer.cxx
@@ -76,9 +76,13 @@ void HcalDoubleEndRecProducer::produce(framework::Event& event) {
     // skip non-double-ended layers
     if (id.section() != ldmx::HcalID::HcalSection::BACK) continue;
 
+    // skip if we couldn't find both ends of the bar
+    auto indices = indices_by_id[id];
+    if (indices.first == -1 || indices.second == -1) continue;
+
     // get two hits_ to reconstruct
-    auto hit_pos_end = hcal_bar.second.at(indices_by_id[id].first);
-    auto hit_neg_end = hcal_bar.second.at(indices_by_id[id].second);
+    auto hit_pos_end = hcal_bar.second.at(indices.first);
+    auto hit_neg_end = hcal_bar.second.at(indices.second);
 
     // update TOA hit with negative end with mean shift
     ldmx::HcalDigiID digi_id_pos(hit_pos_end.getSection(),
@@ -91,7 +95,7 @@ void HcalDoubleEndRecProducer::produce(framework::Event& event) {
 
     double pos_time = hit_pos_end.getTime();
     double neg_time = hit_neg_end.getTime();
-    if (pos_time != 0 || neg_time != 0) {
+    if (pos_time != 0 && neg_time != 0) {
       neg_time = neg_time - mean_shift;
     }
 
--- a/Hcal/src/Hcal/HcalDoubleEndRecProducer.cxx
+++ b/Hcal/src/Hcal/HcalDoubleEndRecProducer.cxx
@@ -78,7 +78,12 @@ void HcalDoubleEndRecProducer::produce(framework::Event& event) {
 
     // skip if we couldn't find both ends of the bar
     auto indices = indices_by_id[id];
-    if (indices.first == -1 || indices.second == -1) continue;
+    if (indices.first == -1 || indices.second == -1) {
+      ldmx_log(warn) << "Couldn't find both ends of the bar "
+                     << " for HcalID section " << static_cast<int>(id.section())
+                     << " layer " << id.layer() << " strip " << id.strip();
+      continue;
+    }
 
     // get two hits_ to reconstruct
     auto hit_pos_end = hcal_bar.second.at(indices.first);
--- a/.github/validation_samples/deep_ecal_gun/config.py
+++ b/.github/validation_samples/deep_ecal_gun/config.py
@@ -75,8 +75,9 @@ import LDMX.Ecal.ecalClusters as ecal_cluster
 # Load the HCAL modules
 import LDMX.Hcal.HcalGeometry
 import LDMX.Hcal.hcal_hardcoded_conditions
-import LDMX.Hcal.digi as hcal_digi
-hcal_digi_reco = hcal_digi.HcalSimpleDigiAndRecProducer()
+import LDMX.Hcal.digi as hcal_digi_and_reco
+hcal_digi = hcal_digi_and_reco.HcalDigiProducer()
+hcal_reco = hcal_digi_and_reco.HcalRecProducer()
 
 
 # Load electron counting and trigger
@@ -115,7 +116,8 @@ p.sequence.extend([
         ecal_veto,
         ecal_mip,
         ecal_veto_pnet,
-        hcal_digi_reco,
+        hcal_digi,
+        hcal_reco,
         hcal_veto,
         trigger,
         dqm.EcalClusterAnalyzer()
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -37,8 +37,9 @@ import LDMX.Ecal.ecalClusters as ecal_cluster
 # Load the HCAL modules
 import LDMX.Hcal.HcalGeometry
 import LDMX.Hcal.hcal_hardcoded_conditions
-import LDMX.Hcal.digi as hcal_digi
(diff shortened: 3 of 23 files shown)


tvami:
2e CI fails bc of the naming convention I'll fix that next

Hcal/src/Hcal/HcalDoubleEndRecProducer.cxx
```diff
@@ -76,9 +76,13 @@ void HcalDoubleEndRecProducer::produce(framework::Event& event) {
     // skip non-double-ended layers
     if (id.section() != ldmx::HcalID::HcalSection::BACK) continue;
 
+    // skip if we couldn't find both ends of the bar
+    auto indices = indices_by_id[id];
+    if (indices.first == -1 || indices.second == -1) continue;
+
```
EinarElen, line 87: Should this be a warning or something logged?

EinarElen (approved):
Looks good to me in general, only thing i can think of is that it might be valuable to keep the simplified reco in CI runs so one can compare the simplified distributions if and when something looks strange in the realistic reco?

Hcal/src/Hcal/HcalDoubleEndRecProducer.cxx
tvami: Done in https://github.com/LDMX-Software/ldmx-sw/pull/1915/changes/37b723b278cffd803547c0dcb90b4dc12e2a6504

tvami:
> only thing i can think of is that it might be valuable to keep the simplified reco in CI 

Done in 
https://github.com/LDMX-Software/ldmx-sw/pull/1915/changes/37b723b278cffd803547c0dcb90b4dc12e2a6504

tomeichlersmith (approved):

PR #1917, opened 2026-01-19
Update rogue-extract.py

### LINKED ISSUE #1916: rogue parsing script doesn't work fully for TS
**Describe the bug**
Erh I seem to have pushed, in place of the latest and greatest, something old and not that useful for parsing TS data to reformat.

**To Reproduce**
Go to the reformatting PR of #1883, follow the instructions, notice that stuff isn't working as advertized (thanks @EBerzin for doing exactly that).

**Desired behavior**
That it works -- I will share the working version of the script asap

bryngemark:

This resolves #1916 by really sharing the working rogue-extract script that actually writes TS output to a file. (Turns out I thought I had downloaded what I had been running on s3df, but I had missed the final `.` at the end of my `rsync`, so, I shared the (local) version of the script I had inherited from @tomeichlersmith! The stuff you keep relearning...)

### DIFF
--- a/Packing/util/rogue-extract.py
+++ b/Packing/util/rogue-extract.py
@@ -4,18 +4,7 @@ import sys
 import numpy as np
 import argparse
 from pathlib import Path
-
-
-# helper class to swap endianness of TS data
 import struct
-
-def swap_endianness(num):
-    # Pack the number as a 32-bit integer in network byte order
-    packed_num = struct.pack('>I', num)
-    # Unpack the number as a 32-bit integer in little-endian byte order
-    unpacked_num = struct.unpack('<I', packed_num)[0]
-    return unpacked_num
-
 
 def main():
     parser = argparse.ArgumentParser(description="just hexdump the Hcal/Ecal data without attempting to decode")
--- a/Packing/util/rogue-extract.py
+++ b/Packing/util/rogue-extract.py
@@ -1,20 +1,30 @@
 #!/usr/bin/env python3
-
-# note that this script requires a setup of Rogue, either an actual install, or using conda or denv.
-# if using denv, take care not to confuse this with the ldmx-sw denv needed for further processing. This script does not run with an ldmx-sw image.
-
 import pyrogue.utilities.fileio
 import sys
 import numpy as np
 import argparse
 from pathlib import Path
 
+
+# helper class to swap endianness of TS data
+import struct
+
+def swap_endianness(num):
+    # Pack the number as a 32-bit integer in network byte order
+    packed_num = struct.pack('>I', num)
+    # Unpack the number as a 32-bit integer in little-endian byte order
+    unpacked_num = struct.unpack('<I', packed_num)[0]
+    return unpacked_num
+
+
 def main():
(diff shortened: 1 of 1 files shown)


Packing/util/rogue-extract.py
```diff
@@ -1,20 +1,30 @@
...
+    parser.add_argument('input', help='input file to extract raw data from')
+    parser.add_argument('--output','-o',  help='output file to write extracted data to')
+    parser.add_argument('--debug', '-d', action='store_true', help='enable debug mode i.e. print some stuff to screen')
     parser.add_argument('--nevent', '-n', help='maximum number of events to decode', type=int)
-    parser.add_argument('--contrib', '-c', choices=['ecal','hcal'], help='contributor to focus on, without this dump all data from subsystem 5 which could be both Ecal and Hcal')
-    parser.add_argument('--subsystem', '-s', choices=[5,2], help='contributor to focus on, without this dump all data from subsystem 5 which could be both Ecal and Hcal')
+    parser.add_argument('--contrib', '-c', choices=['ecal','hcal', 'ts'], help='contributor to focus on, without this dump all data from subsystem 5 which could be both Ecal and Hcal')
+    parser.add_argument('--subsystem', '-s', choices=['5','2'], help='contributor to focus on, without this dump all data from subsystem 5 which could be both Ecal and Hcal')
```
tomeichlersmith, line 16: So... not sure if this needs to actually change the code, I just want to clear something up.

The "subsystem ID" is the one that is in the second byte from the Rogue stream (`data[1]`) and it is _supposed_ to separate the different subsystems. Since Ecal/Hcal used the same firmware block, they ended up sharing the same "subsystem ID" so we had to fallback to the "contributor ID" to separate Ecal/Hcal.

All this to say, for TS, you can (and should) ignore the "contributor ID" since you guys don't have this confusion.

In order to reflect this in the code, I would have the `choices=['ecal','hcal','ts']` be for `--subsystem` and `--contrib` would remain just `['ecal','hcal']` _or_ we could drop the `--contrib` command line argument and just reflect this confusion in the code.

Or or... just leave it as is since it works and Tom keeps promising that he'll write a C++ unpacker for Rogue although I cannot promise a timeline on that due to ICE's occupation of the TC.

tomeichlersmith (commented):
Mainly commenting about a vocabulary confusion.

I don't super like the endianness being swapped in this script, but I am not using or developing the script so I don't think I should impose any rules on it.

Packing/util/rogue-extract.py
bryngemark: Thanks @tomeichlersmith for pointing out the nomenclature confusion. Confirming first with @EBerzin that we don't need a specific contributor ID for TS? Also Elizabeth since you're anyway testing this right now, if you find a more elegant solution you're super welcome to push an updated script as I'm not developing at the moment.

bryngemark:
> I don't super like the endianness being swapped in this script, but I am not using or developing the script so I don't think I should impose any rules on it.

This just reflects that I had been editing starting from the raw rogue output file, manually removing bytes and whatnot, and once it came to using a parsed solution I found that the endianness was swapped in the new output file. I'm not entirely sure what happened there, all I can say is that this works, and I'd defer any changes to when we have a proper unpacker.

EBerzin (approved):

tomeichlersmith (approved):

PR #1927, opened 2026-01-29
Add dEdx as a track property

### LINKED ISSUE #1477: Add local angle / pathlenght information to `Measurement` for tracking

This came up at today's tracking meeting, for dEdx in the tracker we'd need to know the local angles, but that's currently not available at each layer.

Add new member variables that include the local angle information.

Having dx = sensor thickness, but clearly its more correct do calculate it based on the angle. I should also say that we attempted to use the global angles with @danyi211, that gives unphysical values.

**Additional context**

The dEdx dev issue is https://github.com/LDMX-Software/ldmx-sw/issues/1452

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1477

## Check List

On the inclusive sample: makes sense these are mostly straight tracks so the path length is the sensor thickness
Screenshot 2026-01-29 at 15 21 35

So I also ran on the WAB CI hoping for bigger angles thus bigger path length, and indeed

Screenshot 2026-01-29 at 15 22 56

### DIFF
--- a/Tracking/include/Tracking/Event/Measurement.h
+++ b/Tracking/include/Tracking/Event/Measurement.h
@@ -67,6 +67,16 @@ class Measurement {
   };
 
   /**
+   * Set the path length along the track at this measurement.
+   *
+   * @param path_length Path length in mm.
+   */
+  void setPathLength(const float& path_length) { path_length_ = path_length; };
+
+  /// @return The path length along the track at this measurement in mm.
+  [[nodiscard]] float getPathLength() const { return path_length_; };
+
+  /**
    * Set cov(U,U) and cov(V, V).
    *
    * @param cov_uu cov(U, U).
@@ -147,6 +157,8 @@ class Measurement {
   float meas_u_{0.};
   /// Local position in v (mm).
   float meas_v_{0.};
+  /// Path length along the track at this measurement (mm).
+  float path_length_{0.};
   /// The ID of the sensor where the measurement took place.
   int layer_id_{0};
   /// The layer number internal to the tracker.
--- a/Tracking/include/Tracking/Event/Measurement.h
+++ b/Tracking/include/Tracking/Event/Measurement.h
@@ -67,16 +67,6 @@ class Measurement {
   };
 
   /**
-   * Set the path length along the track at this measurement.
-   *
-   * @param path_length Path length in mm.
-   */
-  void setPathLength(const float& path_length) { path_length_ = path_length; };
-
-  /// @return The path length along the track at this measurement in mm.
-  [[nodiscard]] float getPathLength() const { return path_length_; };
-
-  /**
    * Set cov(U,U) and cov(V, V).
    *
    * @param cov_uu cov(U, U).
@@ -157,8 +147,6 @@ class Measurement {
   float meas_u_{0.};
   /// Local position in v (mm).
   float meas_v_{0.};
-  /// Path length along the track at this measurement (mm).
-  float path_length_{0.};
   /// The ID of the sensor where the measurement took place.
   int layer_id_{0};
   /// The layer number internal to the tracker.
--- a/Tracking/src/Tracking/Reco/CKFProcessor.cxx
+++ b/Tracking/src/Tracking/Reco/CKFProcessor.cxx
@@ -387,6 +387,10 @@ void CKFProcessor::produce(framework::Event& event) {
   Acts::VectorMultiTrajectory mtj;
   Acts::TrackContainer tc{vtc, mtj};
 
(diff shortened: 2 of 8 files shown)


danyi211 (approved):

tvami:
OK tests pass now! I think we have a similar situation as in the other tracking PRs, the configs with external physics generators (signal and GENIE) change enough for it to lead to changes in the KS score. The distributions are anyway very compatible
https://github.com/LDMX-Software/ldmx-sw/actions/runs/21501369292/artifacts/5313404187

Tracking/src/Tracking/Reco/CKFProcessor.cxx
```diff
@@ -652,6 +714,11 @@ void CKFProcessor::produce(framework::Event& event) {
   // Add the tracks to the event
   event.add(out_trk_collection_, tracks);
 
+  // Add the updated measurements with path lengths back to the event
+  event.add(on_track_measurements_, updated_measurements);
+  // Drop the original measurements to avoid confusion
+  event.addDrop(measurement_collection_);
```
tomeichlersmith: This `addDrop` is concerning to me for two reasons.

1. The drop/keep rules are meant to be set at the beginning of the run and then stay the same throughout the run so that the branches that end up in the output file do not change as events are processed.
2. I don't like the idea of "replacing" an old branch with a new branch. I think instead of keeping the path-length inside of `Measurement`, they should go into a new object (or just a `std::vector<double>` that has the same indices as the `std::vector<Measurement>`). This appropriately reflects that the producer is adding new information.

Tracking/include/Tracking/Event/Measurement.h
```diff
@@ -147,6 +157,8 @@ class Measurement {
   float meas_u_{0.};
   /// Local position in v (mm).
   float meas_v_{0.};
+  /// Path length along the track at this measurement (mm).
+  float path_length_{0.};
```
tomeichlersmith: See previous comment about "updating" (replacing) objects with new objects in the event.

tomeichlersmith (commented):
The actual tracking stuff seems to make sense. My comment is about the technicality of drop/keep and why the current solution concerns me.

Tracking/src/Tracking/Reco/CKFProcessor.cxx
tvami:
```suggestion
```

Tracking/include/Tracking/Event/Measurement.h
tvami: OK I removed the dropping, I'll keep this one as the new object is of a type of a measurement. For the Measurement before updating this will be at zero, which is actually a good identifier that it's coming from the digitizer instead of the reco. Given it's zeros that will compress well so I'm not worried about size considerations

tvami: I should also say that pathlength is a property of the SimHits so it just brings the measurement to the same level

bloodyyugo:
I think it would be more logical (and cleaner) to make the pathlength (list) be a property of the Track object instead of the Measurement.

tvami:
Hmm, I'll use this for dEdx, so I thought it's more natural to have it in the Measurement (dE is certainly coming from the charge in the "Measurement" hits). While I could do this, I'm a little worried for index mismatch (somehow in your proposed list the nth hit becomes the (n-1)th Measurement).

bloodyyugo:
I just think saving pathlength (or dEdx directly) with the track is nicer since it avoids having to make this separate collection of measurement objects.  Also, there is the issue with multiple tracks using the same Measurement....I think what your code does now is put that Measurement into the hits-on-track collection twice, with a change to just the pathlength. 

Anyway, not a huge concern...just a suggestion.   

> (somehow in your proposed list the nth hit becomes the (n-1)th Measurement).
I'm not sure what mean by this?

tvami:
Yes ok I will do dedx as a list directly, I think that will resolve my worries

tvami:
OK so moved to track level, here is a dEdx distribution from the inclusive sample

Screenshot 2026-02-03 at 21 23 59


and slide from 1.5 years ago from @danyi211 regarding the Sim Level approach (note the units have a factor of 10 diff, and also there is no data cleaning on my plot
Screenshot 2026-02-03 at 21 25 48

I think this shows that the PR's content is physically validated

bloodyyugo (approved):
Looks good!  Any idea about this factor of 10 in the dedx compared to previously?

tvami:
> Any idea about this factor of 10 in the dedx compared to previously?

Yeah the unit was MeV/cm and my plots have MeV/mm

PR #1926, opened 2026-01-29
Add dEdx mass estimator in EN CI, and misc changes

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1925

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [x] I ran my developments and the following shows that they are successful.

Ran the CI and I see the output in the histograms:
Screenshot 2026-01-29 at 10 15 42

PR #1929, opened 2026-01-30
iss1928 GNU parallel options parsing in fire-parallel

### LINKED ISSUE #1928: GNU parallel options parsing in fire-parallel
In several cases it becomes useful to use config files which take multiple options or positional arguments when employing fire-parallel. In these cases, GNU `parallel` provides the option of multiple argument blocks (demarcated by `:::`) and performs a Cartesian product between the provided arguments (so `::: A B ::: 1 2` will be executed four times, A1, A2, B1, B2). Sometimes more precision is necessary, in which case the `--link` option may be used (which makes `--link ::: A B ::: 1 2` be interpreted as A1, B2). However, presently the `fire-parallel` wrapper – while permitting options such as `--link` to be passed in the command line – does not position these options where GNU `parallel` can parse them properly.

When passing arguments such as `--link` to the `fire-parallel` command, they should be positioned where GNU `parallel` can properly parse them. Ideally this should be immediately following the `parallel` command itself, where the `--dry-run` and `--joblog` options already go.

One other possibility is to deal with this in-script: passing only one option (such as an integer) as an argument to the config file, and interpreting this argument and defining variables accordingly all in the Python itself. This would require no changes to `fire-parallel`, only to the Python config files. Nevertheless, this only works with `--link`: other possible options, such as `--bar`, could not be corrected in this way.

**Additional context**
Right now this is how the `fire-parallel` script parses the `--link` command. Instead of producing five samples in parallel, twenty-five jobs are created.
Image

oscarplewis:

This resolves #1928. See there for more details on what exactly is corrected.

## Check List

The following screenshot shows that `--link` is being parsed correctly and the correct jobs are being created by `parallel`.
Screenshot 2026-01-30 at 11 24 07

### DIFF
--- a/Framework/app/fire-parallel
+++ b/Framework/app/fire-parallel
@@ -20,8 +20,11 @@ usage() {
                 the job fully completes (i.e. the parallel default).
                 If this is provided, we instead write the output of each job to
                 its own file and follow parallel's joblog file reporting when jobs
-                complete.
-    --        : signal that the rest of the arguments should be passed to GNU parallel
+                complete
+    --        : delimiter after which all arguments will be passed to GNU parallel
+                directly. Script options must be strictly between '--' and the first
+                instance of '::', with only template values permitted afterwards. 
+                Template values need not be preceded by '--'.
 
   ARGUMENTS
     config.py : configuration script to pass to fire for the parallel jobs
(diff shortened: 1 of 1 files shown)


oscarplewis:
Two PR validation checks failed: the signal and the target_genie checks. The target_genie checks appear similar or identical to the same failed check from the previous successful PR (#1926). The signal check here cannot be explained by the same manner, but may be because of differences in random seeding. No files pertaining to the signal validation were changed in this PR however: the only changed file was the `fire-parallel` script.

tomeichlersmith (changes requested):
This looks good and I think it makes sense. Thank you :)

I also want to point out that you can use the `PARALLEL` environment variable to pass arguments to GNU parallel even with the previous version of `fire-parallel`. https://ldmx-software.github.io/using/batch/fire-parallel.html#passing-options-directly-to-gnu-parallel

My one request is to update the usage description with your `--link` example to show folks how to pass arguments to GNU parallel more clearly.

oscarplewis:
Thanks @tomeichlersmith, I didn't know about the PARALLEL environment variable, but hopefully this will still make things a bit cleaner. I've updated the usage description with two examples that cover using fire-parallel with configs that take multiple arguments and using --link, and changed the description of the `--` delimiter slightly. Let me know if there's anything else you'd like changed or added!

tvami (approved):
Thanks @oscarplewis !

tomeichlersmith (approved):
Looks great, thank you

PR #1932, opened 2026-02-03
Update target conv default settings

### LINKED ISSUE #1931: Update target conv default settings

This block
https://github.com/LDMX-Software/ldmx-sw/blob/8857a15297251639508d81f1be53b2833d2b5769/Biasing/python/target.py#L176-L185
is not right

Should be something like this
https://github.com/LDMX-Software/ldmx-sw/blob/8857a15297251639508d81f1be53b2833d2b5769/Biasing/python/ecal.py

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1931

## Check List

I ran 10000 tries with

```
mySim = target.gamma_mumu('ldmx-det-v15-8gev', generators.single_8gev_e_upstream_tagger())
mySim.description = "Target Conversion Simulation"
p.sequence = [ mySim ]
```
and got
```
[ SimulatorBase ] 10000  warn: Started 10000 events to produce 8 events.
```

### DIFF
--- a/Biasing/python/target.py
+++ b/Biasing/python/target.py
@@ -167,22 +167,22 @@ def gamma_mumu( detector, generator ) :
     sim.setDetector( detector , True )
 
     # Set run parameters
-    sim.description = "gamma -> mu+ mu-, xsec bias 10e9"
     sim.beamSpotSmear = [20., 80., 0.]
-
+    xsec_bias_threshold = 0.625 * generator.energy * 1000.
     tagger_threshold = 0.95 * generator.energy * 1000.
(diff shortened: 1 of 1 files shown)


tvami:
CI test failures are from other PRs

horohoo (approved):
Good catch on this! I wonder if the Xsec bias could be higher than 1e5 for more efficient simulation. Not critical for this PR, but something to think about adjusting in the future.

JYoo001 (approved):

tvami:
>  I wonder if the Xsec bias could be higher than 1e5 for more efficient simulation.

Good point! We'll look into this.

PR #1938, opened 2026-02-04
Include contrib approach in HCal SD

### LINKED ISSUE #1936: Compress HCAL SimHits similar to how contribs are used in the ECAL

```
drwxr-xr-x. 3 vamitamas incandela   84 Feb  4 10:18 ..
-rw-r--r--. 1 vamitamas incandela  95M Feb  4 10:18 v4.5.4_ecal_pn_run160707_event_sim.root
-rw-r--r--. 1 vamitamas incandela 1.3K Feb  4 11:16 cfg_drop_hcal.py
drwxr-xr-x. 2 vamitamas incandela    3 Feb  4 11:16 .
-rw-r--r--. 1 vamitamas incandela  21M Feb  4 11:16 v4.5.4_ecal_pn_run-1_event_sim_drop_hcal.root
```

i.e. 75% of the post-trigger sim-only ecalPN files is the HCAL hits. The ECAL hits are only 3%.
This is because the ECAL has this compression approach

We should do the same in the HCAL

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1936

## Check List

### DIFF
--- a/SimCore/include/SimCore/SDs/HcalSD.h
+++ b/SimCore/include/SimCore/SDs/HcalSD.h
@@ -5,6 +5,7 @@
 /*   DetDescr   */
 /*~~~~~~~~~~~~~~*/
 #include <algorithm>
+#include <map>
 #include <string>
 #include <vector>
 
@@ -83,9 +84,7 @@ class HcalSD : public SensitiveDetector {
   /**
    * Add our hits to the event bus and then reset the container
    */
-  virtual void saveHits(framework::Event& event) override {
-    event.add(COLLECTION_NAME, hits_);
-  }
+  virtual void saveHits(framework::Event& event) override;
 
   virtual void onFinishedEvent() override { hits_.clear(); }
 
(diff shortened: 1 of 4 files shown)


tvami:
I ran the HCAL CI with 10k events:

Standard size: 339MB
After the PR size: 238MB --> 30% reduction

Kinda makes sense that it's less crazy size reduction than the ecal PN as we have less of the PN complexity and that it's not after trigger-skim.

The ECAL PN with same stats

Before PR size: 716MB
After the PR size: 456MB --> 36% reduction

bryngemark (approved):
Looks good, please post some important CI validation plots to the conversation for ease of future reference.

tvami:
Here are all the plots for the HCAL CI

[pr1938_hcal_recon_validation_plots.pdf](https://github.com/user-attachments/files/25101560/pr1938_hcal_recon_validation_plots.pdf)

The first section is the KS failed plots:
The main observations is that we see less hits, but they are more energetic. 

The second section is the KS passed plots:
Really the punchline is that the PE / maxPE is unchanged.

tomeichlersmith (approved):

PR #1937, opened 2026-02-04
Change default settings so only Target + ECAL scoring plane is saved

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1935

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [x] I ran my developments and the following shows that they are successful.

Basic config and ecal PN CIs both run fine.

PR #1947, opened 2026-02-05
fixed PFTrack fatal error in it_pileup config

### LINKED ISSUE #1946: PFTrack, PFlow and PileupFinder errors when running 'validation_samples/it_pileup/config.py'
**Describe the bug**
When running the configuration script found in the validation samples for in-time pileup some routines related to PFlow and PileupFinder run into some errors and their corresponding outputs in 'hist.root' is missing. My guess is that these errors are not too difficult to fix since they seem to be related to incorrect/missing matching references, but I am so new to this so I cannot tell for sure. The bug related to PFTrack has a fix already and a PR will be made very soon.

**To Reproduce**
1. Go to 'ldmx-sw/.github/validation_samples/it_pileup/config.py' 
2. Run script 'denv fire config.py'
    1.a This script requires two .root files generated from 'it_pileup/gen_main.py' and 'it_pileup/gen_pileup.py' to be available in the same directory. You can generate these with 'denv fire gen_main.py; denv fire gen_pileup.py'.
    1.b To run the producer scripts you need to set two params in it which you can do directly in the terminal by typing for example "export LDMX_RUN_NUMBER=13" and "export LDMX_NUM_EVENTS=20" which will eventually result in 10 events.
3. Read terminal output and search for "fatal" or "error" 
4. See error printout for every event
"
[ PFTrack ] 10 fatal: Couldn't find input collection EcalScoringPlaneHitsOverlay_overlay
[ PFlow ] 10 error: Unable to find (one) collection named PFTracks_
[ PileupFinder ] 10 error: Unable to find (one) collection named PFCandidates_
"

**Desired behavior**
That the mentioned routines for PFTrack, PFlow and PileupFinder does not fail so that their corresping outputs in hist.root are available and no error printouts are reported. 

*Lene-Kristian Bryngemark and I have identified a fix for PFTrack and I have soon made a PR for that*

**Screenshots**

**Environment:**
denv ldmx/pro:v4.5.9, 
Ubuntu 24.04.3 LTS

**Additional context**

cat2code:

This partially resolves #1946 : PFTrack fatal error, the other two errors are still present.

## Check List
- [y] I successfully compiled _ldmx-sw_ with my developments.
- [y] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [y] I ran my developments and the following shows that they are successful.

The following fatal error is resolved and is **not showing any longer** in terminal output from  'denv fire config.py'. This requires that the output '.root' file from 'gen_pileup.py' has been generated before running the config file. Credit to @bryngemark for this fix.

**not showing any longer**
'[ PFTrack ] 10 fatal: Couldn't find input collection EcalScoringPlaneHitsOverlay_overlay'


.github/validation_samples/it_pileup/gen_pileup.py
```diff
@@ -9,7 +9,7 @@
 
 from LDMX.SimCore import simulator as sim
 mySim = sim.simulator( "mySim" )
-mySim.setDetector( 'ldmx-det-v15-8gev' )
+mySim.setDetector( 'ldmx-det-v15-8gev', True )
```
tvami: With https://github.com/LDMX-Software/ldmx-sw/pull/1937/ merged this should be 
```suggestion
mySim.setDetector( 'ldmx-det-v15-8gev', include_scoring_planes_minimal = True )
```

tvami (approved):

cat2code:
The latest fix I just pushed in this PR resolves the remaining 2 errors in printout. However I am unsure if this is the preferred way of doing things, so a review is welcome.

tvami (changes requested):
On trunk the variables use snake_case, so the additional changes will have no effect. Also I dont understand how the problem is resolved given that you just added another processor which then does not seem to be used upstream? Or the point is that pfComb wanted to access the hcal clusters but they were never made in the event? In that case I wonder if just the defaults of `pfHcalClusterProducer` should be changed instead of just changing in this config

cat2code:
I wrote the variables based on pro:v4.5.9 and now I understand that the naming I used is obsolete, and for me (not experienced), handling the version update made this more complicated and I don't know what option is the best: should I update just this part or the whole file (config.py)? 

"problem resolved" merely refers to adjusting for desired behavior defined in #1947 and adding the processor resolved the remaining error printouts (pro:v4.5.9). I believe @bryngemark can answer your other thoughts better than I can and how things should ideally be designed, it's almost 6 pm here so it will have to wait until next week. Cheers,

cat2code:
This would be the corresponding snippet adjusted for snake_case

`# HCAL clusters for pfProducer`
`hcal_pf = pf_reco.pfHcalClusterProducer()`
`hcal_pf.min_cluster_hit_mult = 2`
`hcal_pf.cluster_hit_dist = 100.`
`hcal_pf.min_hit_energy = 0.1`
`hcal_pf.do_single_cluster = False`
`hcal_pf.log_energy_weight = True`
`hcal_pf.hit_pass_name = this_pass_name   #run on process+pileup`

tvami:
I'd edit these directly here
https://github.com/LDMX-Software/ldmx-sw/blob/e5dbf6964d16eded2af3e2319f95f7f0cdc8cefa/Recon/python/pf_reco.py#L31-L38

But it actually seems like they match exactly what you had, no no real need to add the variables in the CI config

.github/validation_samples/it_pileup/config.py
```diff
@@ -128,6 +128,15 @@
 cluster.reclustering = True                                                                                                                                                                                
 cluster.rec_hit_pass_name=thisPassName #run on process+pileup       
 
+# HCAL clusters for pfProducer
+hcalPF = pfReco.pfHcalClusterProducer()
+hcalPF.minClusterHitMult = 2
+hcalPF.clusterHitDist = 100.
+hcalPF.minHitEnergy = 0.1
+hcalPF.doSingleCluster = False
+hcalPF.logEnergyWeight = True
```
bryngemark: if these are the same as the defaults you can skip these lines.

.github/validation_samples/it_pileup/config.py
```diff
@@ -128,6 +128,15 @@
...
+hcalPF.input_pass_name = thisPassName   #run on process+pileup
```
bryngemark: one feature with how our input collection lookup is done is that you just have to specify the pass name explicitly if there are several collections with the same name; in this case i think hcal clustering is only run once so it can be found by only collection name, in which case you can skip this line too.

bryngemark (commented):
looks good, can you verify that the complaints about missing collections are gone now or is there still more to be done?

bryngemark:
@cat2code when you're happy with it, push the `Ready for review` button to trigger the validation test suite.

cat2code:
The complaints about missing collections are gone in this latest version. I took the latest version of it_pileup/config.py that has been updated to the enforced coding rules so the latest commit should be consistent with snake_case and the current codebase. 

As both of you saw the default parameters were unnecessary to include and also the pass name so they have been omitted. The only thing in this script that should be different from the latest version of ldmx-sw is the insertions in row 144, 282.

cat2code:
Whoops I missed including hcal_pf in pileup sequence (row 282), now it is as I want it to be.

bryngemark (commented):
i'm a little confused that there are lots of diffs with `trunk` that are just about the case convention, i thought this was already implemented on `trunk`?

it looks like you have a couple of merge conflicts you need to resolve. perhaps this cleans up the file diffs too.

.github/validation_samples/it_pileup/gen_pileup.py
```diff
@@ -12,10 +12,8 @@
 p.logger.term_level = 4
 
 from LDMX.SimCore import simulator as sim
-
-
-my_sim = sim.simulator( "my_sim" )
-my_sim.setDetector( 'ldmx-det-v15-8gev' )
+mySim = sim.simulator( "mySim" )
```
tvami: `mySim` should be `my_sim` in this line and the next too

tvami:
I looked at the output, i.e like

https://github.com/LDMX-Software/ldmx-sw/actions/runs/22954010912/job/66627391602?pr=1947
```
[ PFTrack ] 1 fatal: Couldn't find input collection EcalScoringPlaneHitsOverlay_overlay
[ PFlow ] 1 error: Unable to find (one) collection named PFTracks_
[ PileupFinder ] 1 error: Unable to find (one) collection named PFCandidates_
```

Screenshot 2026-03-11 at 08 18 57

So the issues doesnt seem to be resolved?

tvami:
https://github.com/LDMX-Software/ldmx-sw/pull/2024 supersedes this PR

PR #1949, opened 2026-02-05
Add ruff rules and recipes for python formatting

### LINKED ISSUE #1392: Enforce python formatting

Now that we are enforcing c++ formatting rules, maybe it's time to think about Python formatting as well.

Programs that are widely used and do this are 
 * Ruff: https://docs.astral.sh/ruff/
 * Black: https://github.com/psf/black

Note Ruff would have caught the issue about forgetting to `import os` that  triggered https://github.com/LDMX-Software/ldmx-sw/releases/tag/v4.0.3 and https://github.com/LDMX-Software/ldmx-sw/commit/640596b1541437faf1d9429901696de132127436
(sorry about that btw!)

Keep it the wild west!

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1392

But I'm opening this more as a working discussion.  

## Check List

I plan to add the following to the docs

-----------------------------------------------------------------------------------
# Python Code Style Guide

ldmx-sw uses [Ruff](https://docs.astral.sh/ruff/) for Python code formatting and linting.

Format your Python code:
```bash
just format-python
```

Check for lint issues:
```bash
just lint-python
```

Auto-fix lint issues:
```bash
just lint-python-fix
```

Format only changed files (compared to trunk):
```bash
just format-python-diff
```

## Configuration

Python formatting rules are configured in [`pyproject.toml`](../pyproject.toml).

Key settings:
- **Line length**: 88 characters
- **Quote style**: Double quotes
- **Import sorting**: Enabled with `isort` rules
- **Auto-fixes**: Enabled for most rules

## Linting Rules

Ruff checks for:
- **E, W**: pycodestyle errors and warnings
- **F**: Pyflakes (unused imports, undefined names, etc.)
- **I**: Import sorting and organization
- **N**: PEP 8 naming conventions
- **UP**: Python version-specific upgrades
- **B**: Common bugs and design problems
- **C4**: List/dict/set comprehension improvements
- **SIM**: Code simplification suggestions
- **RUF**: Ruff-specific rules

### Command Line
You can also run ruff directly in the denv:
```bash
# Format files
denv ruff format path/to/file.py

# Check for issues
denv ruff check path/to/file.py

# Auto-fix issues
denv ruff check --fix path/to/file.py
```


justfile
```diff
@@ -188,6 +187,30 @@ format-cpp: format-cpp-all
...
+    fi
+    rm "${py_list}"
+
+# lint the Python source code with ruff
+lint-python:
+    denv ruff check
+
+# lint and auto-fix the Python source code
+lint-python-fix:
+    denv ruff check --fix
```
tomeichlersmith, line 212: Have you checked that this does anything?

Since we keep our Python files in weird places, I'm guessing we will need to update `pyproject.toml` in order to be able to have `ruff` find them all automatically.

tomeichlersmith (commented):
Thank you for kickstarting this, I think it will be helpful.

justfile
tvami: Hey @tomeichlersmith I didnt run any of it, because I wanted to agree on the rules we'll be enforcing before applying anything, thus opening this PR. But if you think the rules are reasonable, then I can go to the next stage

tomeichlersmith: That makes sense, I think the rules are reasonable - we aren't doing anything super complicated in these python modules since they are just for configuration so (hopefully) that makes it easier for us to conform to community standards.

tvami: It does work, I pushed all what it did

tvami:
So the comments in https://github.com/LDMX-Software/ldmx-sw/actions/runs/21758941782/job/62776992260?pr=1949 seem to be the real important ones where `outPath` should really be `out_path` etc, but this needs c++ change so I'd prefer to do it in another PR

tomeichlersmith:
Cool - I'm comfortable applying the linting/formating rules in stages in order to keep them organized. This first pass seems to be a pure format which is just cleaning up whitespace and the like which I think is good to do in bulk.
However, we may need to figure out how to bypass the "remove unused imports" rule since we make heavy use of "unused imports" in order to define the different conditions.

```
  [ fire ] 1 fatal: [ConditionUnavailable] : No provider is available for : HcalGeometry
```
is resolved by loading that condition with
```
import LDMX.Hcal.HcalGeometry
```
but this import is not used in the config files so they are removed by format/linting.

tvami:
> but this import is not used in the config files so they are removed by format/linting.

Can we somehow use it? Instantiate the geometry directly?

tomeichlersmith:
We could but that will just create a unused variable that the linter will probably also complain about

https://github.com/LDMX-Software/ldmx-sw/blob/8b43459b6c6ce827ceef88017ec9f50b3aabe3b1/Ecal/python/EcalGeometry.py#L46

We could refactor how Condition Object Providers are registered with the Process but I really like having the `import` statement "import" a condition. It makes sense to me and it is nice, short, and avoids typos. For these reasons, I'd prefer to disable the unused import lint rule for config scripts.

tvami:
So 
```
ignore = ["F401"] 
```
did the job

LiamBrennan-UCSB (approved):

tomeichlersmith (approved):

PR #1950, opened 2026-02-05
Add the ability to time-shift HCAL simhits before digitizing them

### LINKED ISSUE #1343: Add the ability to time-shift simhits before digitizing them
For the testbeam simulations, we need some way of handling that the beam was asynchronous with the readout electronics. To do this, we can simply add a configurable timeshift in the `DigiProducer` for the entire event. In principle we could also add a shift bar-by-bar but for now this is sufficient. I also believe that there is some plan to do this in the HGCROC emulator at some point, there is at least a TODO note suggesting it.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1343

100% of the credit goes to @EinarElen I just took his changes in the submodule and added here.

## Check List

HCAL CI passes

### DIFF
--- a/Hcal/src/Hcal/HcalDigiProducer.cxx
+++ b/Hcal/src/Hcal/HcalDigiProducer.cxx
@@ -16,12 +16,25 @@ HcalDigiProducer::HcalDigiProducer(const std::string& name,
     : Producer(name, process) {
   /*
    * Noise generator by default uses a Gausian model for noise
-   * i.e. It assumes the noise is distributed around a mean (setPedestal)
+   * i.e. it assumes the noise is distributed around a mean (setPedestal)
    * with a certain RMS (setNoise) and then calculates
-   * how many hits_ should be generated for a given number of empty
+   * how many hits should be generated for a given number of empty
    * channels and a minimum readout value (setNoiseThreshold)
    */
   noise_generator_ = std::make_unique<ldmx::NoiseGenerator>();
+}
+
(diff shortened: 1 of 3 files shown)


tomeichlersmith (approved):

tvami:
I ran this on a clean gold and the signal still comes up. I'm a little worried how much fluctuations there are in the signal. I guess that should be a different issue...

Screenshot 2026-02-06 at 07 10 20

Hcal/src/Hcal/HcalDigiProducer.cxx
```diff
@@ -16,21 +16,34 @@ HcalDigiProducer::HcalDigiProducer(const std::string& name,
     : Producer(name, process) {
   /*
    * Noise generator by default uses a Gausian model for noise
-   * i.e. It assumes the noise is distributed around a mean (setPedestal)
+   * i.e. it assumes the noise is distributed around a mean (setPedestal)
    * with a certain RMS (setNoise) and then calculates
```
bechenard, line 20: Bonus points for fixing the English!

bechenard (approved):
This is good and well coded (I won't quibble for TimeSpread vs time_spread)

Hcal/src/Hcal/HcalDigiProducer.cxx
tvami: Going for the extra credit :P

Issue #1303, opened 2022-05-12
Python Config Validation


tomeichlersmith:
**Is your feature request related to a problem? Please describe.**
It is _very_ common to misspell parameters and occasionally have parameters with incorrect types. We can avoid this issue by only allowing the insertion of new parameter names to occur _once_ at class construction.

**Describe the solution you'd like**
I sketched out a solution that will check the existence and type of parameters when they are attempting to be set.
```python
       : class Base :
    ...:     def __init__(self, **kwargs) :
    ...:         self.__dict__ = kwargs
    ...:     def __setattr__(self, n, v) :
    ...:         if n in self.__dict__ :
    ...:             if isinstance(v, type(self.__dict__[n])) :
    ...:                 self.__dict__[n] = v
    ...:             else :
    ...:                 raise AttributeError(f'{n} was given as type {type(v)} but should be {type(self.__dict__[n])}')
    ...:         elif n == '__dict__' :
    ...:             super().__setattr__(n,v)
    ...:         else :
    ...:             raise AttributeError(f'{n} is not a defined parameter.')
```
Then child classes would be defined so that their constructor defines the parameters that the class uses.
```python
In [13]: class Child(Base) :
    ...:     def __init__(self) :
    ...:         super().__init__(one = 1, two = 2)
    ...: 

In [14]: c = Child()

In [15]: c.one
Out[15]: 1

In [16]: c.one = 2

In [17]: c.two
Out[17]: 2

In [18]: c.one
Out[18]: 2

In [19]: c.three
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-19-e851540052ba> in <module>
----> 1 c.three

AttributeError: 'Child' object has no attribute 'three'
```

## Loopholes
You can get around this solution by using the hole created for setting the `__dict__` member in the `__init__` method.
```python
old = c.__dict__
old['three'] = 3
c.__dict__ = old
```
This is helpful to know about in case the user needs to add a parameter that isn't in the python module or misspelled there for some reason. This is also hard enough to do that I don't see someone stumbling upon it.

tomeichlersmith:
More fleshed out base class:

```python
class Parameters :
    """Python configuration class to help validate existence and type

    The members of this class and its children are dynamically defined
    using the key-word arguments to the constructor. Then any later attempts
    to set its members (aka attributes) will fail if the member does not
    exist or the new value is the wrong type.

    Parameters
    ----------
[...]

PR #1952, opened 2026-02-07
Move beam smearing to gen instead of sim

### LINKED ISSUE #1312: Make beam-spot smearing the default and move to the generator
This came up in the latest SW Dev meeting while talking about validating the ECal PN sample for 8GeV beam particles. It is a common occurrence (especially for new users) to forget the extra config line that enables the standard LDMX beam spot smearing. Maybe we want to make the standard beam spot smearing the default and then allow users to disable it if they so wish?

This would be a very easy change to make, but it has the potential to break many prior configs that rely on _not_ smearing the beam spot. Calling for all thoughts possible! @bryngemark @EinarElen @tylerhoroho-UVA @jpasc27 

### Proposed Changes
Update default beam-spot smear

https://github.com/LDMX-Software/SimCore/blob/f1988b5579dd00ae27ad5c6f089881d8ee5e6248/python/simulator.py.in#L75

to

```python
        self.beamSpotSmear = [ 20., 80., 0. ]
```

and add a new function to make disabling the beam spot a little more ergonomic.

```python
    def disableBeamSpotSmear(self):
        self.beamSpotSmear = []
```

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1312

## Check List

### DIFF
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -7,7 +7,6 @@ from LDMX.Biasing import ecal
 from LDMX.SimCore import generators as gen
 det = 'ldmx-det-v15-8gev'
 mySim = ecal.photo_nuclear(det,gen.single_8gev_e_upstream_tagger())
-mySim.beamSpotSmear = [20.,80.,0.]
 mySim.description = 'ECal PN Test Simulation'
 
 p.sequence = [ mySim ]
--- a/.github/validation_samples/inclusive/config.py
+++ b/.github/validation_samples/inclusive/config.py
@@ -7,7 +7,6 @@ det = 'ldmx-det-v15-8gev'
 mySim.setDetector(det, include_scoring_planes_minimal = True )
 from LDMX.SimCore import generators as gen
 mySim.generators.append( gen.single_8gev_e_upstream_tagger() )
-mySim.beamSpotSmear = [20.,80.,0.]
 mySim.description = 'Basic test Simulation'
 
 p.sequence = [ mySim ]
--- a/.github/validation_samples/it_pileup/gen_main.py
+++ b/.github/validation_samples/it_pileup/gen_main.py
@@ -11,7 +11,6 @@ p.logger.termLevel = 4
 from LDMX.Biasing import ecal
 from LDMX.SimCore import generators as gen
 mySim = ecal.photo_nuclear('ldmx-det-v15-8gev',gen.single_8gev_e_upstream_tagger())
-mySim.beamSpotSmear = [20.,80.,0.]
 mySim.description = 'ECal PN Test Simulation'
 
 p.sequence = [ mySim ]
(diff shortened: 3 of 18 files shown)


tvami:
OK it's a little less transparent than i've hoped for... all the CIs that use a gun have changed seeds to we have lots of changes, all compatible with the original distributions to my eye (not to KS, but oh well...)

tvami:
[pr1952_hcal_recon_validation_plots.pdf](https://github.com/user-attachments/files/25152675/pr1952_hcal_recon_validation_plots.pdf)
[pr1952_deep_ecal_gun_recon_validation_plots.pdf](https://github.com/user-attachments/files/25152677/pr1952_deep_ecal_gun_recon_validation_plots.pdf)

(cant upload the other ones, the PDFs are too big...)

stantler8-arch (dismissed):

LiamBrennan-UCSB (approved):

tomeichlersmith (commented):
This is excellent work 🎉 thank you

My only concern is this breaking different configurations. The Python configuration system silently ignores parameters that are not inspected by the C++, so someone could

1. put `sim.beamSpotSmear = [some,thing,custom]` into their config (or have it still there from a previous config)
2. run and see no warnings or errors
3. assume that beam spot smearing was appropriately applied (it was not, the default values were applied)

The long term solution (I believe) is to update the python side with more fancy parameter checking https://github.com/LDMX-Software/ldmx-sw/issues/1303 (currently blocked by having to do the translation from the current Python2-like implementation to the `@dataclass`-like implemenation for all >300 configuration classes).

@bryngemark Thoughts? Since we are making beam spot smearing the default, I am not as worried but want to check with someone who does more production.

tvami:
> The long term solution (I believe) is to update the python side with more fancy parameter checking https://github.com/LDMX-Software/ldmx-sw/issues/1303 (currently blocked by having to do the translation from the current Python2-like implementation to the @dataclass-like implemenation for all >300 configuration classes).

OK I'd rather work on this and then to add some safety guard for people adding this to the wrong processor

bryngemark:
I agree with @tomeichlersmith's assessment of the situation. I think a python parameter check would be **awesome** of course. Honestly, in the current setup, I think only having the beam spot smearing set by default will save my ass next time I get to sample production. I will have forgotten all about this change at that point. 

Another way to discourage reusing old configs could be to have more standard setup functions that are easy to pull from CI configs, which hide a lot of details and make it more intutive to define sample generation. Just like we have standard beam configurations (single 8 GeV electrons from upstream of the detector, etc) how clunky would it be to set up have a few functions for the most standard components like "LDMXbeam", "ecal PN (this is already basically there), "full reconstruction", ... ? For sure I think every non-expert would appreciate a one-liner for doing standard TS reco 😄 

And maybe who really benefits from this is the newcomer who will just need some samples for a short study.

tvami:
> a few functions for the most standard components like 

Yeah I kinda had this for the tracking

https://github.com/LDMX-Software/ldmx-sw/blob/trunk/.github/validation_samples/ecal_pn/config.py#L94-L95

tomeichlersmith (approved):
Yea I like the `p.sequence.extend(my_module.sequence)` syntax. My only suggestion would be to make it a function rather than a list to avoid having so many global instances of configuration objects floating around but that nit-pick can be for a different PR/Issue.

PR #1956, opened 2026-02-09
Fix PEP8 violations N816, N806 and N803 (naming)

### LINKED ISSUE #1392: Enforce python formatting

tvami:

Part of the python formatting issue https://github.com/LDMX-Software/ldmx-sw/issues/1392

## Check List

I tested the inclusive and the ecal PN CIs


tvami:
OK, I think we should cut a minor release after this is merged, maybe together with https://github.com/LDMX-Software/ldmx-sw/pull/1952 as they both change python conf settings

tvami:
I'll fix the basic test later, but I want the real physics tests to go thru before that

tvami:
Great! The signal failure is known, the rest is just diffs in the logs, due to renamings.

LiamBrennan-UCSB (approved):

tvami:
> The signal failure is known, the rest is just diffs in the logs, due to renamings.

This is not true anymore bc I merged the beamspot smearing PR, but anyway it was true yesterday!

tomeichlersmith (commented):
I think this works and makes sense, but I want to be very clear. **This will break current configuration scripts in a way that my work on #1303 will _not_ resolve.**

While #1303 has a mechanism to help folks rename member variables of configuration classes (stuff like `p.logFrequency` -> `p.log_frequency`), it does **not** have the mechanism to help rename modules within `import` statements.

Personally, I am comfortable with this sort of breaking change since reviewing and updating configs is something that we can do regularly and happens "right away" in the sense that the issue happens immediately when attempting to run. I would like to hear @bryngemark thoughts on this as well before we merge it.

tvami:
> This will break current configuration scripts in a way that my work on https://github.com/LDMX-Software/ldmx-sw/issues/1303 will not resolve.

Yeah that's why I said that we should have a minor release after this

bryngemark (approved):
ugh.

as much as seeing these 200 changed configs inspires dread, thinking of all the habits and configs that will need to change on my end, i think i brought this upon myself and it's a much-needed harmonization. so.

1. thank you, let's rip off this bandaid, and, 
2. pretty please @tvami, can you share both here in the PR (for longer-term documentation) and in the next release update talk (as in, you or whoever will give it), a clear process to apply this variable style change to one's local code prior to any PRs in the coming transition period?

tvami:
> can you share both here in the PR ...  a clear process to apply this variable style change to one's local code prior to any PRs in the coming transition period?

Doing `just lint-python-fix` will do as much as it can automatically, beyond that it will give a list of things that should be changed and it couldnt, then one does a replace for those

PR #1960, opened 2026-02-11
Adding 'ldmx-vertTS-v14-8gev' geometry for studies of TSPads with additional vertical bars.

cjbarton151:
This is a non-issue, intended to add an experimental geometry for a student project and potentially for further development. There should be no conflicts with other parts of ldmx-sw, since only new files are added, with no changes to existing files needed.

## Check List

I would add some plots, but COSMOS is temporarily down. However, I tested the v14i geometry with a v14-compatible config file. I saw something, at least, in TriggerPadTracksX and TriggerPadTracksY, which was different from single values, but didn't look at it too closely. I also checked other TS-related collections and they seemed correct at a glance.


tvami:
If you look at this README https://github.com/LDMX-Software/ldmx-sw/tree/trunk/Detectors/data#readme
the general idea was to have small alternative variations of a geometry be with a descriptive name in front of the `v`, i.e. `ldmx-v14-8gev` --> `ldmx-lyso-v14-8gev` for LYSO (which then had several revisions thus the `r`), or `ldmx-ti-v14-8gev`  for Ti target. I wonder if you could find such short description instead of just having the `i` after the v14, what is the `i` supposed to mean?

bryngemark:
Agree with @tvami , didn’t have time to discuss this with you @cjbarton151 , I propose vertTS as descriptor. _i_ is really just historical, showing that I wasn’t happy until the 9th version! And then it got ported over the years by others working on vertical TS bars stuff.

Detectors/data/ldmx-det-v14-8gevi/detector.gdml
```diff
@@ -0,0 +1,228 @@
...
+      <auxiliary auxtype="DetElem" auxvalue="Top"/>
+    </volume>
+  </structure>
+    
+  <userinfo>
+    <!-- detector version -->
+    <auxiliary auxtype="DetectorVersion" auxvalue="14">
+      <auxiliary auxtype="DetectorName" auxvalue="ldmx-det-v14"/>
+      <auxiliary auxtype="Description" 
+                 auxvalue="The Late 2022 Design Report Geometry."/>
```
tvami, line 111: Please edit this

Detectors/data/ldmx-det-v14-8gevi/detector.gdml
```diff
@@ -0,0 +1,228 @@
...
+        <rotationref ref="identity"/> 
+      </physvol>
+      <auxiliary auxtype="DetElem" auxvalue="Top"/>
+    </volume>
+  </structure>
+    
+  <userinfo>
+    <!-- detector version -->
+    <auxiliary auxtype="DetectorVersion" auxvalue="14">
+      <auxiliary auxtype="DetectorName" auxvalue="ldmx-det-v14"/>
```
tvami, line 109: Please edit this

Detectors/data/ldmx-det-v14-8gevi/detector.gdml
```diff
@@ -0,0 +1,228 @@
...
+    &materials;
+  </materials>
+
+  <solids>
+    <box lunit="mm" name="world_box" x="world_dim" y="world_dim" z="world_dim"/>
+  </solids>
+    
+  <structure>
+    <volume name="World">
+      <materialref ref="Vacuum"/>
```
tvami, line 65: Let's make the world air

tvami (commented):
Beyond the documentation points, I think I have some real comments too:
I'm not sure how this is with the TS, but if somebody wants to use the HCAL and ECAL with this geometry [unclear if this is a TS only project or not] then this
https://github.com/LDMX-Software/ldmx-sw/blob/5e6c79c9928e97935291a770083e3943d17ab5e7/DetDescr/python/EcalGeometry.py#L113
and this
https://github.com/LDMX-Software/ldmx-sw/blob/5e6c79c9928e97935291a770083e3943d17ab5e7/DetDescr/python/HcalGeometry.py#L475
needs to have the new geometry you are introducing [with the new name we discussed]

Detectors/data/ldmx-det-v14-8gevi/detector.gdml
bryngemark: +1 for this, it has changed since, and air is the only thing upstream of the TS

cjbarton151:
Hey Tamas,

We don't foresee the student project needing to be compatible with the ECal or HCal. Even so, I've made all the changes requested. While I was at it, I double-checked after the changes that we were still getting some readout from TriggerPadTracksX and TriggerPadTracksY

image

bryngemark (approved):

bryngemark:
thanks for making these plots, it all looks good!

tvami:
I see the branch https://github.com/LDMX-Software/ldmx-sw/compare/trunk...iss1137 is doing something similar, can it be deleted? That branch is also before we pruned the git history, so if the content is useful (I see some c++ changes) then the new C++ code should be in a new branch anyway

PR #1962, opened 2026-02-13
iss1700 2e tracking sequence

### LINKED ISSUE #1700: PR validation on pileup sample doesn't seem to use overlay collection in tracker dqm
**Describe the bug**
I was checking the validation histograms from iss #1699 and e.g. the recoil track multiplicity closely resembles the distribution in the ecal PN validation suite. 

**To Reproduce**
1. Go to #1699 
2. Click on the validation checks --> view details --> Summary 
3. Scroll down to Artifacts 
4. Download `it_pileup` and `ecal_pn` plots, and compare the Ntracks distribution in the recoil tracker (examples added here)

**Desired behavior**
The validation suite on pileup uses pileup collection input everywhere.

**Plots**
from ecalPN+pileup run
[RecoilDQM_RecoilDQM_N_tracks.pdf](https://github.com/user-attachments/files/20053717/RecoilDQM_RecoilDQM_N_tracks.pdf)
from 1e ecal PN run 
[RecoilDQM_RecoilDQM_N_tracks.pdf](https://github.com/user-attachments/files/20053723/RecoilDQM_RecoilDQM_N_tracks.pdf)

oscarplewis:

This PR introduces the function `setOverlay`  in the `full_tracking_sequence.py` python config as the two-electron (overlay compatible) version of the existing tracking sequence. This enables the full tracking sequence with DQMs to be run on collections which have at minimum the default collections (found in the new `overlay.py`) overlaid. Currently SimParticles overlay is not included in the OverlayProducer which creates some problems with 2e tracking downstream, but everything else is set up.

This PR also moves several hardcoded references to scoring plane and sim particle collections to configurable parameters in the corresponding python files.

Finally, this work resolves #1700 by ensuring that the overlay collections are used for the it-pileup config everywhere relevant and the 2e tracking sequence is employed.

## Check List

Below are (respectively) the RecoilDQM/N_tracks_ and TaggerDQM/N_tracks_ plot for the it_pileup sample according to the new `config.py`. The bins are inclusive to the left, so 0 tracks is in the first bin, 1 in the second, and so on. While these are less than desirable results, the difficulty arises because of non-unique track IDs in the sim collections which should be fixed in a later PR: nevertheless, these plots show that the tracking is correctly sourcing overlay collections now, which it wasn't before.

RecoilDQM:N_tracks
TaggerDQM:N_tracks


.github/validation_samples/it_pileup/config.py
```diff
@@ -17,6 +17,8 @@
 p.run = int(os.environ['LDMX_RUN_NUMBER'])
 p.max_events = int(os.environ['LDMX_NUM_EVENTS']) // 2
 
+p.inputFiles = ['ecal_pn.root']
```
tvami: This is not needed, LL273 already has it

https://github.com/LDMX-Software/ldmx-sw/blob/trunk/.github/validation_samples/it_pileup/config.py#L273

```suggestion
```

tvami (changes requested):
This is great, thanks Oscar! I had a single small change requested

oscarplewis:
Updated PR description with images showing successful tests run on the new `it_pileup/config.py` and with details of the new commits. With this, assuming the validation checks pass, the PR should be ready to merge.

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -121,13 +133,16 @@ void OverlayProducer::produce(framework::Event &event) {
         ldmx_log(trace) << simhit;
         // this copies the hit, its ID and its coordinates directly
         hit_map[simhit.getID()] = simhit;
+        for (int i = 0; i < simhit.getNumberOfContribs(); i++) {
+          ldmx_log(trace) << "   " << simhit.getContrib(i);
+        }
```
tvami: I'd like to avoid an extra loop, since the only thing it does is just a printout

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -86,7 +93,12 @@ void OverlayProducer::produce(framework::Event &event) {
       calo_collection_map;
   std::map<std::string, std::vector<ldmx::SimTrackerHit>>
       tracker_collection_map;
+  std::map<std::string, std::map<int, ldmx::SimParticle>>
+      particle_collection_map;
   std::map<int, ldmx::SimCalorimeterHit> hit_map;
+  int track_id_increment =
+      -1;  // there should only be one SimParticles collection, so it's fine for
```
tvami: Clang broke this up in a weird way, it works if you put comment above
```suggestion
  // there should only be one SimParticles collection, so it's fine for ...
  int track_id_increment = -1;  
```

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -361,6 +430,17 @@ void OverlayProducer::produce(framework::Event &event) {
...
   }
+
+  // and finally for sim particles
+  for (auto &[name, coll] : particle_collection_map) {
+    ldmx_log(debug) << "Writing " << name << " to event bus.";
+    ldmx_log(trace) << "List of particles added: ";
+    for (auto &[track_id, particle] : coll) {
+      ldmx_log(trace) << "Track ID: " << track_id << " --- " << particle;
+    }
+    event.add(name, coll);
```
tvami: Is this a duplicate? L431 has something similar but maybe that's for a different name and coll pair

tvami (changes requested):
Looks good, I have a few questions + suggestions

Recon/src/Recon/OverlayProducer.cxx
oscarplewis:
```suggestion
```
Suggested change to eliminate superfluous loop leftover from debugging

Recon/src/Recon/OverlayProducer.cxx
oscarplewis: This is for the SimParticles branch overlay; LL424-431 are for overlaying the SimTrackerHits collections. Despite the similar names, these are two different operations.

oscarplewis:
I can't determine the cause of the target_genie validation failure, but the failures themselves look relatively minor and the plots look very similar to those plots which passed the validation. I glanced through all the failed plots, and have attached what appeared to me to be the worst offenders below. I can't imagine any change I made that impacted the genie generator, though, so @tvami if you have any ideas that would be appreciated.

[SampleValidation_hardbremdaughters_endZ.pdf](https://github.com/user-attachments/files/25382977/SampleValidation_hardbremdaughters_endZ.pdf)
[SampleValidation_hardbremdaughters_startZ.pdf](https://github.com/user-attachments/files/25382984/SampleValidation_hardbremdaughters_startZ.pdf)

tvami:
We can ignore the genie changes, those are from another PR. Let's see if the tracking plot for 2e PU CI are changed according to our expectation

tvami:
Now only the 2e PU fails, as expected
Screenshot 2026-02-18 at 06 45 15

tvami:
OK so tracking is measuring the PU tracks for sure now
Screenshot 2026-02-18 at 06 48 49

We have more fakes, I believe, but that seems more realistic.

But most importantly the Ntracks is peaking at 2 yaaay
Screenshot 2026-02-18 at 06 52 08

The pulls look worse
Screenshot 2026-02-18 at 06 53 32

But I guess that's down to the 2e tracking and this PR just revealed it.

tvami (approved):

bryngemark:
alright -- i suppose this PR supersedes #1963 ? please see my review in that one then 😄

.github/validation_samples/it_pileup/config.py
```diff
@@ -226,21 +235,24 @@
...
 
-dqm_with_overlay = trigScint_sim_dqm + trigScint_dqm + [trigger_dqm, ecal_digi_verify, ecal_shower_features, ecal_mip_tracking_features, ecal_veto_results] + hcal_dqm
+dqm_with_overlay = trigScint_sim_dqm + trigScint_dqm + [sim_dqm, trigger_dqm, ecal_digi_verify, ecal_shower_features, ecal_mip_tracking_features, ecal_veto_results] + hcal_dqm
 
 p.logger.term_level = 1
 
 # Add full tracking for both tagger and recoil trackers: digi, seeds, CFK, ambiguity resolution, GSF, DQM
-p.sequence.extend(full_tracking_sequence.sequence)
-p.sequence.extend(full_tracking_sequence.dqm_sequence)
+from LDMX.Tracking import full_tracking_sequence_2e
```
bryngemark: like i mentioned in #1963 i would prefer not having a doubled full tracking sequence if we can avoid it, and find a way to add `Overlay` to the 1e one instead.

DQM/src/DQM/TrigScintHitDQM.cxx
```diff
@@ -47,7 +47,7 @@ void TrigScintHitDQM::configure(framework::config::Parameters &ps) {
   hit_collection_name_ = ps.get<std::string>("hit_collection");
   pad_name_ = ps.get<std::string>("pad").c_str();
 
-  trig_scint_passname_ = ps.get<std::string>("trig_scint_passname");
+  trig_scint_passname_ = ps.get<std::string>("pass_name");
```
bryngemark: this style is inconsistent

bryngemark (changes requested):
ok, see review in #1963 for requested changes and some questions.

oscarplewis:
@bryngemark, I intended for #1963 intended to follow this PR, which is the reason for all the very similar commits. If all the content from this PR (which is now renamed per @tvami 's request to better describe what it actually does; the original issue fix for #1700 is now a byproduct of the broader changes) is removed from #1963, then the changes there become more or less limited to the track ID management.

As for your suggested changes, I've fixed the naming inconsistency in the TrigScintHitDQM and made a go at incorporating the `full_tracking_sequence_2e.py` information into the existing `full_tracking_sequence.py`. Now there's a function at the bottom titled `setOverlay()` which iterates through the tracking and DQM lists and updates collection names as needed. 
- I'm uncertain whether it's better (or more consistent with the rest of ldmx-sw) to rename all of the collections with "Overlay" or to only change the sim collections and leave the others to be determined based on pass name - if you have any suggestions that'd be appreciated!

Regarding the suggested change to the logging loops: I don't know enough about the logger to figure that out. I can't figure out if there's a way to pull the user-specified severity level from `ldmx_log` to compare against, but if there is then that's definitely an option.

oscarplewis:
All the commits are here again bc I rebased the branch onto trunk

Recon/src/Recon/OverlayProducer.cxx
tvami: I dont think you applied this? Or at least I still see it on github

oscarplewis:
The signal validation failure is only because the time taken in each step of the process is slightly different. See screenshot of the validation print out below. This failure is leftover from #1966. 

Screenshot 2026-02-19 at 18 34 50

There's a fatal error in the it_pileup now which doesn't seem to be due to this PR, since it is also occurring in #1968. Not sure why this is here.

tvami:
You can ignore the signal indeed, but the it_pileup failure is crucial. Your other PR shares half of this code so not too surprised it's in both. But did you re-initiate the PR validation? I see the it_pileup to be running now

oscarplewis:
> Your other PR shares half of this code so not too surprised it's in both. 

Actually, #1968 shares no code with this PR. It is just a small update to the OverlayProducer to include HcalSimHits in the list of collections that use contribs.

> But did you re-initiate the PR validation? I see the it_pileup to be running now

I did yeah, I turned on the debug logging to get a better picture of what's going on.

oscarplewis:
Okay, the it_pileup fatal error didn't happen during this rerun, so still not sure what's happening there. But it does show results that we wanted.

The N_tracks from reconstruction is peaking at 2 as desired.

RecoilDQM_N_tracks.pdf

Here's the truth N_tracks for comparison.

RecoilDQM_truth_N_tracks.pdf


Here's the pull q/p to compare with the plot Tamas posted above. It looks better this time.

RecoilDQM_pull_qop.pdf

tvami:
Looks great thanks @oscarplewis . @bryngemark do you have any further comments?

bryngemark (commented):
Hello, this mostly looks good, I have two major points.

**About the overlay string:** the reason we have to rename the simhit (and simparticle) collection in overlay (as annoying as it is) is that we as good as always want to have both the no-pileup and the with-pileup collection in the output file, to be able to study performance in pileup. In the mode of overlay where we pull the no-pileup collection from an input file, we could just use different pass names and be done with it (just need to specify the relevant pass name for every downstream processor). But in the overlay-on-the-fly mode, we are creating both collections in the same pass, so we need them to be distinguishable. 

For the reconstructed collections, having separate output collection names is rarely needed, and the rare occasion when it is doesn't warrant propagating a collection name change through as default. So what might be handy is a renaming function that appends the same string to overlay output and digi input. The digi output collections can use the default name such that hit reconstruction etc downstream is unaffected. 

**My second point** is that this PR includes the overlay track ID fix which is still not resolved, as we're actively looking for a solution. While we figure that out, I'd like us to merge only the changes that are unique to this PR and I understood that you had fixed this but a rebasing reintroduced a lot of code changes. Ideas for how to deal with this beyond simply holding off on merging?

oscarplewis:
> For the reconstructed collections, having separate output collection names is rarely needed, and the rare occasion when it is doesn't warrant propagating a collection name change through as default. So what might be handy is a renaming function that appends the same string to overlay output and digi input. The digi output collections can use the default name such that hit reconstruction etc downstream is unaffected.

I took your advice on this and updated the `full_tracking_sequence.py` so that the reconstructed tracking uses pass names to differentiate instead of mass-appending "Overlay" to every collection name in the tracking.

> My second point is that this PR includes the overlay track ID fix which is still not resolved, as we're actively looking for a solution. While we figure that out, I'd like us to merge only the changes that are unique to this PR and I understood that you had fixed this but a rebasing reintroduced a lot of code changes. Ideas for how to deal with this beyond simply holding off on merging?

We could simply cut down this PR to only modify the `full_tracking_sequence.py` and `config.py` files, and merge that, holding onto the other details for a later PR (perhaps #1963). If your thoughts on this are positive, then I'll cut back on a lot of the commits here. 

I don't know if this reduced PR will actually fix #1700 since this seems to require the SimParticle overlaying, but that could be for a later pull request. The 2e tracking sequence could be this PR.

tvami:
@oscarplewis I think you can keep everything except for the 
```

    for (const auto &[track_id, particle] : sim_particles) {
      if (track_id > track_id_increment) {
        track_id_increment = track_id;
      }
```
part (and co)

This is conceptually different than the rest of this PR. The addition of the passnames and collection names doesnt change behavior so you can keep those. Your fixes in the configs and 2e tracking sequence resolve the original issue of the fact that DQM wasnt reading the correct stage of the collections. Now the fact that the collections still dont make 100% sense and you need to shift the track-IDs is conceptiually different problem and can go into your PR https://github.com/LDMX-Software/ldmx-sw/pull/1963

oscarplewis:
> This is conceptually different than the rest of this PR. The addition of the passnames and collection names doesnt change behavior so you can keep those. Your fixes in the configs and 2e tracking sequence resolve the original issue of the fact that DQM wasnt reading the correct stage of the collections. Now the fact that the collections still dont make 100% sense and you need to shift the track-IDs is conceptiually different problem and can go into your PR #1963

Okay, I've removed the SimParticles overlay from the code. This seems to have detrimental effect on the tracking (see below), but the `config.py` is at least reading from the right collections now and the tracking's framework is set up for future modification.

Here are the RecoilDQM outputs:

RecoilDQM:truth_N_tracks
RecoilDQM:N_tracks

And here are the TaggerDQM outputs:
TaggerDQM:truth_N_tracks
TaggerDQM:N_tracks

I'm not certain why this is the case: this only became this way when I removed the SimParticles overlay, but given that RecoilSimHits and TaggerSimHits collections are overlaying properly I would expect the tracking here to be fine. Apparently, though, the tracking rests heavily on the SimParticles collection in a way that I cannot determine myself from looking through all the processors.

Finally, I've updated the PR title and description with the relevant details so it is now accurate. Below I've copied the plots for when the SimParticles overlaying was in effect so the data isn't lost. These below plots DO NOT reflect the present state of the branch to be merged. Respectively, they are the RecoilDQM/N_tracks and TaggerDQM/N_tracks plots.

RecoilDQM_N_tracks
TaggerDQM_N_tracks

tvami:
> , the tracking rests heavily on the SimParticles collection in

this is indeed unexpected. It should only rely on the SimHits which then are transformed to be digis, and then seeds and then the tracks, the SimPartilces should play no role at all. @bloodyyugo does this make sense to you?

tvami:
The `createTruthTrack()` function uses the SimParticles, so I guess mystery solved

oscarplewis:
I found the culprit: [DigitizationProcessor.cxx#L140](https://github.com/LDMX-Software/ldmx-sw/blob/dd5a35af9c5c86f9f0772c21cffeb3ee73e01afc/Tracking/src/Tracking/Reco/DigitizationProcessor.cxx#L140). The DigitizationProcessor merges hits with the same track ID in the same detector cell, meaning that because there is still no track ID management in this PR, the 2e case cannot perform tracking properly because it ends up merging hits too much, resulting in far fewer digitized tracker measurements than should be present. But the case still doesn't look entirely like a 1D case, and so the tracking gets confused.

Here are plots to demonstrate the truth of this. First is the plot for this PR currently; the second plot is what happens when I put back the SimTrackerHit track ID management removed in commit fc480b7. I haven't added back this change into the PR because it will be fixed in #1963 shortly, but this is why the tracking doesn't work now. It's not because of the SimParticles, it's because of the track IDs and the digitization process of hit-merging.

Screenshot 2026-03-05 at 22 41 39

Screenshot 2026-03-05 at 22 43 53

tvami:
Awesome work @oscarplewis

tvami (approved):
OK let merge this and then the next PR asap after this is merged

tvami:
Given how much this PR has become tracking, I added Matt to be a reviewer too

bloodyyugo:
Yeah, this doesn't make much sense.  You give the digi processor a SimTrackerHit collection name, it makes "digitized" hits and then flows from there.   

> > , the tracking rests heavily on the SimParticles collection in
> 
> this is indeed unexpected. It should only rely on the SimHits which then are transformed to be digis, and then seeds and then the tracks, the SimPartilces should play no role at all. @bloodyyugo does this make sense to you?

tvami:
@bloodyyugo yeah in the end Oscar figured it out, it's not the SimParticles, see https://github.com/LDMX-Software/ldmx-sw/pull/1962#issuecomment-4009927387

bloodyyugo:
Ah ok, good!

tvami:
@bryngemark @bloodyyugo please approve if you are happy with the changes

bryngemark (approved):
excellent work @oscarplewis .

Issue #1105, opened 2022-10-18
Update overlay processor to not deal specifically with contribs for Ecal 


bryngemark:
**Is your feature request related to a problem? Please describe.**
Once Ecal moves away from the contribs solution, the overlay producer should also stop treating Ecal differently than other systems. 

**Describe the solution you'd like**
Take out the special treatment of Ecal in the overlay producer like [here](https://github.com/LDMX-Software/ldmx-sw/blob/9ba2f7e27e68fbf46db34ba130fff2dc7c74e0df/Recon/src/Recon/OverlayProducer.cxx#L113): 
``` C++
    auto needsContribsAdded{collName.find("Ecal") != std::string::npos ? true
```
and [here](https://github.com/LDMX-Software/ldmx-sw/blob/9ba2f7e27e68fbf46db34ba130fff2dc7c74e0df/Recon/src/Recon/OverlayProducer.cxx#L138-L141)

``` C++
        if (needsContribsAdded) {
          // this copies the hit, its ID and its coordinates directly
          hitMap[simHit.getID()] = simHit;
        }
```

**Additional context**
This should be linked to any issue about dropping the contribs in Ecal.

bryngemark:
as it happens, instead hcal is also moving to contribs, at least in the shorter term. so now we need to add hcal to the list of systems using contribs in the overlay processor configuration.

bryngemark:
also, while we're changing stuff in the overlay producer I note this thing that's been on my mind for a while:
various limits on trackID to denote "initial" particles pop up here and there, and I'm not sure it plays nicely with multielectron events. for instance, there are cuts on origin id in the CLUE analyzer which would work with mpg generated multielectron events but not really with overlay. for this reason, trackID handling in overlay needs an overhaul -- when I designed it I figured we'd never care to know which pileup electron we were dealing with and all pileup contributions got a nonsensical trackID; I have since changed my mind in light of for example these clustering performance studies.

PR #1963, opened 2026-02-13
iss1954 Track ID management in OverlayProducer, and SimParticles overlay

### LINKED ISSUE #1954: Fix track ID for Overlay

Track ID is not correct when overlay is used.

Fix it.

oscarplewis:

This resolves #1954 by implemented track ID bitwise encoding in the OverlayProducer to ensure that tracks have unique IDs in the overlay collections. The way this works at present is that in the track ID 'int' value, the first four bits after the sign are reserved for the encoding version (0 for no encoding to match up with the 2-electron gun, and 1 for this version). In version 1 the next three bits encode the event index (0 for main events, 1 for overlay events). 

In the process this PR also introduces SimParticles overlay to properly enable the DQMs for 2e samples, following the same track ID encoding steps.

## Check List

As expected, there are now 2 tracks everywhere in the it_pileup validation now that the duplicate track IDs are not causing errors in the overlay. The number of straight tracks and hits in the detector has also increased, understandably. (All plots below are from the it_pileup validation.)

RecoilDQM_N_tracks:
RecoilDQM_N_tracks.pdf

EcalMipTrackingFeatures_n_straight_tracks:
EcalMipTrackingFeatures_n_straight_tracks.pdf

EcalDigiVerify_num_sim_hits_per_cell:
EcalDigiVerify_num_sim_hits_per_cell.pdf

### DIFF
--- a/Recon/src/Recon/OverlayProducer.cxx
+++ b/Recon/src/Recon/OverlayProducer.cxx
@@ -110,9 +110,8 @@ void OverlayProducer::produce(framework::Event &event) {
   std::map<std::string, std::map<int, ldmx::SimParticle>>
       particle_collection_map;
   std::map<std::string, std::map<int, ldmx::SimCalorimeterHit>> hit_map;
-  int track_id_increment =
-      -1;  // there should only be one SimParticles collection, so it's fine for
-           // this to be just an integer rather than a vector
+  int track_id_increment = 100000;  // track_id's for overlay events will be
+                                    // incremented by this value
 
   // start by copying over all the collections from the sim event
 
@@ -203,7 +202,7 @@ void OverlayProducer::produce(framework::Event &event) {
     }
   }  // over particle collections for sim event
 
-  /* ----------- now do the pileup overlay ----------- */
+  /* ----------- now do the overlay ----------- */
 
   // we could shift these by a random number, effectively placing the
   // sim event at random positions in the interval, preserving the
@@ -309,18 +308,51 @@ void OverlayProducer::produce(framework::Event &event) {
             if (this_coll_hit_map.find(overlay_hit_id) ==
                 this_coll_hit_map
                     .end()) {  // there wasn't already a simhit in this id
+              ldmx_log(trace)
+                  << "No existing simhit found for ID " << overlay_hit_id
+                  << "; copying overlay hit to output collection";
               this_coll_hit_map[overlay_hit_id] = ldmx::SimCalorimeterHit();
               this_coll_hit_map[overlay_hit_id].setID(overlay_hit_id);
               std::vector<float> hit_pos = overlay_hit.getPosition();
-              this_coll_hit_map[overlay_hit_id].setPosition(
-                  hit_pos[0], hit_pos[1], hit_pos[2]);
-            }
-            // add the overlay hit (as a) contrib
-            // incidentID = -1000, trackID = -1000, pdgCode = 0  <-- these are
-            // set in the header for now but could be parameters
-            this_coll_hit_map[overlay_hit_id].addContrib(
-                overlay_incident_id_, overlay_track_id_, overlay_pdg_code_,
-                overlay_hit.getEdep(), overlay_time);
+              this_coll_hit_map[overlay_hit_id].setPosition(hit_pos[0], hit_pos[1],
+                                                  hit_pos[2]);
+              std::vector<float> pre_step_pos =
+                  overlay_hit.getPreStepPosition();
+              this_coll_hit_map[overlay_hit_id].setPreStepPosition(
+                  pre_step_pos[0], pre_step_pos[1], pre_step_pos[2]);
+              std::vector<float> post_step_pos =
+                  overlay_hit.getPostStepPosition();
+              this_coll_hit_map[overlay_hit_id].setPostStepPosition(
+                  post_step_pos[0], post_step_pos[1], post_step_pos[2]);
+              this_coll_hit_map[overlay_hit_id].setEdep(overlay_hit.getEdep());
+              this_coll_hit_map[overlay_hit_id].setPathLength(
+                  overlay_hit.getPathLength());
+              this_coll_hit_map[overlay_hit_id].setPreStepTime(
+                  overlay_hit.getPreStepTime() + time_offset);
+              this_coll_hit_map[overlay_hit_id].setPostStepTime(
+                  overlay_hit.getPostStepTime() + time_offset);
+              this_coll_hit_map[overlay_hit_id].setVelocity(overlay_hit.getVelocity());
+            }  // if overlay_hit_id not present
+
+            // add the overlay hit contribs to existing hit,
+            // incrementing track IDs and timestamp as needed
+            int n_contribs = overlay_hit.getNumberOfContribs();
+            ldmx_log(trace)
+                << "Copying and reindexing " << n_contribs
+                << " contributors to the sim hit for ID " << overlay_hit_id;
+            for (int i = 0; i < n_contribs; i++) {
+              ldmx::SimCalorimeterHit::Contrib contrib{
+                  overlay_hit.getContrib(i)};
+              int incident_id = contrib.incident_id_ + track_id_increment;
+              int track_id = contrib.track_id_ + track_id_increment;
+              float time = contrib.time_ + time_offset;
+              this_coll_hit_map[overlay_hit_id].addContrib(incident_id, track_id,
+                                                 contrib.pdg_code_,
+                                                 contrib.edep_, time);
+            }  // loop over contribs in overlay_hit
+            ldmx_log(trace) << "There are now "
+                            << this_coll_hit_map[overlay_hit_id].getNumberOfContribs()
+                            << " total contributors in the output collection";
           }  // if add overlay as contribs
           else {
             calo_collection_map[out_coll_name].push_back(overlay_hit);
@@ -338,7 +370,7 @@ void OverlayProducer::produce(framework::Event &event) {
 
       /* ----------- now do simtracker hits_ overlay ----------- */
 
-      // get the SimTrackerHit collections that we want to overlay
+      // loop over the SimTrackerHit collections that we want to overlay
       for (const auto &coll : tracker_collections_) {
         auto overlay_tracker_hits{
             overlay_event_.getCollection<ldmx::SimTrackerHit>(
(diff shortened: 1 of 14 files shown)


oscarplewis:
Right now the track IDs are all incremented by 100,000, since I figured that would be surpass any real track IDs from the main sample. Future iterations of this workflow with three or more electrons (if we go in that direction) could go up to 200,000, etc. I also considered iterating through all track IDs present in the main sample to find the largest value and incrementing all the overlay track IDs by this value, but the fixed power of ten approach seemed simpler. Any thoughts would be welcome.

Also, there's a few commits here duplicated in another PR. I'll rebase before submitting both so the history doesn't get fouled up, but this is the reason for the shocking number of edited files here (they're all small edits, changing hardcoded references to scoring plane and sim particles collections to configurable ones).

Recon/src/Recon/OverlayProducer.cxx
bryngemark: here you could add `Hcal` to the list, or even better we should change this machinery such that the list of contrib-using simhit collections is passed as part of the configuration, instead of this name parsing business which is kinda ugly to begin with

Recon/src/Recon/OverlayProducer.cxx
```diff
@@ -121,13 +132,16 @@ void OverlayProducer::produce(framework::Event &event) {
         ldmx_log(trace) << simhit;
         // this copies the hit, its ID and its coordinates directly
         hit_map[simhit.getID()] = simhit;
+        for (int i = 0; i < simhit.getNumberOfContribs(); i++) {
```
bryngemark: is there a good way to only perform this loop if we indeed are going to print to log?

Recon/src/Recon/OverlayProducer.cxx
bryngemark, line 195: same question here, if it's possible to do a check of debug level before the loop. i know this is my old code, not yours 🙂 but at a previous stage this producer had its own verbosity management, and in its present state it's a little annoying

bryngemark: check here should be more versatile now that Hcal also uses contribs

Tracking/python/full_tracking_sequence_2e.py
bryngemark, line 1: Does this file simply add the string `Overlay` to all the collection names? Can we instead use an "overlay string" that defaults to empty but can be set to "Overlay" in the call to the (existing) full tracking sequence?

bryngemark (changes requested):
I think this PR is solving several different issues in one go. It's a lot of good work, but (for the future) I would suggest using separate branches for separate issues if possible. For instance it's hard for me to review the stuff that is not overlay related (the topic of the issue mentioned in the PR), and someone else it might have a different expertise and will hesitate to approve the overlay stuff, etc. 

About the overlay parts: thanks for implementing some good additions. 
- the change I would like to see prior to merging is the way we identify systems that use contribs.
- My main question is how we're going to use the large, offset trackID in further analyses. I suppose it means that we have to look both for trackIDs 100001 and 2, for example, depending on how the pileup sample was produced. Thoughts about this?

Recon/src/Recon/OverlayProducer.cxx
oscarplewis: I was looking to see if there's an issue that fits this already, and I saw that #1105 deals with this same thing. I've since made a branch (1105-config-contrib-treatment) which fixes this by creating a python variable `config_collections`, and the `needs_contribs_added` boolean is now a check whether the collection name is in `config_collections`. Is this the kind of fix you had in mind?

Tracking/python/full_tracking_sequence_2e.py
oscarplewis: You're correct - my first attempt at fixing this is in #1962.

oscarplewis:
> * My main question is how we're going to use the large, offset trackID in further analyses. I suppose it means that we have to look both for trackIDs 100001 and 2, for example, depending on how the pileup sample was produced. Thoughts about this?

Well, that first attempt didn't work. It turns out the track IDs in the Ecal hit contribs go up to something like 1E8, so it wouldn't need to be 100k but 1E9, but at that point it runs into the max value for C++ integers. So I don't think that will work after all.

One other idea I had was prepending an integer to the track ID – so the main sample electron would be 11, the pileup 21, and so on. All the other track IDs would follow the same: 49287 in the main sample would become 149287, 2013 in the pileup would become 22013, etc. Of course, this worsens the problem that you brought up since now we're looking for 2 and 21 but both of these tracks are probably used already in each others' samples.

The only other thing I can think of is just going through and reindexing *all* of the track IDs in the overlay producer to align with the convention of the 2e gun, but I'm not certain if that's feasible. 

Any ideas?

Recon/src/Recon/OverlayProducer.cxx
oscarplewis: Update: I made a PR for this since the fix was small. If you have another idea, I'll just delete it and go back to editing. It's #1968.

bryngemark:
> Well, that first attempt didn't work. It turns out the track IDs in the Ecal hit contribs go up to something like 1E8, so it wouldn't need to be 100k but 1E9, but at that point it runs into the max value for C++ integers. So I don't think that will work after all.
> ...
> Any ideas?

ok thanks for looking into this, that's a good catch! 
would it be unimaginably weird to use negative but otherwise sensible track ids for pileup from overlay? this way -1 would be the first pileup electron, and that would be a simple `trackID_*-1`. we'd also need a logic though to index the second pileup electron (pulled from the next event so also with `trackID = 1`) with -2. 

another idea is to create a new variable "beam electron number" or something that is set in the overlay producer. if it is unset (particle gun pileup) one can fall back to using the trackid at analysis level? (i assume it gets clunky to set up a way to set this variable in the multi-particle gun case.) question: are there any other uses of track ID in pileup than the beam electron index?

EinarElen:
Hi all, 

is it not better to use some bit-level encoding for the track IDs? One could use ~30 bits for track ID stuff

oscarplewis:
> is it not better to use some bit-level encoding for the track IDs? One could use ~30 bits for track ID stuff

Hi @EinarElen, thanks for the input. I'm afraid I don't understand what you mean though, would you mind expanding on this? Do you have in mind something like (in binary) 1000000... for overlay events and 0000000... for main events?

EinarElen:
Hi,
no worries and, to be clear, i havent checked any numbers but I would think that this should be possible to do without running out of track IDs but one would have to check.

For trackIDs, we have 32 bits of information available in an int. All of our valid track IDs are positive and they increase monotonically from 0 so the bits in the integer that corresponds to the sign and the largest powers of two are, effectively, unused. What one can do then is to use N of those bits (possibly including the sign bit but then one would have to be a bit careful) to store some information (similar in spirit to your original 100'000 separation but more information dense).

The downside of this is two-fold, the trackIDs become harder to understand if one just looks at the raw track IDs in a dataset since they now need to be decoded (effectively just masking those bits) but the bigger issue is the same thing as with your original proposal, if we use too many bits for metadata then we reduce the range available for actual track IDs and risk collisions. Since you mentioned track IDs in Ecal hit contribs can reach ~1E8, that's roughly 27 bits worth of range, so we'd only have about 4–5 bits to play with for metadata (sample index, encoding version, etc.) without hitting the sign bit (which we can do, its just a bit more annoying to deal with).  That should still be enough to distinguish a handful of overlay samples.
For example, if you reserve the top 4 bits (below the sign bit) you get 16 possible "overlay tags" and still have 27 bits (~134 million) for the track ID itself. Decoding at analysis level is straightforward, just a bitshift or mask to extract the tag and the original ID separately.

If one does an encoding like this, it is worth thinking about versioning. If there is space in the metadata to include an encoding version then that is usually good. That way, if the scheme ever needs to change (e.g. more overlay samples, or a different bit allocation), downstream analysis code can check the version and decode accordingly rather than silently misinterpreting the IDs. But it has to be weighed against not running out of trackIDs.

oscarplewis:
I went back to look at where I got that number to find out how many bits it was using, but I had mistaken the detector IDs for the track IDs. Looking at the contributing track IDs in the (e.g.) HcalSimHits and EcalSimHits collections in my 1k Ecal PN sample I was using for testing, only at most 14 bits are used to store the data on track information, which leaves us with plenty of room for versioning and an overlay index. These 14 bits correspond to the order of 1E4, which means my original +100 000 idea actually worked and I just didn't realize, but I think @EinarElen's idea is better in the long run. We can probably set aside 4 bits for versioning (doubling as the "encoded" flag for analysis) and another 3 for overlay track indexing without touching the sign bit, and still have plenty of room to spare. I'll write up a commit that does this and submit it here to see if any terrible things happen in the validation checks because of this change :)

oscarplewis:
Track ID encoding version 1 has been written to the branch. This sets aside the first four bits after the sign bit in the track ID integers for encoding versioning, and the next three after that for denoting which sample in the overlay the track ID belongs to. This has all been compiled and tested, but I'm holding off deploying the PR because there will need to be some changes specific to #1962 when it gets merged.

The first image below is a plot of the contributing track IDs in the EcalSimHitsOverlay branch; the two spikes clearly indicating the success of the encoding, the left-most one clustered around the original track IDs and the right-most one clustered around the overlay IDs. The second image shows debug printouts that verify this conclusion, as well as exhibiting that the encoding fails for negative values (which I don't think should be a problem since those are nonsensical and shouldn't be used anyways?).

EcalSimHitsOverlay_overlay_EcalSimHitsOverlay_overlay track_id_contribs_
Screenshot 2026-03-07 at 13 18 35

tvami:
This certainly looks very good
Screenshot 2026-03-16 at 08 19 18

tvami (approved):
Great, thanks a lot @oscarplewis !

bryngemark (approved):
Thanks @oscarplewis, I really like this implementation (and @EinarElen, you're a star, great solution).

I have been sifting through the in-time pileup validation plots. Here is a summary of my findings:
- as @tvami points out ntracks looks better now, with 2 tracks instead of 1 all kinds of places, that's great
- there is an entirely new population showing up in the fakes and duplicates plots
- fakes have vanised from "SeedTaggerDQM" plots
- tagger track phi and p resolution degrade in pileup -- expected but good to flag to tracking people
- same for truth matching probability, that really degrades in some collections (e.g. tagger seed)
-  it looks like in recoil tracking, resolution improves a bit? Perhaps related to having an additional high-momentum track to populate these, now that pileup is included 
- there also seems to be more antiparticles (if i understood the PID plot correctly: the -1 bin has higher counts now)
- the TS "beam electron energy fraction", beam_efrac is broken by this or some other PR since gold. I kind of suspect it's this PR though, and that the cause is using the track id in TrigScintRecHitProducer to identify beam electrons 
- the ECal cluster analyzer plots show no significant differences. I'd have assumed that the origin ID would be affected by the trackID changes -- but maybe instead this points to that the present origin logic is not picking up the pileup electron trackID, which I suppose is reasonable (origin ID maxes out at 6 or so I think).

All of this is not to say that these fixes have to go into this PR, but that we have a few follow-up issues to create:
- beam_efrac
- originID
- any "downstream" producer that uses trackID (e.g recoil electron identification in signal selection workflows? truth electron hit producer?) should be updated 
And that this trackID change should be widely advertised in the next release talk.

@oscarplewis I don't know what your plans are but you'd be welcome to dig into the follow-up issues with your current familiarity with the trackID numbering scheme.

bryngemark:
@EinarElen a follow-up question, ideas for how to use this mask with samples from a multi-particle gun? here the `N` beam electron trackIDs are all the first `N` IDs in the list. should we deal with this at the "analysis" (reconstruction producer) level?

tvami:
Made the issue: https://github.com/LDMX-Software/ldmx-sw/issues/2002

oscarplewis:
Thanks @bryngemark for all your observations and help with this PR. I think the TrigScint beam efrac might be due to [this line](https://github.com/LDMX-Software/ldmx-sw/blob/15c4582cacb634cdaa856a6b09fa30b8fb5b8f59/TrigScint/src/TrigScint/TrigScintDigiProducer.cxx#L82) in the TrigScintDigiProducer, right now the producer uses a hardcoded "SimParticles" collection, so for these validation runs it wouldn't be referencing a collection with the encoded track IDs: because of the logic of the section I linked this means no hits are actually added to the beam efrac variable during calculations. If there's no other problems going on there implicitly, then this should be a straightforward fix for #2002. 

Here's the plot in question, for others' reference (each pad shows a similar trend):
TrigScintTracks_beamEfrac.pdf

As for the origin IDs, I don't exactly know what those are used for, but the current logic in the SimCalorimeterHits object is to set them all to -1 by default unless overrided in the `addContrib` call. There's not even a getter function for the contrib origin IDs. So this might be a reason why the cluster plots look the same? But as I said, I don't know how the origin IDs are intended to function so I couldn't say for certain.

PR #1968, opened 2026-02-19
OverlayProducer contribs collections identification

### LINKED ISSUE #1105: Update overlay processor to not deal specifically with contribs for Ecal

oscarplewis:

Previously, the SimCalorimeterHit collections only used contribs in the Ecal, so there was a hardcoded collection name parsing line in the OverlayProducer which determined if the word "Ecal" was in the collection - if so, then contribs were added separately. Now, however, both the HCal and ECal use contribs and it would be better if these weren't hardcoded into the producer. 

This PR instead creates a variable in the Python configuration called `contrib_collections_`, which contains a list of collection names that use contribs (by default, these are "EcalSimHits" and "HcalSimHits"). Instead of parsing the collection name, now the OverlayProducer checks whether the collection name being overlaid is present in the `contrib_collections_` list. 

This resolves #1105. 

(There's also a random config that was lumped up in the ruff check, so I've included that as well, but I can discard it if need be.)

## Check List

Below are the trace level debug printouts I added when collections are using contribs. We can see that both the Ecal and Hcal sim hits collections use the contribs, as desired.

Screenshot 2026-02-19 at 11 44 13

Screenshot 2026-02-19 at 11 43 58

### DIFF
--- a/Biasing/test/resim_ecal_conv_deep.py
+++ b/Biasing/test/resim_ecal_conv_deep.py
@@ -15,7 +15,7 @@ det = 'ldmx-det-v14-8gev'
 # Please make sure that the simulator here matches the simulator in the original config
 #mysim = ecal.deep_photo_nuclear(det, generators.single_8gev_e_upstream_tagger(),
 #bias_threshold = 5010., processes=['conv','phot)'], ecal_min_z = 300.)
-mysim = ecal.deep_photo_nuclear(det, generators.single_8gev_e_upstream_tagger(), 
+mysim = ecal.deep_photo_nuclear(det, generators.single_8gev_e_upstream_tagger(),
     bias_threshold = 3000., processes=['conv','phot)'], ecal_min_z = 400.)
 mysim.description = "ECal Deep Donversion Test Re-Simulation"
 #mysim.actions.extend([filters.TargetBremFilter()]),
--- a/Recon/include/Recon/OverlayProducer.h
+++ b/Recon/include/Recon/OverlayProducer.h
@@ -2,6 +2,7 @@
 #define RECON_OVERLAYPRODUCER_H
 
 //---< C++ StdLib >---//
+#include <algorithm>
 #include <map>
 #include <string>
 #include <vector>
@@ -96,6 +97,11 @@ class OverlayProducer : public framework::Producer {
    * combining sim and pileup
    */
   std::vector<std::string> tracker_collections_;
+
+  /**
+   * List of SimCalorimeterHit collections which keep track of hit contribs.
+   */
+  std::vector<std::string> contrib_collections_;
 
   /**
    * Pileup overlay events input pass name
--- a/Recon/include/Recon/OverlayProducer.h
+++ b/Recon/include/Recon/OverlayProducer.h
@@ -15,6 +15,7 @@
 #include "Framework/Configure/Parameters.h"
 #include "Framework/EventFile.h"
 #include "Framework/EventProcessor.h"
+#include "Framework/Exception/Exception.h"
 #include "Framework/RandomNumberSeedService.h"
 #include "SimCore/Event/SimCalorimeterHit.h"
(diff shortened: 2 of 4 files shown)


oscarplewis:
The signal failure in the validation checks is leftover from #1966.

Screenshot 2026-02-21 at 20 08 39

I don't know what's going on with the it_pileup fatal error right now but #1962 (which shares no commits with this PR) also demonstrated this error, though it went away after rerunning the job.

Screenshot 2026-02-21 at 20 12 01

tvami:
We'd need to understand how the `EcalAbstractID` is constructed. @oscarplewis can you reproduce this problem locally?

oscarplewis:
TL;DR, a patch has been pushed to the branch and the fatal error should now be fixed. The HCal and ECal SimCalorimeterHit collections were being mashed together accidentally.

I've reproduced the error locally: it turns out that for whatever reason there is an entry in the `EcalDigis` collection that's registering as an HCal hit. The subdetector ID is 6, which corresponds to the HCal in the `DetectorID.h` initialization. This raises a flag when the EcalRecProducer tries to read in the hit and checks that its subdetector ID actually is in the ECal on [line 55](https://github.com/LDMX-Software/ldmx-sw/blob/e5dbf6964d16eded2af3e2319f95f7f0cdc8cefa/Ecal/src/Ecal/EcalRecProducer.cxx#L55) (which has the ID of 5 here). This doesn't come up in the EcalDigiProducer because it only ever uses a different constructor for the ldmx::EcalID class (from the individual layer, module, cell ids, as opposed to a single raw integer).

Screenshot 2026-02-25 at 16 58 06


The question then becomes: "why is there an HCal hit in the ECal collection?" This is because my change from hardcoded string parsing to checking if collections were present in a list of SimCalorimeterHit collections that used the contrib method failed to include this update in the `hit_map` variable to which all contrib-using hits were added. Hence the HcalSimHits collection was never created, and the EcalSimHits collection ended up with both the Ecal and Hcal hits.

I've fixed this by reinitializing `hit_map` to have multiple dimensions such that it includes information for all contrib-using collections and not just one. See the three most recent commits for updates. I've also implemented a "spellcheck" which checks that the contrib-using collections are actually collections in the `calo_collections_` list to avoid spelling problems spelling disaster for the overlay process.


Screenshot 2026-02-25 at 18 05 17

oscarplewis:
Four validation checks failed other than the Python ruff check:

Screenshot 2026-02-27 at 12 11 29

The signal failure is still leftover from PR #1966, and the it_pileup only barely fails by a single plot (as expected, since nothing in the reconstruction or analysis uses contribs). The only difference between logs seems to be in EcalVetoBDT scores.

Neither the target_genie or deep_ecal_gun validation checks had anything in their logs or failed plots that was significantly different from before, mostly just random differences between where points were in the same overall trend/curve.

bryngemark (approved):
Looks good. Great detective work @oscarplewis :)

tvami (approved):
Looks good, thanks!

PR #1970, opened 2026-02-26
Part-4 of ruff formatting

tvami:

Part-4 of ruff formatting

## Check List


ComparePlots/_file.py
```diff
@@ -33,7 +35,7 @@ def __repr__(self) :
         return 'File { Histograms labeled '+self.hist_kwargs['label']+' }'
 
 
-    def from_path(filepath, legendlabel_parameter = None) :
+    def from_path(self, legendlabel_parameter = None) :
```
tvami: I'm not sure why it did this

DetDescr/python/hcal_geometry.py
```diff
@@ -324,7 +323,7 @@ def make_v14(self):
         back_hcal_scint_length = 2000.0
         back_hcal_dx = back_hcal_scint_length
         # back_hcal_dy = back_hcal_scint_length
-        back_hcal_dz = back_hcal_num_layers * back_hcal_layer_thick
+        back_hcal_num_layers * back_hcal_layer_thick
```
tvami: this looks wrong

DetDescr/python/hcal_geometry.py
```diff
@@ -343,7 +342,7 @@ def make_v14(self):
             + side_hcal_num_layers[2]
             + side_hcal_num_layers[3]
         ) * 2
-        side_hcal_layer_thick = (
+        (
```
tvami: same here

SimCore/python/generators.py
```diff
@@ -253,16 +253,26 @@ class genie(simcfg.PrimaryGenerator) :
...
+        if direction is None:
+            direction = [0.0, 0.0, 1.0]
+        if beam_size is None:
+            beam_size = [0.0, 0.0]
+        if position is None:
+            position = [0.0, 0.0, 0.0]
+        if abundances is None:
+            abundances = []
+        if targets is None:
+            targets = []
```
tvami, line 275: I'm not sure why this is better?

DetDescr/python/hcal_geometry.py
tomeichlersmith: It probably removed this because it's not used later but left the calculation in case the calculation has some side-effect that the static analysis cannot understand. I think we should see this and understand that the whole line should be removed.

SimCore/python/generators.py
tomeichlersmith: This is better because then the `list` objects (which are "bigger" in some sense than fundamental types `str`, `bool`, `int`, `float`) are not created until _after_ the function is called (i.e. only when necessary). If the default is written in the function signature like it was previously, then it will be created when the function is defined (i.e. at module import) and Python will be carrying around those objects even if the function is never called.

For our purposes, I don't think these performance-focused lints are very important except maybe in `ComparePlot`. #1303 is already doing some stuff to intentionally avoid the extra `field(default_factory=...)` boilerplate which goes around this "fix".

ComparePlots/_file.py
tomeichlersmith: This is bad-form on our part. We should decorate this function with `@staticmethod` to inform Python (and us) that it is not a function that should be called on a instance of this object.

In Python, member functions are just functions where the first argument is the instance that is being called (by tradition, called `self` so that other Python developers are used to it). The linter is applying this tradition to attempt to avoid confusion. I think instead we should use `@staticmethod` and keep the `filepath` name.

Detectors/util/ecal_layer_stack.py
```diff
@@ -235,11 +237,11 @@ def __init__(self, name, thickness, sensitive = False) :
     def __str__(self) :
         return f'{self.thickness:.2f} mm {self.name}'
 
-    def air(t) :
-        return Layer('Air',t)
+    def air(self) :
```
tomeichlersmith: These `self` changes have the same `staticmethod` comments as before

Framework/python/ldmxcfg.py
```diff
@@ -635,7 +637,7 @@ def __setattr__(self, key, val):
         super().__setattr__(key, val)
 
 
-    def add_library(lib) :
+    def add_library(self) :
```
tvami: @tomeichlersmith this should be static too?

Packing/python/rawio.py
```diff
@@ -39,11 +39,11 @@ def __init__(self, raw_file) :
         super().__init__(f'IO_{raw_file}','packing::RawIO','Packing')
         self.raw_file = raw_file
 
-    def source(file_name) :
-        return RawIO(RawDataFile(file_name, False))
+    def source(self) :
```
tvami: and this too?

tvami:
https://github.com/LDMX-Software/ldmx-sw/actions/runs/22694423855/job/65797509143?pr=1970

"Found 360 errors." 
--> This was at 8000 originally, I think I'll do one more PR beyond this one and hopefully that can have this at zero

LiamBrennan-UCSB (approved):

Packing/python/rawio.py
tomeichlersmith: correct

Framework/python/ldmxcfg.py
tomeichlersmith: yep!

tomeichlersmith (approved):

PR #1979, opened 2026-03-06
Change Track object to put momentum & position in TrackState; change GeV-->MeV

### LINKED ISSUE #1847: Tracks and TrackStates should not use Tracking coordinates

bloodyyugo:
I made many changes to the Track event object: 

- TrackState struct now contains:  momentum and position vectors and the momentum+position covariance at a given track state
- Track no longer has it's own momentum or position but has getters: e.g. getMomentum(TrackStateType) 
- Track has a getter specifically for the target:  getMomentumAtTarget()
- The perigee parameters/covariance/reference at the target are saved in Track for use in code that e.g. makes refined tracks (GSF for example)
- all momenta and positions are saved in the LDMX frame and are in MeV (with exception of the perigee parameter q/p, still in GeV)
- - this includes the perigee reference position so be careful!  Need to rotate to ACTs frame if using there

A lot of files have changed but for the ones outside of Tracking changes are minor.  

This resolves issue #1847  and overrules pull request #1855 .  

## Check List
- [X ] I successfully compiled _ldmx-sw_ with my developments.
- [X ] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).
- [X ] I ran my developments and the following shows that they are successful.


SimCore/include/SimCore/BiasOperators/XsecBiasingOperator.h
```diff
@@ -125,8 +125,7 @@ class XsecBiasingOperator : public G4VBiasingOperator {
    *
    * @param[in,out] header RunHeader to write configuration to
    */
-  // NOLINTNEXTLINE
-  virtual void RecordConfig(ldmx::RunHeader& header) const = 0;
+  virtual void recordConfig(ldmx::RunHeader& header) const = 0;
```
tvami: sorry I know this is annoying, the clang will make this change. But actually it's fixed in `trunk` and now has the `// NOLINTNEXTLINE` above it, https://github.com/LDMX-Software/ldmx-sw/blob/trunk/SimCore/include/SimCore/BiasOperators/XsecBiasingOperator.h#L128

Same for the `biasedXsec`

tvami: I guess what I'm suggesting is to rebase to trunk before the auto-commit https://github.com/LDMX-Software/ldmx-sw/pull/1979/changes/777bec90ce001e5227c890d2e1400b407f145cf4 is applied and then it wont be here anymore

Tracking/include/Tracking/Event/Track.h
```diff
@@ -274,6 +288,12 @@ class Track {
...
   int pdg_id_{0};
 
+  // Track charge
+  int q_{0};
+
+  // Track time
+  double time_{-666.0};
+
   // Track States
   std::vector<TrackState> track_states_;
```
tvami, line 299: You'll need to change the `ClassDef(Track, 5);` to `ClassDef(Track, 6);` in the end of the file

https://github.com/LDMX-Software/ldmx-sw/blob/777bec90ce001e5227c890d2e1400b407f145cf4/Tracking/include/Tracking/Event/Track.h#L301

Tracking/include/Tracking/Event/Track.h
```diff
@@ -274,6 +288,12 @@ class Track {
   // pdgID
   int pdg_id_{0};
 
+  // Track charge
+  int q_{0};
```
tvami: given that we are changing the class def, I think this would be a good opp to move away from a single letter variable and call this
```suggestion
  int charge_{0};
```

Tracking/include/Tracking/Sim/TrackingUtils.h
```diff
@@ -287,7 +273,99 @@ inline bool sourceLinkEquality(const Acts::SourceLink& a,
...
+  r_6.block<3, 3>(0, 0) = r;
+  r_6.block<3, 3>(3, 3) = r;
+  const Eigen::Matrix<double, 6, 6> cov_ldmx = r_6 * cov_acts * r_6.transpose();
+
+  // Step 5: Flatten upper triangle -> 21 elements.
+  // Scale momentum rows/cols from GeV to MeV:
+  //   pos-pos (i<3, j<3): x1       [mm^2]
+  //   pos-mom (i<3, j>=3): x1000   [mm*MeV]
+  //   mom-mom (i>=3, j>=3): x1e6   [MeV^2]
+  const double me_v = Acts::UnitConstants::MeV;
```
tvami: I know this wasnt you, but this should be changed to `mev`

Tracking/src/Tracking/dqm/TrackingRecoDQM.cxx
```diff
@@ -73,9 +73,9 @@ void TrackingRecoDQM::analyze(const framework::Event& event) {
                                                  ecal_sp_passname_));
   }
 
-  if (event.exists("TargetScoringPlaneHits", target_sp_events_passname_)) {
+  if (event.exists("target_scoring_plane_hits", target_sp_events_passname_)) {
```
tvami: https://github.com/LDMX-Software/ldmx-sw/pull/1962/changes#diff-410afeeb5f354726ae3ac34d3e65854409c7e10966466db677f19955171f2328

will fix this

Tracking/src/Tracking/Reco/CKFProcessor.cxx
```diff
@@ -487,132 +491,94 @@ void CKFProcessor::produce(framework::Event& event) {
...
-      ldmx::Track trk = ldmx::Track();
-      ldmx::Track::TrackState ts_at_target;
+      // Build the output Track
+      ldmx::Track trk;
 
-      // Extrapolate to the target
-      bool success = trk_extrap_->trackStateAtSurface(
-          track, target_surface_, ts_at_target, ldmx::TrackStateType::AtTarget);
+      // Extrapolate to the target surface
+      auto opt_target = trk_extrap_->extrapolate(track, target_surface_);
```
tvami, line 498: what's `opt` short for?

tvami (changes requested):
Great work, thanks Matt! I have a few smaller comments, nothing substantial

SimCore/include/SimCore/BiasOperators/XsecBiasingOperator.h
tvami: I guess it already lead to a conflict
Screenshot 2026-03-06 at 14 05 14

Tracking/src/Tracking/Reco/CKFProcessor.cxx
bloodyyugo: These are optional return parameters.

Tracking/src/Tracking/dqm/TrackingRecoDQM.cxx
bloodyyugo: Ah, now I see I was supposed to be looking at that PR too!

tvami: Haha, yeah so I dont know what's the best way to fix it, with the ("target_scoring_plane_hits",) I think this wont actually run. So prob we should merge #1962 and then this auto-rename wont happen anymore

tvami:
```
In file included from input_line_9:6:
/home/runner/work/ldmx-sw/ldmx-sw/Tracking/include/Tracking/Event/Track.h:160:27: error: use of undeclared identifier 'q_'
  void setCharge(int q) { q_ = q; }
                          ^
/home/runner/work/ldmx-sw/ldmx-sw/Tracking/include/Tracking/Event/Track.h:161:37: error: use of undeclared identifier 'q_'
  double getCharge() const { return q_; }
                                    ^
```

Tracking/include/Tracking/Event/Track.h
```diff
@@ -139,25 +157,13 @@ class Track {
...
 
-  void addDedxMeasurement(float path_length) {
-    dedx_measurements_.push_back(path_length);
-  }
-  std::vector<float> getDedxMeasurements() const { return dedx_measurements_; }
-
-  /// d_0 z_0 phi_0 theta q/p t
-  // void setPerigeeParameters(const Acts::BoundVector& par)  {perigee_pars_ =
-  // par; } Acts::BoundVector getPerigeeParameters() {return perigee_pars_;}
+  void setCharge(int q) { q_ = q; }
```
tvami:
```suggestion
  void setCharge(int q) { charge_ = q; }
```

Tracking/include/Tracking/Event/Track.h
```diff
@@ -139,25 +157,13 @@ class Track {
...
+  double getCharge() const { return q_; }
```
tvami:
```suggestion
  double getCharge() const { return charge_; }
```

tvami (changes requested):

tvami:
Pulls at the ECAL are wider:
Screenshot 2026-03-20 at 15 08 05

This might just be the fact that the DQM was resized?
Screenshot 2026-03-20 at 15 08 37

But this should be the same, no?
Screenshot 2026-03-20 at 15 09 26

Tracking/include/Tracking/Reco/TrackExtrapolatorTool.h
```diff
@@ -271,37 +271,14 @@ class TrackExtrapolatorTool {
...
         std::cout << "[TrackExtrapolatorTool]   Surface location: ("
                   << surf_loc(0) << ", " << surf_loc(1) << ", " << surf_loc(2)
                   << ")\n";
       }
 
-      if (debug_)
-        std::cout << "[TrackExtrapolatorTool]   Getting parameters...\n";
-      // Parameters
-      ts.params_ =
-          tracking::sim::utils::convertActsToLdmxPars((*opt_pars).parameters());
```
tvami, line 291: do we not need this? Could this be the reason for the extrapolated locations to be off?

bloodyyugo: This is moved into tracking::sim::utils::makeTrackState, where it uses the acts2LdmxRotation() rotation matrix so it should be fine, but let me make sure the rotation matrix is correct.

bloodyyugo: ...actually, all convertActsToLdmxPars  does is put the acts perigee parameters into a vector, so this can't be it.  We don't save the perigee parameterization for track states anymore (they are saved for Track, atTarget only) so this isn't need here.

tvami: OK, any other idea what shifted the X project at the ECAL?

tvami:
@bloodyyugo did you have the chance to have a look what's causing at the ecal propagation change?

Tracking/include/Tracking/Reco/TrackExtrapolatorTool.h
bloodyyugo: It looks like the issue is in the truth X position, since the track position plots (trk_ecal_loc0) pass the tests.  I'll take a look.  

And yeah, momentum is now in MeV so you need to expand the range on these.

bloodyyugo: I fixed the truth track @ ecal location...I was just setting the truth atECal track state as just the position & momentum of the scoring plane hit but that's not exactly where we extrapolate the track to.  That should be working now.  

Any plot with momentum though will be broke because of the GeV --> MeV conversion.

tvami: Great, I've triggered the tests again

tvami:
Tests fail with 
```
 fire ] 2 fatal: [InvalidArg] : Histogram target_truth_loc0 not found in pool.
```

I'm not sure if this is due to the merge of the trunk or really was just forgotten to be booked

@bloodyyugo if you change this, dont forget to pull since I made a commit in here. Otherwise I can check it tomorrow

tvami:
OK I think I fixed it! Let's wait for the new gold to finish and then we can test again

tvami:
Looks perfect now

Screenshot 2026-03-26 at 08 42 09

tvami (approved):

tomeichlersmith (approved):
Thank you :heart: I'm still behind on understanding all the Tracking stuff, but this looks good and will make the `Track` an easier object to use by us non-Tracking folks.

PR #1989, opened 2026-03-11
Factor out ClusterFinder to Recon

### LINKED ISSUE #976: Refactor TemplatedClusterFinder and move to Tools
ecal::TemplatedClusterFinder is actually a more-general tool than currently used.  Sophie has copied it into Hcal for use there, but it would be better to put a clean separate version into Tools and use it from both Ecal and Hcal

The cluster finder doesn't care about anything specific to EcalHit or HcalHit anyway, so it can be refactored to run off CalorimeterHit.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/976

## Check List

ECAL PN CI runs


tomeichlersmith (approved):
thank you, I like this removal of repetition

sophiemiddleton (approved):

PR #1990, opened 2026-03-11
Move TruthSeeding to use realistic propagator

### LINKED ISSUE #1957: Use the real extrapolator in TruthSeedProcessor

This line
https://github.com/LDMX-Software/ldmx-sw/blob/63b0954ddeaa2d31ab3a83609f24ebe795869e16/Tracking/include/Tracking/Reco/TruthSeedProcessor.h#L262-L264

lit says
```
  // Track Extrapolator Tool :: TODO Use the real extrapolator!
  std::shared_ptr<tracking::reco::TrackExtrapolatorTool<LinPropagator>>
      trk_extrap_;
```

Move to the one that has the material and magnetic field included

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1957

## Check List

ECAL CI runs fine

### DIFF
--- a/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
+++ b/Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
@@ -12,13 +12,39 @@ void TruthSeedProcessor::onNewRun(const ldmx::RunHeader& rh) {
   gctx_ = Acts::GeometryContext();
   normal_ = std::make_shared<std::normal_distribution<float>>(0., 1.);
 
-  Acts::StraightLineStepper stepper;
+  // Custom transformation of the interpolated bfield map
+  auto transform_pos = [](const Acts::Vector3& pos_) {
+    Acts::Vector3 rot_pos;
+    rot_pos(0) = pos_(1);
+    rot_pos(1) = pos_(2);
+    rot_pos(2) = pos_(0) + DIPOLE_OFFSET;
+    return rot_pos;
+  };
+
+  auto transform_b_field = [](const Acts::Vector3& field,
+                              const Acts::Vector3& /*pos_*/) {
+    Acts::Vector3 rot_field;
+    rot_field(0) = field(2);
+    rot_field(1) = field(0);
+    rot_field(2) = field(1);
+    return rot_field;
+  };
+
+  // Setup the interpolated bfield map
+  const auto map = std::make_shared<InterpolatedMagneticField3>(
+      loadDefaultBField(field_map_, transform_pos, transform_b_field));
+
+  // Setup the stepper and navigator
+  const auto stepper = Acts::EigenStepper<>{map};
   Acts::Navigator::Config nav_cfg{geometry().getTG()};
+  nav_cfg.resolveMaterial = true;
+  nav_cfg.resolvePassive = true;
+  nav_cfg.resolveSensitive = true;
   const Acts::Navigator navigator(nav_cfg);
 
-  linpropagator_ = std::make_shared<LinPropagator>(stepper, navigator);
+  propagator_ = std::make_unique<TruthPropagator>(stepper, navigator);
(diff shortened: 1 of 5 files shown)


tvami:
I should prob have the 
https://github.com/LDMX-Software/ldmx-sw/actions/runs/22984384933/job/66731865262?pr=1990
```
  > [ TruthSeedProcessor ] 7397  warn: Propagation to target surface failed — track may have exited the B-field map.
```
moved lower in logging, maybe to `debug`, or no msg at all. @bloodyyugo any preference?

tvami:
@bloodyyugo what do you think about this change?

bloodyyugo:
I changed the TruthSeedProcessor quite a bit in #1979 ...I'm not sure but there may be a clash.  

I think these errors come up for transporting the particle from it's origin, through the tagger, to the target ... that's obviously a long way.  We don't see the for the ecal propagation since we use the sim hits at the ecal scoring plane and then propagate to 240.5mm (the ecal face?) which is much shorter.  I'm not sure why we don't just use the tagger-target scoring plane for the target truth...do you?    

Otherwise, this looks ok and more strictly correct, I think, so good!

tvami:
> I'm not sure but there may be a clash.

Yeah I'll fix the conflicts after we merge https://github.com/LDMX-Software/ldmx-sw/pull/1979

> tagger-target scoring plane for the target truth...do you?

There is no tagger SP so we cant do this

bloodyyugo:
> > tagger-target scoring plane for the target truth...do you?
> 
> There is no tagger SP so we cant do this

No?  Only on the recoil side?

tvami:
> > > tagger-target scoring plane for the target truth...do you?
> > 
> > 
> > There is no tagger SP so we cant do this
> 
> No? Only on the recoil side?

Right, the recoil has it, and then also the Magnet has a bunch (which we also dont use, and usually doesnt even save [and also I think there is a bug in the indexing for the magnet SP haha...]

Tracking/src/Tracking/Reco/TruthSeedProcessor.cxx
```diff
@@ -162,28 +195,21 @@ void TruthSeedProcessor::createTruthTrack(
...
+    return;
+  }
+
   // Create the seed track object.
   Acts::Vector3 ref = target_surface->center(geometryContext());
-  Acts::Vector3 ref_ldmx = tracking::sim::utils::acts2Ldmx(ref);
-  trk.setPerigeeLocation(ref_ldmx(0), ref_ldmx(1), ref_ldmx(2));
 
-  auto prop_bound_vec = (prop_bound_state.value()).parameters();
+  trk.setPerigeeLocation(ref(0), ref(1), ref(2));
```
tvami: @bloodyyugo I should revert this, right? This is something you changed recently, right?

SimCore/G4DarkBreM
tvami, line 1: @tomeichlersmith can you remind me what I need to do to get rid of this change?

tomeichlersmith: I just pushed to resolve. I did
```
# 1. sync the submodule reference with trunk
git checkout trunk -- SimCore/G4DarkBreM
# 2. commit this change
git commit -m "checkout trunk -- SimCore/G4DarkBreM"
# 3. update the local clone of the submodule to be the trunk reference
git submodule update
```

tvami: Thanks Tom!

tvami:
@bloodyyugo I'm ready for your review for this

bloodyyugo (approved):
This looks fine with me

tvami:
While the KS tests flag many of the plots, looking at them they are all compatible, and look good

tomeichlersmith (approved):
looks good and makes sense to me :)

PR #1996, opened 2026-03-12
Python config validation mega-PR

### LINKED ISSUE #1303: Python Config Validation

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1303

## Check List

Ecal CI runs


tvami:
well yes, testing with the ECAL PN CI means ECAL PN CI passes... nothing else does lol
Will fix in the next commit

tvami:
@oscarplewis I just merged your changes from trunk into this. Can you please double-check that I didnt undo any of your dev while doing that?

oscarplewis:
> I just merged your changes from trunk into this. Can you please double-check that I didnt undo any of your dev while doing that?

It doesn't look like it, no. I wasn't able to look through every instance where I changed a hardcoded reference to SimParticles/SimHits/ScoringPlaneHits to a configurable reference, but unless you think there's good reason that went back to being hardcoded in these commits it seems like everything should be fine merging. The major changes (it_pileup/config.py and full_tracking_sequence.py) weren't touched.

tvami:
Will test with new gold when they are ready

tomeichlersmith:
It seems like there are two failure modes from the PR Validation:


#### Missing DQM histograms
e.g. 
image

This could be related to the configuration changes if they are being unintentionally renamed.

#### Actual KS Failure
These end up in the `fail` sub-directory and are all tracking related. My guess is that since this branch has a messy history, it is still running the old tracking? I'm not sure.

tvami:
OK so for the "missing"

They arent, they are just renamed such that in includes the namespace 
Screenshot 2026-03-23 at 10 45 56

vs gold
Screenshot 2026-03-23 at 10 47 01

tomeichlersmith:
this is downstream of the update that has the default `instance_name` just be the `class_name`.

https://github.com/LDMX-Software/ldmx-sw/blob/9effae5e1941440c9bdcbdc49cf2269758d6ab7c/Framework/python/_processor.py#L94-L110

we could either (a) accept this change in naming convention or (b) add some name-cleaning into the `processor` decorator to update the `instance_name` default to something else.

The `instance_name` is really only there for us - the Framework does not require it to function - so we can do whatever we want with it.

tvami:
I was thinking it's fine to have the namespace included, but I'd be also fine to have the cleaning. Probably easier at this point just to have the renamed version.

tvami:
Found the issue with the tracking, it was collection name issue, fixed in https://github.com/LDMX-Software/ldmx-sw/pull/1996/changes/4cf1df06cace170e2855206033b58f6b5d5e5447

tvami:
@tomeichlersmith the KS issue is gone now!

tomeichlersmith:
sweet :sunglasses: I remembered that its pretty easy to trim off the namespace from the class name so I pushed that and am going to relaunch the PR Validations in ~20min after the basic tests pass.

I hope I'm just being overly cautious or maybe I'm just addicted to the green checkmarks.

tomeichlersmith:
I don't know why changing the default `instance_name` appears to be preventing further overwriting of the specific `instance_name`. The errors are coming from the fact that we use two instances of `dqm::HCalDQM` in the CI configs and - since their names are both `HCalDQM`, they try to create the same directory in the output histogram file, causing the run to end (via a seg fault in ROOT's standard way). This was not a problem before since both of those processors have specific `instance_name`s passed to their constructors in the CI configs. I am building locally to see what is up

tomeichlersmith:
The issue was that many DQM processors want to dynamically change the `instance_name` if the user hasn't overwritten the default and since I changed the default, that broke the dynamic change.

I replaced the places where the default is checked for with `self.__dataclass_fields__["instance_name"].default` which is longer but avoids hard-coding the default in two places. I also added a way for folks to change the default instance name in the decorator if they wish.

LiamBrennan-UCSB (approved):

tomeichlersmith (approved):
😎 now the only reason the PR Validation is failing (for the three that are already done) is because of the log comparison which I am much more comfortable with.

PR #1998, opened 2026-03-13
Add tspads+materials into the v15-lyso geom

### LINKED ISSUE #1992: Include TS light pipes + SiPMs in ldmx-lyso-r4-v15-8gev as well
The standard `v15` geometry has now t pipes + SiPMs for the TS, i.e. PR https://github.com/LDMX-Software/ldmx-sw/pull/1888

The same thing should be added in `ldmx-lyso-r4-v15-8gev` as well

I think we have to have these consistent

cjbarton151:

Resolves #1992 and incorporates changes from #1899 

While porting things from standard v15 to the lyso v15, I also noticed a couple minor issues in standard v15 (constants outside of the `constants.gdml` file and a missing element definition, which was being populated automatically by Geant4), and that's why there are changes in that directory as well.

I also tweaked the envelopes a bit; the TSPad3 envelope from v15 was too tall and the lyso target envelope was too slim, so I made the first shorter and the second thicker. I also had to greatly increase the width of the target envelope to account for the light pipes and SiPMs. These changes didn't affect the positions of the objects inside of the envelopes at all.

## Check List

Hard to explicitly verify the materials change without doing some analysis, but everything compiles and there are no _new_ warnings when loading the .gdml files.

Running the upcoming `tspads.mac` through `g4-vis` shows that the geometry changes have taken effect:

image
image

(also, note that the TSPad1+2 envelope is still quite large - hence why it doesn't show up in the image along the beam direction)

### DIFF
--- a/Detectors/data/ldmx-det-v15-8gev/constants.gdml
+++ b/Detectors/data/ldmx-det-v15-8gev/constants.gdml
@@ -10,7 +10,8 @@
 
 <!-- 
     Target
---> 
+-->
+
 <!-- position -->
 <constant name="target_z" value="0.0*mm" />
 <!-- Tungsten X0 = .3504 cm.  Target thickeness = .1X0 -->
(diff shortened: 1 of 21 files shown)


tvami:
Looks good, let's see what the CIs say regarding change of physics

tomeichlersmith (approved):
looks good to me :)

tvami (approved):

PR #2001, opened 2026-03-19
VisAttributes refactor

### LINKED ISSUE #1978: Repairing VisAttributes in v15
**Describe the bug**
When using `g4-vis`, the vast majority of all objects are displayed using the default color, white. The target is red, which indicates that the functionality for including color isn't completely gone, but it seems it's either not being read properly or is being neglected for other objects.

**To Reproduce**
Use `g4-vis` to display v15 geometry

**Desired behavior**
It'd be nice if we could have different colors for each of the submodules, and perhaps even associate certain colors with certain materials or something.

**Screenshots**
This is an outdated picture, but demonstrates my point:
Image

**Additional context**
It'd be nice to decide on/confirm which colors should be used for each of the other submodules/materials.

cjbarton151:
Cleaned up VisAttributes and put them in their own accessory file. Created new VisAttributes for each material (not yet implemented). Added VisAttributes to almost every volume in the geometry.

This will resolve #1978 and add some improvements.

VisAttributes, like materials (but to a lesser extent), were defined across a number of files and somewhat inconsistently. These are now collected into a new accessory file, `visattributes.gdml`, and imported into `detector.gdml` in the same way that the materials are.

Additionally, in many cases, the VisAttributes tag was set for the envelope volume of a module, but unfortunately these VisAttributes aren't inherited by the daughter volumes.* So I've taken the liberty of giving (most) volumes their own <visattributes> tag, and setting it according to their region.

Finally, I've changed the names of the existing 'regional' VisAttributes from for example `HcalVis`->`HcalRegionVis` in anticipation of a future update.

\
\* It would be possible, with a pretty large refactor of the AuxInfoReader, to make things like color and linestyle inheritable by daughter volumes, but in my opinion the juice is not worth the squeeze.
<br/>

## Check List

Before (note the orange coloration of the HCal envelope:
image
After:
image


cjbarton151:
Currently, VisAttributes are defined for each module (HcalRegionVis, TaggerRegionVis etc.), what I'll refer to as 'module-based attributes'. I would like to add functionality for these to be defined by material instead (material-based attributes). Any user is, of course, free to edit the attributes by hand for their own visualization needs, but I think having a toggle option for 'materials mode' would be useful.

The big problem here is that our current GDML implementation (and anything XML-based) is entirely descriptive, not prescriptive, which is to say we can't give arguments directly to the geometry files before building the geometry. `g4-vis` itself (or rather, OpenGL) can take a macro full of commands to alter the attributes of individual volumes, but I can't think of a clean way to 'toggle' multiple volumes, either at the .cxx stage or at the .mac stage.

Another important fact to note is that, if a VisAttribute of a volume is already set, and then it is set again, the second set of attributes will override the first in a pretty normal way, to the best of my knowledge.

MY current idea for how to implement a 'materials mode':

1. Define an Auxiliary tag such as "MaterialsMode" in `detector.gdml` (before loading `visattributes.gdml`) and have it set by default to 'false' or 'true'
2. Add a second set of material-based attributes to every volume, much like I did with the module-based attributes in the first commit. The material-based attributes would come first, so for example the target would have tungsten attributes followed by target region attributes

At this point, the material-based attributes would be set, then immediately overwritten by the module-based attributes. However, we could:

3. Link the Auxiliary tag from step 1 to a bool inside `AuxInfoReader`
4. Add a rule to AuxInfoReader that, if 'materials mode' is true, the AuxInfoReader would ignore any attributes with "RegionVis" in the name (all module-based attributes have this string in their name)
5. Otherwise ('materials mode' false) the AuxInfoReader acts as normal, and the module-based attributes override the material-based attributes before visualization

I'm not in love with this solution because it isn't scaleable and it sometimes overrides values that were already assigned, but - the next-best solution that I see is the one involving giant macro files, which is even more brittle. I'd like to ask for advice from any interested/knowledgeable parties: can you think of a better way to do this? If not, is this a feature worth adding, or is this not of general interest?

tvami (approved):
I think this is awesome @cjbarton151 my only comment is that it should be also done for the v15-lyso version as well too, it could be another PR if you prefer that tho.

tvami:
>  I'd like to ask for advice from any interested/knowledgeable parties: can you think of a better way to do this? If not, is this a feature worth adding, or is this not of general interest?

My take is that any solution is better than the current stage, and that I dont want to take too much of your time on this. So if the options are to improve this vs that you add the mini-HCAL to the ESA VST geometry, I'd pick the second :)

tomeichlersmith (approved):
I second tamas's response - I think this is a step in the right direction and I'm happy they are being used / thought about at all.

tvami:
@cjbarton151 feel free to merge whenever you are happy with the PR

cjbarton151:
I just couldn't bring myself to implement this without making it plausibly extensible, so... Rather than just adding a boolean flag to decide between 'material' and 'region' color modes, I made an optional ColorMode tag that can be interpreted by the AuxInfoReader, and (for now) the only valid options are 'Region' and 'Material'. To keep backwards compatibility, the coloring scheme defaults to 'Region' mode if the tag is unset (this is the one which was originally implemented).

ECal and HCal only with 1/4 cutaway:

image


And a close-up of the target region in material mode, with all volumes (and with the cutaway):


image

tvami:
This is looking really cool @cjbarton151 !

In the ruff formatting PR the `makeTSassemblyfiles.py` got touched, so now it's leading to a conflict. Please fix that

SimCore/src/SimCore/Geo/AuxInfoReader.cxx
```diff
@@ -105,6 +107,13 @@ void AuxInfoReader::assignAuxInfoToVolumes() {
                                              "' was not found!");
         }
       } else if (aux_type == "VisAttributes") {
+	//Same as when registering visattributes, must match mode (is in use)
+	auto match = std::find(universal_visattributes.begin(),universal_visattributes.end(),aux_val);
+	if(match==universal_visattributes.end() && color_mode_!="" && aux_val.find(color_mode_)==std::string::npos)
+	  {
+	    std::cerr << "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" << aux_val <<std::endl;
```
tvami: do we need this? :D 
Probably would be better to have it in the ldmx_log system

SimCore/src/SimCore/Geo/AuxInfoReader.cxx
```diff
@@ -252,6 +261,38 @@ void AuxInfoReader::createRegion(const G4String& name,
...
+  std::vector<std::string> valid_modes = {"Region","Material"};
+
+  auto match = std::find(valid_modes.begin(),valid_modes.end(),mode);
+
+  if(match == valid_modes.end())
+    {
+      std::cerr << "Color mode setting doesn't match any available modes!"
+		<< std::endl << "Currently available modes:" << std::endl;
+      for(auto& valid_mode : valid_modes)
+	std::cerr << valid_mode << std::endl;
```
tvami: Same comment here for the logging system

SimCore/src/SimCore/Geo/AuxInfoReader.cxx
cjbarton151: Lol, I forgot to remove that

SimCore/src/SimCore/Geo/AuxInfoReader.cxx
cjbarton151: I can move it but I'm afraid I'm out of time for tonight. You can cut the minor release without this PR

SimCore/src/SimCore/Geo/AuxInfoReader.cxx
```diff
@@ -105,6 +107,15 @@ void AuxInfoReader::assignAuxInfoToVolumes() {
                                              "' was not found!");
         }
       } else if (aux_type == "VisAttributes") {
+        // Same as when registering visattributes, must match mode (is in use)
+        auto match = std::find(universal_visattributes.begin(),
+                               universal_visattributes.end(), aux_val);
+        if (match == universal_visattributes.end() && color_mode_ != "" &&
+            aux_val.find(color_mode_) == std::string::npos) {
+          std::cerr << "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA" << aux_val
```
tvami: you still have this

tvami:
```
/home/runner/work/ldmx-sw/ldmx-sw/SimCore/src/SimCore/Geo/AuxInfoReader.cxx:111:32: error: 'universal_visattributes' was not declared in this scope; did you mean 'universal_visattributes_'?
  111 |         auto match = std::find(universal_visattributes.begin(),
      |                                ^~~~~~~~~~~~~~~~~~~~~~~
      |                                universal_visattributes_
/home/runner/work/ldmx-sw/ldmx-sw/SimCore/src/SimCore/Geo/AuxInfoReader.cxx: In member function 'void simcore::geo::AuxInfoReader::createVisAttributes(const G4String&, const G4GDMLAuxListType*)':
/home/runner/work/ldmx-sw/ldmx-sw/SimCore/src/SimCore/Geo/AuxInfoReader.cxx:302:3: error: 'universal_visattributes' was not declared in this scope; did you mean 'universal_visattributes_'?
  302 |   universal_visattributes = {"InvisibleNoDau", "InvisibleShowDau", "NoDau",
      |   ^~~~~~~~~~~~~~~~~~~~~~~
      |   universal_visattributes_
```

tvami:
```
 [ fire ] 0 fatal: [MissingInfo] : Referenced VisAttributes 'SpRegionVis' was not found!
    at /home/runner/work/ldmx-sw/ldmx-sw/SimCore/src/SimCore/Geo/AuxInfoReader.cxx:125 in assignAuxInfoToVolumes
```

cjbarton151:
https://github.com/LDMX-Software/ldmx-sw/blob/e32e9ca34277ae1d05ba99465a596fcfd93515c3/SimCore/src/SimCore/ParallelWorld.cxx#L17-L35

When adding the parallel worlds, a new instance of `AuxInfoReader` is created, destroying any information including the string which dictates the current material mode. I'd highly recommend doing this differently. Anyways, since it is currently the way that it is, plus the scoring planes aren't real and usually don't need to be visualized anyways, I'll revert them to their default mode of having just one universal visattribute.

tvami (approved):
great thanks

PR #2005, opened 2026-03-23
iss2004 TS beam_efrac fix in overlay

### LINKED ISSUE #2004: TS beam_efrac overlay fix
The TS "beam electron energy fraction" was broken by #1963, and now doesn't show results at all due to the track ID encoding updates in the OverlayProducer.

The TS beam_efrac variable should function properly.

**Additional context**
Here is the TS beam_efrac plot from the it_pileup validation for #1963.

oscarplewis:

This resolves #2004 by removing the hardcoded reference to the SimParticles collection in the TrigScintDigiProducer and replacing it with a configurable value in the python file. This lets the TS digi reference the correct collection in the case of overlay samples like with the it_pileup validation, and thus fixes the problem from #1963 where the beam_efrac variable was being mishandled following the overlay process.

## Check List

Before there was no value for the beam_efrac variable. The it_pileup validation output this plot:

Now there is a value being output:
Screenshot 2026-03-23 at 13 20 20


tvami (approved):

tomeichlersmith (approved):
Thank you @oscarplewis I love this shit - "move slow and fix things" is my programming mantra

oscarplewis:
Well this is what the it_pileup outputted. (The only plots that failed are the beam_efrac ones, which is great.) The first plot is this PR, the second plot is the one where the error was introduced... it doesn't look a lot like it used to, so I'm going to hold off merging this until I can double check that things are going okay here.

TrigScintTracks_beamEfrac.pdf

tvami:
@oscarplewis I think you changes here are correct nevertheless, even if there is something more to be done to get back the old behavior. I'm expecting really 50% for this variable for the 2e case but maybe I'm misunderstanding what this variable is doing, @bryngemark please correct me

tvami:
(also you'll need to resolve the conflict @oscarplewis )

oscarplewis:
Upon further investigation, it looks like this variable calculates (for each TS module) the fraction of energy deposition which is associated with a SimParticle having PDG ID of 11 and a genStatus of 1. Because of the way the OverlayProducer handles SimParticles at the moment, both the main and pileup electrons have a genStatus of 1, and so they're both getting flagged as "beam electrons." Thus it seems like *all* energy (excepting noise) in the TS is being flagged as associated with a beam electron, and so the beam_efrac is 1. 

I'm not so certain this is a flaw. In the real 2e case, we would have 2 electrons coming from the beam and so we expect the beam_efrac to be 1 if there's no other particles or noise there. So it seems to me like this is a good sign? I'm not certain I have a full grasp on everything going on here though so I will readily accept correction if someone else knows more.

PR #2014, opened 2026-04-04
Ignore products, mix improvements in keep-drop rules

### LINKED ISSUE #1297: patch drop keep rules and add some safe guards
While writing up notes on the drop keep rules for the website, I wrote an example and tried testing it.

```python
p.keep = [ 'drop .*ScoringPlane.*', 'keep EcalScoring.*']
```

This _did not work_. All of the scoring planes were dropped from the output file which I think we can patch. The same affect can still happen with

```python
p.keep = [
  'drop MagnetScoringPlane.*',
  'drop HcalScoringPlane.*',
  # ... list all non-Ecal scoring plane
]
```

but it kinda defeats the purpose of supporting regex at all.

Besides patching this "exception-based" dropping, I also think we could add some safe guards into the drop keep rules. For example,
- Never allow `EventHeader` to be dropped, it is necessary to be re-read by Framework so it should not be removable
- Warn/Error if all collections are being dropped, why even run?

Whomever works on this should also update the [associated doc page](https://github.com/LDMX-Software/ldmx-software.github.io/blob/trunk/src/config/drop-keep.md).

### LINKED ISSUE #2003: Drop collections before processing

We have tools to drop collections in the end, but what if I dont care about anything with passname "test" and just want my new "reco" pass name to survive?

Instead of passing the new pass name to all the processors, I'd like to have a drop_input_collections function that doesnt even read in the branches with a given pass name.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/2003
and
resolves https://github.com/LDMX-Software/ldmx-sw/issues/1297

## Check List

I produced 20 events with the inclusive CI and then run this
```
p = ldmxcfg.Process("test2")
.
.
.
p.histogram_file = "hist2.root"
p.input_files = ["events.root"]
p.output_files = ["events2.root"]
p.keep = [ 'drop .*ScoringPlane.*', 'keep EcalScoring.*']
p.keep.append("ignore test")
```

--> No issues making the DQM plots from the `test2` processes. 
--> EcalScoringPlane is kept and no other ScoringPlane is kept

Docs PR is in https://github.com/LDMX-Software/ldmx-software.github.io/pull/65

### DIFF
--- a/Framework/include/Framework/Event.h
+++ b/Framework/include/Framework/Event.h
@@ -156,6 +156,17 @@ class Event {
   void addDrop(const std::string &exp);
 
   /**
+   * Add an ignore rule to the list of regex expressions to ignore on input.
+   *
+   * If a branch name matches one of the stored regex expressions, it will
+   * not be added to the list of known products when reading the input tree.
+   * This prevents processors from accessing these collections at all.
+   *
+   * @param exp regex to match
+   */
+  void addIgnore(const std::string &exp);
+
+  /**
    * Adds an object to the event bus
    *
    * @throws Exception if there is an underscore in the collection name.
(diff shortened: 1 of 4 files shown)


tomeichlersmith (approved):
Excellent work, thank you!

LiamBrennan-UCSB (approved):

PR #2017, opened 2026-04-09
Updated v15 with LYSO to new visualization standards.

### LINKED ISSUE #1978: Repairing VisAttributes in v15

cjbarton151:

Resolves #1978 again, by importing new visualization features to `ldmx-lyso-r4-v15-8gev`.

Working on this revealed a couple minor discrepancies, which I'll address in follow-up comments. If they're significant enough, I can create a new issue and a new PR.

## Check List

The visualization works much like it does in v15, but the 'target' is green instead of red (arbitrary choice) since it's made of LYSO:

image

### DIFF
--- a/Detectors/data/ldmx-lyso-r4-v15-8gev/detector.gdml
+++ b/Detectors/data/ldmx-lyso-r4-v15-8gev/detector.gdml
@@ -2,13 +2,15 @@
 <!DOCTYPE gdml [
 <!ENTITY constants SYSTEM "constants.gdml">
 <!ENTITY materials SYSTEM "materials.gdml">
-
+<!ENTITY visattributes SYSTEM "visattributes.gdml">
 ]>
-<gdml xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="http://service-spi.web.cern.ch/service-spi/app/releases/GDML/schema/gdml.xsd">
+<gdml xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="gdml.xsd">
 
+
+  
   <define>
 
-    &constants;  
+    &constants;
 
     <!--
       Position of the tagger and recoil tracker envelopes with respect to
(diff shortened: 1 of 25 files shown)


cjbarton151:
When running the `target_pn_lyso` CI test with runnumber=150, there is a series of errors/warnings during event 3:

image

Searchable text, for future users who might be searching closed issues:
`Propagator ERROR Propagation reached the step count limit of 10000 (did 10000 steps)`
`GSF WARNING Performed steps with errors MagneticFieldError:1 | ` ...
`warn: GSF re-fit failed: stepping error occurred in all components`

Is this expected behavior? Admittedly, I haven't looked much into the implementation of the magnetic field in ldmx-sw...

tvami:
> Is this expected behavior? Admittedly, I haven't looked much into the implementation of the magnetic field in ldmx-sw...

Yeah this is known, we'll fix it eventually. I think there is an issue about it, but might be not obviously titled

tvami (approved):

tvami:
At one point we should figure out how not to copy the xsd to each geometry but have a common directory to take it from. Or maybe copy to each of these directories at compiler time from a central directory?

cjbarton151:
> At one point we should figure out how not to copy the xsd to each geometry but have a common directory to take it from. Or maybe copy to each of these directories at compiler time from a central directory?

Yeah... One of those things I keep telling myself I'll do when there are no geometry PRs open, then I forget and open another geometry related PR 😅

tvami:
@cjbarton151 can you make an issue about the XSD? And then we can merge this, unless @tomeichlersmith has a strong opinion

tomeichlersmith:
no strong opinion here, `git` already does some de-duplication of objects that have the same hash so I don't think we are carrying around too many copies of the xsd anyways

tvami:
@tomeichlersmith please sign in that case

tomeichlersmith (approved):
apologies, switching too fast between different things

PR #2020, opened 2026-04-16
run-validation should merge instead of checkout

### LINKED ISSUE #1991: `run-validation` should `merge` instead of checkout

When we trigger tests with "ready for review", the branch in the PR is merged with trunk and is tested like that.
When we trigger tests with `run-validation`, it checks out the branch the PR is from, which could be way way behind trunk, and tests like that.

I'd like the `run-validation` to do the same as "ready for review" approach

Keep as is

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1991

## Check List

### DIFF
--- a/.github/workflows/pr_validation.yml
+++ b/.github/workflows/pr_validation.yml
@@ -40,10 +40,7 @@ jobs:
       id: get_pr
       if: github.event_name == 'issue_comment'
       run: |
-        PR_DATA=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
-          "https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.issue.number }}")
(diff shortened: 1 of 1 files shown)


tvami:
So GENIE not being reproducible doenst make me happy.

But then here are the results
Screenshot 2026-04-16 at 09 02 26

Screenshot 2026-04-16 at 09 02 54

The label is different, but I think it's working? Not sure how else to check given that we are in a clean gold now

tvami:
OK I think I got myself convinced that the plot above shows that this **did not work** but that is because for it to work, we actually need to merge this :D

tvami:
Real testing in https://github.com/LDMX-Software/ldmx-sw/pull/2022

tvami:
It works, see https://github.com/LDMX-Software/ldmx-sw/pull/2022#issuecomment-4261941599

PR #2021, opened 2026-04-16
Timings and other checks in the CI

### LINKED ISSUE #1873: Tweak CI log character count to notice big differences
The CI log comparisons are super helpful to catch changes to our code which do not show up in validation histograms. One example could be a change that makes event processing or simulation run slower -- or faster, for good reasons or after accidentally dropping something (a real example is misaligned beam missing several detector components, although by now our validation histograms would catch that particular type of mistake).

I'd like a procedure that times more processors and uses that info to judge whether something is taking significantly longer or shorter than "usual", and doesn't bother to warn about small changes that are inevitable when a job is run in a different context than gold (different machine, queue, resource availability...)

How to judge what is normal is of course a big question. Can we store an ensemble average from gold.log as they are continuously updated? Until then, maybe a fixed, generous tolerance like 30% could do.

We can also leave it as is and trust that the developer will read the diff closely enough to notice when something is really off, before deciding to ignore the warning. I suspect that the system we have now is crying wolf a little too often for that to work.

**Additional context**
I started this discussion in #1870 which I have since closed since it was partially (the part I cared about at the moment) resolved.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1873

## Check List

### DIFF
--- a/.github/actions/check/check.sh
+++ b/.github/actions/check/check.sh
@@ -38,12 +38,53 @@ __main__() {
     end_group
   fi
 
-  # check character count of logs
+  # check character count of logs, allowing up to 5% difference to avoid
+  # false positives from minor run-to-run output variations
   ngold=$(wc --chars gold.log | cut -f 1 -d ' ')
   nnew=$(wc --chars output.log | cut -f 1 -d ' ')
   if (( ngold != nnew )); then
-    error "Different character counts in logs"
-    rc=1
+    if (( ngold > 0 )); then
+      char_diff_pct=$(( (nnew - ngold) * 100 / ngold ))
+      char_abs_pct=${char_diff_pct#-}
+      if (( char_abs_pct > 5 )); then
+        error "Log character count differs by ${char_abs_pct}%: gold=${ngold}, new=${nnew}"
+        rc=1
+      else
+        warn "Log character count differs by ${char_abs_pct}% (gold=${ngold}, new=${nnew}); within tolerance"
+      fi
+    fi
+  fi
+
+  # compare total wall-clock run time if gold.time is available in the archive
+  # gold.time has one line per sample: "<sample> <seconds>"
+  # timing.txt has one line for this job: "<sample> <seconds>"
+  local _timing_tolerance=10
+  if tar -tf ${_archive} gold.time timing.txt &>/dev/null 2>&1; then
+    tar xzf ${_archive} gold.time timing.txt
+    local _sample new_s gold_s
+    read _sample new_s < timing.txt
+    gold_s=$(awk -v s="${_sample}" '$1 == s {print $2}' gold.time)
+    if [[ -z "${gold_s}" ]]; then
+      warn "No gold timing entry for sample '${_sample}'; skipping timing check."
+    else
+      echo "Timing for ${_sample}: gold=${gold_s}s, new=${new_s}s"
+      if (( gold_s > 0 )); then
+        local timing_diff_pct timing_abs_pct
+        timing_diff_pct=$(( (new_s - gold_s) * 100 / gold_s ))
+        timing_abs_pct=${timing_diff_pct#-}
+        if (( timing_abs_pct > _timing_tolerance )); then
+          if (( new_s > gold_s )); then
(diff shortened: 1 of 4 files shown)


tvami:
I think this is working
```
Warning: Log character count differs by 0% (gold=7359530, new=7402024); within tolerance
Warning: No gold.time found; skipping timing check. Re-generate gold data to enable timing comparisons.
Error: Process completed with exit code 1.
```


and
```
diff gold.log output.log
Warning: No gold.time found; skipping timing check. Re-generate gold data to enable timing comparisons.
```

tomeichlersmith (approved):
I'm a little concerned that we will see this change drastically outside of changes to our software (just due to the host that is runnig our jobs), but I think the code looks good and the idea is sound.

tvami:
Yeah we can adjust the threshold. I picked 10% because for the PU sample I had notes from the past that say
"""
it_pileup: 75m —> reaching 67m doesnt mean anything
"""
because it takes more than 67m to make the inputs and then the overlay itself happens in the last 10 minutes

tvami:
I have to say I really like my dev that sends this validation results msg automatically :P

Also the GENIE, I'll fix in https://github.com/LDMX-Software/ldmx-sw/pull/2028

.github/actions/check/check.sh
```diff
@@ -38,12 +38,53 @@ __main__() {
     end_group
   fi
 
-  # check character count of logs
+  # check character count of logs, allowing up to 5% difference to avoid
```
bryngemark: is 5% a hunch or optimized? in my experience a lot of the time the difference is trailing zeroes or similar precision effects (which largely but not fully cancel) but i suppose addressing that requires a more sophisticated line-by-line comparison.

the concern here is that if we're too generous we will miss real differences that we wanted log comparisons to catch. if there is a lot of printout and only one processor changes, this difference could drown. do we have a feeling for how large the character count is overall?

bryngemark (commented):
cool, thanks for working on this! i think the approach is very reasonable and it will be interesting to see how often we (still) get false timing alarms purely from hardware differences. if the chance is less than, say 10%, rerunning validation should mostly resolve that case. 

perhaps we can implement rerunning once already now, for the case that it's only the timing check that fails? (ok let's walk through this. with 10% probability of fakes and one automatic rerun, we'd then have to investigate, including running manually, once in a hundred jobs. so an interested student could even find the distribution of run times and set the tolerance threshold such that we only have to rerun some acceptable fraction _f_ of validation jobs, at our current PR pace, _f_ should probably be somewhere in the percent-permil range.)

tvami:
> perhaps we can implement rerunning once already now,

I'd defer this to a future dev, but also I'm not even 100% convinced it's needed to be automatized, as running the tests again is quite simple. Anyway, I suggest we gain some experience with this and make a decision about that later

.github/actions/check/check.sh
tvami: It's a hunch. Overall char counts differ CI by CI. I think the kaon is the biggest and that has 8300573 chars. That is 8k/event. So 400 char is off per event then this is triggered. We could make it an order of magnitude smaller, prob better to have false alarms then to miss something

tvami:
```suggestion
  # check character count of logs, allowing up to 0.5% difference to avoid
```

.github/actions/check/check.sh
```diff
@@ -38,12 +38,53 @@ __main__() {
...
+  # false positives from minor run-to-run output variations
   ngold=$(wc --chars gold.log | cut -f 1 -d ' ')
   nnew=$(wc --chars output.log | cut -f 1 -d ' ')
   if (( ngold != nnew )); then
-    error "Different character counts in logs"
-    rc=1
+    if (( ngold > 0 )); then
+      char_diff_pct=$(( (nnew - ngold) * 100 / ngold ))
+      char_abs_pct=${char_diff_pct#-}
+      if (( char_abs_pct > 5 )); then
```
tvami:
```suggestion
      if (( char_abs_pct > 0.5 )); then
```

tvami (commented):
Move to 0.5% for the text threshold

bryngemark (approved):
Looks good, let's try this in production and improve as needed, as you suggest @tvami .

PR #2022, opened 2026-04-16
Update README to point to 4.7.1

tvami:

This is the actual testing of https://github.com/LDMX-Software/ldmx-sw/pull/2020


tvami:
Yep this works, now (on 10 events) the labels are fine
Screenshot 2026-04-16 at 10 01 17

PR #2029, opened 2026-04-19
Eventbuilder for VST data

sophiemiddleton:

## Check List

Screenshot 2026-04-19 at 16 30 17


tvami:
Is this ready for the review Sophie?

It does not compile, with
```
/home/runner/work/ldmx-sw/ldmx-sw/EventBuilder/src/EventBuilder/EventBuilder.cxx: In member function 'virtual void eventbuilder::EventBuilder::produce(framework::Event&)':
/home/runner/work/ldmx-sw/ldmx-sw/EventBuilder/src/EventBuilder/EventBuilder.cxx:227:71: error: 'getSubsystemName' is not a member of 'packing::LDMXRoRHeader'
  227 |                     std::string subsys_name = packing::LDMXRoRHeader::getSubsystemName(
      |                                                                       ^~~~~~~~~~~~~~~~
/home/runner/work/ldmx-sw/ldmx-sw/EventBuilder/src/EventBuilder/EventBuilder.cxx:334:63: error: 'getSubsystemName' is not a member of 'packing::LDMXRoRHeader'
  334 |             std::string subsys_name = packing::LDMXRoRHeader::getSubsystemName(
      |                                                               ^~~~~~~~~~~~~~~~

```

sophiemiddleton:
Let me check, perhaps I didnt push all the code (there was an issue with my branch due to a failed attempt to push a large data file).

sophiemiddleton:
its likely I just didnt add the Packing package from that printout, so perhaps its an easy fix..

EventBuilder/include/EventBuilder/FragmentBuffer.hh
```diff
@@ -0,0 +1,95 @@
+// FragmentBuffer.hh (revised)
```
tvami:
```suggestion
```

EventBuilder/include/EventBuilder/FragmentBuffer.hh
```diff
@@ -0,0 +1,95 @@
...
+#ifndef FRAGMENTBUFFER_H
+#define FRAGMENTBUFFER_H
```
tvami:
```suggestion
#ifndef EVENTBUILDER_FRAGMENTBUFFER_H
#define EVENTBUILDER_FRAGMENTBUFFER_H
```

this should be applied to all the header files

EventBuilder/include/EventBuilder/FragmentBuffer.h
tvami, line 1: The header files should just end with `.h` instead of the `.hh`

EventBuilder/exampleConfigs/run_eventbuilder.py
```diff
@@ -0,0 +1,12 @@
+from LDMX.Framework import ldmxcfg
+p = ldmxcfg.Process('builder')
+p.run = 1
+p.max_events = 10000
+p.verbose_parse=True
+from LDMX.EventBuilder import eventbuilder
+import sys
+p.sequence = [
+    eventbuilder.from_dat_file(sys.argv[1])
```
tvami, line 9: How big is the input you used to test it? Wonder if we can upload it to the ci-data repo

EventBuilder/python/eventbuilder.py
```diff
@@ -0,0 +1,41 @@
...
+    instance_name : str
+        Instance name given to the Producer.
+    """
+
+    def __init__(self, dat_file=None, output_name='PhysicsEventData', verbose_parse=True,
+                 instance_name='EventBuilder'):
+        super().__init__(instance_name, 'eventbuilder::EventBuilder', 'EventBuilder')
+        self.dat_file = dat_file or ''
+        self.output_name = output_name
+        self.verbose_parse = verbose_parse
```
tvami, line 27: This is not the current way to have classes, please have a look at any class now in `trunk`

EventBuilder/run_eventbuilder.py
tvami, line 1: This seems like a copy of the file that's in the example config, so this should be deleted

Framework/src/Framework/EventSummary.cxx
```diff
@@ -0,0 +1,29 @@
...
+
+void EventSummary::print(Option_t*) const {
+  std::cout << "EventSummary {" << " eventNumber: " << event_number_
+            << ", timestamp: " << timestamp_ns_ << " ns"
+            << ", nSystems: " << nsystems_
+            << ", payloadSize: " << payload_size_ << " bytes"
+            << ", errorFlags: 0x" << std::hex << error_flags_ << std::dec;
+  std::cout << " }" << std::endl;
+}
+
```
tvami, line 28: You should also have the `ostream`, for example
https://github.com/LDMX-Software/ldmx-sw/blob/4d6752b58cd0d69bde042052ea725c6ea6de2a93/Hcal/src/Hcal/Event/HcalHit.cxx#L15

EventBuilder/src/EventBuilder/EventBuilder.cxx
```diff
@@ -0,0 +1,478 @@
...
+  }
+  if (ps.exists("dat_file")) {
+    m_input_file = ps.get<std::string>("dat_file");
+  } else {
+    const char* env_input = std::getenv("EVENTBUILDER_INPUT");
+    if (env_input) m_input_file = env_input;
+  }
+  if (ps.exists("output_name")) m_output_name = ps.get<std::string>("output_name");
+  
+  std::cerr << "[EventBuilder] configure(): dat_file='" << m_input_file << "' output_name='" << m_output_name
```
tvami: You should not use `cerr` this should instead be in the ldmx_log() system

EventBuilder/src/EventBuilder/EventBuilder.cxx
```diff
@@ -0,0 +1,478 @@
...
+    if (m_verbose_parse || produce_call_count <= 3) {
+        std::cerr << "[EventBuilder] produce() call #" << produce_call_count << std::endl;
+    }
+
+    // Start timing for this event
+    m_event_start_time = std::chrono::steady_clock::now();
+
+    // Read ONE event per produce() call
+    // Loop through frames in the file until we form a complete event
+    const long long coherence_window_ns = 5000000;  // 5 ms window for collecting fragments from same physics event
```
tvami: Should this be configurable?

EventBuilder/src/EventBuilder/EventBuilder.cxx
```diff
@@ -0,0 +1,478 @@
...
+        current_event_errors = 0;  // Reset for next event
+        
+        assembled_event_fragments.clear();
+        return;
+    }
+    
+    // No more events - print final summary statistics
+    auto final_time = std::chrono::steady_clock::now();
+    auto total_time_s = std::chrono::duration_cast<std::chrono::seconds>(final_time - m_start_time);
+    double final_events_per_sec = (total_time_s.count() > 0) ? (double)m_total_events_built / total_time_s.count() : 0.0;
```
tvami: the implicit conversion `(double)m_total_events_built` is very strongly discouraged, please use casting isntead

tvami (changes requested):
Thanks Sophie for this work! I've made a handful cosmetic / technical comments. It would be good to have the input you used to do some real testing. I'd likely also need to have a second read as there is quite some casting ongoing that I'm not getting yet

tvami:
How do you interpret 
https://private-user-images.githubusercontent.com/4170148/580488937-f87b01af-6d35-487b-8e2c-aa1912b519da.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NzY3MjE3MzEsIm5iZiI6MTc3NjcyMTQzMSwicGF0aCI6Ii80MTcwMTQ4LzU4MDQ4ODkzNy1mODdiMDFhZi02ZDM1LTQ4N2ItOGUyYy1hYTE5MTJiNTE5ZGEucG5nP1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI2MDQyMCUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNjA0MjBUMjE0MzUxWiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9MDQ0OTg5ZGY1NGJlZjRiOTkxZjVlZTk1ZDNmOTZjOGE4ZTE0OTIxYmU3ZjIyNGU4YjlhMDY2OWZkYzg5NzMzMiZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QmcmVzcG9uc2UtY29udGVudC10eXBlPWltYWdlJTJGcG5nIn0.aFudbFNpLdRx7uiLDtol4Aoq8pfn2b_V6r1U4CSArU4

btw? What's all the zeros? And the periodic bumps?

sophiemiddleton:
Thanks, ill look at theae tomorrow. The cose has evolved through many iterations and in three different environments hence the mismatches in ldmx style. I will correct the few points and push

sophiemiddleton:
With regards to the plot. I checked using just a subsystem packer of Toms and the shape was the same. I need to decode for it to make any sense really. Thats the next step....

EventBuilder/exampleConfigs/run_eventbuilder.py
sophiemiddleton: It can be, it was too large to push, i know that but we could use it for validation still

EventBuilder/src/EventBuilder/EventBuilder.cxx
sophiemiddleton: It can be, will add

EventBuilder/src/EventBuilder/EventBuilder.cxx
sophiemiddleton: Sure, this code is a little messy, i was debating putting something in a separate class to tidy up the main code.

EventBuilder/src/EventBuilder/EventBuilder.cxx
sophiemiddleton: Thanks, i was not familar with that. Ill loom into using it

EventBuilder/run_eventbuilder.py
sophiemiddleton: Yes, should be deleted

EventBuilder/exampleConfigs/run_eventbuilder.py
tvami: This is the repo I had in mind: https://github.com/LDMX-Software/ci-data

sophiemiddleton: yes, we can put something there

sophiemiddleton:
I have mostly completed the requests above (minus a few). For your casting doubts, I will try to improve things to remove implicit casting. Hopefully that helps alleviate some of the concerns.

EventBuilder/python/eventbuilder.py
sophiemiddleton: I think this requires a full update of the software stack (?) . Tom mentioned it. I think it might be better to do that once this is merged. Since that was my plan. Unless its more seamless than I'm assuming?

EventBuilder/exampleConfigs/run_eventbuilder.py
sophiemiddleton: what's the best way to add the .dat here. I'm guessing manual upload?

tvami: Yeah, but what's the size? If it's too big we could consider skimming to less events

sophiemiddleton:
@tvami - was there anything else you needed me to fix here?

My next steps are to work with detector teams to decode the events and make plots to see if they make sense (this links to what was discussed on Wednesday)

tvami:
I was hoping somebody else (@tomeichlersmith maybe?) would also chime in with the review before I have a second look. You finished with the casting points, right?

EventBuilder/src/EventBuilder/EventBuilder.cxx
```diff
@@ -0,0 +1,472 @@
...
+        packing::LDMXRoRHeader ror_header;
+        int pos_before_ror = m_reader.tell();
+        
+        // Attempt to read as RoR header
+        try {
+            ror_header.read(m_reader);
+            // Successfully parsed RoR header
+            fragment.header.subsystem_id = static_cast<uint64_t>(ror_header.subsystem());
+            fragment.header.contributor_id = static_cast<uint64_t>(ror_header.contributor());
+            fragment.header.timestamp = ror_header.timestamp();
```
tomeichlersmith, line 149: I'm not really sure where to put this, but I want to emphasize something here. The "time stamp" that was stored in the readout header does **not** record a time in time units (like nanoseconds). It is some combination of PulseID and BunchID delivered to us by the beam timing system. This makes it easier to match (no clock sync, we can check for perfect equality) at the expense of having a more abstract time stamp (more complicated to find out an absolute datetime or even the relative ordering of events if that is of interest).

EventBuilder/python/eventbuilder.py
tomeichlersmith: I'm not sure what you mean by "full update of the software stack". I think you can (and should) update the implementation of this class to be in sync with the other stuff on `trunk` before it is merged. I am willing to help you do this if you are unsure on what to do from looking at the other python config classes on trunk.

tomeichlersmith (changes requested):
I have two larger thoughts and I'm choosing to not delve much deeper. Partially due to laziness, but also because I don't want to get stuck in my own perfectionist mindset. I think the better solution would be merge this (as long as it compiles on `trunk`) and then we can start iteratively developing it as folks use it and we test it more thoroughly.

EventBuilder/src/EventBuilder/EventBuilder.cxx
sophiemiddleton: yes, I agree. The term timestamp is not exact. I'm happy to call it something else

EventBuilder/python/eventbuilder.py
sophiemiddleton: I just meant, if I change my syntax alone then it will probably not be correct. I need to pull the latest code or latest environment or something too (instructions might be useful, so I dont become incompatible)

tomeichlersmith: Ah sure, here is my recipe for updating to be in sync with `trunk`. I don't think there will be issues because I tried it locally and it worked and the CI is compiling fine and the CI uses the newest environment.
```
# 1. download trunk from GitHub to make sure its up to date
git switch trunk
git pull origin trunk
# 2. merge the trunk branch with yours
git switch eventbuilder
git merge trunk
# 3. there was an update to the SimCore/G4DarkBreM submodule so you'll need to also run
git submodule update
```
As for updating to the latest environment, you can `just pull ldmx/dev:latest` from within your clone of ldmx-sw.

Personally, I prefer to use `git rebase` instead of `git merge`, but the difference between those two is lost after we squash-merge the PR into trunk anyways.

EventBuilder/src/EventBuilder/EventBuilder.cxx
tomeichlersmith: No I just wanted to be clear since I noticed that you had a `window_ns` variable floating around (implying that you were treating the timestamps as actual times in nanoseconds or something like that). Having the ability to event-align based off "close enough" rather than exactly equal is still good, just the window may not be in units of nanoseconds since the timestamps aren't in units of time.

EventBuilder/python/eventbuilder.py
sophiemiddleton: OK, let me try this later today - will let you know if I need more information

tvami:
@sophiemiddleton do you need help with this?

sophiemiddleton:
No, I just got busy in the past few weeks. I'll put some time aside to try to get this done in the coming week

EventBuilder/include/EventBuilder/FragmentBuffer.h
```diff
@@ -0,0 +1,94 @@
...
+        for (auto it = it_begin; it != it_end; ++it) {
+            timestamps_in_window.push_back(it->first);
+            for (const auto& frag : it->second) {
+                subsystems_found.insert(frag.header.subsystem_id);
+            }
+        }
+
+        // Assemble if we have at least 3 subsystems (standard LDMX requirement)
+        // This ensures we don't create partial/incomplete events #FIXME - make this configurable or more flexible in the future
+        if (subsystems_found.size() < 3) {
```
mgignac: @sophiemiddleton -- would it be possible to relax / make this configurable? I'm working with this branch now and ran into an issue associated with this strict requirement when the ESA25 test beam only readout two subsystems. I have a fix in place already -- just looking for permission to push those changes to this branch

sophiemiddleton: sure this should definitely be configurable

EventBuilder/python/eventbuilder.py
tvami: @mgignac since you are editing this branch, would you mind doing this transition too?

mgignac: Sure, no worries -- I'm running a bit more validation first, then will circle back to this probably tomorrow, but I'll take care of it.

EventBuilder/exampleConfigs/run_eventbuilder.py
mgignac: @tvami , is this a solved issue from #2138 now? It looks like you uploaded run 111? I guess we can worry about running the event building in the CI in another PR.

EventBuilder/python/eventbuilder.py
mgignac: Alright, i think I managed to merge with trunk & migrate class structure this to be in sync with trunk

EventBuilder/exampleConfigs/run_eventbuilder.py
tvami: Yes indeed @mgignac

tvami (approved):
Thanks a lot both Sophie and Matt!

tomeichlersmith (approved):

PR #2027, opened 2026-04-17
Warning message fix in OverlayProducer track ID encoding

oscarplewis:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?

The SimCalorimeterHits in the ECal currently set all origin IDs to a default value of -1, which the track ID encoding in the OverlayProducer can't handle. This results in a warning message. However, since this is the case for *all* origin IDs, this message shows up far too often to be appropriate. Accordingly, this PR resolves #2026 by changing the log level from `warn` to `trace` and by adding a note that this should be expected for origin IDs when track encoding SimCalorimeterHits.

*Did this commit on web, so I'm waiting on the validation runs to show that compilation succeeds and the development works.*

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [x] I ran my developments and the following shows that they are successful.

Gold.log
Screenshot 2026-04-16 at 21 32 25

Output.log
Screenshot 2026-04-16 at 21 32 46

PR #2032, opened 2026-04-21
Fix origin ID handling in OverlayProducer

### LINKED ISSUE #2002: Fixes to the Overlay producer
From https://github.com/LDMX-Software/ldmx-sw/pull/1963#pullrequestreview-3973989814

- the TS "beam electron energy fraction", beam_efrac is broken by this or some other PR since gold. I kind of suspect it's this PR though, and that the cause is using the track id in TrigScintRecHitProducer to identify beam electrons 
- the ECal cluster analyzer plots show no significant differences. I'd have assumed that the origin ID would be affected by the trackID changes -- but maybe instead this points to that the present origin logic is not picking up the pileup electron trackID, which I suppose is reasonable (origin ID maxes out at 6 or so I think).

All of this is not to say that these fixes have to go into this PR, but that we have a few follow-up issues to create:
- beam_efrac
- originID
- any "downstream" producer that uses trackID (e.g recoil electron identification in signal selection workflows? truth electron hit producer?) should be updated 
And that this trackID change should be widely advertised in the next release talk.

Fix these

cat2code:

This partially resolves #2002 of one of the listed follow-up issues, the one related to origin_id. This fix is also somewhat related to #2027. 

## Check List

Since the merged issue, #1963, track ID is encoded in a neat way using a bitwise encoding, which in turn affects "downstream" producers that expects track ID to be in the old format, or variables that are otherwise just related to the new encoding paradigm. Origin ID stores the information of what electron is the origin for a specific hit in the ECal example. This information is relevant for me to use in a current thesis project where I train a neural network to separate electron showers in the ECal so that was why I was looking at this with @bryngemark. 

### This PR does three things:

1. Adds origin_id to the OverlayProducer::produce() and thus passes it on to also be included the contrib struct for the overlayed event. 
    * Previously origin_id was not passed on in the overlay producer when events were overlayed so in the resulting overlayed event origin_id was reset to the default "nonsensical" value of -1 (since it is a default member in the contrib struct).
2. It suggests that origin_id should not be encoded in the new bitwise encoding format and implements for that by removing it from SimCalorimeterHit::encodeTracks(). The integer value of origin_id are no longer encoded and preserves the integer values that are perhaps easier to understand by a glance. 
    * I think this technically means that the warning/error case in #2027 should not be possible to occur any longer...
3. Removes the variable overlay_event_index from OverlayProducer.cxx.
    * The variable "overlay_event_index" defined in this script was not iterated I think and when I was about to add an iterator operation over the overlay events loop I found it more clean to use the iteration variable "i_ev" instead that is defined in the for loop (it is only relevant in the scope of the for loop anyways). They hold the same information, but they differ by off by one error so this explains why you can see "i_ev+1". This implementation is arguably less readable but the usecase of overlay_event_index is only restricted to the scope of this for loop anyways so I concluded that using a variable that goes out of scope outside of the loop made more sense. 
(4. Extra! I also added a ldmx_log(error) guard if one would try to overlay more than 7 events which is not supported with the bitwise encoding implementation (3 bits are reserved for the overlay_event_index)).

I believe above lays out all the relevant parts in the follow up issue of origin_id from #2027. This is merely one suggestion of a fix and I recognize that one could implement this in a different way it depends on how you want to handle the origin_id member in the end. I made these changes to fix origin_id for my own usecase for my thesis and for that this was satisfactory. 

### Some plots and validation

Here are three 3D plots that I used to validate my changes and investigate the output events.root files.

Plot 1 below shows the value of origin_id of the dominant reconstructed energy contribution for all hits for one specific event in the ECal. Here you see 1 signal event overlayed with 2 pileup events. 

3_events_ecal_3d_origin_id

Plot 2 below is the same as above but we are looking at one pileup event before an overlay (so this has not been touched by the overlay producer it is a simple simulation). 

1_event_ecal_3d_origin_id

Plot 3 below shows a lot of information found in leaves for one specific hit in the ECal. It is a mixed hit where we can see 4 unique track ID’s and I decoded the track ID of the one with the dominant energy contribution. You can see that event_index and origin_id differ by value 1 since they hold basically the same information it is just off by one error between them. 

ecal_3d_decoded_mixed_event

### DIFF
--- a/.github/actions/check/check.sh
+++ b/.github/actions/check/check.sh
@@ -38,12 +38,53 @@ __main__() {
     end_group
   fi
 
-  # check character count of logs
+  # check character count of logs, allowing up to 0.5% difference to avoid
+  # false positives from minor run-to-run output variations
   ngold=$(wc --chars gold.log | cut -f 1 -d ' ')
   nnew=$(wc --chars output.log | cut -f 1 -d ' ')
   if (( ngold != nnew )); then
-    error "Different character counts in logs"
-    rc=1
+    if (( ngold > 0 )); then
+      char_diff_pct=$(( (nnew - ngold) * 100 / ngold ))
+      char_abs_pct=${char_diff_pct#-}
+      if (( char_abs_pct > 0.5 )); then
+        error "Log character count differs by ${char_abs_pct}%: gold=${ngold}, new=${nnew}"
+        rc=1
+      else
+        warn "Log character count differs by ${char_abs_pct}% (gold=${ngold}, new=${nnew}); within tolerance"
+      fi
+    fi
+  fi
+
+  # compare total wall-clock run time if gold.time is available in the archive
+  # gold.time has one line per sample: "<sample> <seconds>"
+  # timing.txt has one line for this job: "<sample> <seconds>"
+  local _timing_tolerance=10
+  if tar -tf ${_archive} gold.time timing.txt &>/dev/null 2>&1; then
+    tar xzf ${_archive} gold.time timing.txt
+    local _sample new_s gold_s
+    read _sample new_s < timing.txt
+    gold_s=$(awk -v s="${_sample}" '$1 == s {print $2}' gold.time)
+    if [[ -z "${gold_s}" ]]; then
+      warn "No gold timing entry for sample '${_sample}'; skipping timing check."
+    else
+      echo "Timing for ${_sample}: gold=${gold_s}s, new=${new_s}s"
+      if (( gold_s > 0 )); then
+        local timing_diff_pct timing_abs_pct
+        timing_diff_pct=$(( (new_s - gold_s) * 100 / gold_s ))
+        timing_abs_pct=${timing_diff_pct#-}
+        if (( timing_abs_pct > _timing_tolerance )); then
+          if (( new_s > gold_s )); then
+            warn "Timing regression for ${_sample}: new=${new_s}s vs gold=${gold_s}s (+${timing_diff_pct}%, tolerance ${_timing_tolerance}%)"
+          else
+            warn "Timing anomaly for ${_sample}: new=${new_s}s vs gold=${gold_s}s (${timing_diff_pct}%, tolerance ${_timing_tolerance}%)"
+          fi
+        else
+          echo "Timing within ${timing_abs_pct}% of gold (tolerance ${_timing_tolerance}%)"
+        fi
+      fi
+    fi
+  else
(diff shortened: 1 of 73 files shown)


oscarplewis:
I think the implementation of the multiple PU overlay as well as the origin IDs is great, and I agree with the logic behind not encoding the origin IDs, but I think it would be useful to add a note into the SimCalorimeterHit and SimTrackerHit header files that explains which track IDs are encoded and which are not.

My only concern is with how the correction to origin ID is made: the default is currently -1, and so when the contribs are overlaid there's a line (edit: it's line 380) that sets the origin ID to:
```cpp
contrib.origin_id_ + i_ev + 1
```
This does correct it, but seeing as `contrib.origin_id_ + 1` is just 0 anyways with the default, and in the future it may be that the origin ID is set before reaching the OverlayProducer, it might be better to just reset this to `i_ev` without including the default value.

Otherwise I think this change is great.

bryngemark (approved):
I think this looks good. @cat2code thanks for the plots showing that originID is always set to 1, 2, 3 etc.

@oscarplewis i understand your thought about `-1+1 = 0` but in fact now with the passing of `originID` to `addContrib`, the `originID` is never set to -1 (while it is if that arg is left empty). The reason to increment the existing `originID` by the overlay event number (`i_ev+1`) rather than set it to `i_ev+2` is that this way, one can allow pulling pileup from e.g. a multiparticle gun file or some other input with two or more electrons in the events already (unlikely that I'd do it personally but it's an idea someone might have) and you would still get the origin ID right. Since we no longer see nonsense `originID`s and thus avoid setting it to 0, I think this is a safer implementation.

bryngemark:
> and in the future it may be that the origin ID is set before reaching the OverlayProducer, it might be better to just reset this to `i_ev` without including the default value.
> 
Clarifying here that `originID` is set at the sim stage already when contribs are made so it was the overlay producer implementation that reset it to `-1`.

tvami:
@cat2code is this PR in draft mode on purpose?

cat2code:
I marked the PR ready for review I am still new to proper github issue practices perhaps I should have set it ready for review from the start @tvami  :) It seems like you guys might be happy with the PR in the current state. 

Thanks for the input, everyone!

tvami (approved):

oscarplewis:
> Clarifying here that `originID` is set at the sim stage already when contribs are made so it was the overlay producer implementation that reset it to `-1`.

Ah wonderful, I didn't understand that, my apologies. I think this is good then! Thanks @bryngemark for the clarification.

oscarplewis (approved):

PR #2033, opened 2026-04-22
Add PN cascade genealogy

### LINKED ISSUE #1313: Add an option to record more details about Photonuclear interactions
As discussed on the August 9 SW meeting, there are several times where it would be useful to know more details about what happened inside the PN interaction. Example use cases would include 
- Noticing edge cases in the cascade, allowing us to understand if a final state is unphysical 
- Knowledge of what part of the model is responsible for a challenging form of final states (e.g. the "tungsten bomb" interactions seem to come from the de-excitation code, not the cascade code). 
- Seeing where differences between e.g. 8 GeV and 4 GeV final states are coming from 
- Work like @LisaALoman's primary vs secondary kaon production would be possible to verify directly  
- Seeing what difference bugs like https://github.com/LDMX-Software/geant4/issues/11 has 

There is already inside of Geant4 a `CascadeHistory` class which we can build on. Only issue is that it isn't really accessible from outside of Bertini and that it only really covers the cascade-part. But since we have the ability to change the PN models in SimCore, we can work around this. 

I'm envisioning something similar to the TrackMap that maps cascade particles to the eventual final state particles with full G4 ids (a bit similar to a Pythia event record).

### LINKED ISSUE #1378: Add some additional information about PN interactions in the event header
We already track and store some information like 
```
  eventHeader.setFloatParameter("total_photonuclear_energy",
                                event_info->getPNEnergy());
  eventHeader.setFloatParameter("total_electronuclear_energy",
                                event_info->getENEnergy());
  eventHeader.setFloatParameter("db_material_z",
                                event_info->getDarkBremMaterialZ());
```

I think the following properties (at least) would be useful to be able to check and are currently not available at all

- Target nucleus (Z,A) for the PN interaction 
PN interactions behave really quite differently for different targets, so being able to understand what effect our materials have on the expected rates for different event types would be useful. Since we are wrapping Bertini, we can always access this. 
- The number of resample attempts that were used (if resampling was used, otherwise just 1)
Necessary for getting a better understanding of how to do EoT-estimates

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1313 and resolves https://github.com/LDMX-Software/ldmx-sw/issues/1378

## Check List

The new histograms are
Screenshot 2026-04-21 at 20 10 13
Screenshot 2026-04-21 at 20 10 21
Screenshot 2026-04-21 at 20 10 47
Screenshot 2026-04-21 at 20 11 21
Screenshot 2026-04-21 at 20 11 59

This actually seems to indicate with a lot of "tungsten bombs"

### DIFF
--- a/SimCore/python/user_actions.py
+++ b/SimCore/python/user_actions.py
@@ -26,11 +26,6 @@ class PhotonuclearTracker(UserAction):
     target nucleus information, immediate secondaries, and descendant genealogy.
 
     The collected interactions are saved to the event as "PhotonuclearInteractions".
-
-    Attributes
-    ----------
-    enabled : bool, optional
-        Enable or disable tracking (default: True)
     """
 
-    enabled: bool = True
+    pass
--- a/SimCore/python/user_actions.py
+++ b/SimCore/python/user_actions.py
@@ -15,3 +15,22 @@ def user_action(class_name: str, module_name: str = "SimCore"):
         post_init=lambda self: _register.library(self.module_name),
         required_base=UserAction,
     )
+
+
+@user_action("simcore::PhotonuclearTracker")
+class PhotonuclearTracker(UserAction):
+    """Configuration for tracking detailed photonuclear interaction information
+
+    This UserAction captures comprehensive details about photonuclear (PN)
+    interactions during simulation, including incident photon kinematics,
+    target nucleus information, immediate secondaries, and descendant genealogy.
+
+    The collected interactions are saved to the event as "PhotonuclearInteractions".
+
+    Attributes
+    ----------
+    enabled : bool, optional
+        Enable or disable tracking (default: True)
+    """
+
+    enabled: bool = True
--- a/.github/validation_samples/ecal_pn/config.py
+++ b/.github/validation_samples/ecal_pn/config.py
@@ -7,12 +7,16 @@ p.max_tries_per_event = 10000
 
 from LDMX.Biasing import ecal
 from LDMX.SimCore import generators as gen
+from LDMX.SimCore.user_actions import PhotonuclearTracker
 
 
 det = "ldmx-det-v15-8gev"
 
 my_sim = ecal.photo_nuclear(det, gen.single_8gev_e_upstream_tagger())
 my_sim.description = "ECal PN Test Simulation"
+
+# Enable detailed photonuclear interaction tracking
+my_sim.actions.append(PhotonuclearTracker())
 
 p.sequence = [my_sim]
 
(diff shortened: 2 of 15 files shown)


tvami:
Something went wrong with the gold for the kaons and target_genie, not related to this PR

SimCore/python/user_actions.py
```diff
@@ -15,3 +15,22 @@ def user_action(class_name: str, module_name: str = "SimCore"):
...
+
+    The collected interactions are saved to the event as "PhotonuclearInteractions".
+
+    Attributes
+    ----------
+    enabled : bool, optional
+        Enable or disable tracking (default: True)
+    """
+
+    enabled: bool = True
```
tomeichlersmith: So, my only comment is about this `enabled` parameter. I feel like this is redundant. As far as I can tell, if I don't add `PhotonuclearTracker` to my `sim.actions`, then the PN interaction won't be tracked. Is this leftover from prior development where the action was always created by SimCore?

On the Python side, you can just add `pass` to the `class` body for it to be empty. (It won't actually be an empty class because the `@user_action` decorator adds stuff.)

On the C++ side, this will just simplify implementation by removing the `if(enabled_)` checks.

tomeichlersmith (commented):
Looks good to me. It is a lot of information, but it is also opt-in which I think makes it less worrisome.

SimCore/python/user_actions.py
tvami: Sounds good, will do!

tvami:
On the physics side: are the plots according to your expectations? I was a little surprised how big the showers are (and that the compactness is peaking at zero)

tomeichlersmith:
This does align with my expectations although I would like to have @EinarElen weigh in as well. The "tungsten bombs" to my understanding was not a bug in the nuclear cascade but a bug in translating the nuclear cascade into "real" particles that "escape" the nucleus and become stuff that propagates through the rest of the Geant4 simulation.

At the end of the day, these cascade-based simulations are trying to simulate the nucleus by respecting how busy it is in there. There are a lot of quarks and a lot of energy that bounces around before specific stuff "escapes" into particles that we can observe with our detector.

tvami:
Screenshot 2026-04-23 at 08 23 47
Screenshot 2026-04-23 at 08 23 52
Screenshot 2026-04-23 at 08 23 59
Screenshot 2026-04-23 at 08 24 05

Higher stats physics validation ^

SimCore/python/user_actions.py
tvami: Done in https://github.com/LDMX-Software/ldmx-sw/pull/2033/commits/81abb9e8002ab891910b1113b5c7c980920e58fe

tomeichlersmith (approved):

bryngemark (approved):
Can't say I'm an expert on this modeling but I approve this on account of adding information such that said modeling can be better studied by someone who is or wants to become one.

PR #1900, opened 2025-12-13
TS+light pipes+Trackers Geometry for ESA (ldmx_reduced_v3)

Layan-Sarayra:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
This partially resolves #iss1880, it's the TS+light pipes+Tracker subsystems geometry for ESA slice test.

## Check List
- [X] I successfully compiled _ldmx-sw_ with my developments.
- [X] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).
- [X] I ran my developments and the following shows that they are successful.

I duplicated `ldmx-reduced-v2` and called it `ldmx-reduced-v3`, and made changes to the following files:
constants.gdml
detector.gdml
materials.gdml
target.gdml
trig_scint.gdml
ldmx-sw/.github/validation_samples/reduced_ldmx/config.py

I built the geometry for ldmx-reduced-v3 based on the CAD layouts Majd Ghrear designed (I annotated in red, but all the designs are Majd's work!)

Key changes between ldmx-reduced-v2 and ldmx-reduced-v3:
1. Positioning: Things are still with respect to target (target=0mm)
      - TS1 : −381.0 mm
      - TS2 : −304.8 mm (TS2 is 76.2 mm downstream of TS1)
      - Tracker 1 (tagger): −240.5 mm (64.3 mm downstream of TS2)
      - Tracker 2 (recoil): −140.5 mm (100 mm downstream of tracker 1)
      - TS3 : −76.2 mm (64.3 mm downstream of tracker 2)
      - Target : 0 mm (76.2 mm downstream of TS3)
      - Increased the mother envelope area in target.gdml to fit target and TS pad 3
      
CAD-layout-updated
image (0)
      
2. Trigger Scintillator (Pads 1 and 2, and same pattern for Pad 3 in target.gdml):
      - Each pad now has: 12 bars per layer × 2 layers = 24 bars
      - Bars are 3 mm (y) × 2 mm (z)
      - Neighboring bars in y: 0.15 mm gap.
      - Layers are shifted by 1.55 mm in y.
      - Z: layer centers are 4 mm apart (2 mm thickness + 2 mm gap).
      - 0.15 mm Y-gap, 1.55 mm Y-shift, 2 mm Z-gap → 4 mm center separation

Screenshot 2025-09-03 at 2 25 57 PM

3. Added light pipes:
      - Material defined/added to materials.gdml 
      - 12 bars per layer x 2 layers = 24 bars
[...]

PR #2038, opened 2026-04-30
ESA slicetest geometry ldmx-reduced-v3: 3 TS modules + LYSO module + light pipes + 2-module tracker

Layan-Sarayra:

## Check List

This PR supersedes [#1900](https://github.com/LDMX-Software/ldmx-sw/pull/1900). The original PR became difficult to update after trunk migrated `ldmx-reduced-v2` to `ldmx-reduced-v3`, leading to large GDML rename/add conflicts with my branch. This PR reapplies my changes cleanly on top of current trunk.

I'm updating `ldmx-reduced-v3` detector geometry for TS+Tracker+Target subsystems:

1. Target becomes LYSO module (with light pipes). Target is still placed at z = 0 and used as the reference point; upstream elements are at negative z, matching the CAD layout and ESA setup.
2. Trigger Scintillator pads 1/2/3 with light pipes.
3. Trackers: removed tagger tracker for this slicetest and placed two recoil tracker modules (2 planes each: axial+stereo) with 100 mm spacing (peaks near z = −240.5 and −140.5 mm, matching the CAD design).

I verified that we can run full sim and reco for these subsystems and get the expected results:

1. TS pads 1/2/3 have SimHits (x/y/z/t/..) and produce Digi/Clusters/Track.
2. LYSO module has SimHits + TargetDigis + TargetClusters with nonzero values.
3. Recoil tracker has SimHits at the expected module centers.

I’ll attach a small subset of plots (for 10k events) to keep the PR light, and link the full set below.
ts_simhits_z
ts_simhits_time
clusters_pe

lyso_edep
trackers_z

Full root file: `/sdf/group/ldmx/users/layan/my-ldmx/ldmx-reduced-v3-dev-test-runs/reduced_v3_new_branch_test_35.root`
Config used: `/sdf/group/ldmx/users/layan/my-ldmx/run_Config.py`
Plots: https://drive.google.com/drive/folders/16of9bijwEBIECApoRQ68jQHbmv8HZ3Xi?usp=sharing 

Notes (not blocking this PR, can be opened as separate issues):

- TS timing cuts: earlier CI failures indicated tracking/cluster timing needed re-tuning for this geometry; the pad-level time selection is geometry-dependent, so I kept this PR focused and we can create a follow-up for making the pad timing parameters auto derived per geometry.
- Light pipes are currently passive; in future work we may define how to handle energy deposited in LPs.


tvami:
The CI fails with
```
  Unrecognized Exception: unordered_map::at
```

Detectors/data/ldmx-reduced-v3/recoil.gdml
```diff
@@ -10,17 +10,22 @@
...
+    <constant name="recoil_envelope_dz_esa" value="220*mm"/>
+
+    <!-- Trim downstream empty recoil envelope to avoid overlaps with target/TS3 -->
+    <constant name="recoil_cut_start_z" value="40*mm"/>
+    <constant name="recoil_cut_box_z"   value="2000*mm"/>
+    <variable name="recoil_cut_center_z" value="recoil_cut_start_z + recoil_cut_box_z/2"/>
+
     <!-- layering start position in envelope volume -->
-    <variable name="recoil_envelope_up_face_z" value="-(recoil_envelope_dz/2)" />
+    <variable name="recoil_envelope_up_face_z" value="-(recoil_envelope_dz_esa/2)" />
```
bryngemark, line 21: Why do we need to specify `_esa` here instead of just changing the variable value?

bryngemark (commented):
Thanks @Layan-Sarayra for pushing this forward, and for adding all this material to show your validation work! Overall it looks sensible, a few details:

- It looks like we're using vacuum in this geometry, can we use air instead? We've seen unphysical effects in Geant4 from having a vaccum/material boundary.
- (as we discussed separately) can you update the CI config for this geometry to handle the pad time used in TS clustering, so we get TS tracks in the validation histograms, please? 
  - As you mention it would be neat to have a geometry dependent fix for this, but until then...

Detectors/data/ldmx-reduced-v3/recoil.gdml
Layan-Sarayra: In constants.gdml, `recoil_envelope_dz` is a derived length describing the nominal recoil envelope based on the layer spacings and clearances:

`recoil_envelope_dz = tracker_envelope_clearance*2 + recoil_delta_l14*3 + recoil_delta_l45 + recoil_delta_l56` 

For the v3 reduced geometry, I only needed to shrink/trim the recoil mother volume to match the CAD design and fix downstream empty volume issues I faced, without changing the global recoil tracker envelope defined in constants.gdml, so I introduced a local override (i.e. `_esa`) inside recoil.gdml and used it only for the envelope box reference. This keeps the slice test specific thing localized and explicit, instead of redefining a global derived quantity and potentially making the constants internally inconsistent (since the deltas would no longer correspond to the envelope length).

tvami:
The changes in the tracking cxx files should be not needed, we dont run the Seeder and GSF on the reduced geometry (we have a dedicated package for that). I think the segfault is gone now bc you added
```
if (not event.exists(track_collection_name_, pass_name_)) return;
```

in the `TrigScintClusterDQM` which just means the TS tracks were not made and this is a workaround. I dont know if we expected the TS tracks to work (I'd assume yes given the TS tracking robustness). @bryngemark are you ok with this change?

Layan-Sarayra:
Thank you for the feedback! I reverted the changes in the three cxx files. For the TS tracking: I verified that TS tracks are being produced. The `event.exists(...)` check is just a guard against a missing collection, not a workaround for TS tracking failing so I’ll keep it as we discussed, in case there is a warning message.

Tracking/include/Tracking/geo/TrackingGeometry.h
tvami, line 1: This needs to change back too, I think, right now there is compilation error
```
/home/runner/work/ldmx-sw/ldmx-sw/Tracking/src/Tracking/Reco/GSFProcessor.cxx: In member function 'virtual void tracking::reco::GSFProcessor::produce(framework::Event&)':
/home/runner/work/ldmx-sw/ldmx-sw/Tracking/src/Tracking/Reco/GSFProcessor.cxx:283:68: error: potential null pointer dereference [-Werror=null-dereference]
  283 |       acts_examples::IndexSourceLink idx_sl(hit_surface->geometryId(), imeas);
      |                                             ~~~~~~~~~~~~~~~~~~~~~~~^~
```

Layan-Sarayra:
I checked this more carefully, and undoing the change in `Tracking/include/Tracking/geo/TrackingGeometry.h` brings back the same error crashing reduced_ldmx validation test the first time:

```
[ GEANT4 ] 1 debug: Stripping off GDML names of materials, solids and volumes ...
[ TrackersTrackingGeometry ] 1  info: x_length = 84.953 y_length = 480 z_length = 240
[ TrackersTrackingGeometry ] 1  info: x_length = 2 y_length = 480 z_length = 240
Unrecognized Exception: unordered_map::at
error: Recipe `fire` failed on line 107 with exit code 127
```
But, removing the changes from the cxx files while keeping this patch is what caused the clang/compile issue in normal Tracking code, because GSFProcessor and the others assume `getSurface()` always returns a valid pointer (what my initial fixes addressed). 

Reverting the header fixes the compile warning, but brings back the reduced validation crash.

I can either revert all of these Tracking fixes and leave the original reduced_ldmx crash unresolved, or bring back the paired .cxx guards so the TrackingGeometry.h change is handled consistently.

Detectors/data/ldmx-reduced-v3/recoil.gdml
bryngemark: mmm ok I see what you mean. Everything in a geometry directory is specific to it though so everything in the reduced geometry is supposed to be for the ESA case. But don't bother changing this, I was just confused if there might be other places the old variable was used or something.

bryngemark:
> The changes in the tracking cxx files should be not needed, we dont run the Seeder and GSF on the reduced geometry (we have a dedicated package for that). I think the segfault is gone now bc you added
> 
> ```
> if (not event.exists(track_collection_name_, pass_name_)) return;
> ```
> 
> in the `TrigScintClusterDQM` which just means the TS tracks were not made and this is a workaround. I dont know if we expected the TS tracks to work (I'd assume yes given the TS tracking robustness). @bryngemark are you ok with this change?

I think it's fine to handle non-existing input collections gracefully and let the CI flag if this means that a collection is suddenly missing (or appearing). One should be very aware though that any failure depends on whatever gold is, and maybe gold has no tracks for some other reason. Simply, we can't (and shouldn't) rely on the CI tests as the only validation of our developments, they are really there to catch unintended side effects. So that the DQM analyzer doesn't seg fault is in my view in itself an improvement.

Layan-Sarayra:
I removed magnet scoring planes since there's no magnet. This is my last change, it's now ready for review.

tvami:
OK I understand what the issue is with the tracker (it's that the recoil is placed at negative z which by the code becomes the tagger tracker for the IDs...), let's just merge this and I can fix it in the next PR.

tvami (approved):

tvami:
>  I can fix it in the next PR.

The fix is here: https://github.com/LDMX-Software/ldmx-sw/pull/2048

tvami:
Nice! With my PR in trunk the reduced geom CI passes too!

cjbarton151:
I checked out the geometry with the visualizer and noticed some discrepancies between the build as intended and the build as implemented. A summary can be found in the attached PDF.

There were 2 main issues which needed to be addressed.

1) in `trig_scint.gdml`, the <loop> for the light pipes was still missing a "from" parameter. Since this loop used the same variable for iterating as the first loop, it was starting at the incorrect value and not iterating properly. I'm still not convinced that the "from" parameter should be optional. This isn't the first time this issue has come up in ldmx-sw, and I'm considering making "from" required for \<loop>s by editing the local XSD file, but I'm not sure if I want to open that can of worms again (the local XSD files, that is).

2) It seems that somehow, the light pipes were at the appropriate y value in both geometries, but the scintillating bars were down-shifted. @Layan-Sarayra has done a lot of work in accounting for all of the constants used to define these bar positions, but I haven't and I don't really want to either! So I just copypasted the correct y value definition from one of the sets of light pipes and applied it everywhere. If someone else wants to go in and make sure all the constant names are accurate and descriptive, they're welcome to, but the main point is that the bars and pipes seem to both be in the correct positions now.

One other tiny fix - you can see in slide 6 of the attached PDF that TSPad 2 was halfway in, halfway out of the trig_scint envelope. I moved the bars and pipes upstream by half a bar width to nudge them fully inside the envelope.

Finally, a couple disclaimers: I made slight edits to the visual properties of these objects for easier viewing in slides 9-17, but didn't include these in the commit because they may not be of general interest and we already have enough other things going on. Also, due to issues with our computing cluster, I wasn't able to do my final final checks before pushing the commit, so hopefully I didn't miss something silly like a comment only for myself again. Lastly, the light pipes along the LYSO target do not line up with the bars and as far as I know are just a copy of the pipes placed for the TS pads. This may not be 1000% accurate but I left it as is for now because again, there's already been a lot of activity in this PR.

[iss1880-19May.pdf](https://github.com/user-attachments/files/28012412/iss1880-19May.pdf)

tvami:
@bryngemark @cjbarton151 I think this PR is ready to be merged, please approve if you agree!

cjbarton151 (approved):
Yeah, it looks buttoned up to me!

cjbarton151:
@Layan-Sarayra when you have a moment, please clean up/delete the other iss-1880 branches you created :^)

PR #2040, opened 2026-05-04
Fixing SingleSubsystemUnpacker bug

### LINKED ISSUE #2036: Reader in SingleSubsystemUnpacker breaks for large files
**Describe the bug**
Using `int` for the output of `reader_.tell()` in the unpacker and in `frame_header.read()` breaks the reader for large file sizes.

**To Reproduce**
Run `SingleSubsystemUnpacker` on `Run_182_20251213_174649.dat`. The file has 292569 total frames, but the unpacker only sees the first ~57k. 

**Desired behavior**
The unpacker should make it through the full file, can be done by replacing `int` with `std::streampos` or `std::streamoff`.

**Screenshots**
From some initial debugging:
[FRAME 57672] size=48 channel=0 flags=0 error=0
[FRAME 57673] size=258036 channel=0 flags=2 error=0 
Stream failed after reading frame_header at frame 57674: eof: 0 tell: -1

**Environment:**
```denv_workspace="/sdf/group/ldmx/users/eberzin/ldmx-src/ldmx-sw"
denv_name="ldmx"
denv_image="ldmx/dev:latest"
denv_mounts="/sdf/group/ldmx/users/eberzin/ "
denv_shell="/bin/bash -i"
denv_network="true"
apptainer version 1.2.5-1.el8
DISPLAY=localhost:19.0
DENV_NAME=ldmx
DENV_RUNNER=apptainer
DENV_IMAGE=ldmx/dev:latest
```

EBerzin:

 This resolves #2036

## Check List

We can now read the full number of TS frames in `Run_182_20251213_174649.dat`. This is `frame_count_` at the end of the file: `Total Frames: 41123`.

### DIFF
--- a/Packing/app/rogue_extract.cxx
+++ b/Packing/app/rogue_extract.cxx
@@ -109,7 +109,8 @@ int main(int argc, char* argv[]) {
     while (r) {
       r >> frame_header;
       frame_count++;
-      const int frame_end = r.tell() + static_cast<std::streamoff>(frame_header.size());
+      const int frame_end =
+          r.tell() + static_cast<std::streamoff>(frame_header.size());
       if (frame_header.channel() != 0) {
         // non-data channel in StreamWriter, skip
         r.seek(frame_end);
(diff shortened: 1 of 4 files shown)


tomeichlersmith (approved):
Looks good, this is what I would do too 👍

tvami (approved):

PR #2041, opened 2026-05-04
iss 2039:  Realistic digitization for trackers

### LINKED ISSUE #2039: Realistic digitization for trackers
Currently, in the trackers,  we simply smear the SimHit positions to make our hits for tracking.  The goal for this issue is to implement a charge-collection and digitization simulation to the readout chain.  This will require a number of additions/changes to the Tracking package: 

-  new containers:  RawSiStripHit (adc values of pulse readout, strip and layer id), FittedSiStripHit (amplitude, t0, strip and layer id)
- digi simulation:  energy --> charge in Si --> charge collection by strips --> strip digitization
- strip clustering into measurement objects used in tracking (clusters saved as Measurement)

bloodyyugo:
I am updating _ldmx-sw_ to include realistic digitization into the tracker simulation.  This resolves #2039 . 

The simulation part takes the charge deposition for each SimHit, converts to electrons and simulates the spread and collection of the electrons on each strip of an LDMX Si sensor.  The charge collected on each strip is shaped (in time) into a pulse (ala the APV25) which is then sampled by an "adc" to give a (currently and settable) 3-sample ADC pulse.  The ADC-vs-sample#  pulse on a strip is stored as an Event/RawSiStripHit. 

Also added code to convert the Event/RawSiStripHit to an Event/FittedStripHit to extract the pulse amplitude and t0.  Fitted strips are combined into clusters (Event/Measurement) that are then used downstream in seeding and tracking reconstruction.  These steps should be identical to what's done in real data (though not the test-beam data, that had a funny trigger/readout structure AFAIK).  

I also added a bunch of plots to the digi-dqm for debugging.  

This probably needs some tuning (at least) but the structure is there.  The truth-smear hit making will still stay in the code base; with this branch there is a switch in the full_tracking_sequence.py to choose between truth-smear/full digi (default set to full digi).  

Here are some plots for the full-digi pertaining to the strips and clusters for both trackers:
[comp-lorentz-dqm.pdf](https://github.com/user-attachments/files/27377284/comp-lorentz-dqm.pdf)
(the two histograms are with/without adding a lorentz-effect to the charge diffusion through the sensor, it's a tiny effect).  

..and here are some tracking plots comparing truth-smear and full-digi on a track-by-track basis: 
[smear-digi-comparison-100jobs.pdf](https://github.com/user-attachments/files/27377372/smear-digi-comparison-100jobs.pdf)


tomeichlersmith (commented):
For some reason I can't post in-line comments right now[^1] 

[^1]: Since GitHub is being continuously enshittified since Micros**t bought it. https://mrshu.github.io/github-statuses/ Mark my words, our free CI will be tightened slowly but surely over the coming years. We might want to get ahead of it and move to something less enshittified (e.g. Codeberg) and self-host the machines that run our CI or actually pay for a CI service.

### Housekeeping
- Since you deleted some files and our CI formatter isn't perfect, please manually apply the formatting: `just format-cpp-diff` will error out (due to missing files) but still apply the formatting. Commit and push the results.
- The basic tests are failing because it compiles with stricter rules enabled (basically errors out if any warnings are emitted). You can use `just configure-force-error` to mimic this behavior or just go through a clean build and absolve warnings.

### By File
SiStripConstants.h
The other digi emulations have these constants passed in through the Python configuration interface. This makes it a lot easier to tune them later on since you can change them without having to re-compile. That being said, I also think it makes sense to have these sensitive constants be _hard_ to change to avoid accidental breaking later on. Do you see these changing a lot? If so, they should probably be in the Python configuration. If not, then leaving them here is fine.

SiStripDigitizer.h
Not sure about `SensorParams` - is this necessary? (do you use `digitizer.params()` a lot?)

SiStripDigitizer.h:213
I don't think you need this to be a pointer. You can just use the default constructor and then use `generator_.seed` when you have access to a seed in `onNewRun`

FittedSiStripHit.h, Measurement.h, RawStripHit.h:159, Track.h
Jump in ClassDef version (presumably from while you were developing), drop this back down to just one increment above whats on trunk.

BFieldXYZUtils.h
Use Framework's Exception (unless good reason not to?), also other places like PulseShape.cxx

full_tracking_sequence.py
Convert these to functions instead of global variables, then we can call it like `full_tracking_sequence(use_truth_smearing = False)` (or `True`) without having to change the installed module.

In processors
Use `const auto&` rather than `const auto`. The `&` prevents copying the data from the `event` bus on each event.
(THis is so common even on `trunk`, I feel like I want to figure out how to write a custom `clang-tidy` rule or something.)

tvami:
Screenshot 2026-05-12 at 12 34 44

The real digitizer has a better track quality? I'm a little surprised by that.

tvami:
Screenshot 2026-05-12 at 12 36 25

This is quite a significant low momentum peak now with the digitizer, is this something you expected? has the fake rate increase too with this?

tvami:
given the comment of 

> (the two histograms are with/without adding a lorentz-effect to the charge diffusion through the sensor, it's a tiny effect).

I'm not actually sure where the bigger changes in the low-p part are coming from

bloodyyugo:
> The real digitizer has a better track quality? I'm a little surprised by that.

Yeah, it's we need to look at the unbiased pulls but it might be that the cluster positions uncertainties are too large.  I'm using the same hardcoded per-nStrips values we use in HPS for now, but we can always change them.  

 switch (cl.n_strips) {                                                                                              
    case 1: sigma_u = SENSE_PITCH_MM / sqrt(12.0);  // 8.7 µm  — binary resolution                                    
    case 2: sigma_u = SENSE_PITCH_MM / 5.0;         // 6.0 µm  — best charge-sharing                                  
    case 3: sigma_u = SENSE_PITCH_MM / 3.0;         // 10.0 µm                                                        
    case 4: sigma_u = SENSE_PITCH_MM / 2.0;         // 15.0 µm                                                        
    default: sigma_u = SENSE_PITCH_MM;              // 30.0 µm — full pitch                                           
  }   


For truth smearing the amount smeared was also used as the hit uncertainty.

bloodyyugo:
This needs to be studied....for now I'm just trying to get the framework set up.  

> This is quite a significant low momentum peak now with the digitizer, is this something you expected? has the fake rate increase too with this?

bloodyyugo:
I think I've addressed all of Tom's comments (thanks).  I've fixed a few other things: 

- -this discrepancy between the number of strips in a cluster between axial and stereo layers was due to ldmx-ACTS coordinate system mix up...it's fixed now.  
- -I wasn't handling charge deposited on a strip from two different SimParticles correctly...it was just keeping the first and tossing the rest.  Now we use a strip-charge map, go through all the SimParticles and sum the charge collected correctly.

bloodyyugo:
I'm not sure why SimCore_config is failing here.  Running this test locally it passes.

tomeichlersmith (approved):
From a technical point of view, I think this looks okay. With how big this PR is, I fear there are latent bugs, but I think we have to start using it to find them.

tvami:
```
    File "config.py", line 47, in <module>
      from LDMX.Tracking import full_tracking_sequence
    File "/home/runner/work/ldmx-sw/ldmx-sw/install/python/LDMX/Tracking/full_tracking_sequence.py", line 396
      sequence = digi_sequence + [
                                 ^
  SyntaxError: '[' was never closed
  Configuration Error [Python] : Execution of python script failed.
```

bloodyyugo:
Fixed the bracket in full_tracking_sequence.py plus made a couple of changes to exampleConfigs so that they run now (these were unrelated to my changes, but left over from when sim.beamSpotSmear was moved to gen).

bloodyyugo:
So I see most of these validation jobs are failing due to the change in the API (trying to do full_tracking_sequence.sequence).  We could either have them return the a default call to full_tracking_sequence() (using __getattr__) or fix these configs to use the new API.   

Also, current default is to do digitization...should we change so that's it's the old truth smearing?

tvami:
> We could either have them return the a default call to full_tracking_sequence() (using getattr) or fix these configs to use the new API.

No strong opinion on this.

> Also, current default is to do digitization...should we change so that's it's the old truth smearing?

Here I'd prefer to have the default to stay the truth smearing until the digitization details are smoothened out

tvami (approved):
Thanks Matt!

PR #2044, opened 2026-05-14
Move local XSD files to a central location, delete redundant copies

cjbarton151:

No associated issue, just a minor housekeeping PR. Having local XSD files for the v15 geometries was first implemented back in #1888, but this was a classic case of having one PR doing too many things. Since the focus was on the geometry updates, I implemented the local XSD files very lazily and never went back to clean it up... UNTIL NOW! 📢📢📢

Currently, each geometry with local XSD files requires them to be in the same directory as the GDML files. The GDML files vary between geometries - but the XSD files do not. This resulted in identical XSD files being copy-pasted into each new geometry. This is now fixed with a central 'xsd_files' directory at the same level as the geometry directories, and the pathing is handled in the same way as for the GDML files.

## Check List

I didn't take any screenshots, but I ran the visualizer with each of the geometries that rely on local XSD files, and also 2 geometries that do not rely on local XSD files. They all executed with no problems. Fun fact, though: the visualizer can load even without any valid XSD files! But it outputs at least 30,000 lines of validation errors when it does so.


tvami (approved):
Great, thanks a lot @cjbarton151 !

tomeichlersmith (approved):
good ole `sed` running from within CMake 😎 thank you

PR #2047, opened 2026-05-14
Changes needed for gcc15 and boost updates (available in Ubuntu 26)

tvami:

When moving to the new Ubuntu in https://github.com/LDMX-Software/dev-build-context/compare/ubuntu_26.04?expand=1
we have stricter requirements in CMake which then doesnt allow to load unused stuff. This PR deletes the `system` component

## Check List

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,8 +63,7 @@ find_package(GENIE)
 
 find_package(
   Boost REQUIRED
-  COMPONENTS system
-             log
+  COMPONENTS log
              filesystem
              thread
              chrono
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -61,6 +61,7 @@ find_package(Geant4Interface)
 
 find_package(GENIE)
 
+cmake_policy(SET CMP0167 NEW)
 find_package(
   Boost REQUIRED
   COMPONENTS log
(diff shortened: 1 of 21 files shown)


LiamBrennan-UCSB (approved):

tomeichlersmith (approved):

tomeichlersmith:
A few more patches were needed to support compiling with GCC15

- scattered potentiall-null-dereference warnings that are resolved with some combination of size-checking and using `.at()` instead of `[]` which does the bounds checking for us
- copy cmake policy and c++ standard into G4DarkBreM
- label Acts and Eigen3 includes as SYSTEM (like ROOT) since they are producing a lot of warnings
  - a better solution would be to update the versions of those dependencies, Acts is on 46.7 right now and Eigen3 release version 5 (the successor to v3.4 since they removed the major number since it wasn't ever changing).

tomeichlersmith:
I'm stuck on silencing the `maybe-uninitialized` warnings that are produced within Eigen when Acts uses one of its templates. :/

tomeichlersmith:
The last complexity is with the clang tools. The updated Clang has updated tools which make slightly different format/tidy decisions than were made before. This can produce some conflicts and will probably require a full-ldmx-sw clang-format and/or clang-tidy at some point in the future. The current CI that attempts to run the clang tools on all files successfully runs but does not error out when warnings or errors are generated (even with the older Clang): https://github.com/LDMX-Software/ldmx-sw/actions/runs/27081773302/job/79928918210

I think updating that CI workflow and doing a format/tidy pass over all of ldmx-sw is its own, separate issue relative to getting the basic tests working. For this reason, I'm going to rebase this branch and remove the auto-applied clang format commit that is generated when I push.

PR #2049, opened 2026-05-18
Bertini history recording

tvami:

First step for GEANT update: brining Bertini out of GEANT and into ldmx-sw.
This is the work of @EinarElen I'm just finishing it up.

## Check List

### DIFF
--- a/SimCore/include/SimCore/Event/CascadeHistory.h
+++ b/SimCore/include/SimCore/Event/CascadeHistory.h
@@ -1,0 +1,86 @@
+/**
+ * @file CascadeHistory.h
+ * @brief Container for the complete history of a Bertini intranuclear cascade
+ */
+
+#ifndef SIMCORE_EVENT_CASCADEHISTORY_H
+#define SIMCORE_EVENT_CASCADEHISTORY_H
+
+#include <vector>
+
+#include "SimCore/Event/CascadeStep.h"
+#include "TObject.h"
+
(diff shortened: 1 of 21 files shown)


SimCore/include/SimCore/Event/CascadeHistory.h
```diff
@@ -0,0 +1,86 @@
...
+  int incident_track_id_{-1};
+  int target_a_{0};
+  int target_z_{0};
+  double incident_energy_{0.0};    // [MeV]
+  double excitation_energy_{0.0};  // [MeV]
+  int residual_a_{0};
+  int residual_z_{0};
+  std::vector<CascadeStep> steps_;
+
+  ClassDef(CascadeHistory, 3);
```
tomeichlersmith:
```suggestion
  ClassDef(CascadeHistory, 1);
```

tomeichlersmith (approved):
Looks pretty good, I am going to bother Einar on slack to see if this means we can swap out underlying Geant4 or if there is more development left to do.

cjbarton151 (approved):
Honestly, it looks fine. Good documentation, nothing obvious missing from the headers or the python integration. I'm not convinced that the uses of` std::move` are increasing efficiency much in a couple of cases, but that's not really a problem. I think that for the upgrade, we also need to be cognizant of changes to the XS databases, but that's a separate issue.

PR #2051, opened 2026-05-19
Increase navigation layer clearance

### LINKED ISSUE #1328: Restore larger navigation layers in the recoil tracker
While checking the geometry of the recoil tracker I noticed that the size of the navigation layers is the same size of the sensitive layers. This might lead to loss of low momenta tracks that curve a lot. 
Increase the size of the navigation layers envelope.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1328

## Check List


bloodyyugo (approved):
This should be harmless and might increase low-p acceptance

mgignac (approved):

PR #2052, opened 2026-05-20
Make amplitude-weighting of cluster centroids variable

### LINKED ISSUE #1969: Move away from amplitude-weighted TS cluster centroids
Currently, TS clustering weights cluster centroids by hit amplitude. This is common in detectors with charge sharing, but probably does not apply to the TS. To study this, it would be nice to be able to select whether centroid values should be amplitude weighted or not. 

I would like the addition of a boolean variable to select amplitude weighting or not. 

Immediately removing amplitude weighting completely, without studying it. 

**Additional context**
Having more well-defined centroid values will simplify creating and using a LUT for TS tracking.

kvarnla:

This resolves issue #1969.

## Check List

In order to implement LUT-method tracking in the TS, cluster centroids of integer and half-integer values are required. This means making amplitude weighting in centroid calculations variable. These are the changes made:
- Added "ampl_weighting_" as a boolean variable to TrigScintClusterProducer.cxx, TrigScintClusterProducer.h, and trig_scint.py. 
- Changed "centroid_" from double to float in TrigScint/include/TrigScint/Event/TrigScintCluster.h
- Set default "ampl_weighting_" to True in trig_scint.py

These are the cluster centroids comparing ampl_weighting as True (first image) and False (second image). For pad 1:

pad 1 true AW
pad 1 false AW

Amplitude weighting as True (first image) and False (second image). For pad 2:

pad 2 true AW
pad 2 false AW

Amplitude Weighting as True (first image) and False (second image). For pad 3:

pad 3 true AW
pad 3 false AW

(The wave-like shape in the "True" cases is bin-related only). In the case of all three pads, the number of clusters made is the same, as expected since this has not been changed. The only changing parameter is the centroid values. The mean and standard deviation values remain (nearly exactly) the same. After clustering, we observe the flag's effect on track-making (first plot shows ampl_weighting True and second False):

image

image

Though the number of tracks formed decreases slightly, this is not surprising. Centroid values are used in building tracks so some change is expected. The effect is acceptable with regards to its future use in an alternate LUT-based track-forming method. And of course track residuals go from:

image
 to the following: 
image

Since tracks can only be made from integer and half-integer centroid values, the range of possible residual values is strictly marked at the two specific peaks rather than distributed over many values when ampl_weighting =True.


tvami:
@kvarnla there is a merge conflict in `TrigScint/python/trig_scint.py`

Screenshot 2026-05-20 at 07 31 10

tvami:
Since you changed the TrigScintCluster object you should increase the ClassDef

https://github.com/LDMX-Software/ldmx-sw/blob/a1139f670972e08e88ed52cba11caee25c84671b/TrigScint/include/TrigScint/Event/TrigScintCluster.h#L181

to
```
ClassDef(TrigScintCluster, 4);
```

tvami:
> Since you changed the TrigScintCluster object you should increase the ClassDef
> 
> https://github.com/LDMX-Software/ldmx-sw/blob/a1139f670972e08e88ed52cba11caee25c84671b/TrigScint/include/TrigScint/Event/TrigScintCluster.h#L181
> 
> ```
> ClassDef(TrigScintCluster, 4);
> ```

Speaking of this tho, what's the reason to go to `float` from `double`? I thought @EinarElen was a big supporter of doing it as a `double`

bryngemark (approved):
This looks good to me and is a good first step towards enabling using a pattern lookup table (where allowing all possible values for the centroid becomes highly impractical).

bryngemark:
> 
> Speaking of this tho, what's the reason to go to `float` from `double`? I thought @EinarElen was a big supporter of doing it as a `double`

That was probably me, it seems like major overkill to use double precision to represent an integer or half-integer value. Don't recall the discussion about preferring `double`, do you remember the argument @tvami ?

tvami:
I dont really remember the details. I remember I was on the side of doing `float`-s and lost the argument haha. Maybe @tomeichlersmith remembers this?

Anyway, the `ClassDef` needs to change with the current changes.

tomeichlersmith:
The biggest reason to prefer `double` is precisely because its precision is so overkill that we don't have to worry about it. `float` has little enough precision that we can observe floating point errors rather easily.

```cpp
#include <iostream>
#include <iomanip>

int main() {
    {
        float one_third = 1./3;
        std::cout << std::setprecision(100) << one_third << std::endl;
    }

    {
        double one_third = 1./3;
        std::cout << std::setprecision(100) << one_third << std::endl;
    }
    
    return 0;
}
```
produces
```
0.3333333432674407958984375
0.333333333333333314829616256247390992939472198486328125
```

https://godbolt.org/z/TcTEPrdn5

My personal stance is that we should use `double` for everything *except* event objects. Having the extra decimal points during the calculation helps avoid propagating floating point errors, but - once the calculation is done - writing down an equivalent `float` (as long as our units are chosen appropriately) will be sufficient for the precision we care about and save us space in the output event file by reducing the "width" of any floating-point "column" in the output tree.

tomeichlersmith:
I found a better example online after thinking about this a bit more. Its a function that should return an exact value of `0.0` and mocks doing a bunch of floating-point calculations (like how our simulation works).

```cpp
#include <iostream>

// https://stackoverflow.com/questions/249467/what-is-a-simple-example-of-floating-point-rounding-error
template<typename Float>
Float example(int k) {
    Float result = k / Float(10);
    for (int i{0}; i < k; i++) {
        result -= 0.1;
    }
    return result;
}

int main() {

    std::cout << example<float>(10000) << std::endl;
    std::cout << example<double>(10000) << std::endl;

    return 0;
}
```
produces
```
0.0971123
-1.58871e-10
```

If we want our result written to the output file to be precise to say, 5 significant figures, then `float` fails but `double` passes. 10k floating point subtractions feels like a reasonable ballpark for what we could see in a simulation. We'd need to get up to 100M for `double` to achieve the same order-of-magnitude error.

bryngemark:
Love it, thanks @tomeichlersmith for this (reminder) explanation and check! Then I agree we should leave  them at double precision.

kvarnla:
Thank you for the input @tvami and @tomeichlersmith ! I've switched the centroids back to `double`-s now and left the `ClassDef` at 3.

cjbarton151 (approved):
Looks good to me! :^)

tvami (approved):

PR #2053, opened 2026-05-21
Updated CLUE clustering to pull RoC values from CSV file, if available.

### LINKED ISSUE #1885: Look up ecal layer weights in CLUE
Currently, CLUE uses hardwired layer weights (for v14). These should be found from the current geometry conditions instead. 

These should be found from the current geometry conditions instead. 

None

cjbarton151:

This resolves #1885 

When calculating... Something (not sure what and it doesn't say in the code), the CLUE algorithm has been relying on a hardcoded array of containment radii defined in [CLUE.h](https://github.com/LDMX-Software/ldmx-sw/blob/4dfa4987f4a2cd01a8b38c5b6e60c4f059587e3e/Ecal/include/Ecal/CLUE.h). These values are outdated and, like most hardcoded values, at risk of becoming outdated again even if we update them.

With this PR, if possible, the hardcoded array will be replaced with values pulled from a CSV file stored in [Ecal/Data](https://github.com/LDMX-Software/ldmx-sw/tree/dfe42ac6b49fd835c84143bf0446d70b59faafbf/Ecal/data). To do this, the path to the selected CSV file is passed through the EcalClusterProducer, just like the rest of the parameters used in CLUE. If there is no CSV file path defined, the program defaults to using the hardcoded values. If the file path is somehow faulty, the program exits with an error. I considered downgrading this to a warning, followed by defaulting to the hardcoded values, but I think an exception is more appropriate.

Only two disclaimers this time:
1. the array in question is a vector of doubles. The file reading algorithm is designed for floats. As far as I can tell, this doesn't affect anything in the results.
(as an aside: it turns out that there's some thorny issue related to localization that messes up `stod` and `stof` when operating in a country where decimal places are denoted by ',' instead of '.' and I spent way too much time diving into this rabbit hole)
2. The RoC array as implemented in CLUE is a 1D array, while in the EcalVetoProcessor (from which I adapted the CSV file reading) uses a 2D array. Each row in the CSV file is valid for a different containment region. I made the assumption that the RoC values used in CLUE are only for the minimum angle range, from 0 to 10 degrees.

## Check List

Ran without crashing, RoC values were updated (when setting ECalClusterProducer.nbr_of_layers to a value >1, which I don't think any of the CI tests do anyway)

### DIFF
--- a/Ecal/include/Ecal/CLUE.h
+++ b/Ecal/include/Ecal/CLUE.h
@@ -17,6 +17,8 @@
 #include <memory>
 #include <set>
 #include <stack>
+#include <stdlib.h>
+#include <fstream>
 
 #include "DetDescr/EcalID.h"
 #include "Ecal/Event/EcalHit.h"
@@ -88,7 +90,7 @@ class CLUE {
   // number we're working on
   std::vector<std::vector<const ldmx::EcalHit*>> clustering(
       std::vector<std::shared_ptr<Density>>& densities, bool connectingLayers,
-      int layerTag = 0);
+      int layerTag = 0, std::string roc_file_name="");
 
   std::vector<std::shared_ptr<Density>> setupForClue3D();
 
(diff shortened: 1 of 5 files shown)


tvami (approved):
Great thanks a lot!

Issue #1880, opened 2025-11-20
ESA full slice test geometry


bryngemark:
**Is your feature request related to a problem? Please describe.**
We need an adequate detector geometry for simulations to match the beam data we will take in the ESA slice test. This is an already ongoing effort; this issue is meant to track related sub-issues.

**Describe the solution you'd like**
Every subsystem implements their system `.gdml` to what's considered an appropriate level of detail. Much like we normally do, the overall `detector.gdml` positions them, here according to technical drawings. 

A note: we will need to revise in later versions of this geometry based on _in-situ_ measurements (eventually derived from beam data) to take actual positions and alignment into account. 

So I imagine
1. technical-drawing version (perfect = baseline)
2. post-installation version: measured position version 
3. post-beam version: includes alignment corrections

I'd like to have all three as it allows us to study the impact on data/MC agreement of large effects (positioning, translational alignment) and smaller (things like rotations).

**Describe alternatives you've considered**
None.

**Additional context**
@Layan-Sarayra @cjbarton151 @SanjitMasanam please (in some cases create and) link your particular reduced geometry issues to this one. Who will take on the overall relative positioning of subdetectors?

Layan-Sarayra:
On my branch [iss1880-TS-light-pipes-trackers ](https://github.com/LDMX-Software/ldmx-sw/tree/iss1880-TS-light-pipes-trackers)  you can find my work on the TS+Tracker subsystem.

I copied ldmx-reduced-v2 and called it ldmx-reduced-v3, and made changes to the following files:

constants.gdml
detector.gdml
materials.gdml
target.gdml
trig_scint.gdml
ldmx-sw/.github/validation_samples/reduced_ldmx/config.py
and created a new file: ts_light_pipes.gdml

I built the geometry for ldmx-reduced-v3 based on the CAD layouts Majd Ghrear designed and are attached below.

Key changes between ldmx-reduced-v2 and v3:

1. Trigger Scintillator (Pads 1 and 2, and same pattern for Pad 3 in target.gdml):
      - Each pad now has: 12 bars per layer × 2 layers = 24 bars
      - Bars are 3 mm (y) × 2 mm (z)
      - Neighbouring bars in y: 0.15 mm gap.
      - Layers are shifted by 1.55 mm in y.
      - Z: layer centers are 4 mm apart (2 mm thickness + 2 mm gap).
[...]

PR #2055, opened 2026-05-22
Cleaned up and centralized materials definitions and visattributes.

cjbarton151:

Following the successful inclusion of the tracking system in the slice test geometry, as outlined in the original plan for #1880, I've tidied up the GDML files to match the reworked format. All materials definitions are standardized and collected in one place, as are all visattributes, and the visattributes have been moved to a separate GDML file. The color mode tag has been included and set to region-by-region coloring. God willing and the creek don't rise, this is the last time we'll have to do this cleanup, and the tidiness should propagate through future geometry versions.

Work is not done on the visattributes side of things; they aren't properly implemented for nearly any volume. However I would like to take care of visattributes for all of the volumes at once - just as soon as we have a full slice test geometry.

## Check List

Passed the `reduced_ldmx` CI test and successfully rendered with `g4-vis`

### DIFF
--- a/.github/validation_samples/cascade_history/config.py
+++ b/.github/validation_samples/cascade_history/config.py
@@ -29,6 +29,7 @@ p.sequence = [my_sim]
 
 import os
 
+
 p.max_events = int(os.environ["LDMX_NUM_EVENTS"])
 p.run = int(os.environ["LDMX_RUN_NUMBER"])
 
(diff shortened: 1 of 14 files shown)


tvami (approved):
Thanks CJ

tomeichlersmith (approved):

Issue #1807, opened 2025-08-11
CI Refactor Meta-Issue


tomeichlersmith:
I am writing this for long-term notes on how to cleanup the CI and make it less error prone. The general goal would be to make it closer to something folks can run locally to make it easier to replicate the CI results and also make it easier to test-run the CI locally to quicken development cycles when adding new CI workflows.

## Drop GitHub "Local" Actions
These are the things stored in `.github/actions/`, and I want to drop them because they are difficult to replicate locally without other tooling. I would be more in favor of keeping the scripts in `.github` (or maybe a `ci` directory for extra clarity) and running then scripts from `just`. The main stumbling block here is that the scripts in `.github/actions/` rely on `.github/actions/common.sh` to define some helper shell functions.

## Drop `compare.py`
Two things are stopping us from using `ComparePlots` in CI:
1. Letting the file determine which plots are made - we can use `keys` from `uproot.File` to loop through all the objects and retrieve the list of histograms.
2. KS Test - we can use SciPy: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.kstest.html

Make sure to fail the CI in the case of missing histograms ("gold" has X but test does not).

## Drop Tar-Ball Build Artifact
This is almost identical to manually creating a layer of a container image. I think our general workflow should be
1. Test compile ldmx-sw with strict static analysis
2. Build ldmx-sw into an image
3. Tests where we run ldmx-sw use the built image

The benefit of this workflow is that the intermediate image can benefit from the extra container tooling that is built around containers (caching, push/pulling, etc.). Steps (1) and (2) could happen at the same time, I just put them in sequence here to emphasize that Step 2 is only necessary if we are running Step 3. Step 3 then does not need to download artifacts and instead would just run the built image with `denv`.

The intermediate ldmx-sw builds would go into a new image registry. I'm thinking something named like `ldmx/build` to distinguish it from the front-facing images used for development or production. We could also look at hosting the ldmx-sw build images on GitHub's Container Registry (ghcr) to potentially improve push/pull rates on the runners themselves.

This could also go along with other refactoring of the image layer stack. Rough outline below since I'm dumping my notes here.

Image | Base | Comments
---|---|---
ldmx/base | ubuntu | necessary dependencies (ROOT, Geant4,...)
ldmx/dev | ldmx/base | additional dev tooling (clang-tidy, etc...)
ldmx/build | ldmx/dev | debug build of ldmx-sw
[...]

PR #2058, opened 2026-05-25
Change the EN config so the threshold is at 100 MeV

tvami:

We said we want to study ω ∈[0.1,4] GeV for that the recoil threshold needs to change accoring to this PR

## Check List

### DIFF
--- a/Biasing/python/target.py
+++ b/Biasing/python/target.py
@@ -50,7 +50,7 @@ def electro_nuclear(detector, generator):
     sim.biasing_operators = [bias_operators.ElectroNuclear("target", 1e5)]
 
     # Configure the sequence in which user actions should be called.
-    recoil_thresh = 0.975 * generator.energy * 1000.0
+    recoil_thresh = 0.9875 * generator.energy * 1000.0
     tagger_threshold = 0.95 * generator.energy * 1000.0
     sim.actions.extend(
         [


tvami:
> 
target_ti_en | ✅ PASS
-- | --

woow that's surprising... I guess there are too few events in the q = 100-200 MeV region to show up in the test? Or our tests are not working...

tvami:
OK so it's not the way the CI was triggered. Interesting.

tomeichlersmith:
Something is broken about the CI, there are plots that fail the KS-test from that run.
```
tom@ty-lee:~/Downloads$ tar tzf target_ti_en_recon_validation_plots.tar.gz fail/
fail/
fail/RecoilDQM_unbiased_pull_u_l10.pdf
fail/RecoilGSFDQM_trk_ecal_loc1-truth_ecal_loc1.pdf
fail/TaggerDQM_pull_theta_vs_p.pdf
fail/RecoilDQM_unbiased_res_u_l0.pdf
fail/TaggerGASDQM_res_d0_vs_p.pdf
fail/RecoilGSFDQM_p_err_vs_p_8hits.pdf
```

tomeichlersmith:
The check is failing to run since it can't find `common.sh`
```
.github/actions/check/check.sh: line 10: /../common.sh: No such file or directory
```
but this is ignored by `set +e` and since it never writes anything to `check_output.txt`, the rest of the check assumes it passes.

I probably failed to bring this up in a previous review and I apologize, but `check.sh` is meant to be run by GitHub as an action. GitHub defines some environment variables and runs the action script from a specific location (so we can deduce `common.sh`). I'm not super happy with this setup since its difficult to test locally (see https://github.com/LDMX-Software/ldmx-sw/issues/1807), so I'm happy to move away from GitHub "local actions" and instead just have scripts that are run from `just`. We could even have a CI-specific justfile if we wanted e.g. https://just.systems/man/en/modules.html or https://just.systems/man/en/imports.html

tvami:
Thanks for finding this Tom. I'm a little confused on where this went wrong as clearly it had worked at a certain point haha. Anyway, I'll try to fix this next

tomeichlersmith (approved):

tvami:
OK I'm happy with the CI msg now!

therwig (approved):

tvami:
Screenshot 2026-05-27 at 17 46 55

Not much changes when doing this threshold change for the physics category point of view

PR #2060, opened 2026-05-26
Adding a simplified HCal to the reduced-v3 geometry

cjbarton151:

This is the next step in #1880 

The HCal should be the final subsystem component which needs to be added for a full slice test geometry. This implementation features a highly simplified, but mostly dimensionally accurate, version of the slice test HCal.

The composition and dimensions of the objects directly in the beamline should be fairly accurate. This includes the 6 quad bars at ESA, plus two sides of the plastic case containing them (the sides with surfaces perpendicular to the beam direction). The other 4 sides of the casing, the 80/20 supports, the cables/electronics, and the table are not included.

In `constants.gdml`, I added new constants for the slice test HCal, but left old constants for the baseline HCal intact, because there are still pieces of `ecal.gdml` and `scoring_planes.gdml` which rely on them.

## Check List

### Geometry validation
(along beam view)
image
(side view)
image
The orange box is the boundaries of the HCal envelope, the thin bars on the left and right side are the plastic housing, and the six bars in the center are the quad bars.
Side view of entire slice test geometry:
image

### Physics validation
Well, I don't really know what to expect the data to look like, but we're getting entries in all the hits collections associated with the HCal, at least, when running the reduced CI.
image
image
And from the DQM histos:
image

### DIFF
--- a/Detectors/data/ldmx-reduced-v3/constants.gdml
+++ b/Detectors/data/ldmx-reduced-v3/constants.gdml
@@ -1,4 +1,4 @@
-<!-- Constant defining the clearance between volumes --> 
+s<!-- Constant defining the clearance between volumes --> 
 <constant name="clearance" value="0.001*mm" />
 
 <!-- Define center position and identity rotation -->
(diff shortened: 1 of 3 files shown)


cjbarton151:
I would also like to emphasize that the z position of the HCal is arbitrarily chosen to have its center 1.0 m downstream. This may be (probably should be) changed once final checks and adjustments are made.

tvami (approved):

tomeichlersmith (approved):
makes sense to me, I think the Ecal housing is pretty incorrect but that is another issue/pr (or something we never deal with until another version of the Ecal is assembled that can actually collect data to compare to the sim)

I have no ideas/opinions about correct placement along z. @erik-wallin ?

tvami:
>  I think the Ecal housing is pretty incorrect but that is another issue/pr (or something we never deal with until another version of the Ecal is assembled that can actually collect data to compare to the sim)

Yes I pinged @SanjitMasanam about this we should have the housing in here that actually was in the prototype

PR #2061, opened 2026-05-26
Move CI scripts to just recipes, and use them in CI

tvami:

Resolves the issue seen in https://github.com/LDMX-Software/ldmx-sw/pull/2058#issuecomment-4545217124

## Check List


tvami:
@tomeichlersmith my understanding is that none of this is tested until it's merged. Please let me know if you are fine with the changes

tomeichlersmith (approved):
Yea I think this makes sense

PR #2064, opened 2026-05-27
Iss1325 truthseedprocessor class separations

### LINKED ISSUE #1325: Seeding and Truth information improvements
There are some issues with the seeding in the tagger and recoil volumes that need to be investigated. However the current processors are rather confusing in terms of functionalities and implementations. Additionally, we should have a better handle on the technical efficiency. 

--- Implementation changes
The `TruthSeedProcessor` should made clear that it provides both truth tracks and seeds. The way seed tracks are provided for recoil and tagger tracker will differ, so I think:
- Make 1 base class for the `TruthSeedProcessor`
- Make 2 derived classes for the Tagger and Recoil respectively. 

--- Setup a truth track analysis
Check the Nhits distribution on the electron beam
Survival rate of electrons to the target surface
...

vdharanikota0:
Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1325

This is a draft, still debugging but code compiles and runs.

Split TruthSeedProcessor into two classes: TruthSeedProcessor and TruthTrackProcessor

- TruthSeedProcessor: handles scoring plane hit selection, hit counting,
   starts extrapolation, and seed smearing
- TruthTrackProcessor: finishes extrapolation to target surface, applies stricter momentum
  and hit cuts, and applies ecal bound state to tracks

Issues:
-histograms are not lining up in TaggerDQM and RecoilDQM
-run time is long since both classes are performing extrapolations, want input on how the processes in these classes should be split


Tracking/python/tracking.py
```diff
@@ -344,46 +301,104 @@ class TruthSeedProcessor(Processor):
...
         The name of the recoil seeds collection.
-    recoil_truth_collection : str
-        The name of the recoil truth collection.
     field_map: str
         Magnetic field map
     """
 
     debug: bool = False
-    pdg_ids: list[int] = [11]
+    pdg_ids: list[int] = field(default_factory=lambda: [11], metadata={"entry_type": int, "dimension": 1})
```
tvami: was this changed by you? I dont see why we need it

Tracking/python/tracking.py
```diff
@@ -344,46 +301,104 @@ class TruthSeedProcessor(Processor):
...
-    n_min_hits_recoil: int = 7
     z_min: float = -9999.0
     track_id: int = -9999
-    pz_cut: float = -9999.0
     p_cut: float = 0.0
-    p_cut_max: float = 100000.0
     p_cut_ecal: float = -1.0
+    recoil_sp: bool = True
+    n_min_hits_tagger: int = 11
+    beamOrigin: list[float] = field(default_factory=lambda: [-883.0, -21.745876, 0.0], metadata={"entry_type": float, "dimension": 1})
```
tvami: again I dont think you need the field stuff

Tracking/python/tracking.py
```diff
@@ -344,46 +301,104 @@ class TruthSeedProcessor(Processor):
...
+    n_min_hits_tagger: int = 11
+    n_min_hits_recoil: int = 7
+    pz_cut: float = -9999.0
+    p_cut: float = 0.0
+    p_cut_max: float = 100000.0
+    skip_tagger: bool = False
+    skip_recoil: bool = False
+    particle_hypothesis: int = 11
+    seedSmearing: bool = False
+    d0smear: list[float] = field(default_factory=lambda: [0.01, 0.01, 0.01], metadata={"entry_type": float, "dimension": 1})
```
tvami: same comments on here too

tvami:
Screenshot 2026-05-28 at 17 33 37

There are conflicts that you'll need to resolve

tvami:
@bloodyyugo what do you think about this?

bloodyyugo:
Can you describe in more detail what these two classes are doing?  It seems like TruthSeedProcessor is making truth tracks, then TruthTrackProcessor is also making truth tracks and also making (maybe) smeared truth tracks.  I am also getting confused by the term "seed" here (and in the old TruthSeedProcessor tbh).

tvami:
The old "TruthSeedProcessor" was doing both "truth seeds" and "truth tracks", now with this change these are decoupled similar to the way we do this at reco level.

bloodyyugo:
From what I can tell (in the old processor, which has always been hard to parse) the tracks with the "seed" label are just the truth tracks (maybe smeared) with a non-zero covariance matrix.  Honestly, I'm not sure why we save both "truth" and "seed" collections though it may be from PF doing studies.  

The beamelectrons are handled a bit differently (calls taggerFullSeed, the true position at the target SP is saved as as the atTarget scoring plane and sim hit IDs are added as measurement IDs). Non-beam stuff (in the tagger, recoil is ok) never get their hits added to the track, which seems like an oversight.

tvami:
> though it may be from PF doing studies.

PF actually wanted to break these up, see:
https://github.com/LDMX-Software/ldmx-sw/issues/1325

bloodyyugo:
Right, but he's referring to having a base class and then child recoil/tagger classes taking care of (I guess) the different track states get created and handling "beamelectrons" and whatever needs to be handled differently.   

And I can see how that would be sensible; createTruthTrack and seedFromTruth methods are agnostic, you could just include a loop over a single supplied scoring-plane list in the base class and maybe have a method call to "doSpecificTrackerStuff" (which does different things for recoil/tagger classes)...

But,  I'm just not sure the PR here makes things that simpler.  It adds 3 extra collections to the event (from the new TruthSeedProc), which (I think?) contain exactly the same info as the truth_track output of TrackTruthProc apart from the helical parameters.  It does simplify the scoring plane loops, I guess.

vdharanikota0:
@bloodyyugo @tvami Yes, the older TruthSeedProcessor class would output tagger_seed_collections and recoil_seed_collections, treating “seeds” as tracks that had smearing applied to them. But now, the new TruthSeedProcessor class treats “seeds” as initial candidates for tracks that have not undergone extrapolations. And the TruthTrackProcessor takes the hit maps and seed attributes from the TruthSeedProcessor and makes the final tracks. I see where this definition of “seed” is not completely clear. 

I agree with what you are saying, as separating the classes does not add much optimization other than with the scoring plane hits. Maybe a better way to address the problems in the TruthSeedProcessor class would be to organize its functions and logical progression of creating a track better, as there are unused functions and dead code throughout the class that makes it confusing to read through, like the recoilFullSeed function that never gets used. As you were saying, making respective functions for recoil, tagger, beam electrons and decluttering the “produce” function would make sense. Let me know what you think of this.

PR #2066, opened 2026-05-29
Tracker Decoding

mgignac:

Implements decoding of the track ESA25 test-beam data. In addition, separates out truth and reconstructed hits into separate classes.  This is the first PR of a few that will implement the workflow for analyzing the test-beam data. It has intentionally been broken into smaller components for easier review.

## Check List


Packing/src/Packing/SingleSubsystemUnpacker.cxx
```diff
@@ -78,9 +84,13 @@ void SingleSubsystemUnpacker::produce(framework::Event& event) {
...
-    // ror_header has global RoR information
-    event.getEventHeader().setIntParameter("RoR Timestamp",
-                                           ror_header.timestamp());
+    // Store the full 64-bit RoR timestamp as two 32-bit halves since
+    // EventHeader::setIntParameter only accepts int.
+    uint64_t ts = ror_header.timestamp();
+    event.getEventHeader().setIntParameter("RoR Timestamp LSB",
+                                           static_cast<int>(ts & 0xFFFFFFFFU));
+    event.getEventHeader().setIntParameter("RoR Timestamp MSB",
+                                           static_cast<int>(ts >> 32));
```
mgignac, line 94: @tomeichlersmith -- this is one point where I could use some input from you. The timestamp is currently a 32 bit integer, which is insufficient for the test beam data (64 bit int). Do we want to change the data structure? As implemented now, it's a bit of a workaround, but I suspect this isn't the preferred long-term solution....

Tracking/include/Tracking/Event/RawSiStripHit.h
```diff
@@ -158,7 +104,7 @@ class RawSiStripHit {
   uint8_t  filter_{0};
 
   /// Class declaration needed by the ROOT dictionary.
-  ClassDef(RawSiStripHit, 3);
+  ClassDefOverride(RawSiStripHit, 4);
```
tvami, line 107: Why is this Override instead of just bumping up the ClassDef? Same question for the other class too

mgignac: I thought this was necessary since `RawSiStripHit` derives from `SiStripHit` which has a ClassDef already.   With `-Winconsistent-missing-override` on (this build enables it via ADDITIONAL_WARNINGS=ON), that produces a warning — and the CI's configure-force-error job turns warnings into errors, so it'd fail the build. Without it, I ran into the following build time error:

```
  warning: 'Streamer' overrides a member function but is not marked 'override'
    [-Winconsistent-missing-override]
     ClassDef(SimSiStripHit, 1);
     ...
     note: overridden virtual function is here
     ClassDef(SiStripHit, 1);
     ```
Happy for some guidance here though if this isn't correct :)

File: Packing/src/Packing/SingleSubsystemUnpacker.cxx (hunk 1 above)
Comment (tomeichlersmith) [reply Packing/src/Packing/SingleSubsystemUnpacker.cxx:94]: I think changing `LDMXRoRHeader` is reasonable, I did not notice that inconsistency. For storing this timestamp in the `EventHeader`, I would leave it in two parts as-is. I have [long term schemes](https://codeberg.org/tomeichlersmith/ldmx-rntuple-experimentation) that will update `EventHeader` and make this easier anyways.

Comment (mgignac) [reply Packing/src/Packing/SingleSubsystemUnpacker.cxx:94]: Sounds good, many thanks Tom!

File: Tracking/include/Tracking/Event/RawSiStripHit.h (hunk 2 above)
Comment (tvami) [reply Tracking/include/Tracking/Event/RawSiStripHit.h:107]: OK makes sense, thanks!

**Review by bryngemark — COMMENTED** [review]
A detail here, would it be possible to stick to the name convention `[SomeSystem]SimHit` instead of separating `Sim` and `Hit`? It makes applying drop/keep rules with wildcards a lot easier (and we've been bitten by having a different naming convention in the trackers before so harmonizing would be great). 

If relevant elsewhere in this work (I haven't checked at all), other systems also use the same conventions for `...Digi` and `...RecHit` so using that convention would be great.

**Comment by mgignac** [comment]
Hi @bryngemark, 
Happy to discuss -- let me just make sure I understand. You want use to drop the SiStrip part of the naming convention, so we have `TrackerSimHit` and `TrackerRawHit` (and base abstract class `SiStripHit` to `TrackerHit`. We would, however, like to keep separate classes for the real (raw) data and simulated data, as their class structures are entirely different and it wouldn't make sense to have one for both. We should make sure this make sense as the SiStrip hits get processed downstream to produce clusters and measurements. @bloodyyugo , any thoughts on this?
Cheers,
Matthew

**Review by tomeichlersmith — COMMENTED** [review]
Please remove claude as a co-author from these commits. If it did so much that you cannot remove its contributions, then I would not want to merge this. But what I suspect is the case is that you just had it do some routine `git` shenanigans in which case I _really_ want to remove the contribution. Not only do I want to avoid giving free advertising to Anthropic, I also don't want anyone to believe that ldmx-sw is accepting slop contributions.

I want to again offer my help if you want to learn more about `git` - I require less water per day, you don't have to pay me a subscription fee, and I don't instruct the US government to bomb a children's school!

**Comment by mgignac** [comment]
Done.

File: Tracking/python/full_tracking_sequence.py [hunk 3]
```diff
@@ -178,7 +178,7 @@ def tagged(name):
             depletion_voltage=70.0,
             noise_electrons=1000.0,
             threshold_electrons=3000.0,
-            out_raw_collection=tagged("RecoilRawSiStripHits"),
+            out_raw_collection=tagged("RecoilSimSiStripHits"),
```
tomeichlersmith: These collection names are the things that are looked at by the drop/keep rules. I don't think we need to rename the classes unless you want to (e.g. for ecal/hcal, the class is named `SimCalorimeterHit` while the branch name is `EcalSimHits` (or `HcalSimHits`)). It would just be aligning the in/out collection names so that the name of the output branches following the `___SimHits` convention.

tomeichlersmith (commented):
I think this looks good - my only comment is about the event object naming stuff to hopefully clarify @bryngemark comment.

Tracking/python/full_tracking_sequence.py
mgignac: I don't have any strong opinions on this, and something we can also change without too much impact later on if we decide we don't like the choice made here. I'm fine moving this to `RecoilSimHits` to make things a bit more uniform unless @bloodyyugo has objections?

tomeichlersmith (approved):

Tracking/python/full_tracking_sequence.py
bryngemark: yes, all i was asking for here is to consider `SimHit`a word so to speak, so you can use SiStrip all you want, but it would be more streamlined with other systems to call it `...SiStripSimHits` than to call it `...SimSiStripHits`

bryngemark:
besides the naming thing my only input is thank you for moving this along, i'm really looking forward to having decoding in place for all subsystems and with your work we're getting there!

Tracking/python/full_tracking_sequence.py
mgignac: Ah ok. I will have another PR shorty and may reinstate the `SiStrip` into the name but keep the phrase `SimHit`. So something like `SiStripSimHit` would probably do the trick. But I'll leave that change for the next PR.

tomeichlersmith: Apologies for jumping to merge too quickly!

PR #2068, opened 2026-06-04
Add WAB target filter

kawaterrysit:

This PR adds a target bremsstrahlung filter for WAB studies.

The filter selects events containing bremsstrahlung photons produced in the target and allows configurable angular requirements through brem_theta_min and brem_theta_max(Minimum and Maximum bremsstrahlung photon theta angle), as well as dralmin and dralmax(Minimum and Maximum angle between WAB photon and electron). This enables filtering of WAB events during simulation.

Notes
This PR is being opened as a draft for initial review and feedback on the implementation.
Validation studies are still in progress and will be added in a future update before the PR is marked as ready for review.

## Check List
- [x ] I successfully compiled _ldmx-sw_ with my developments.
- [x ] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

**Preliminary Validation**
The figure below shows the total energy deposited in the target for an 8 GeV electron beam with the WAB target filter enabled.

_Applied selection:_

Maximum recoil electron energy: 1.5 GeV
Minimum photon energy: 1.0 GeV
Minimum bremsstrahlung photon polar angle (brem_theta_min): 0.5 rad
Maximum WAB electron-photon separation (dralmax): 0.5
wab_filter_ldmx_sw

**Motivation**
The figure below shows the recoil energy distribution of the effective LYSO nucleus from the WAB MadGraph generation with an 8 GeV electron beam.

The long-term goal of this study is to reproduce, as closely as possible, the generator-level behavior using only the information available in ldmx-sw. The target bremsstrahlung filter introduced in this PR is intended to provide a framework for selecting events with kinematic properties similar to those generated in the MadGraph WAB sample.

Initial comparisons indicate noticeable differences between the generator-level lyso nucleus recoil-energy distribution and the corresponding distributions obtained from the current ldmx-sw implementation. Further validation and investigation are ongoing.
nucleus_T_overlay_chi1

**Comparison with Inclusive Background**
The figure below shows the total energy deposited in the target for the inclusive 8 GeV electron background sample provided by the standard single_8gev_e_upstream_tagger() generator, with no WAB target-filter selections applied.

The resulting distribution is very similar to the filtered WAB-target-filter sample shown above. This preliminary comparison suggests that the current filter requirements do not significantly change the target energy deposition distribution. Further validation is still in progress.
inclusive_bkg


Biasing/include/Biasing/TargetBremFilter.h
```diff
@@ -68,6 +68,14 @@ class TargetBremFilter : public simcore::UserAction {
   /// Brem gamma energy treshold
   double brem_energy_threshold_{2500};
 
+  /// Maximum and Minimum brem gamma angle relative to beam axis [rad].
+  double brem_theta_min_{-999.0};  // rad
+  double brem_theta_max_{999.0};   // rad
+
+  /// Maximum and Minimum electron-photon separation in eta-phi space.
+  double dral_max_{999.};
+  double dral_min_{-999.};
```
tvami: this should be 0 for the default. But also it's better not to have these initialized here (ie. just do ` double dral_min_;` here and have the `Biasing/python/filters.py`'s content give the defaults.

Biasing/src/Biasing/TargetBremFilter.cxx
```diff
@@ -129,12 +138,48 @@ void TargetBremFilter::stepping(const G4Step* step) {
...
+          auto momentum = secondary_track->GetMomentum();
+          double theta = std::atan2(std::sqrt(momentum.x() * momentum.x() +
+                                              momentum.y() * momentum.y()),
+                                    momentum.z());
+          bool pass_brem_theta =
+              theta >= brem_theta_min_ && theta <= brem_theta_max_;
+
+          // Maximum and Minimum angle between outgoing Brem photons and
+          // electrons
+          auto gamma_mom = secondary_track->GetMomentum();
```
tvami, line 172: shouldnt we check the PDG ID for this? to make sure it's a gamma?

Biasing/src/Biasing/TargetBremFilter.cxx
```diff
@@ -129,12 +138,48 @@ void TargetBremFilter::stepping(const G4Step* step) {
...
+          auto electron_mom = track->GetMomentum();
```
tvami, line 173: same thing here

Biasing/src/Biasing/TargetBremFilter.cxx
kawaterrysit: Actually, this part makes me some kind of confused. In the original implementation, the code seems to have only checked that the eBrem process existed and that the secondary passed the energy threshold, but it didn't explicitly require the secondary to be a photon?

Biasing/src/Biasing/TargetBremFilter.cxx
kawaterrysit: For `electron_mom`, I think `track` should already be guaranteed to be the primary electron from the earlier checks at the start of `stepping()`:

'''cpp
if (track->GetParentID() != 0) return;
if (pdg_id != 11) return;
'''

Do you think it would be useful to add another explicit check here for safety?

Biasing/src/Biasing/TargetBremFilter.cxx
tvami: ok I guess the `eBrem` can only make photons, probably it's fine as is then. You could test it by just printing out the IDs for a few events

Biasing/src/Biasing/TargetBremFilter.cxx
tvami: ok sounds good, no need then! thanks

tvami:
With the current setting I have
```
[ SimulatorBase ] 65  warn: Started 60445231 events to produce 8 events.
```

I guess WABs are rare and there is a reason why we have been doing MadGraph

tomeichlersmith (approved):
This looks good to me, I am curious to see if it affects the physics distributions much if at all. I'm unsure if the 0.0-0.1 radian range is heavily populated after we make our requirement that enough energy goes into the photon but since the defaults are not applying these additional filters that is another study.

tvami:
The diffs are because I increased the 
```
ldmx_log(info) << "ACCEPTED: brem candidate with theta=" << last_theta
                   << ", dR=" << last_dral;
```

I think we could keep this and have a new gold (which is anyway kinda due [since we moved to the new OS])

tvami (approved):

Issue #2035, opened 2026-04-22
Move TS tracking to use LUT


kvarnla:
**Is your feature request related to a problem? Please describe.**
- TSTrackProducer creates tracks based on a certain acceptable distance between clusters that makes it probable that they belong to the same electron. In the case of a module misalignment, these distances will be different and possibly unknown. 

- TSFirmwareTracker uses a LUT method to produce tracks based on a fixed set of acceptable cluster combinations, also based on distance between clusters. 

- Simulations show that more cluster combinations than these exist in good tracks, and good tracks are rejected as a consequence.

**Describe the solution you'd like**
By using a LUT method which creates tracks based on how frequently their cluster combinations appear, the misalignment is immediately learned from the data. Also this way, the track making efficiency is potentially increased. Additionally, using a LUT is usually faster than running matching loops to create tracks. 

**Describe alternatives you've considered**
In the case of a misalignment, one could apply an offset (if the misalignment can be learned from the data), to counteract the misalignment. In this case, one would have to analyze the frequencies of different cluster combinations to establish the misalignment. In effect, this would mean creating a frequency-based LUT. 

**Additional context**
Using a LUT in software tracking would harmonize this with firmware tracking.

bryngemark:
@kvarnla ricardo's PR is merged, are you ready to make a PR?

PR #2070, opened 2026-06-06
Iss2035 add LUT method tracking to TSTrackProducer

kvarnla:
I am updating _ldmx-sw_, here are the details.​​

As part of my bachelor thesis project, I have implemented a method of TS track-making using a lookup table to accommodate potential misalignment. To do this, a boolean variable "lut_tracking" has been added to TrigScint/src/TrigScint/TrigScintTrackProducer, so that this method and the (currently existing) method using max_delta can be 'switched between.' It reads a LUT file, a text file, in order to realize the new method; the LUT file is constructed using two new analyzers as described below. The LUT file is a list of cluster combinations from the three pads which appear "often enough" to be considered as originating from beam electrons. ​ ​

I have added the following files: ​

- **ClusterTripletMaker.cxx** and **ClusterTripletMaker.h** : analyzer to write the clusters in pads 1, 2, and 3 to a text file ("clusters.txt"). ​
- **PatternLUTMaker.cxx** and **PatternLUTMaker.h** : analyzer to take the "clusters.txt" as input and output a LUT text file ("LUT.txt"). It computes the vertical distances between cluster centroids in pads 1 and 2, and between clusters in pads 2 and 3, and groups them by these combinations ("patterns"). The configurable "lut_threshold" is the minimum frequency that a such combination must appear in order to be written to the LUT.txt file. I have set it to default 0.0008 since this returns the highest track-making efficiency with the lowest rate of fake tracks for sets of <1 million events. ​
- **runClusterstxt.py** and **runLUTana.py** : exampleConfigs for the respective analyzers above. ​

I have changed the following files: ​

- **TrigScintTrackProducer.cxx** and **TrigScintTrackProducer.h** : to introduce lut_tracking and another variable (lut_file) to specify the LUT file. ​
- **trig_scint.py** : to include the new variables in 'TrigScintTrackProducer' and added definitions for both new analyzers. ​(Additionally: I believe TrigScintTrackProducer was defined twice in trig_scint.py for some reason. I have removed one block). 
- **runLUTtracking.py** : an exampleConfig to demonstrate configuring TrigScintTrackProducer to run both lut-method tracking and max_delta definition tracking. ​

 
### What are the issues that this addresses?​​

This resolves issue #2035.​
​

## Check List​

- [X] I successfully compiled _ldmx-sw_ with my developments.​
- [X] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).​
​
- [X] I ran my developments and the following shows that they are successful.​
​ ​

Excerpt of runClusterstxt.py output, which utilizes ClusterTripletMaker.cxx to write the following clusters.txt file: 

image

The columns here refer to event number, pad 1 cluster centroid, pad 2 cluster centroid, and pad 3 cluster centroid. 

Excerpt of runLUTana.py output, which takes the runClusterstxt.py output and uses PatternLUTMaker.cxx to write the following LUT.txt:​
​
image

The columns here refer to pad 1 cluster, pad 2 cluster, and pad 3 cluster for those rows in clusters.txt which pass the LUT threshold when grouped into patterns in PatternLUTMaker. 

Here is track-making as performed in runLUTtracking.py for both lut-method tracking and current max_delta method tracking. The number of tracks made by lut-method tracking is dependent on the LUT threshold, here set at 0.0008: 

For the max_delta definition tracking, as currently implemented in TrigScintTrackProducer:
image

For lut-method tracking, as proposed here (for a lut_threshold of 0.0008):
image


cjbarton151:
Hey Lucia - I thought I'd make a few intertwined, higher-level suggestions that wouldn't fit properly into the line-by-line code review.

Currently, your commit implements the LUT production in `TrigScint/ClusterTripletMaker`. I would argue that the LUT construction will probably be done once for each geometry*, and so it doesn't need to be integrated into the main bus. Maybe it should exist in  `Tools/ClusterTripletMaker` and  `Tools/PatternLUTMaker` and be called with its own dedicated config script, also stored in Tools. You can see `Tools/HgcrocEmulator` and the associated config script for an example on how to do this. Then, the `LUT.txt` files can be stored in `TrigScint/data`, each one with a slightly more descriptive name for each geometry and threshold e.g. `LUT_v15_8gev_10000.txt` or `LUT_reducedv3_8gev_10000.txt`, for events with a threshold of 1/10000.

*You might be wondering 'why would I only need to make one LUT for each geometry, when the probability threshold can be varied?' Well, I see an opportunity to streamline the functionality by including the probability of each combination as a 5th parameter in the LUT, then do the threshold cut once on process start in the `TrigScintTrackProducer`. But we can leave this task for another contributor. For now, I think that we probably won't be playing around with too many different thresholds; we'd probably just choose a seemingly optimal threshold for any given geometry and stick with it, so having one LUT.txt per geometry should in practice suffice.

@bryngemark may have different thoughts on the matter, so I think it would be useful to hear a second opinion before making large-scale changes to your pull request.

TrigScint/exampleConfigs/runClusterstxt.py
```diff
@@ -0,0 +1,49 @@
...
+    cluster.ampl_weighting = False
+    cluster.clustering_threshold = 3.0
+
+from LDMX.TrigScint.trig_scint import ClusterTripletMaker
+
+triplets = ClusterTripletMaker("tripletmaker")
+triplets.output_collection = "clusters.txt"
+
+p.sequence = [
+             #*truth_hits,
```
bryngemark: if you want these commented here as an indication that including them is optional, i suggest either 
- adding a comment saying something like "uncomment to make truth tables" (but i think you need some more stuff for that to work, right? you'd need to define `truth_hits` and use them as input collection to the later processors (at least digi))
- adding a more elaborate setup which does all of the needed steps to use truth hits, and toggle with some boolean, doing {if [that boolean], define stuff, p.sequence.append(truth_hits), then digi etc}

TrigScint/exampleConfigs/runLUTana.py
```diff
@@ -0,0 +1,15 @@
+#config file to make LUT for tracking using PatternLUTMaker
+
+from LDMX.Framework import ldmxcfg
+
+p = ldmxcfg.Process("LUTmaker")
+p.input_files =  ["Clusters.root"] #necessary to define but unused, the real input file is defined below
```
bryngemark: I see what @cjbarton151 means here: you're not really using `ldmx-sw` capabilities to manage ROOT tree input and output objects for this, but rather an auxiliary file structure where you have a messy input list and an ordered output list. So in that sense he is right that it's more of a tool and I think moving these files (stuff not reading/writing TS info from/to a tree) to `Tools` is a good idea. Reach out to one of us if this seems too complicated and we can discuss

TrigScint/exampleConfigs/runLUTtracking.py
```diff
@@ -0,0 +1,27 @@
...
+
+tstp_tracks.delta_max = 1.0
+tstp_tracks.verbosity = 1
+tstp_tracks.output_collection = "tstpTracks"
+
+lut_tracks.delta_max = 1.0
+lut_tracks.verbosity = 1
+lut_tracks.lut_tracking = True
+lut_tracks.lut_file = "LUT.txt"
+lut_tracks.output_collection = "lutTracks"
```
bryngemark, line 18: do these settings differ from the default ones specified in `trig_scint.py`? if not, i suggest omitting them

Tools/include/Tools/ClusterTripletMaker.h
```diff
@@ -0,0 +1,62 @@
+/**
+ * @file ClusterTripletMaker.h
+ * @brief Writes cluster combinations to text file
+ * @author Lucia Kvarnstrom, Lund University
```
bryngemark: typo: ö ;)

Tools/include/Tools/ClusterTripletMaker.h
```diff
@@ -0,0 +1,62 @@
...
+  int verbose_{0};
+
+  // specific pass name
+  std::string pass_name_{""};
+
+  // input cluster collections
+  std::vector<std::string> cluster_input_collections_;
+
+  // output text file
+  std::string output_collection_{"clusters.txt"};
```
bryngemark: can you name it something else than `collection` since it's not going to be on a ROOT TTree ?

Tools/include/Tools/PatternLUTMaker.h
```diff
@@ -0,0 +1,64 @@
...
+
+ private:
+  // verbosity
+  int verbose_{0};
+
+  // input text file of clusters
+  std::string input_collection_;
+
+  // output LUT file name
+  std::string output_collection_;
```
bryngemark: same comment, rename from collection if possible

Tools/src/Tools/ClusterTripletMaker.cxx
```diff
@@ -0,0 +1,116 @@
...
+  const auto clusters_pad2{event.getCollection<ldmx::TrigScintCluster>(
+      cluster_input_collections_.at(1), pass_name_)};
+
+  const auto clusters_pad3{event.getCollection<ldmx::TrigScintCluster>(
+      cluster_input_collections_.at(2), pass_name_)};
+
+  const size_t num_clusters = std::min(
+      {clusters_pad1.size(), clusters_pad2.size(), clusters_pad3.size()});
+
+  if (verbose_ > 1) {
```
bryngemark: do we need this additional verbosity check or is it enough to let the logger level (debug) handle this?

Tools/src/Tools/PatternLUTMaker.cxx
```diff
@@ -0,0 +1,115 @@
...
+  // Example --- horizontal tracks whose pad 1-2 and pad 2-3 delta values
+  //             are both 0 make up ~80% of tracks out of 10.000 events,
+  //             while a track with deltas (+22,-22), meaning pad 1 cluster in
+  //             e.g. bar 4, pad 2 cluster in bar 26, and pad 3 cluster in bar 4
+  //             only appears once in 10.000 events (0.01%). The straight
+  //             tracks are written to the LUT and the single "anomaly" is not.
+  // Primary motivation here is for use in the case of an unknown TS
+  // misalignment.
+
+  if (verbose_) {
```
bryngemark: same here, is info level verbosity enough?

Tools/src/Tools/PatternLUTMaker.cxx
```diff
@@ -0,0 +1,115 @@
...
+    ldmx_log(info) << "Total number of track candidates: " << totalLines << "\n"
+                   << "Number of track candidate types: " << groups.size()
+                   << "\n"
+                   << "LUT Threshold: " << lut_threshold_ * 100 << "%\n";
+  }
+
+  for (auto& g : groups) {
+    float p12 = g.first.first;
+    float p23 = g.first.second;
```
bryngemark: are these used outside of the printout? otherwise, don't define them as separate variables, but instead access them directly in the printout (it saves resources to not define stuff that is not used)

Tools/src/Tools/PatternLUTMaker.cxx
```diff
@@ -0,0 +1,115 @@
...
+    int count = g.second.size();
+
+    double frac = static_cast<double>(count) / totalLines;
+
+    if (verbose_) {
+      ldmx_log(info) << "(" << p12 << "," << p23 << ") appears " << count
```
bryngemark: suggest to move this to `debug` level, which is usually used for detailed low-level info that appears once per event. 
`info` is used for more generally interesting things like settings etc that happens once per run, and if you have something that happens e.g. for every hit or cluster i.e. many times per event, use level `trace`. odd stuff that shouldn´t happen is `warning`, and then you have `fatal` for really serious errors

Tools/src/Tools/PatternLUTMaker.cxx
```diff
@@ -0,0 +1,115 @@
...
+  while (infile >> ev >> p1 >> p2 >> p3) {
+    float p12 = p2 - p1;
+    float p23 = p3 - p2;
+
+    groups[{p12, p23}].push_back({ev, p1, p2, p3});
+
+    totalLines++;
+  }
+
+  return;
```
bryngemark:
```suggestion
  int ev;
  float p1, p2, p3;
  while (infile >> ev >> p1 >> p2 >> p3) {
    float p12 = p2 - p1;
    float p23 = p3 - p2;
    groups[{p12, p23}].push_back({ev, p1, p2, p3});
    totalLines++;
  }
  return;
```

TrigScint/src/TrigScint/TrigScintTrackProducer.cxx
```diff
@@ -153,107 +176,136 @@ void TrigScintTrackProducer::produce(framework::Event &event) {
...
+              (centroid >= vert_bar_start_idx_ &&
+               cluster1.getCentroid() >= vert_bar_start_idx_ &&
+               seed.getCentroidX() == cluster1.getCentroidX())) {
+            // use geometry y overlap scheme to see if this is really a match in
+            // x should be done in a map
+
+            if (centroid >= vert_bar_start_idx_ &&
+                seed.getCentroidY() < cluster1.getCentroidY()) {
+              // impossible combination
+              if (verbose_ > 1) {
```
bryngemark: remove verbosity check and promote logger level to `warning`

TrigScint/src/TrigScint/TrigScintTrackProducer.cxx
```diff
@@ -153,107 +176,136 @@ void TrigScintTrackProducer::produce(framework::Event &event) {
...
+                // use geometry y overlap scheme to see if this is really a
+                // match
+                // in x
+
+                if (centroid >= vert_bar_start_idx_ &&
+                    (seed.getCentroidY() < cluster2.getCentroidY() ||
+                     cluster1.getCentroidY() >
+                         cluster2.getCentroidY())) {  // impossible
+                  if (verbose_ > 1) {
+                    ldmx_log(debug)
```
bryngemark: same here

TrigScint/src/TrigScint/TrigScintTrackProducer.cxx
```diff
@@ -333,10 +385,15 @@ void TrigScintTrackProducer::produce(framework::Event &event) {
         // no need to start pulling constituents from tracks that are
         // ridiculously far apart
         if (((fabs(track.getCentroid() - next_track.getCentroid()) <
-            3 * max_delta_) && (track.getCentroid()<vert_bar_start_idx_)) //for the horizontal bars
-            || ((fabs(track.getCentroidX() - next_track.getCentroidX()) < 2*max_delta_vert)
-          && (track.getCentroidY() == next_track.getCentroidY()) && (track.getCentroid()>=vert_bar_start_idx_))) { 
-            //and for the vertical bars, check if they are in the same quad and close enough
+              3 * max_delta_) &&
+             (track.getCentroid() <
+              vert_bar_start_idx_))  // for the horizontal bars
```
bryngemark:
```suggestion
              3 * max_delta_) && (track.getCentroid() < vert_bar_start_idx_))  // for the horizontal bars
```

TrigScint/src/TrigScint/TrigScintTrackProducer.cxx
```diff
@@ -364,11 +424,19 @@ void TrigScintTrackProducer::produce(framework::Event &event) {
...
-            && (track.getCentroid()>=vert_bar_start_idx_)){ //specific case for the vertical bars
-              continue; //currently we can't do more here
-            } else if (((tracks_.at(idx)).getResidual() <(tracks_.at(idx_comp)).getResidual() && (track.getCentroid()<vert_bar_start_idx_)) ||
-            ((tracks_.at(idx)).getResidualX() <(tracks_.at(idx_comp)).getResidualX() && (track.getCentroid()>=vert_bar_start_idx_))) {
+            if (((fabs((tracks_.at(idx)).getResidualX() -
+                       (tracks_.at(idx_comp)).getResidualX())) <
+                 0.01)  // it should be equal
+                &&
+                (track.getCentroid() >=
+                 vert_bar_start_idx_)) {  // specific case for the vertical bars
```
bryngemark:
```suggestion
            if (((fabs((tracks_.at(idx)).getResidualX() - 
               (tracks_.at(idx_comp)).getResidualX())) < 0.01) // should be equal
                && (track.getCentroid() >= vert_bar_start_idx_)) {  // specific case for the vertical bars
```

TrigScint/src/TrigScint/TrigScintTrackProducer.cxx
```diff
@@ -683,7 +752,8 @@ void TrigScintTrackProducer::matchXYTracks(
         sx1 = x_conv_factor_ / 2.;  // 1 bar width
         sx2 = sx1;
         x = (x1 + x2) / 2.;
-        sx = fabs(x1 - x2) / 2;  // rely on x precision being one single pad width
+        sx = fabs(x1 - x2) /
+             2;  // rely on x precision being one single pad width
```
bryngemark:
```suggestion
        sx = fabs(x1 - x2) / 2;  // rely on x precision being one single pad width
```

bryngemark (commented):
This looks good overall, most of my comments are on formatting etc. I have two general requests:
1. add more comments explaining what the code does, either at the top of files/functions or as you go along, as needed
2. condense vertical white space a bit. there is no need for blank lines between sets of variables etc, and it makes the code long and hard to read (you have to scroll). NOTE if this white space was added by the formatting script then nevermind

TrigScint/src/TrigScint/TrigScintTrackProducer.cxx
tvami: This I think was the clang tidy, the way out of it is to comment above the line
```suggestion
        // rely on x precision being one single pad width
        sx = fabs(x1 - x2) / 2; 
```

tvami:
I have a new gold incoming in https://github.com/LDMX-Software/ldmx-sw/actions/runs/27962615284
I recommend testing this PR after it's finished

cjbarton151:
@kvarnla I rebased you onto trunk and fixed the immediate issues so that your branch can compile, but your example script for generating the LUT is not working for me. Perhaps you could straighten it out?

tvami:
To fix the DarkBrem:

# 1. sync the submodule reference with trunk
```
git checkout trunk -- SimCore/G4DarkBreM
```
# 2. commit this change
```
git commit -m "checkout trunk -- SimCore/G4DarkBreM"
```
# 3. update the local clone of the submodule to be the trunk reference
```
git submodule update
```

kvarnla:
Thank you @cjbarton151, all example config scripts should be working now. Is it okay that they're still located in  `TrigScint/exampleConfigs/` or should they be moved somewhere within `Tools` also?

cjbarton151:
I don't see any reason why the example configs can't continue living at their current location. However, I'm still getting a seg fault when trying to run the configs over a sample input file - perhaps my sample input file is missing something? Or maybe you can include the script you use to generate the first input file in the example configs as well?

tvami:
@kvarnla did you see my msg here ?
https://github.com/LDMX-Software/ldmx-sw/pull/2070#issuecomment-4790550000

kvarnla:
@cjbarton151 I added the script I used to generate event files under `TrigScint/exampleConfigs/eventGen.py`.

@tvami yes, though when I ran those commands it seems like nothing happens(?), it simply says "nothing added to commit" at the `git commit -m "checkout trunk -- SimCore/G4DarkBreM"` step, which I don't understand since it states there are changes to `SimCore/G4DarkBreM` in this PR. Should they not be included?

cjbarton151:
@kvarnla Thanks for the additional script! I was able to run that and `runClusterstxt.py` without issue, but `runLUTana.py` always returns an empty `LUT.txt` file. After some testing, it seems that, as far as I can tell, `PatternLUTMaker::analyze()` never actually runs. Perhaps you could check this yourself by downloading a fresh copy of ldmx-sw and checking out your branch. `onProcessStart()` and `onProcessEnd()` seem to work just fine.

tvami:
Yay, I'm glad this https://github.com/LDMX-Software/ldmx-sw/pull/2070/commits/faf05440f55e58aa280c137719c486b885965624 worked now! Thanks!

kvarnla:
@cjbarton151 Ah I think the issue is that because `PatternLUTMaker.cxx` is not reading any root events, `PatternLUTMaker::analyze` always reads 0 events. I moved the section previously within `PatternLUTMaker::analyze` to `PatternLUTMaker::onProcessStart` so that LUT.txt isn't empty. `PatternLUTMaker::analyze` is now empty, but it should be working. 

@tvami got it now! Glad it's working :)

cjbarton151 (approved):

cjbarton151:
OK, from a fresh download I was now able to run everything without issue 🙌

I only tested the workflow with the default statistics of 10 events, so my testing didn't include the possibility of rare or unusual events which cause problems. However, from your own testing as part of your semester project, I think we've already covered that.

I added a few more comments to the scripts for clarity, so future users have a bit more guidance. I also added an error handler to `TrigScintTrackProducer.cxx` that catches bad/empty LUT files. But, with that, once this passes the validation tests, I think it's good to go :^)

tvami (approved):

bryngemark (approved):
Thanks all for pulling this across the finish line!

PR #2071, opened 2026-06-10
Updates to the ESA25 TS Decoding Workflow

Layan-Sarayra:

The ESA25 TS decoding scripts had become out of sync with the current ldmx-sw configuration API. This PR is doing two things:
1. Updates the relevant files the decoder calls, including updated imports, logger configuration, input/output file handling, etc.
2. Adds an optional output directory argument to `TrigScint/util/runReformatting_ESA2025sliceTest.sh`. The default is still reserved, but if an output directory is provided, the script instead creates `rootFiles/` there, which avoids writing large decoded ROOT outputs into the shared `/sdf/group/ldmx/data/esa25` directory.

## Check List

I tested the pipeline and it runs successfully:
An example of specified output directory:
`TrigScint/util/runReformatting_ESA2025sliceTest.sh -V -f /sdf/group/ldmx/users/layan/my-ldmx/esa25-ts/run082/fTest_082 -e 50000 -n 30 -r 14 -c channelMap_4modules_14lanes.txt -o /sdf/group/ldmx/users/layan`

Result:
```
[ Process ] 2083  info: Closing file /sdf/group/ldmx/users/layan/rootFiles/decoded_Run_082_20251212_084217_linearize.root

**********************************************************

Done reformatting 50000 events from /sdf/group/ldmx/data/esa25/Run_082_20251212_084217.dat in 174 seconds: 
txt -> dat in 0 s, 
dat -> raw root in 7 s, 
raw root -> QIE in 32 s, 
QIE -> linear charge in 35 s.
linear charge -> testbeam hits in 48 s.
testbeam hits -> clusters in 37 s.
event analysis in 15 s.
hit analysis in 0 s.

*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-*-
```
ADCs and TDCs distribution from the decoded root file:
adcs_dist
tdcs_dist


TrigScint/exampleConfigs/runRawUnpackerZCCM.py
```diff
@@ -1,30 +1,26 @@
...
-RAWfileName=sys.argv[1]
+p = ldmxcfg.Process("raw")
+
+RAWfileName = sys.argv[1]
 p.output_files = [sys.argv[2]]
-log_name=p.output_files[0].replace(".root", "_toRoot.log")
-if len(sys.argv) > 3 :
-    p.maxEvents = int(sys.argv[3])
-else :
-    p.maxEvents = 4_000_000
```
bryngemark, line 13: i see that the now deleted default probably got butchered in a formatting step somewhere, but why shouldn't the maximum number of events be settable here?

bryngemark (approved):
Thanks @Layan-Sarayra for doing this cleanup! It is super helpful since these scripts are not included in a CI (which they probably should be, with some mini input testbeam data file).

I can only comment on the TS side of this but it looks good, pending resolution to the `max_events` question. I approve already now since I'm sure we'll agree on something after you explain.

Packing/include/Packing/LDMXRoRHeader.h
```diff
@@ -48,6 +48,8 @@ class LDMXRoRHeader {
    */
   static const std::unordered_map<std::string, int> CONTRIBUTOR_ID;
 
+  bool valid() const { return valid_; }
```
tomeichlersmith: These `valid` changes and its implementation in LDMXRoRHeader.cxx will conflict with #2066 where @mgignac implemented the same thing. I'm unsure if you both organically came to this solution, but it works, this is just a fair warning to you both depending on which PR gets merged first.

Packing/src/Packing/SingleSubsystemUnpacker.cxx
```diff
@@ -31,22 +31,27 @@ void SingleSubsystemUnpacker::produce(framework::Event& event) {
...
-  while (reader_ and not reader_.eof()) {
+  while (reader_ and not reader_.eof()) { 
     reader_ >> frame_header;
 
     // store location of end-of-frame for skipping this frame
     // if we fail any of the filter checks
     const auto frame_end =
         reader_.tell() + static_cast<std::streamoff>(frame_header.size());
 
-    if (frame_header.channel() != 0 or frame_header.probablyYaml()) {
```
tomeichlersmith: This `channel() != 0` check should go back in. I'm not sure which runs you are looking at, but all of the data in all of the runs was put into channel 0 while some of the runs had the YAML configuration put into channel 255. I like having both checks because, in the future, I would really like to avoid the `probablyYaml` check which just looks for a magic byte which could be mimiced by real data accidentally.

tomeichlersmith (changes requested):
One explicit change and a warning about a future potential merge conflict (depending on whose PR gets merged first).

TrigScint/exampleConfigs/runRawUnpackerZCCM.py
Layan-Sarayra: It's still settable here (lines 11-12) just updated to the current config API

Packing/include/Packing/LDMXRoRHeader.h
Layan-Sarayra: Interesting! I didn’t realize @mgignac was working on the same fix. I made these changes locally a couple of months ago and only now had a chance to open the PR. I’m happy to remove this part from my end if it’s better to keep it in #2066.

tomeichlersmith (approved):
The tests are failing since the `latest` container image used by the tests is updated to v6.0.1 but that image requires a few patches to ldmx-sw https://github.com/LDMX-Software/ldmx-sw/pull/2047

I'm willing to merge this without the tests passing or we can wait until that other PR is merged and re-request the tests here.

PR #2076, opened 2026-06-15
Apply clang-format with new CLANG version

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
This is the format part of https://github.com/LDMX-Software/ldmx-sw/issues/2072

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [x] I ran my developments and the following shows that they are successful.

Issue #2072, opened 2026-06-11
get clang tools working again


tomeichlersmith:
With the updated Ubuntu came an updated Clang with updated format/tidy rules that affect us. This issue is just a reminder that the commits pushed by the clang tool CI will be confusing until we align our `trunk` with the updated Clang, probably through a large format+tidy pass.

Additionally, the CLANG+LTO CI is failing during the Framework tests which is odd and I'm unsure on where its going wrong.

tomeichlersmith:
Apparently, after merge, the CLANG+LTO CI test passed, so I'm going to drop that from my to do list. https://github.com/LDMX-Software/ldmx-sw/actions/runs/27370300408

tomeichlersmith:
Good news, the clang format is very fast even across all of the files (< 4s) and is the only one that actually touches a lot of the files.
```
tom@zuko:~/code/ldmx/ldmx-sw$ time just format-cpp

real	0m3.787s
user	0m0.209s
sys	0m0.040s
tom@zuko:~/code/ldmx/ldmx-sw$ git diff --name-status | wc -l
178
```

The clang-tidy took ~40min to go through all of the files and produced changes in the following
```
# I did a `git restore` before this so these changes are ONLY from the clang-tidy
tom@zuko:~/code/ldmx/ldmx-sw$ git status
On branch 2072-trunk-job-fail-if-change-made
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   Biasing/include/Biasing/EcalDarkBremFilter.h
	modified:   Biasing/include/Biasing/MidShowerDiMuonBkgdFilter.h
	modified:   Biasing/include/Biasing/MidShowerNuclearBkgdFilter.h
	modified:   Biasing/include/Biasing/TargetDarkBremFilter.h
	modified:   Biasing/src/Biasing/Utility/StepPrinter.cxx
	modified:   Conditions/include/Conditions/GeneralCSVLoader.h
	modified:   Ecal/src/Ecal/EcalTrackFinderProcessor.cxx
	modified:   SimCore/include/SimCore/G4User/UserAction.h
	modified:   SimCore/include/SimCore/MagneticFieldStore.h
	modified:   Tools/HLS_Arbitrary_Precision_Types/include/ap_fixed_base.h
	modified:   Tools/HLS_Arbitrary_Precision_Types/include/ap_impl/ap_common.h
[...]

PR #2077, opened 2026-06-15
fix clang format/tidy attempt ii

tomeichlersmith:

This is work towards #2072 but does not resolve it yet.

While here, I update the `UserActions` to have the correct naming convention for their functions. Mainly as a test to see how the clang tools applied only to the change files works. I'm unsure if these changes conflict with the format pass in #2076 

The actual CI changes are adding in `set -e` to the `run:` commands to see if that's why they aren't failing (showing a red x) when clang finds format/tidy errors.

### DIFF
--- a/Biasing/include/Biasing/Utility/PartialEnergySorter.h
+++ b/Biasing/include/Biasing/Utility/PartialEnergySorter.h
@@ -33,7 +33,7 @@ namespace utility {
  * the threshold.
  *
  * ## Necessities for Filters Run in Sequence
- * - *If* your derived class re-defines ClassifyNewTrack, you should avoid
+ * - *If* your derived class re-defines classifyNewTrack, you should avoid
  *   changing the classification and return the current track classification.
  * - *If* your derived class re-defines stepping, you should avoid
  *   change the track status.
@@ -62,7 +62,7 @@ class PartialEnergySorter : public simcore::UserAction {
    *
    * @param[in] event unused
    */
-  void BeginOfEventAction(const G4Event* event) override;
+  void beginOfEventAction(const G4Event* event) override;
 
   /**
    * Classify a "new" track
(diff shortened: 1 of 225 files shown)


Biasing/include/Biasing/Utility/PartialEnergySorter.h
```diff
@@ -62,7 +62,7 @@ class PartialEnergySorter : public simcore::UserAction {
    *
    * @param[in] event unused
    */
-  void BeginOfEventAction(const G4Event* event) override;
+  void beginOfEventAction(const G4Event* event) override;
```
tvami, line 65: Oh I thought these were GEANT functions that we just override and that's why we had to keep them not follow the convention

tomeichlersmith: You'll find the Geant4 overrides in SimCore/G4User - our `UserAction` is custom and is an attempt to unify Geant's stepping, tracking, stacking, event, and run actions into one interface. We could choose to undo this change and keep it as an exception to our naming rules, but I'd prefer to avoid that.

tvami: OK sounds good

tvami:
Yeah so https://github.com/LDMX-Software/ldmx-sw/pull/2077/changes/ed3dfd4a92700f89c97d4d081ac869ed8c2925db will conflict now

tomeichlersmith:
So the merge conflicts still present themselves, but since formatting is a deterministic procedure, we can format on this branch (done by the clang-format CI job) and then default any conflict to "ours" which is the branch we are on.

```
git switch trunk
git pull # make sure trunk is updated
git switch -
git pull # make sure we get format from CI
git merge trunk --strategy-option=ours
```

tomeichlersmith (commented):
ok, I think this is ready

tvami (approved):

PR #2080, opened 2026-06-17
2072 clang tidy

tomeichlersmith:

I think with some additional testing I was able to resolve a part of #2072 and get the scheduled jobs on `trunk` to fail **if** changes would be made to source files by format/tidy/lint.

The requirement that changes be made to source files is one that I'm not happy with. I'm hoping we can enable more clang-tidy checks and some of them do not have a "auto-fix" option that change the file, but - for now at least - this will at least alert us when there are "auto-fixes" available for files on `trunk`.

I made a small change to a header file to trigger the apply clang tools on a changed file to see how it performs in CI. It should update the destructor to use the correct style for the local variable name `lower_case` instead of `camelBack`.

### DIFF
--- a/.github/workflows/apply-clang-tools.yml
+++ b/.github/workflows/apply-clang-tools.yml
@@ -32,5 +32,5 @@ jobs:
         just format-cpp-diff
         just tidy-cpp-diff
         echo $?
-        git diff
+        git diff --exit-code
         echo $?
--- a/.github/workflows/apply-clang-tools.yml
+++ b/.github/workflows/apply-clang-tools.yml
@@ -1,5 +1,6 @@
 name: Apply clang tools
 on:
+  push: # TEMP, do not merge
   pull_request:
       types: [opened, ready_for_review]
   workflow_dispatch:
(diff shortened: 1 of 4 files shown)


.github/workflows/apply-clang-tools.yml
```diff
@@ -25,23 +25,19 @@ jobs:
...
-        just tidy-cpp-all
-        git diff # fail job if format/tidy made a change
-    - name: auto-commit changes
-      if: github.ref != 'refs/heads/trunk'
-      run: |
         git config user.name 'github-actions[bot]'
         git config user.email 'github-actions[bot]@users.noreply.github.com'
         git add .
-        git commit -m "Apply clang-format and clang-tidy" || exit 0
+        git commit -m "apply clang-format and clang-tidy" || exit 0
```
tvami, line 36: I like doing these as uppercase, it's a "semi sentence" to me... like bullet points in a talk, that's also uppercase but no full stop in the end. But ok that's just style

tvami (approved):

PR #2083, opened 2026-06-18
Full simulation with Genie

### LINKED ISSUE #2059: Full simulation with HepMC inputs
Right now the GENIE (and other HepMC-based) simulations start at the target and do not include upstream electron. 

I'd like a situation similar to how the signal is simulated: we have an incoming electron beam, goes into the target volume, picks the kinematics of the electron and produces the EN event like that. With that the GENIE outputs would not be primaries and could be labeled with processType = `electronNuclear`

This is needed for full realism.

tvami:

A big change where Genie is under GEANT and similarly to the signal we lead the 8 GeV primary generator thru the upstream material, and then hand the electron at the correct energy at the target volume to Genie. With this these particles are now secondaries and can be one-to-one compared to the GEANT based EN processes.

Resolves #2059 

## Check List

Source is not at 0,0,0 anymore:

Screenshot 2026-06-17 at 19 29 33

Mult:
Screenshot 2026-06-17 at 19 41 17

Tagger DQM is filled now:
Screenshot 2026-06-17 at 19 42 34

Implementation is such that the old way of doing it is still possible


tvami:
/run-validation

 * this I expect to differ quite a lot

.github/validation_samples/target_genie/config.py
```diff
@@ -4,34 +4,40 @@
...
+# silently recompute all cross sections on the fly and the job appears to hang.
+# See .github/validation_samples/target_genie/README.md.
+my_sim.genie_nuclear.spline_file = (
+    f"{os.environ['CI_DATA']}/target_genie/gxspl_emode_{my_sim.genie_nuclear.tune}.xml"
+)
+my_sim.genie_nuclear.message_threshold_file = "Messenger_ErrorOnly.xml"
 
-my_sim.generators = [genie]
+# Bias EN cross-section in target region so interactions occur at usable rate
+my_sim.biasing_operators = [ElectroNuclear(volume="target_region", factor=1e6)]
```
therwig, line 40: It is fantastic to have an implementation of this capability in ldmx-sw - thank you!  (Will be great for many reasons, but (selfishly) with the upstream electrons now included we will finally have the TS hits necessary to implement a real electron pT trigger simulation!)

I am curious to understand this part a little better; the eN 'event generation' part is now handled by genie, but the biasing (and interaction cross section / probability) in each of the geometry materials is still decided by g4? 
Since I read some of the code but not all (yet..?) I am thinking there are a number of potential subtleties like this and it would be very useful to have you present the work in a swan meeting (and/or sw-dev) to discuss in detail. But I'm sure you were already planning this :)

tvami: > But I'm sure you were already planning this :)

I was planning on having this merged quickly, cut a release and have a high stat sample and compare it to the GEANT based high stat sample, and show that at the next SWAN :D 

> a real electron pT trigger simulation

This sounds amazing!

> the eN 'event generation' part is now handled by genie, but the biasing (and interaction cross section / probability) in each of the geometry materials is still decided by g4?

The cross section is also calculated by Genie per material, per isotope, at the correct energy, and then to keep the event, it is scaled by the biasing factor.

SimCore/src/SimCore/GenieElectroNuclearProcess.cxx
```diff
@@ -0,0 +1,604 @@
...
+  // 64 is arbitrary but distinct from standard EM subtypes and the DarkBrem
+  // subtype (63).
+  SetProcessSubType(64);
+
+  // Read configuration (targets/abundances are optional — empty means
+  // auto-discover)
+  static const std::vector<int> no_targets;
+  static const std::vector<double> no_abundances;
+  targets_ = params.get<std::vector<int>>("targets", no_targets);
+  abundances_ = params.get<std::vector<double>>("abundances", no_abundances);
```
tomeichlersmith: I would have guessed that passing `{}` instead of `no_targets` would have worked. If that didn't work, I would consider that a bug with the Parameters class.

SimCore/src/SimCore/RunManager.cxx
```diff
@@ -197,6 +201,21 @@ void RunManager::TerminateOneEvent() {
   };
 
   reactivate_dark_brem(G4Electron::Definition()->GetProcessManager());
+
+  // Reactivate GENIE electronNuclear process if it was deactivated
+  // (only_one_per_event mode) — same pattern as DarkBrem above
```
tomeichlersmith: Since they are both looping through the electron process manager, maybe just have one for loop and two `if` statements inside?

I also have no idea why I did this lambda-function-which-is-immediately-called pattern (maybe to leave the possibility open to use the same loop over other particles that also have the same process?). I now think it is overly complicated and we can drop it.

tomeichlersmith (changes requested):
Honestly, this looks great. My only requests are a few technical things.

.github/validation_samples/target_genie/config.py
therwig: Thanks, that makes sense!  OK, I will sign off and look forward to hearing more :)

therwig (approved):

SimCore/src/SimCore/GenieElectroNuclearProcess.cxx
tvami: I tried that first, the issue is that it's a vector so it doesnt know if it's a vector of `double`s or `int`s if I just do the `{}`

SimCore/src/SimCore/RunManager.cxx
tvami: Haha ok will check this!

SimCore/src/SimCore/GenieElectroNuclearProcess.cxx
tomeichlersmith: But the template argument to `get` informs the compiler if its a `double` or `int` vector, can you share the compiler error? I'm fairly certain it should work since we are doing something similar with `std::vector<std:string>` in 

https://github.com/LDMX-Software/ldmx-sw/blob/bd8057beb458032cc0dada7027db9e25f1c79569/Framework/src/Framework/Process.cxx#L45

tvami:
While I work on Tom's suggestion, here is the KS test summary (with my own importance filter so the file doesnt get bigger than the gihub max for uploads)

One change is that Genie was considering everybody to be primary, now that's not the case anymore.

And then ofc we now fill the tagger plots, the target volume is bigger, etc.

Seems like the electron is softer than earlier, but that might just be statistics?

The electron is more fiducial in the ECAL.

Figure 140: less zero track events makes me happier

Figure 159: this confuses me tho... @tomeichlersmith any input on this?

[iss2083_target_genie_recon_validation_plots.pdf](https://github.com/user-attachments/files/29102036/iss2083_target_genie_recon_validation_plots.pdf)

tomeichlersmith:
Now that we aren't producing the GENIE event inside the target, we are susceptible to upstream objects softening the electron and causing it to curve further towards negative x in the magnetic field. I'm surprised if that affect causes the full shift of the distribution, so I'm guessing that our settings for the beam are a little off.

I originally got the settings for the electron by shooting a positron backwards from the target after removing the upstream detector components. https://github.com/LDMX-Software/ldmx-sw/pull/1396 but there have been many changes since then so we may need to update those settings?

tvami:
> we may need to update those settings?

Yeah prob we should check again. I see the config to run, but then how do you extract the final number? Can you maybe put the analysis code too? I also wonder if we should make that into a unit test?

tomeichlersmith:
The analysis I used way back then is included in the tar-ball I uploaded, I'm trying it out now.

tomeichlersmith:
alright, kinda quick and dirty, I did not update the legend, but the beam still seems to be well centered on the target. edit: this is with v14, running with v15 next.

position-and-momentum-trunk-1395-comp

I'm now unsure on where the large deflection is coming from.

tomeichlersmith:
yea, I'm much more confident its not the configuration of the generator now
position-and-momentum-trunk-1395-comp

tvami:
OK I think the beam smear actually makes sense, you do the beamspot from around the origin as coming from the primary. But this study was great to show that v15 didnt change much, thanks Tom!

SimCore/src/SimCore/GenieElectroNuclearProcess.cxx
tvami: This indeed works

tomeichlersmith (approved):

PR #2084, opened 2026-06-18
VST Ecal support box + global-z of the HCAL

### LINKED ISSUE #1880: ESA full slice test geometry

tvami:

Resolves #1880

## Check List

It's not 100% perfect ECAL, we had the motherboards in a special shape that I just approximate as a square. 
Same for the support planes. But this is much closer to reality than what we had earlier. 
And with this the HCAL is positions right.

I had to change quite some much to make the g4-vis work on POD from my Mac (hopefully works on other machines too, tho I understand from CJ that linux based ssh didnt have any issues)

I also changed the Si color to blue.

Screenshot 2026-06-17 at 18 38 22
Screenshot 2026-06-17 at 18 38 27
Screenshot 2026-06-17 at 18 38 30

Screenshot 2026-06-17 at 19 49 24

Screenshot 2026-06-17 at 19 50 25


Detectors/data/ldmx-reduced-v3/ecal_support_box_solids.gdml
```diff
@@ -0,0 +1,28 @@
+<!--
+  ECAL support-box part solids (2025 Winter test beam), ENTITY-included into
+  ecal.gdml <solids>. Generated by step_tools/generate_support_box.py
+  (do not hand-edit). See the structure fragment for the assembly.
+-->
+<box lunit="mm" name="front_0_box" x="457.2" y="457.2" z="0.8"/>
+<box lunit="mm" name="front_1_box" x="457.2" y="457.2" z="0.8"/>
```
cjbarton151, line 7: Here and on subsequent lines: these two solids are functionally identical, and it's OK to just have one and use it multiple times for multiple volumes. On the other hand, it doesn't really hurt anything to define multiple solids with the same dimensions, either.

Detectors/data/ldmx-reduced-v3/ecal.gdml
```diff
@@ -1,6 +1,8 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no" ?>
 <!DOCTYPE gdml [
 <!ENTITY constants SYSTEM "constants.gdml">
+<!ENTITY support_box_solids SYSTEM "ecal_support_box_solids.gdml">
```
cjbarton151, line 4: There are multiple ways to call GDML files within GDML files. While this isn't typically the method our particular group uses to call assemblies, it should be perfectly valid and I don't see the harm in it.

cjbarton151 (approved):

cjbarton151:
Also you were right, that really is quite a workaround required for Mac OS! 😂

Detectors/data/ldmx-reduced-v3/ecal_support_box_solids.gdml
tvami: OK in that case, I'd prefer to keep as is!

justfile
```diff
@@ -413,6 +413,52 @@ rootbrowse FILE:
...
+    # The denv/apptainer container's HOME is the workspace and it does NOT mount
+    # your real ~/.Xauthority, so the container cannot find your X11 cookie. Drop a
+    # wildcard (FamilyWild) cookie for the current $DISPLAY into the workspace, where
+    # the container picks it up automatically as ~/.Xauthority. FamilyWild matches
+    # regardless of the container's (differing/empty) hostname.
+    #
+    # This cluster forwards X through more than one host, each with a different
+    # cookie for the same display number, so pick the cookie that sshd stored for
+    # *this* host -- that is the one the live
+    # forwarding validates against. Fall back to the plain $DISPLAY lookup.
```
tomeichlersmith, line 446: I have no changes to request except that this appears to resolve a long standing issue for `denv` that I could not figure out. https://github.com/tomeichlersmith/denv/issues/125 Thank you!

tomeichlersmith (approved):

justfile
tvami: Haha nice!

tvami:
For documentation: here is the comparison of what failed the KS:
[pr2084_reduced_ldmx_recon_validation_plots.pdf](https://github.com/user-attachments/files/29098675/pr2084_reduced_ldmx_recon_validation_plots.pdf)

The significant change is in the HCAL plots, Section 1.7, and I guess the Al support structure added does shield the HCAL more, max PE is almost halved. I dont think this can be cause by the HCAL being closer, right? Although it was a significant distance and the electron could interact with the Air in between... (but I think it's the support box)

cjbarton151:
It might be useful to check the truth information for the x distribution of energy depositions in the quad bars, as well as activity in the first bar vs the bars behind it. If this is a quenched shower then I would expect the hits in the current version to be spread out much more along the quads, and possibly to be reduced by even just two or three layers of quad bar material.

PR #2087, opened 2026-06-19
complete clang-tidy identifier namings

### LINKED ISSUE #2072: get clang tools working again

tomeichlersmith:

This resolves #2072 by completing the to-do's from clang-tidy. There are still warnings emitted by clang-tidy, but they reference dead code in `EventDisplay` or Acts internals that we don't have access to.
The periodic `clang-tidy` run over all of the files will only fail if clang-tidy is able to fix the warnings/errors it detects which is not always the case, so we should still periodically check the output from that weekly run.

### DIFF
--- a/.github/workflows/pr_validation.yml
+++ b/.github/workflows/pr_validation.yml
@@ -187,7 +187,7 @@ jobs:
 
   post-validation:
     needs: [check-comment, validation]
-    if: always() && (github.event_name == 'pull_request' || (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '/run-validation')))
+    if: (!cancelled()) && (github.event_name == 'pull_request' || (github.event_name == 'issue_comment' && github.event.issue.pull_request && contains(github.event.comment.body, '/run-validation')))
     runs-on: ubuntu-latest
     steps:
       - name: Resolve PR number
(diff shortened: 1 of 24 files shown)


tvami (approved):
Looks good! But maybe we should merge it after the gold of v4.8.1 is done?

tomeichlersmith:
That works for me :saluting_face:

tvami:
> Timing regression for inclusive: new=7109s vs gold=3573s (+98%, tolerance 10%)

I have hard time believing this is true

tvami:
/run-validation 

* let's try again

tvami:
> > Timing regression for inclusive: new=7109s vs gold=3573s (+98%, tolerance 10%)
> 
> I have hard time believing this is true

We should also fix the fact that while this msg is here the CI still flagged it as "PASS"

tvami:
Not sure what's up with the inclusive but it's coming up here too
https://github.com/LDMX-Software/ldmx-sw/pull/2091#issuecomment-4773946227

tomeichlersmith:
Since we don't control the runners, we have no idea how burdened they are at any one time. I am not surprised the timing is off by so much and is switching between different validation samples because our jobs are probably sharing the machine hosting the runner with other runners. If that machine is slowed for _any_ reason, our job will be slowed down.

While I don't think we should stop keeping track of the time, I don't think it should be a failure condition unless we have our own runners that we control so we can be certain performance issues are not the fault of the host.

tvami:
Yeah ok, I think we can merge this

PR #2101, opened 2026-07-27
upgrade ACTS from v36 to v47

### LINKED ISSUE #2100: upgrade ACTS from v36 to v47
It's been about two years since we re-upped the ACTS release.  This issue if for updating from v36 to v47 (current as of 7/24/26).

bloodyyugo:
Upgrade ACTS versions for 36-->47.  Closes #2100  

----------------------------------------------------------
Algorithm changes that may effect tracks: 

- CKF branch handling
    The CKF received improvements to outlier rejection and branch pruning
    across these 11 major versions. This can shift the
    efficiency/fake-rate tradeoff even with identical inputs.

- GainMatrixSmoother → smoothTrack()
    The Kalman backward (RTS) smoother was rewritten as a free function.
    The math is identical in principle, but the implementation details
    (handling of degenerate states, track-end boundary conditions, numerical
    precision) may differ. Most likely source of small systematic shifts in
    momentum resolution or pull distributions.

-   EigenStepper / RKN propagator
    Step-size control and Jacobian transport have had stability improvements.
    Effects appear mainly in track-to-truth residuals for long extrapolations
    (e.g., tagger perigee → ECAL scoring plane).

-   GSF Bethe-Heitler handling
    AtlasBetheHeitlerApprox was renamed PolynomialBetheHeitlerApprox — the
    polynomial coefficients are the same. However, the GSF received active
    development between v36 and v47 in component reduction strategies, weight
    re-normalization, and handling of near-degenerate Gaussian mixtures. For
    LDMX, where the GSF drives electron momentum measurement, this is the
    highest-risk area for physics-level differences vs. trunk.

-------------------------------------------------------------
Interface changes etc: 

- In v46, the calibrator and measurementSelector
delegates moved out of CombinatorialKalmanFilterExtensions into a new
Acts::TrackStateCreator object. The SourceLinkAccessorDelegate also moved there.

- DetectorElementBase -> Acts::SurfacePlacementBase.
- transform() override -> localToGlobalTransform() (matches new base class).
- surface->assignDetectorElement() -> surface->assignSurfacePlacement().
- thickness() is no longer a pure virtual in the base; override removed.
- isSensitive() added as a required override (returns true for sensor elements).
- Acts::ViewConfig now uses C++ designated initializers: {.color = {r, g, b}}.
- Added tagger_start_surface_ at ACTS x=-617mm (1mm inside the tagger
  volume outer boundary at x=-618mm, 1.5mm upstream of the L1 doublet at
  x=-615.5mm).
- Recoil tracking volume extended upstream so its low-x edge is at x=-1mm,
  placing the target plane (x=0) clearly inside the volume for the CKF Navigator
  (previously the target was on or outside the volume boundary, causing Navigator
  failures for recoil tracks).

- BetheHeitlerApprox changed from a local stack object to a shared_ptr:
    Old: BetheHeitlerApprox bh = Acts::makeDefaultBetheHeitlerApprox();
    New: auto bh = std::make_shared<Acts::AtlasBetheHeitlerApprox>(
             Acts::makeDefaultBetheHeitlerApprox());

- GsfOptions.propagatorPlainOptions now requires an explicit cast:
    gsf_options.propagatorPlainOptions =
        static_cast<Acts::PropagatorPlainOptions>(propagator_options);

- A separate GsfExtrapPropagator (EigenStepper + VoidNavigator) was added for
  post-fit track extrapolations (to target, ECAL). VoidNavigator is required
  because the ECAL scoring plane is outside the tracking geometry volume and the
  standard Navigator would stop propagation at the volume boundary.

- GainMatrixSmoother removed; Acts::smoothTrack() used instead, with its return
  value checked. 

- Acts::FreeSquareMatrix renamed (TrackingUtils.cxx)
  Old: Acts::FreeSquareMatrix
  New: Acts::FreeMatrix
  (defined in Acts/Definitions/TrackParametrization.hpp, same as before)

- Acts::GeometryContext default constructor made private
  The no-arg Acts::GeometryContext() constructor is private in v47. Code that
  held Acts::GeometryContext as a member and relied on implicit default
  initialization would fail to compile.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,6 +71,31 @@ find_package(
              atomic
              regex)
 
+# Find Eigen3 (required by ACTS)
+find_package(Eigen3 CONFIG REQUIRED)
+
+# GNUInstallDirs must be included before ACTS is configured as a sub-project;
+# ACTS only calls include(GNUInstallDirs) when it is the top-level project, so
+# CMAKE_INSTALL_INCLUDEDIR would otherwise be empty and its install() calls fail.
+include(GNUInstallDirs)
+
+# Find or fetch ACTS - must be at top level so both Ecal and Tracking can use it
+find_package(Acts 46.0.0 QUIET)
+if (NOT Acts_FOUND)
+  message(STATUS "Did not find Acts >= 46.0.0, downloading and compiling v46.8.1")
+  include(FetchContent)
+  FetchContent_Declare(
+      Acts
+      URL https://github.com/acts-project/acts/archive/refs/tags/v46.8.1.tar.gz
+      URL_HASH MD5=671ef266b008d70e5cdbe2b829f09881
+      DOWNLOAD_EXTRACT_TIMESTAMP TRUE
+  )
+  FetchContent_MakeAvailable(Acts)
+  FetchContent_GetProperties(Acts)
+  include_directories(SYSTEM "${Acts_SOURCE_DIR}/Core/include")
+else()
+  message(STATUS "Found Acts ${Acts_VERSION}")
+endif()
 
 # FunctionalCoreTest.cxx doesnt comply with clang-tidy but that's ok
 set_source_files_properties(Framework/test/FunctionalCoreTest.cxx PROPERTIES COMPILE_OPTIONS "-Wno-null-dereference")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,14 +80,14 @@ find_package(Eigen3 CONFIG REQUIRED)
 include(GNUInstallDirs)
 
 # Find or fetch ACTS - must be at top level so both Ecal and Tracking can use it
-find_package(Acts 46.0.0 QUIET)
+find_package(Acts 47.0.0 QUIET)
 if (NOT Acts_FOUND)
-  message(STATUS "Did not find Acts >= 46.0.0, downloading and compiling v46.8.1")
+  message(STATUS "Did not find Acts >= 47.0.0, downloading and compiling v47.0.0")
   include(FetchContent)
   FetchContent_Declare(
       Acts
-      URL https://github.com/acts-project/acts/archive/refs/tags/v46.8.1.tar.gz
-      URL_HASH MD5=671ef266b008d70e5cdbe2b829f09881
+      URL https://github.com/acts-project/acts/archive/refs/tags/v47.0.0.tar.gz
+      URL_HASH MD5=833c02edee49827f25be74c80d8ee654
       DOWNLOAD_EXTRACT_TIMESTAMP TRUE
   )
   FetchContent_MakeAvailable(Acts)
--- a/Tracking/include/Tracking/Reco/TrackExtrapolatorTool.h
+++ b/Tracking/include/Tracking/Reco/TrackExtrapolatorTool.h
@@ -85,17 +85,8 @@ class TrackExtrapolatorTool {
 
     if (result.ok())
       return *result->endParameters;
-    else {
-      std::cout << "[TET DIAG BTP] propagation failed: "
-                << result.error().message() << "\n";
-      std::cout << "[TET DIAG BTP]   start pos=("
-                << pars.position(gctx_).transpose() << ")\n";
-      std::cout << "[TET DIAG BTP]   direction=("
-                << pars.direction().transpose() << ")\n";
-      std::cout << "[TET DIAG BTP]   target surface type="
-                << target_surface->name() << "\n";
+    else
       return std::nullopt;
-    }
   }  // end of extrapolate()
 
   /** Method to extrapolate to a target surface given a track
@@ -131,16 +122,6 @@ class TrackExtrapolatorTool {
         Acts::TrackExtrapolationStrategy::firstOrLast);
 
     if (!stateResult.ok()) {
-      // Temporary diagnostic: print why and dump type flags of all states
-      std::cout << "[TET DIAG] findTrackStateForExtrapolation failed: "
-                << stateResult.error().message()
-                << "  nStates=" << track.nTrackStates() << "\n";
-      for (const auto& tsDiag : track.trackStatesReversed()) {
-        std::cout << "  isMeas=" << tsDiag.typeFlags().isMeasurement()
-                  << " isHole=" << tsDiag.typeFlags().isHole()
-                  << " hasFiltered=" << tsDiag.hasFiltered()
-                  << " hasSmoothed=" << tsDiag.hasSmoothed() << "\n";
-      }
       return std::nullopt;
     }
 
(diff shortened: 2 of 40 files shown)


CMakeLists.txt
```diff
@@ -71,6 +71,31 @@ find_package(
...
+
+# GNUInstallDirs must be included before ACTS is configured as a sub-project;
+# ACTS only calls include(GNUInstallDirs) when it is the top-level project, so
+# CMAKE_INSTALL_INCLUDEDIR would otherwise be empty and its install() calls fail.
+include(GNUInstallDirs)
+
+# Find or fetch ACTS - must be at top level so both Ecal and Tracking can use it
+find_package(Acts 47.0.0 QUIET)
+if (NOT Acts_FOUND)
+  message(STATUS "Did not find Acts >= 47.0.0, downloading and compiling v47.0.0")
```
tomeichlersmith, line 85: While I like this fallback in order to support older versions of the image, I think this PR should have a companion PR where the development image is updated to Acts v47. Is it backwards compatible? Could Acts v47 be used to compile what's on `trunk` right now?

tomeichlersmith (commented):
> For LDMX, where the GSF drives electron momentum measurement, this is the
highest-risk area for physics-level differences vs. trunk.

Do you plan to do an initial study of these differences and post them here before the PR is merged? Or do you think it is too detailed (and the differences are likely improvements) so it will wait for future work?

CMakeLists.txt
tomeichlersmith: (These questions will just motivate how we document the development image and not necessarily whether we update the Acts within it.)

bloodyyugo: I've just been looking at the validation plots (and really just the signal sample).  The biggest differences I see:  

- tagger:  the uncertainties for d0/z0/phi are smaller with v47, residuals the same, so the pulls are also broader.  I don't think this is real...it's probably due to the change in the "origin" point which was moved closer so that it is in tracker volume.  The pulls at the target X/Y pass the KS tests (though X vs n_hits fails but for some reason this plot only goes up to 12 hits, which must just be a bug).

- recoil:  we get KS failurs for some dup and fake plots and this is expected (though honestly they look ok to me).  Also see failures for target position X (both truth and reco?) but they are pretty close by eye.  The locations projected to the ecal fail and look bad (even the gold plots look bad, now they look worse).  We will have to figure this out.  

-  GSF:  apart from the changes that feed down from the CKF tracks, the q/p error is much larger than before which effects the pulls (I think making them better).  

-  ECal Tracking!   I forget to look at these earlier!  There are quite a few failures in EcalSPTrackCompare (particularly the deltas) and EcalTrackAnalyzer.  Most of the plots fail actually.

bloodyyugo: On your question on the image, no trunk as of now will not compile with v47...interface and class-name changes.  What is the right thing to do here?  I can't remember what we did when we updated to v36....

tomeichlersmith: It will just be nice to document it. If the answer is that the developer needs to either (a) merge their updates with the updated `trunk` or (b) pin their environment on an old image, that's okay I just wanted to know.

bloodyyugo:
The most recent commit fixes the Ecal tracking.  There were two issues, first the starting surface needed to be in the tracking volume (same thing we saw for recoil and tagger); also the surfaces given to ACTs didn't have a "sensitive" surface defined, so the tracking didn't know which surfaces the  hits were supposed to be on (and what it should be swimming track candidates to).  

So now we see tracks for ECal tracking...a few of the plots still fail the validation tests but looking at them they all look very close.  

For the trackers, the only thing that sticks out to me is the ecal_pulls_of_loc0 (so, x position pull at the ecal).  It's much broader for acts47 than the trunk.  I've attached the plot RecoilGASDQM_ecal_Pulls_of_loc0.pdf.  

Screenshot 2026-07-29 at 2 16 09 PM


Note, I think it's best to look at the "GAS" plots (i.e. after getting rid of highly redundant tracks).  The raw ones, e.g. "RecoilDQM" I expect to be different because the  changes to "outlier rejection and branch pruning" while there have been many changes to the GSF algorithm.  For the recoil GAS tracks, only plots related to the loc0 pulls fail.  For the tagger, the errors on the track parameters (not the parameter values themselves) change a bit so the _error and _pull plots fail the KS tests.  It is a bit weird that this only shows up in the tagger and not the recoil though.

CMakeLists.txt
tomeichlersmith: The companion PR in dev-build-context is https://github.com/LDMX-Software/dev-build-context/pull/198
I have not looked at the necessary patch files for making older versions of ldmx-sw compile with this updated image which I do believe is necessary to avoid confusion within our developer community.

Tracking/include/Tracking/Reco/TrackExtrapolatorTool.h
```diff
@@ -112,88 +109,54 @@ class TrackExtrapolatorTool {
...
-    // get first and last track state on surface
-    if (debug_)
-      std::cout
-          << "[TrackExtrapolatorTool]   Getting outermost track state...\n";
-    auto outermost = *(track.trackStatesReversed().begin());
-    if (debug_)
-      std::cout
-          << "[TrackExtrapolatorTool]   Getting innermost track state...\n";
-    auto begin = track.trackStatesReversed().begin();
-    std::advance(begin, track.nTrackStates() - 1);
```
tvami, line 124: this `advance` is not needed anymore?

Tracking/python/full_tracking_sequence.py
```diff
@@ -164,7 +164,7 @@ def tagged(name):
             depletion_voltage=70.0,
             noise_electrons=1000.0,
             threshold_electrons=3000.0,
-            out_raw_collection=tagged("TaggerSimHits"),
+            out_raw_collection=tagged("TaggerSimSiStripHits"),
```
tvami, line 167: This means moving to the version that goes thru the correct digitization?

Tracking/include/Tracking/Reco/TrackExtrapolatorTool.h
bloodyyugo: Acts::findTrackStateForExtrapolation automatically finds the track state closest to the surface we are extrapolating, so this whole snippet of code isn't needed.

Tracking/python/full_tracking_sequence.py
bloodyyugo: I think there were some naming clashes so this (and for Recoil) fixes it.

tvami (approved):
Thanks @bloodyyugo !

tomeichlersmith (approved):

tvami:
Hmm should have paid more attention to the CI
https://github.com/LDMX-Software/ldmx-sw/issues/2114

Issue #2104, opened 2026-08-05
memory leaks detected


tomeichlersmith:
**Describe the bug**
In a recent ASAN/UBSAN run https://github.com/LDMX-Software/ldmx-sw/actions/runs/30274309712 the following memory leaks were detected.

```
==16==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 304 byte(s) in 1 object(s) allocated from:
    #0 0x7f0a1643b73f in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:109
    #1 0x7f0a15a2dc09 in new_ldmxcLcLRunHeader /home/runner/work/ldmx-sw/ldmx-sw/build/Framework/FrameworkDict.cxx:280

Indirect leak of 41 byte(s) in 1 object(s) allocated from:
    #0 0x7f0a1643b73f in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:109
    #1 0x55f73c1cfe41 in std::__new_allocator<char>::allocate(unsigned long, void const*) /usr/include/c++/15/bits/new_allocator.h:151
    #2 0x55f73c1cfe41 in std::allocator<char>::allocate(unsigned long) /usr/include/c++/15/bits/allocator.h:203
    #3 0x55f73c1cfe41 in std::allocator_traits<std::allocator<char> >::allocate(std::allocator<char>&, unsigned long) /usr/include/c++/15/bits/alloc_traits.h:614
    #4 0x55f73c1cfe41 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_S_allocate(std::allocator<char>&, unsigned long) /usr/include/c++/15/bits/basic_string.h:142
    #5 0x55f73c1cfe41 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::_M_create(unsigned long&, unsigned long) /usr/include/c++/15/bits/basic_string.tcc:164
```

This happened after merging #2101 so maybe they are related to the changes from that PR? Its not obvious if they are...

Issue #1894, opened 2025-12-04
"Potential overlap in geometry"


tvami:
**Is your feature request related to a problem? Please describe.**

With https://github.com/LDMX-Software/ldmx-sw/pull/1893
now we see the diagnostic msg from GEANT4.

For example
https://productionresultssa7.blob.core.windows.net/actions-results/d47b5319-c12b-4ba9-9ce9-c3cc2f60f991/workflow-job-run-eb76e797-f017-5917-a82e-f42298cd97d5/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-12-04T15%3A22%3A16Z&sig=KzEh44AYSxTkzE0DJckdk7W54iaVG0HO0e3%2Bf4ijjDA%3D&ske=2025-12-05T01%3A57%3A36Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-12-04T13%3A57%3A36Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-12-04T15%3A12%3A11Z&sv=2025-11-05

Show a LOT of 
```
025-12-04T06:47:54.5170515Z [ GEANT4 ] 1776  warn: 
2025-12-04T06:47:54.5171239Z -------- WWWW ------- G4Exception-START -------- WWWW -------
2025-12-04T06:47:54.5171802Z *** G4Exception : GeomNav1002
2025-12-04T06:47:54.5172254Z       issued by : G4Navigator::ComputeStep()
2025-12-04T06:47:54.5172976Z Track stuck or not moving.
2025-12-04T06:47:54.5173444Z           Track stuck, not moving for 10 steps
2025-12-04T06:47:54.5174576Z           in volume -av_1_impr_6_nohole_motherboard5_assembly_pv_0- at point (259.873,193.43,312.532) (local point (79.3109,1.666,397.533))
2025-12-04T06:47:54.5175669Z           direction: (-0.398084,-0.626739,0.669871) (local direction: (0.398084,0.669871,-0.626739)).
2025-12-04T06:47:54.5176356Z           Potential geometry or navigation problem !
2025-12-04T06:47:54.5176954Z           Trying pushing it of 1e-07 mm ...Potential overlap in geometry!
2025-12-04T06:47:54.5177355Z 
2025-12-04T06:47:54.5177524Z *** This is just a warning message. ***
[...]

PR #2108, opened 2026-08-20
Revive Sim from Scoring Plane

tomeichlersmith:

This does not address an open issue. It brings back a generator lost to the sands of time where the primary particles for a new simulation are copied from the scoring plane hits of a previous simulation.

The implementation here is different than the implementation a few years ago however. Instead of having the generator read the input file, I have the input file be read by the framework. This allows the "stage one" simulation products to be copied into the output file along side the "stage two" products so final analysis only needs to look at one events file. Moreover, it makes implementing the generator very easy once we include a callback to access objects from the event bus _before_ the simulation event is begun. This callback has the signature
```
void prepEvent(const framework::Event& event);
```
- named `prepEvent` to try to emphasize its only for special stuff before the event is simulated
- given a `const` reference to the `Event` because generators should only be `event.get` in this function. `event.add` should only be called by the Producer after the simulation has successfully completed (i.e. was not aborted).

## Check List

I've included two example configurations to help guide folks interested in using this feature in the future and they offer a nice testing ground. The example configurations are not general - they do the two stages I'm designing this feature for. Simulating the tracker-only ("no-cal") detector and doing tracking reconstruction so we can simulate a lot of events quickly before making selections on the reconstructed tracks and only doing the longer simulation of the calorimeters for "dangerous" events.

First stage, only simulate the tracker and do tracking reconstruction.
```
just fire SimCore/exampleConfigs/stage-one-sim-no-cal.py
```
This produces `stage-one-no-cal.root` which is referenced in the second configuration.
```
just fire SimCore/exampleConfigs/stage-two-sim-cal.py
```
This works but there are some concerns.

### works
I have written [a short script](https://github.com/user-attachments/files/31265721/main.py) that does not make a pretty plot but textually confirms that the primary particles in the stage-two "simcal" collection started at z=240mm while the primary particles in the stage-one "simtrk" collection started at the beam origin z=-883mm.

```
simtrk
Primary Particle Origin z:
[[-883], [-883], [-883], [-883], [-883]]
simcal
Primary Particle Origin z:
[[240, 240, 240, 240], [240, 240, 240, 240], [...], [240], [240, 240, 240]]
```

### concerns
The main concern is I get the following warning from Geant4 regularly. Not on every event and not for every primary created.
```
-------- WWWW ------- G4Exception-START -------- WWWW -------
*** G4Exception : GeomNav1002
      issued by : G4Navigator::ComputeStep()
Track stuck or not moving.
          Track stuck, not moving for 10 steps
          in volume -av_1_impr_25_nohole_motherboard5_assembly_pv_0- at point (-255.583,186.106,605.128) (local point (78.7847,1.666,390.209))
          direction: (-0.643129,0.149852,0.750952) (local direction: (-0.643129,-0.750952,0.149852)).
          Potential geometry or navigation problem !
          Trying pushing it of 1e-07 mm ...Potential overlap in geometry!

*** This is just a warning message. ***
-------- WWWW -------- G4Exception-END --------- WWWW -------
```
As the warning states, this probably is a geometry overlap issue. I am using `v14` in this example since that is where the `no-cals` detector is.


tomeichlersmith:
Associated documentation that should be merged when this feature is merged and included in an ldmx-sw release: https://github.com/LDMX-Software/ldmx-software.github.io/pull/68

tomeichlersmith:
I've patched the lint errors that correspond to the code I've written here. The other lint errors correspond to other files (mainly in Tracking/exampleConfigs) which have line lengths that are too long and what not. I'm guessing the python lint failed and we accidentally ignored it or something.

tvami:
We've seen the "concerns" before and it's fixed for v15 + it actually has something that's about the scoring planes, if you move to v15 + switch off the SP these overlaps are gone. @cjbarton151 can say more about it

cjbarton151:
If you're interested you can follow the discussion on overlaps/stuck particles [here](https://github.com/LDMX-Software/ldmx-sw/issues/1894), but tl;dr is that to the best of my knowledge, remaining issues are either due to G4 not playing nice with complex solids generated from CAD imports, or due to boolean solids confusing the overlap checker in v10.

tomeichlersmith:
The ASAN/UBSAN failure is the same failure as #2104 so I'm not fixing it here.

SimCore/exampleConfigs/stage-one-sim-no-cal.py
```diff
@@ -0,0 +1,30 @@
...
+
+p = ldmxcfg.Process("simtrk")
+p.run = 1
+p.max_events = 10
+p.output_files = ["stage-one-no-cal.root"]
+
+from LDMX.SimCore import simulator, generators, sensitive_detectors
+
+sim = simulator.Simulator("simtrk")
+sim.set_detector("ldmx-det-v14-8gev-no-cals", True)
```
tomeichlersmith: Switch to v15

SimCore/python/generators.py
```diff
@@ -223,6 +223,33 @@ class Genie(PrimaryGenerator):
...
+        GDML, but it is probably easiest to just run this generator on a few events
+        with logging set to trace and see a few plane numbers.
+        If an empty list is used, then we use all hits regardless of plane ID number.
+        '31' is the front plane (lowest-z in xy plane) of the Ecal scoring planes
+        and '41' is the front plane of the Hcal scoring planes.
+    """
+
+    coll_name: str = "EcalScoringPlaneHits"
+    pass_name: str = ""
+    select_planes: list[int] = []
```
tomeichlersmith: update default to be front Ecal scoring plane

SimCore/src/SimCore/Generators/FromScoringPlane.cxx
```diff
@@ -0,0 +1,92 @@
...
+    auto momentum{earliest_hit->getMomentum()};
+    auto energy{earliest_hit->getEnergy()};
+    particle->Set4Momentum(momentum[0], momentum[1], momentum[2], energy);
+    // probably not correct from a relativistic perspective
+    particle->SetProperTime(earliest_hit->getTime());
+    // label this particle as a primary for purposes of serialization
+    UserPrimaryParticleInformation* uppi = new UserPrimaryParticleInformation;
+    uppi->setHepEvtStatus(1);
+    particle->SetUserInformation(uppi);
+    // can I copy in the old track_id?
```
tomeichlersmith: try this, not sure if I _want_ to do it this way but could be nice

tomeichlersmith (commented):
just reviewing to draft a to-do list for me

SimCore/exampleConfigs/stage-one-sim-no-cal.py
tomeichlersmith: I'm creating the `ldmx-det-15-8gev-no-cals` detector and going to try to use symlinks to ensure that it is truly a clone of `ldmx-det-v15` but with "no cals". If the symlinks work, I'll update `ldmx-det-v14-8gev-no-cals` to be similar.

tomeichlersmith: symlinks worked but now I'm seeing this message

```
Setting detector for tracking to ldmx-det-v14-8gev
```

which is concerning

tomeichlersmith: just forgot to update the config, so that's not a concern anymore

cjbarton151: In PR #[1899](https://github.com/LDMX-Software/ldmx-sw/pull/1899) I mentioned this in the opening comment (but didn't fix it lol). I stopped investigating because it didn't seem to have a tangible effect on results

tomeichlersmith: There are no updates to the tracking detectors between v14 and v15, so it shouldn't have an effect. I just noticed and was concerned for a future detector change that should have an effect.

SimCore/src/SimCore/Generators/FromScoringPlane.cxx
tomeichlersmith: I tried it and it doesn't do what I would want it to do (i.e. still have the new primaries be the lowest track IDs and count up from there), so I'm not copying in the track Id and leaving this finding in a comment

tomeichlersmith:
I'm still seeing those stuck/overlap issues with the v15 detector. What's concerning to me is that I don't see these warnings when running the normal simulation which makes me think there is something specifically wrong with how I'm constructing the primaries from the scoring plane hits.

tomeichlersmith:
It's a small sample, but I'm still not seeing any trends. I did `StepPrinter(process_name = "Primary")` to see if the warnings were all originating from some kind of primary (e.g. I was thinking that maybe the low-energy primaries would cause issues), but that was not the case. Many warnings are not originating from primaries and some of them are originating from primaries with several hundred MeV of energy.

I am going to proceed forward and ignore the warnings now especially since #1894 appears to fully list the issues I'm seeing as well.

tvami (approved):
Look great, thanks!

PR #2115, opened 2026-08-25
Fix G4Box cast and null-check trackers for VST geom

tvami:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
Resolves https://github.com/LDMX-Software/ldmx-sw/issues/2114

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [x] I ran my developments and the following shows that they are successful.

PR #2117, opened 2026-08-27
Tracking waveform analysis & clustering

mgignac:
I am updating _ldmx-sw_, here are the details:

- Added ESA DAQ map  — electronics (feb, hybrid) translated to Acts surface/geometry
- SiStripWaveformFitProcessor — real-data pulse fitter, address via DAQ map, per-channel pedestal noise
<strike>- recoil.gdml geometry fix — corrected stereo/axial z-order + stereo tilt sign</strike>
- Gutted SiStripWaveform and haromized with simulation workflow in  FittedSiStripHit. Added noise field.
- cluster de-hardcoding — added real per-channel noise & per-sensor strip counts, MC path unaffected
- Added/updated a number of example configs:, raw_to_measurements (with ability to stop at a given stage), plot_*)
- Loosened the waveform-builder gates from 5σ/4 high, 3σ/5 low and clustering 4/3/4 to loose-A 4σ/3, 2.5σ/3 and clustering 3/2/3 — recovering ~3× the hits and track candidates
  
Updates discussed during August 27th Tracking Meeting:
https://indico.fnal.gov/event/74602/#2-esa25-data-analysis-update

## Check List


tvami:
For the geom update we really should have
https://github.com/LDMX-Software/ldmx-sw/pull/2115
in, otherwise the CI doesnt run so your change is not being tested

mgignac:
I still have a bit more work to do before this should go in, so that shouldn't be a problem. I'll wait until #2115 is merged in before removing the draft PR from here.

mgignac:
The geometry changes are a bit more involved than I thought, due to how sensors are mapped in `TrackingUtils.cxx` -- so I'm removing those from this PR and will follow up with one dedicated to that. Core of the issue is that we are recycling the ldmx recoil geometry in the ESA (and future cosmic ray) test stand, but the hardware setup was fundamentally different.

Tracking/src/Tracking/Reco/StripClusterProcessor.cxx
```diff
@@ -102,7 +117,14 @@ void StripClusterProcessor::produce(framework::Event& event) {
...
       // -------------------------------------------------------------------
       using namespace tracking::digitization;
-      const int n_int = N_READOUT_STRIPS / 2;  // integer division = 383
+      // Centre-strip offset: N/2 (integer division).  For MC this is the fixed
+      // N_READOUT_STRIPS constant; for real data, if a DAQ map was supplied,
+      // use that sensor's real strip count so the local origin sits at its
+      // centre.
+      int n_strips = N_READOUT_STRIPS;
+      auto it_ns = layer_n_strips_.find(layer_id);
+      if (it_ns != layer_n_strips_.end()) n_strips = it_ns->second;
```
tvami, line 126: I dont know if I agree to have this diverge here for MC and data. Why not have both rely on the DAQ mapping? Maybe you could add a warning if the MC has a different number of strips than the DAQ map, but I dont think we should silently just override and move on

Tracking/src/Tracking/Reco/StripClusterProcessor.cxx
```diff
@@ -102,7 +117,14 @@ void StripClusterProcessor::produce(framework::Event& event) {
       // µm, etc.
       // -------------------------------------------------------------------
       using namespace tracking::digitization;
-      const int n_int = N_READOUT_STRIPS / 2;  // integer division = 383
+      // Centre-strip offset: N/2 (integer division).  For MC this is the fixed
```
tvami: > Centre

Just a s side note that we have been using American English in ldmx-sw, while here it's just a comment, for variable names etc it's good to keep it in mind

Tracking/src/Tracking/Digitization/StripClusterer.cxx
```diff
@@ -120,10 +123,11 @@ std::vector<StripClusterer::ClusterCandidate> StripClusterer::findClusters(
       const double amp = hit.getAmplitude();
 
       // Accumulate cluster quantities.
+      const double noise = hitNoise(hit);
```
tvami, line 126: if there is no match, the default will be noise = 0 and then later you divide with a derived quantity. I'd exit early with a warning that this channel doesnt have noise in the mapping

tvami (changes requested):

tvami:
/run-validation
 * I'm surprised the previous failed for HCAL

Tracking/src/Tracking/Reco/StripClusterProcessor.cxx
mgignac: Part of the issue is that the test-beam sensors are HPS ones and differ in channel count than what LDMX will have. I agree it would be better to have a DAQ map for MC simulation too, but this would require rehauling that workflow which I'd prefer to defer to @bloodyyugo and/or a follow up PR, otherwise this one gets too large.

Tracking/src/Tracking/Reco/StripClusterProcessor.cxx
mgignac: Noted -- fixed comment non-the-less.

Tracking/src/Tracking/Digitization/StripClusterer.cxx
mgignac: The code already had protection in place for noise values of zero (not part of this PR):
https://github.com/LDMX-Software/ldmx-sw/blob/trunk/Tracking/src/Tracking/Digitization/StripClusterer.cxx#L145

tvami (approved):

tvami:
/run-validation
 * we actually have new golds now

bloodyyugo (approved):

PR #2121, opened 2026-09-08
Hcal cosmic teststand geometry at ESA

WildeHedin:

This code adds a gdml file called ldmx-hcal-teststand that describes some of the HCal components of the first proposed design of the teststand that will be used to measure cosmic muons. The geometry includes two HCal sections that each consist of 8 layers of 8 scintillator bars and a scintillator cover. Mounting plates have not been included. In addition the geometry has been defined in the file hcal_geometry.py.

When simulating cosmic muons through the geometry and using the processors HcalDigiProducer and HcalRecProducer, some strange discrepancies occur in the energy detected in the scintillator bars. This does not occur when using HcalSimpleDigiAndRecProducer. I have not had time to figure out why. This should be considered if the geometry is ever used. For more details check out my thesis: [https://lup.lub.lu.se/student-papers/search/publication/9246704](url)

## Check List

This is the teststand design that the geomety is based on:
Skärmbild 2026-02-28 162234
These two images show the detector geometry when using the Geant4 visualizer OpenGL:
IMG_0787 (1)
prototyp_geometry_side
This image show the detected energy in the scintillator bars dependent on the cosmic muons incoming angle (using HcalSimpleDigiAndRecProducer):
touseforPRangle


tvami:
There is a naming convention for the geometries

https://github.com/LDMX-Software/ldmx-sw/tree/trunk/Detectors/data#naming-convention

and given that this will at one point become the full cosmic test stand, not just HCAL, I'd really suggest to remove hcal in the name. 

With that in mind, I think some of these GDMLs should be merged into less files, so it's easier to add the other subdetectors later

DetDescr/python/hcal_geometry.py
```diff
@@ -539,3 +624,6 @@ class HcalGeometry:
     v2_prototype: HcalReadoutGeometry = field(
         default_factory=HcalReadoutGeometry.make_v2_prototype
     )
+    cosmic_prototype: HcalReadoutGeometry = field(
```
cjbarton151, line 627:
```suggestion
    teststand_prototype: HcalReadoutGeometry = field(
```

Detectors/data/ldmx-teststand/detector.gdml
```diff
@@ -0,0 +1,67 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE gdml [
+<!ENTITY constants SYSTEM "constants.gdml">
+<!ENTITY materials SYSTEM "materials.gdml">
```
cjbarton151, line 4: A few months ago I made one big `materials.gdml` file and moved it to a separate directory. It's now shared between most of the current geometries. I'd appreciate it if you could update that file with any new materials you added, then delete this file. If you don't feel like messing around with that, I can tidy it up in a separate commit

cjbarton151: You can look at the beginning lines of the gdml files in another current geometry, for example ldmx-det-v15-8gev, to see how the shared directory is implemented (in this case you could just replace "materials.gdml" with "shared_gdml_files/materials.gdml")

Detectors/data/ldmx-teststand/gdml.xsd
```diff
@@ -0,0 +1,225 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cjbarton151, line 1: The XSD files are all now shared as well, and in this case it's automatic, so you can safely delete all .xsd files in this commit.

Detectors/data/ldmx-teststand/detector.gdml
cjbarton151: You can do the same with visattributes.gdml, and update the VisAttributes in hcal.gdml to match the VisAttributes available in the shared file

cjbarton151 (changes requested):

tvami:
> There is a naming convention for the geometries
> 
> https://github.com/LDMX-Software/ldmx-sw/tree/trunk/Detectors/data#naming-convention
> 
> and given that this will at one point become the full cosmic test stand, not just HCAL, I'd really suggest to remove hcal in the name.
> 
> With that in mind, I think some of these GDMLs should be merged into less files, so it's easier to add the other subdetectors later

Sorry I was not explicetly suggesting a name earlier, but based on the README this should be called maybe `ldmx-esa-cosmic-v1`

tvami:
@mgignac suggested that this should be `ldmx-esa-cosmic26-v1` which I think is a good addition to the name

cjbarton151 (approved):
Looks good to me! Visualization works, and I ran a test config with LHE files from the cosmic muon generator and I'm getting nonzero output in HcalSimHits. Thanks for being persistent and continuing work on this with us :^)

bryngemark (approved):
Thanks @WildeHedin for sharing this work back to LDMX! Looks good to me.

PR #2122, opened 2026-09-08
pr needs history to be created (i think)

tomeichlersmith:
this updates the workflow to fetch the entire history if we are running on trunk while keeping it only to the commit under test otherwise


tomeichlersmith:
The workflow running on this PR correctly only used a fetch-depth of 1

https://github.com/LDMX-Software/ldmx-sw/actions/runs/34242188755/job/102114904910?pr=2122#step:3:4

tvami (approved):

PR #2128, opened 2026-09-11
Propagate beamEfrac to rechits in TS

### LINKED ISSUE #1336: New TS digi+rechit pipeline doesn't propagate the fraction of hit energy deposited by beam electrons
When we designed the first TS digi (in effect, just a sim hit reconstruction layer), we introduced a variable `beamEfrac` to keep track of which fraction of a hit's energy  was actually coming from beam electrons. This number is then propagated through clustering and tracking, and is a handy method for, at very little extra cost in terms of memory, assess the quality/purity of the tracks found, and to investigate properties of fake tracks. 

This metric was dropped in the `RecHitProducer`, following the logic that this will never ever be an actual reconstructed observable. Fair enough. 

I do think it's still important to be able to assess performance of our reconstruction even if it has an improved level of realism (digi with e.g. finite time resolution), and would not have to revert to the old "digi" which will not give the same answer.

**Proposed solution** 
To implement keeping this one simulation-level info float attached to our rechit collection.

**Alternative solutions considered** 
To make a new truth info collection that can be used, by truth-reco association. I think this will be disadvantageous to our memory/disk space footprint.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1336

## Check List

### DIFF
--- a/TrigScint/src/TrigScint/TrigScintQIEDigiProducer.cxx
+++ b/TrigScint/src/TrigScint/TrigScintQIEDigiProducer.cxx
@@ -5,6 +5,7 @@
 #include "Framework/Exception/Exception.h"
 #include "Framework/Logger.h"
 #include "Framework/RandomNumberSeedService.h"
+#include "SimCore/Event/SimParticle.h"
 
 namespace trigscint {
 
@@ -23,6 +24,10 @@ void TrigScintQIEDigiProducer::configure(
   input_collection_ = parameters.get<std::string>("input_collection");
   input_pass_name_ = parameters.get<std::string>("input_pass_name");
   output_collection_ = parameters.get<std::string>("output_collection");
+  sim_particles_coll_name_ =
+      parameters.get<std::string>("sim_particles_coll_name");
+  sim_particles_passname_ =
+      parameters.get<std::string>("sim_particles_passname");
 
   // QIE specific parameters initialization
   maxts_ = parameters.get<int>("maxts");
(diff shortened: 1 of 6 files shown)


bryngemark:
Thanks for working on this long-standing feature request @tvami ! The code looks good, do you have any plots demonstrating that the beamEfrac setting works, for instance comparing beamEfrac in rechits using the "direct" vs digi emulation path?

tvami:
yes, 100% correlated
Screenshot 2026-09-11 at 06 59 07

TrigScint/src/TrigScint/TrigScintQIEDigiProducer.cxx
```diff
@@ -94,12 +102,33 @@ void TrigScintQIEDigiProducer::produce(framework::Event& event) {
   // loop over sim hits and aggregate energy depositions for each detID
   const auto sim_hits{event.getCollection<ldmx::SimCalorimeterHit>(
       input_collection_, input_pass_name_)};
+  const auto particle_map{event.getMap<int, ldmx::SimParticle>(
```
bryngemark: we need a graceful exit if this is real data and no simparticle/simhit collection exists; the testbeam pipeline uses the same class

bryngemark (commented):
we don't have a testbeam data CI yet, right? we need to make sure this doesn't break the real data path.

tvami:
> we don't have a testbeam data CI yet, right

no, I can work on that next. The reason I havent done that yet is the files are big so we need to figure out how to slim them down to just a few events we use in the testing

cjbarton151 (commented):
Looks good; I'll give it another review once the comment is resolved

TrigScint/src/TrigScint/TrigScintQIEDigiProducer.cxx
tvami: Done in https://github.com/LDMX-Software/ldmx-sw/pull/2128/changes/5e91ad2ad36854fe864de4fb3a907f303b70d254

tvami:
> we don't have a testbeam data CI yet, right? we need to make sure this doesn't break the real data path.

I tested this with the new CI https://github.com/LDMX-Software/ldmx-sw/pull/2138
 --> it did not break it @bryngemark

cjbarton151 (approved):

bryngemark (approved):
Thanks for these checks, looks good to me!

PR #2130, opened 2026-09-11
Fix production problems (coming from preemption at SLAC)

### LINKED ISSUE #2129: Fix problems seen while EN production on SDF

When a job on the preemptable QOS is killed mid-write, the output file keeps its `LDMX_Events` tree but loses the `LDMX_Run` header, and fire accepts such a file as valid input and then crashes: 
```
Warning in <TFile::Init>: file .../ldmx_v4.8.1_GENIE_G21_11b_00_000_en_ti_run7478_event.root probably not closed, trying to recover 
```
→ 
```
Error in <TTreeReader::TTreeReader>: No TTree called LDMX_Run was found in the selected TDirectory.
```
 → 
```
*** Break *** segmentation violation in tracking::reco::TruthSeedProcessor::createTruthTrack(...)
```

With `skip_corrupted_input_files` never firing because a missing `LDMX_Run` is not counted as corruption.

Move  `LDMX_Run`  to file open? 

**Additional context**

EN production on SDF

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/2129

## Check List

### DIFF
--- a/Framework/include/Framework/EventFile.h
+++ b/Framework/include/Framework/EventFile.h
@@ -234,9 +234,11 @@ class EventFile {
    * overwritten. Process calls it as soon as the headers are known and
    * again on close, so a killed job still leaves a usable file.
    *
+   * @param[in] completed true when closing cleanly. Only the final call
+   * should pass true.
    * @throw Exception if call this function on a non-output file.
    */
-  void writeRunTree();
+  void writeRunTree(bool completed = false);
 
   /**
    * Update the RunHeader for a given run, if it exists in the input file.
--- a/Framework/include/Framework/EventFile.h
+++ b/Framework/include/Framework/EventFile.h
@@ -230,6 +230,10 @@ class EventFile {
   /**
    * Write the map of run headers to the file as a TTree of RunHeader.
    *
+   * Safe to call more than once; any run tree already in the file is
+   * overwritten. Process calls it as soon as the headers are known and
+   * again on close, so a killed job still leaves a usable file.
+   *
    * @throw Exception if call this function on a non-output file.
    */
   void writeRunTree();
@@ -322,6 +326,9 @@ class EventFile {
 
   /// Map of run numbers to RunHeader objects (owned via shared_ptr)
   std::map<int, std::shared_ptr<ldmx::RunHeader>> run_map_;
+
+  /// The run tree, owned by file_ once written
+  TTree* run_tree_{nullptr};
 
   enableLogging("EventFile")
 };
--- a/Framework/include/Framework/RunHeader.h
+++ b/Framework/include/Framework/RunHeader.h
@@ -53,6 +53,10 @@ namespace ldmx {
  *
  * ## v5
  * Include a member to track the ldmx-sw version
+ *
+ * ## v7
+ * Add the completed_ member so a consumer can tell a complete file from one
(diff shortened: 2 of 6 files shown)


tomeichlersmith (commented):
I am very nervous about this "solution" because, while it allows you to specifically avoid the pre-emption issues, it will also enable other types of erroneous runs to be "valid" and readable. I don't really want to get into the situation where jobs that are killed produce files that are the same as valid files. This will just encourage users to ignore errors/warnings which I don't believe will be helpful.

Maybe a better solution is to have a "resume from file" run mode? so that a pre-empted run can be completed in a deterministic manner? As I mentioned in the issue, I'm not confident that handling pre-emption by a specific job control system at a specific cluster should be in the purview of ldmx-sw.

tvami:
>  I don't really want to get into the situation where jobs that are killed produce files that are the same as valid files. This will just encourage users to ignore errors/warnings which I don't believe will be helpful.

I could have a follow-up PR that by default will not read file where `isComplete` is false. That should address your worry.

>  I'm not confident that handling pre-emption by a specific job control system at a specific cluster should be in the purview of ldmx-sw.

While I agree with this statement in general, this specific system is the one that we use for big productions so that does mean we need to deal with it somehow and getting out of the preemption would require money which we dont have.

tomeichlersmith (approved):
Ok, I understand the reasoning and while I'm not happy about it, I agree that it is a necessary feature to make do with our current resources. I like the idea of having a check to force the user to acknowledge that they are processing an incomplete file.

tvami:
Thanks, please feel free to edit the issue I just made: https://github.com/LDMX-Software/ldmx-sw/issues/2135

jmmans:
How long is the user to be allowed to add to the run tree?

jmmans:
In general, is the pre-emption issue for both simulation and reco or is this primarily a simulation issue?  I would assert that it is quite dangerous to allow uncontrolled splitting of reco outputs without a strong database backend tracking provenance.

tvami:
> I would assert that it is quite dangerous to allow uncontrolled splitting of reco outputs without a strong database backend tracking provenance.

I totally agree with this!

But for making the GEN events, having 70% of the events produced is better than 0% (I have had preemptions where 30 min more would have finished everything). But also 20% statistics is better than nothing too. We just need to count event after the fact.

But yes this is not to be used for RECO.

tomeichlersmith:
> How long is the user to be allowed to add to the run tree?

In `beforeNewRun`, so for simulation/production runs, just at the beginning. If there is more than one run, then periodically throughout program execution.

jmmans:
For GEN/SIM, I would be VERY comfortable with allowing this.  This is effectively the "origin" step.  Any assumption that a given file has exactly a certain number of events is foolish.  If there is an input file, I would be uncomfortable with this, but with no input file...  pre-empt away.  :-)

tvami:
Now that you brought this up, maybe it should be moved to `Process::newRun`?

tvami:
@tomeichlersmith @jmmans is https://github.com/LDMX-Software/ldmx-sw/pull/2130/changes/f04cb1a4ea1ec937b75614da2e4a7eb2b1cc0bfc better?

LiamBrennan-UCSB (approved):

PR #2137, opened 2026-09-14
ESA25 TS Analysis

mgignac:
I am updating _ldmx-sw_, here are the details.

### What are the issues that this addresses?
Decoding for the TS -- work done by @Layan-Sarayra and @EBerzin. Now needed for next steps for TS+Tracker data analysis, so I'm pushing on this and helping to get it merged in.

## Check List
- [x] I successfully compiled _ldmx-sw_ with my developments.
- [x] I read, understood and follow the [coding rules](https://ldmx-software.github.io/developing/coding-rules.html).

- [x] I ran my developments and the following shows that they are successful.

Issue #1341, opened 2020-12-17
DMQ analyzer for digi collections


awhitbeck:
Need a DQM analyzer for new digi collections

bryngemark:
The priority on this potentially increased now that there will very soon be an automatic validation procedure for PRs, involving running the DQM plots and comparing them to a set of reference plots created upon making stable releases.

I think the old digi plots still work for RecHits, since they have the same set of variables. It's mostly a matter of also making histograms of ADC, TDC and CID values for the QIE digis. And any other plots that could be helpful for checking that nothing is broken, of course.

GNiramay:
Ok. I think we can also add a few more histograms for bar IDs and perhaps the linearized charges if required.

PR #2138, opened 2026-09-14
Introduce ESA VST CI

### LINKED ISSUE #1940: Add a new CI that test the VST testbeam data

We have a lot of dev now with the ESA VST data

I'd like a CI that tests on the top of that. Prob store the input in the `ci-data` repo similar to what we do for the WAB CI

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1940

## Check List


TrigScint/include/TrigScint/QIEAnalyzer.h
```diff
@@ -48,6 +48,8 @@ class QIEAnalyzer : public framework::Analyzer {
 
   // plotting stuff
   int n_ev_{200};
```
bryngemark, line 51: i think in effect this was always used for `n_event_displays`

.github/validation_samples/esa_vst/config.py
```diff
@@ -0,0 +1,103 @@
...
+    trk_unpack,
+    trk_decoder,
+    trk_peds,
+    trk_waveforms,
+    trk_dqm,
+    ts_unpack,
+    ts_decoder,
+    ts_readout,
+    ts_hits,
+    ts_clusters,
```
bryngemark, line 86: this sequence stops at TS clustering, but there is no DQM for clusters invoked (nor for reconstructed hits). if we'd want to make sure tracking also works, this is an implicit clustering check, but we'd need to add all three pads.

bryngemark: for tracking to work we might want to make use of the fiber time shift possibility, since there were some shifts in different fibers (@EBerzin and @Layan-Sarayra know everything about this). even better, a data-derived gain and pedestal file instead of the generic 2e6 gain etc (it could be for this run or some other, it is stable to the level needed to find hits) but check with the two TS data experts what checks are needed/useful. we also need to know the pad-dependent (time-of-flight dependent) cluster pad time wrt to event t=0.

one reason to do this extra properly is that generally, CI tests are a great way to learn how to run different workflows.

bryngemark (commented):
Excited to see this! Great step forward. I have some implementation questions as you can see, where input from others might help.

TrigScint/include/TrigScint/QIEAnalyzer.h
tvami: Yes but not only. It is also the event-axis binning of `h_tdc_fire_chan_vs_event_` given that I generated gold with this setup, I'd refer to keep this as is.

.github/validation_samples/esa_vst/config.py
tvami: This sounds good, but I also suggest to do it in a follow-up PR. There are other thing in ongoing PRs that we should add to this CI anyway: https://github.com/LDMX-Software/ldmx-sw/pull/2137 https://github.com/LDMX-Software/ldmx-sw/pull/2117 https://github.com/LDMX-Software/ldmx-sw/pull/2029

Do you agree @bryngemark ?

tvami:
Also @bryngemark 

> this sequence stops at TS clustering, but there is no DQM for clusters invoked (nor for reconstructed hits).

That is the same point as in 

https://github.com/LDMX-Software/ldmx-sw/issues/1341

right?

Tracking/python/dqm.py
```diff
@@ -992,3 +992,46 @@ def __post_init__(self):
...
+    """
+
+    raw_hits_collection: str = "RawSiStripHits"
+    subtracted_hits_collection: str = "TrackerHits"
+    waveforms_collection: str = "TrackerWaveforms"
+    input_pass_name: str = ""
+    n_hybrids: int = 4
+
+    def __post_init__(self):
+        self.histogram("n_raw_hits", "Raw hits per event", 100, 0, 40000)
```
mgignac: Some of these ranges should probably be adjusted, otherwise the histograms will only populate one bin? Especially for ESA, capping at 1000 would be plenty. Similarly for number of waveforms, maybe capping around 60 is more suitable (4 hybrids x ~10 electrons max, typically).

mgignac (approved):
Looks good enough to me-- we still have quite some work to do, and it will be very helpful to be able to track changes as that work progress with this. Thanks for implementing!

Tracking/python/dqm.py
tvami: I looked at the histograms and adjusted accordingly in https://github.com/LDMX-Software/ldmx-sw/pull/2138/changes/1b30e81a7f988327ce3ce1f7e3f4d0caa9b44bd0

TrigScint/include/TrigScint/QIEAnalyzer.h
tvami: I did this in https://github.com/LDMX-Software/ldmx-sw/pull/2138/changes/fb390a087dd0719e1690a7efbb021a5da21c5636

tomeichlersmith (approved):

PR #2140, opened 2026-09-15
Give standalone processors dependency includes

### LINKED ISSUE #2106: Processor from file doesn't handle Geant4 includes
`processor_from_file` lets us list libraries through `needs`, but this currently only adds `-l<name>` to the hand-written compiler command. It does not add the include directories or compile definitions required by those libraries.

For example, this processor includes a public header from `SimCore_G4User`:

```cpp
#include "SimCore/G4User/UserAction.h"

class ProcessorFromFileGeant4Reproducer : public framework::Analyzer {
 public:
  ProcessorFromFileGeant4Reproducer(const std::string& name,
                                    framework::Process& process)
      : framework::Analyzer{name, process} {}

  void analyze(const framework::Event&) override {}
};

DECLARE_ANALYZER(ProcessorFromFileGeant4Reproducer);
```

```python

p = ldmxcfg.Process("processor_from_file_geant4_reproducer")
p.max_events = 0
p.sequence = [
    ldmxcfg.processor_from_file(
        "processor_from_file_geant4_reproducer.cxx",
        class_name="ProcessorFromFileGeant4Reproducer",
        needs=["SimCore_G4User"],
    )
]
```

Compilation fails with:

```text
SimCore/G4User/UserAction.h:20:10: fatal error:
G4EventManager.hh: No such file or directory
```

The emitted command links `SimCore_G4User`, but it does not contain the Geant4 include directory:

```text
g++ ... -lSimCore_G4User -lFramework \
  -I/usr/local/include/root \
  -I<ldmx-sw install>/include \
  -L<ldmx-sw install>/lib
```

Adding `-I/usr/local/include/Geant4` by hand fixes this case. The same problem will occur for any library whose public headers depend on headers outside the LDMX and ROOT include directories.

A quick fix would be to allow appending / prepending stuff to the compiler invocation through python, but a more long term solution would be to use the cmake targets directly?

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/2106

## Check List


tvami:
@EinarElen is this what you had in mind?

PR #2148, opened 2026-09-16
Use target position in seed fit

### LINKED ISSUE #1912: Include target location in the tracking

In some cases we are running into the issue that we dont have the 5 needed hit to form a track. 

Including the vertex position in the target from the tagger tracker could help that situation.

We could explore iterative tracking with less and less hits.

tvami:

Resolves https://github.com/LDMX-Software/ldmx-sw/issues/1912

## Check List

I kept the default off tho for now


Tracking/src/Tracking/Reco/SeedFinderProcessor.cxx
```diff
@@ -170,6 +198,21 @@ void SeedFinderProcessor::produce(framework::Event& event) {
     }
   }
 
+  // Constraints that enter the fit: tagger tracks, else the beam spot
```
bloodyyugo, line 201: What is this beamspot constraint?

bloodyyugo (approved):

Tracking/src/Tracking/Reco/SeedFinderProcessor.cxx
tvami: the 20 by 80 rectangle (well it's configurable, but that's the default)

bloodyyugo: Is this useful?

tvami: It's better than the full plane :D

mgignac (approved):

PR #2156, opened 2026-09-17
Add ESA25 vertical-slice-test detector (ldmx-esa25-v1) + teststand tracker support

mgignac:
I am updating _ldmx-sw_, here are the details

Adds the standalone silicon-tracker test-stand geometry used for the ESA25 slice test, plus the minimal DetDescr/SimCore/Tracking plumbing to reconstruct it.

Core changes:
- New detector Detectors/data/ldmx-esa25-v1/ (hardware-correct recoil z-order/same stereo sign for both stations).
-  SD_TRACKER_TESTSTAND = 8 subdetector id (DetDescr) + TrackerID interpreter and python binding.
- TrackerSD.teststand() sensitive detector (SimCore).
- TrackingUtils::getSensorID branch for the test-stand tracker.
-  ldmx-esa25-v1 added to the ECAL/HCAL geometry-provider valid lists.

Why the sensor-parity change:

The test-stand modules are mounted hardware-correct: the stereo sensor sits upstream, the axial sensor downstream. Acts numbers a station's two sensors by z, so sensor 0 = upstream = stereo — i.e. sensor = layerID % 2, the opposite parity to the recoil/v14 branch, which assumes axial-upstream. Without this branch, hits map to the wrong Acts surface and y is meaningless.

Why a new detector instead of patching ldmx-reduced-v3:

reduced-v3 encodes the v14 z-order convention (axial upstream) and is used by production sim/reco. The real ESA hardware has the opposite z-order; "fixing" it in reduced-v3 would silently change a shared production geometry and break everything relying on its current convention. Giving the test stand its own detector + subdetector id keeps the hardware-correct z-order fully isolated and leaves v14/reduced-v3 untouched.

## Check List


DetDescr/src/DetDescr/DetectorIDBindings.cxx
tvami: according to the bot this is not needed, can you check if it's true?

tvami:
@mgignac please delete the reduced geom, and add this new and renamed version to both the "reduced" sim based CI and the real data VST CI. Thanks!

DetDescr/src/DetDescr/DetectorIDBindings.cxx
mgignac: It seems this is actually needed -- removing the include fails the
  build with BOOST_PYTHON_MODULE ... error: expected constructor, destructor, or type conversion. As per the CI job message, I've marked it as needed and seems the bot is no longer flagging it here.

tvami (approved):

tomeichlersmith (approved):

PR #2157, opened 2026-09-17
TrigScint Geometry Provider for ESA25

mgignac:

TS geometry-provider + measurement-producer feature for the ESA25 slice test — the TS-side counterpart to the tracker geometry work. 

What it adds (all new files):

- DetDescr — TrigScintGeometry class (.h/.cxx + trigscint_geometry.py): a TS geometry description modeled on HCal/ECal.
- TrigScint — TrigScintGeometryProvider.cxx (conditions/geometry provider, "following closely HCal"), ts_daq_map.py + ts_daqmap_esa25_slice_test.json (the TS DAQ map), and edits to TrigScintRecHitProducer.h.
- Tracking — TrigScintMeasurementProducer (.h/.cxx + ts_measurements.py + example config): produces the TrigScintMeasurements.
- TS Calibration files from esa25 
- CMake wiring (Tracking + TrigScint + a TS dependency in the Tracker CMake).

## Check List


tvami:
@mgignac this is looking pretty good, please add the new element to the VST CI so the bot is also testing it, thanks!

mgignac:
@tvami I've added these `ldmx::Measurement` creation to the CI (which relies on geometry providers), so we should be testing these new elements. I had to start to tackle issue #2160 as well to integrate these geometry providers into the CI, but it raises another issue since the DQM histograms arising from that workflow through the EventBuilder doesn't work with this workflow. I would suggest we address that issue in another PR before closing #2160.

tvami:
/run-validation
 * after the CI edits

Tracking/src/Tracking/Reco/TrigScintMeasurementProducer.cxx
```diff
@@ -0,0 +1,127 @@
...
+
+      const auto p0 = geom.getBarPosition(geo_module, b0);
+      const auto p1 = geom.getBarPosition(geo_module, b1);
+      const double x = (1. - f) * p0.X() + f * p1.X();
+      const double y = (1. - f) * p0.Y() + f * p1.Y();
+      const double z = (1. - f) * p0.Z() + f * p1.Z();
+
+      ldmx::Measurement meas;
+      meas.setGlobalPosition(static_cast<float>(x), static_cast<float>(y),
+                             static_cast<float>(z));
```
tvami: x,y,z could just be a float in the first place and then you dont need to cast it, right?

tvami (approved):
I had a small comment about casting, otherwise it looks good to me

Tracking/src/Tracking/Reco/TrigScintMeasurementProducer.cxx
mgignac: Good catch — done, x/y/z are now float and the casts are gone

tomeichlersmith (approved):

PR #2159, opened 2026-09-18
Epeets_cosmics_display_work

EmrysPeets:
I am updating _ldmx-sw_ to add a browser display and modular geometry for the ESA cosmic test stand.

This builds on `iss2119-add-hcal-teststand-geometry` and implements the ECal and other ESA subsystem geometries from `ldmx-reduced-v3` within the cosmic test stand. The stand layout and support renderings come from the newer `CosmicTestStand.stp` CAD model. The STP-derived geometry guides detector placement, while the existing GDML provides the ECal’s internal materials and sensitive layers.

The HCal is divided into three stations: one at the top of the stand and two surrounding the ECal. Each station has its own commented GDML file, with 104 individually placed scintillator bars across ten crossed layers. The bar counts are inferred from the CAD dimensions and still need confirmation against the hardware.

The package includes:
- Separate subsystem GDML files so individual components can be updated.
- An interactive browser display with detector visibility controls and an illustrative cosmic-ray trajectory.
- A README with installation instructions, example images, and instructions for sharing the display as a single offline HTML file.
- Bar positions, copy numbers, source hashes, and validation results.

The CAD supports are included only in the display. The ray markers show geometric intersections, not simulated detector hits. This update will be combined with Takumi Britt's tracking + TS work to 

## Check List

- [ ] I successfully compiled _ldmx-sw_ with my developments. The transport checks used a prebuilt LDMX runtime; I have not completed a fresh C++ build.

All nine GDML files pass schema validation, and ROOT loads both detector entry points. Checks confirm unique bar IDs, containment within station envelopes, and no bar-to-bar overlaps. Both entry points completed a one-event Geant4 transport check with sensitive detectors disabled. The viewer also builds successfully in a fresh Python environment.

Cosmic stand with segmented HCal

This remains a draft pending review of detector alignment, whole-detector overlaps, and HCal readout integration. The existing readout-position model does not describe this three-station layout. Independently loadable subsystem files also repeat material definitions, which currently produce Geant4 warnings.


tvami:
Does this mean https://github.com/LDMX-Software/ldmx-sw/pull/2121 's content is already in here? i.e. this supersedes that?

tvami:
@EmrysPeets please clean this up, remove all your pngs and other files that are documentation. You can add those to https://github.com/LDMX-Software/ldmx-software.github.io under the event display. Also rename the geometries to have a single geometry called `ldmx-esa-cosmic26-v1`. Hannah's PR is merged, so please rebase this branch to trunk

